PR #2, opened 2017-10-19
Proof of concept for HipChat channel.

gkillough:
DO NOT MERGE! 

Opening for feedback.


src/main/java/com/blackducksoftware/integration/hub/alert/channel/SupportedChannels.java
```diff
@@ -22,15 +22,8 @@
  */
 package com.blackducksoftware.integration.hub.alert.channel;
```
gkillough: If we like this approach, this will eventually be an enum. I'm leaving them as strings for ease of refactoring right now.

src/main/java/com/blackducksoftware/integration/hub/alert/config/DataSourceConfig.java
```diff
@@ -41,7 +41,8 @@
     @Bean
     public DataSource dataSource() {
         final EmbeddedDatabaseBuilder builder = new EmbeddedDatabaseBuilder();
-        final EmbeddedDatabase dataSource = builder.setType(EmbeddedDatabaseType.H2).addScript("db/create-notification-schema.sql").addScript("db/create-notification-db.sql").build();
```
gkillough: We still need to determine where each database is configured.

src/main/java/com/blackducksoftware/integration/hub/alert/Application.java
```diff
@@ -56,7 +56,7 @@
 import com.google.gson.GsonBuilder;
 
 @EnableAutoConfiguration(exclude = { BatchAutoConfiguration.class })
-@EnableJpaRepositories(basePackages = { "com.blackducksoftware.integration.hub.alert.datasource.repository" })
+@EnableJpaRepositories(basePackages = { "com.blackducksoftware.integration.hub.**.datasource.repository" })
```
jrichardBD: Is the plant to have multiple 'datasource.repository' packages? Why wouldnt we put the classes in the same package?

gkillough: We will be creating different *datasource.repository* packages in each channel, so we want to pick up each of those. I would be alright with switching to use the same package for all channels in this case because it's focused on DB operations and not channel operations. @psantos1113 requested separate packages, so let's discuss with him.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/rest/ChannelRestConnectionFactory.java
```diff
@@ -0,0 +1,71 @@
...
+public class ChannelRestConnectionFactory {
+    private static final Logger logger = LoggerFactory.getLogger(ChannelRestConnectionFactory.class);
+
+    public static RestConnection createUnauthenticatedRestConnection(final String stringUrl) {
+        final URL url = getUrlFromString(stringUrl);
+        return createUnauthenticatedRestConnection(url);
+    }
+
+    public static RestConnection createUnauthenticatedRestConnection(final URL url) {
+        return createUnauthenticatedRestConnection(url, new PrintStreamIntLogger(System.out, LogLevel.INFO), 5 * 60 * 1000);
```
gkillough: TODO: Use a better logger

src/main/java/com/blackducksoftware/integration/hub/alert/channel/rest/ChannelRestConnectionFactory.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.blackducksoftware.integration.exception.IntegrationException;
+import com.blackducksoftware.integration.hub.rest.RestConnection;
+import com.blackducksoftware.integration.hub.rest.UnauthenticatedRestConnection;
+import com.blackducksoftware.integration.log.IntLogger;
+import com.blackducksoftware.integration.log.LogLevel;
+import com.blackducksoftware.integration.log.PrintStreamIntLogger;
+
+public class ChannelRestConnectionFactory {
```
gkillough: If we decide to go this route, this should probably live in hub-common.

gkillough:
Closing because changes have been squashed and merged.

PR #7, opened 2017-10-25
Ps realtime refactor

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestNotificationProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+import com.blackducksoftware.integration.hub.notification.processor.NotificationCategoryEnum;
+
+@Component
+public class DigestNotificationProcessor {
+
+    public List<AbstractChannelEvent> processNotifications(final DigestTypeEnum digestType, final List<NotificationEntity> notificationList) {
+        final DigestRemovalProcessor removalProcessor = new DigestRemovalProcessor();
+        final List<NotificationEntity> processedNotificationList = removalProcessor.process(notificationList);
+        if (notificationList == null) {
+            return new ArrayList<>(0);
```
gkillough: For efficiency, we could do this before trying to process the notifications. We should also use Collections.emptyList() wherever possible.

PR #8, opened 2017-10-25
Added HipChat test endpoint and abstracted channel config rest calls.

gkillough:


src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/EmailConfigController.java
```diff
@@ -40,87 +37,57 @@
...
-                restModel.getMailSmtpTimeout(), restModel.getMailSmtpFrom(), restModel.getMailSmtpLocalhost(), restModel.getMailSmtpEhlo(), restModel.getMailSmtpAuth(), restModel.getMailSmtpDnsNotify(), restModel.getMailSmtpDsnRet(),
-                restModel.getMailSmtpAllow8bitmime(), restModel.getMailSmtpSendPartial(), restModel.getEmailTemplateDirectory(), restModel.getEmailTemplateLogoImage(), restModel.getEmailSubjectLine());
+        final EmailConfigEntity databaseModel = new EmailConfigEntity(restModel.getId(), restModel.getMailSmtpHost(), restModel.getMailSmtpUser(), restModel.getMailSmtpPassword(), restModel.getMailSmtpPort(),
+                restModel.getMailSmtpConnectionTimeout(), restModel.getMailSmtpTimeout(), restModel.getMailSmtpFrom(), restModel.getMailSmtpLocalhost(), restModel.getMailSmtpEhlo(), restModel.getMailSmtpAuth(),
+                restModel.getMailSmtpDnsNotify(), restModel.getMailSmtpDsnRet(), restModel.getMailSmtpAllow8bitmime(), restModel.getMailSmtpSendPartial(), restModel.getEmailTemplateDirectory(), restModel.getEmailTemplateLogoImage(),
+                restModel.getEmailSubjectLine());
         return databaseModel;
     }
```
bamandel: The databaseModelsToRestModels method below this can be moved into the parent class as it just loops and calls the above method.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/DatabaseEntity.java
```diff
@@ -40,6 +40,10 @@
     @Column(name = "id")
     private Long id;
 
+    public DatabaseEntity(final Long id) {
+        this.id = id;
+    }
+
```
bamandel: Should we require that anything extending the DatabaseEntity class need to have an ID? What about the cases when we have a table that has a many to many relationship?

gkillough: We should probably rename DatabaseEntity to be more specific to configuration, but not in this pull request.

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/EmailConfigController.java
gkillough: Resolved

PR #9, opened 2017-10-25
Alert ui

bamandel:
This will probably need to be merged after the changes to AlertProperties, otherwise none of the notification functionality will work.


package.json
```diff
@@ -0,0 +1,31 @@
...
+  "version": "0.0.1",
+  "description": "Test",
+  "keywords": [
+    "rest",
+    "blackduck",
+    "spring",
+    "data",
+    "react"
+  ],
+  "author": "Brian Mandel",
```
gkillough: Not sure we want to commit this.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/SettingEntity.java
```diff
@@ -0,0 +1,74 @@
...
+    @Column(name = "setting_name")
+    private String settingName;
+
+    @Column(name = "setting_value")
+    private String settingValue;
+
+    @Column(name = "setting_type")
+    private String settingType;
+
+    public SettingEntity(final String settingName, final String settingValue, final String settingType) {
```
gkillough: We might need to be able to pass a nullable id here.

src/main/java/com/blackducksoftware/integration/hub/alert/hub/HubServiceWrapper.java
```diff
@@ -42,6 +43,8 @@
 
     private final AlertProperties alertProperties;
 
+    private boolean isHubConnectionValid = false;
```
gkillough: This should only be set through the constructor.

PR #10, opened 2017-10-26
Store global config

jrichardBD:


src/main/java/com/blackducksoftware/integration/hub/alert/config/CommonConfig.java
```diff
@@ -48,15 +57,39 @@
...
         this.jobBuilderFactory = jobBuilderFactory;
         this.stepBuilderFactory = stepBuilderFactory;
         this.taskExecutor = taskExecutor;
         this.notificationRepository = notificationRepository;
         this.transactionManager = transactionManager;
+        this.taskScheduler = taskScheduler;
+    }
+
+    public void scheduleJobExecution(final String cron) {
+        if (StringUtils.isNotBlank(cron)) {
```
gkillough: Should we have more validation on the cron expression? Even if we just split on spaces and counted the fields, that would make me feel a little better about this.

jrichardBD: I agree we should be validating them, but would you be ok with that validation being in the controllers? So if the User tries to configure gibberish we can send an error back .

psantos1113: The post and put into the rest controller should validate the input values and return a correct error code.  I agree with @jrichardBD

PR #12, opened 2017-10-27
Refactoring controllers

jrichardBD:


src/main/java/com/blackducksoftware/integration/hub/alert/web/ObjectTransformer.java
```diff
@@ -0,0 +1,160 @@
...
+    public <T> List<T> transformObjects(final List<?> objects, final Class<T> newClass) throws IntegrationException {
+        final List<T> newList = new ArrayList<>();
+        for (final Object object : objects) {
+            final T newObject = transformObject(object, newClass);
+            newList.add(newObject);
+        }
+        return newList;
+    }
+
+    public <T> T transformObject(final Object object, final Class<T> newClass) throws IntegrationException {
```
jrichardBD: Any ideas on how to make this cleaner?

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/ConfigActions.java
```diff
@@ -0,0 +1,94 @@
...
+            } catch (final Exception e) {
+                throw new IntegrationException();
+            }
+        }
+        return null;
+    }
+
+    public abstract Map<String, String> validateConfig(R restModel);
+
+    public void customTriggers(final R restModel) {
```
gkillough: I understand why this needs to be empty having seen the rest of the PR, but we should definitely have a comment explaining it. I would also rename it to something more descriptive, along the lines of respondToConfigChanges(...).

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/ConfigController.java
```diff
@@ -22,92 +22,79 @@
...
-            if (response == null) {
-                final D updatedEntity = repository.save(modelEntity);
-                response = createResponse(HttpStatus.ACCEPTED, updatedEntity.getId(), "Updated.");
+    public ResponseEntity<String> putConfig(final R restModel) throws IntegrationException {
+        if (configActions.doesConfigExist(restModel.getId())) {
+            final Map<String, String> validationResults = configActions.validateConfig(restModel);
+            if (validationResults.isEmpty()) {
+                configActions.customTriggers(restModel);
+                final D updatedEntity = configActions.saveConfig(restModel);
+                return createResponse(HttpStatus.CREATED, updatedEntity.getId(), "Updated");
```
gkillough: I initially chose to use HttpStatus.CREATED here as well, but I believe you changed it to HttpStatus.ACCEPTED later, which is a more accurate response. Unless you have a reason for changing this back, I would say this is a mistake.

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/ConfigController.java
```diff
@@ -22,92 +22,79 @@
...
-        final List<R> restModels = new ArrayList<>();
-        for (final D databaseModel : databaseModels) {
-            restModels.add(databaseModelToRestModel(databaseModel));
-        }
-        return restModels;
+    protected ResponseEntity<String> createResponse(final HttpStatus status, final String id, final String message) {
+        return createResponse(status, configActions.objectTransformer.stringToLong(id), message);
     }
 
     protected ResponseEntity<String> createResponse(final HttpStatus status, final Long id, final String message) {
```
gkillough: One of these should be private; based on your changes, I believe it should be this one.

PR #13, opened 2017-10-30
Slack channel

bamandel:
There is more than just the Slack channel in this PR. When creating it I noticed a lot of duplicate code, so I tried creating parent classes. I'm still verifying everything works but also want everyones opinion on it.


src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/SlackConfigEntity.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "slack_config", schema = "configuration")
+public class SlackConfigEntity extends DatabaseEntity {
+    private static final long serialVersionUID = 4872590789715928839L;
+
+    @Column(name = "channel_name")
+    private String channelName;
```
gkillough: Is this the name that will be applied to the bot? If so, I think it should always just be Hub Alert (using the constant). Are there other basic fields that would be useful to configure other than these?

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/SlackConfigActions.java
```diff
@@ -0,0 +1,49 @@
...
+public class SlackConfigActions extends ConfigActions<SlackConfigEntity, SlackConfigRestModel> {
+
+    @Autowired
+    public SlackConfigActions(final SlackRepository slackRepository, final ObjectTransformer objectTransformer) {
+        super(SlackConfigEntity.class, SlackConfigRestModel.class, slackRepository, objectTransformer);
+    }
+
+    @Override
+    public Map<String, String> validateConfig(final SlackConfigRestModel restModel) {
+        return Collections.EMPTY_MAP;
```
gkillough: We should use Collections.emptyMap(); I know it seems trivial, but this way, backward compatibility can't be broken.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -126,36 +107,10 @@ private int sendMessage(final HipChatConfigEntity config, final String apiUrl, f
 
     @Override
     public String testMessage(final HipChatConfigEntity config) {
-        final int responseCode = sendMessage(config, HIP_CHAT_API, "Test Message", HIP_CHAT_FROM_NAME + " Tester");
+        final int responseCode = sendMessage(config, HIP_CHAT_API, "Test Message", AlertConstants.ALERT_APPLICATION_NAME + " Tester");
         return String.valueOf(responseCode);
     }
 
-    private String createHtmlMessage(final ProjectData projectData) {
```
jrichardBD: I think we should leave the message formatting up to each channel for now rather than trying to abstract them for now

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
```diff
@@ -22,19 +22,82 @@
...
+    public DistributionChannel(final Gson gson, final JpaRepository<C, Long> repository, final Class<E> clazz) {
         super(gson, clazz);
+        this.repository = repository;
     }
 
     public abstract void sendMessage(final E event, final C config);
 
     public abstract String testMessage(final C config);
 
+    protected String createHtmlMessage(final ProjectData projectData) {
```
gkillough: I'm still not a huge fan of abstracting this.

jrichardBD: I could be wrong but it looks like slack recommends json as the message body 
https://api.slack.com/web

src/main/java/com/blackducksoftware/integration/hub/alert/channel/slack/SlackChannel.java
```diff
@@ -0,0 +1,118 @@
...
+        sendMessage(htmlMessage, config);
+    }
+
+    @Override
+    public String testMessage(final SlackConfigEntity config) {
+        final String message = "Test from Alert application";
+        return String.valueOf(sendMessage(message, config));
+    }
+
+    private int sendMessage(final String htmlMessage, final SlackConfigEntity config) {
```
gkillough: I'd like to see an example of what a notification generated by the Hub looks like.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
psantos1113: I agree.  It isn't the responsibility of this class to define the implementation of the message only the API that channels must provide.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
```diff
@@ -22,19 +22,82 @@
...
+        }
+        return htmlBuilder.toString();
+    }
+
+    protected String createPlaintextMessage(final ProjectData projectData) {
+        final String htmlMessage = createHtmlMessage(projectData);
+        return StringEscapeUtils.escapeHtml4(htmlMessage);
+    }
+
+    public void handleEvent(final E event) {
```
psantos1113: This does not belong here.  The work being done with filtering will address the routing of the messages.  The handleEvent message should remain abstract.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
```diff
@@ -22,19 +22,82 @@
...
+        final List<C> configurations = repository.findAll();
+        for (final C configEntity : configurations) {
+            sendMessage(event, configEntity);
+        }
+    }
+
+    @Override
+    public void receiveMessage(final String message) {
```
psantos1113: This method may make sense here given that handleEvent will be abstract and the individual channel can determine what to do with the event.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
psantos1113: I agree

src/main/java/com/blackducksoftware/integration/hub/alert/channel/slack/SlackChannel.java
```diff
@@ -0,0 +1,118 @@
...
+        final String slackUrl = config.getWebhook();
+        final RestConnection connection = ChannelRestConnectionFactory.createUnauthenticatedRestConnection(slackUrl);
+        if (connection != null) {
+            final String jsonString = getJsonString(htmlMessage, config.getChannelName(), config.getUsername());
```
psantos1113: Do you need to include an API token to post to the slack channel.  Is there any authorization?

src/main/java/com/blackducksoftware/integration/hub/alert/web/model/ConfigRestModel.java
```diff
@@ -43,6 +43,10 @@ public String getId() {
         return id;
     }
 
+    public void setId(final String id) {
```
psantos1113: Why do you need the setId method?  Do you call it elsewhere in your code?  The id field is auto generated when the entity is created by the database.

jrichardBD: You need the setId for the Put and Post methods otherwise you'll be creating a new config in the database every time

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/SlackConfigEntity.java
bamandel: channel_name would be the name of the chat room in slack you want to send the notification to. Slack uses the name channel for this, like we do.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
bamandel: I take this message and insert it into json in my slack_channel. But I will now go and remove the HTML elements and make my own specific channel message.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/SlackConfigEntity.java
bamandel: Talked to Paulo and we believe we should leave the bot name up to the user but default to hub_alert when necessary

PR #17, opened 2017-11-02
Add unit tests

psantos1113:


src/test/resources/com/blackducksoftware/integration/hub/datasource/repository/notificationData.xml
```diff
@@ -0,0 +1,140 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gkillough: Do we still need this file? I'm fine with checking it in in case we ever figure out how DbUnit is supposed to work.

PR #22, opened 2017-11-07
Alert ready for styling

bamandel:
Added the necessary changes to webpack for styling as well as adding tabs and css modules.


src/main/js/App.js
```diff
@@ -6,20 +6,43 @@ import GlobalConfiguration from './component/GlobalConfiguration';
...
+    constructor() {
+        super();
+        this.state = {index: 0};
+    }
+
+    render() {
+        return (
+            <div>
+                <div className={styles.header}>
+                    <h1>Blackduck Software Alert</h1>
```
gkillough: This should be Black Duck Alert

bamandel: We don't want something like Blackduck Alert? I didn't think we wanted to separate the name Blackduck.

PR #23, opened 2017-11-09
Alert styling

bamandel:
Prettier UI (Still not complete but usable)


src/main/css/main.css
```diff
@@ -1,8 +1,24 @@
-h1, h2, h3 {
-    text-align: center;
+html {
```
gkillough: For this and the one below it, I'm not sure we want to modify the _html_ tag's default styling. I would prefer to create wrappers within it. Using wrappers will prevent us from edge cases where the browser corrects margin and width based on its own settings.

PR #24, opened 2017-11-10
Notification filters

gkillough:
There are a lot of changes, but many of them are refactoring or adding DB tables. We will discuss the major changes in the code review.


src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestNotificationProcessor.java
```diff
@@ -51,74 +58,115 @@
...
-                categoryData = categoryBuilderMap.get(categoryKey);
-            }
-            int count = 1;
-            final Map<String, Object> dataSet = new HashMap<>();
-            if (categoryKey == NotificationCategoryEnum.HIGH_VULNERABILITY || categoryKey == NotificationCategoryEnum.MEDIUM_VULNERABILITY || categoryKey == NotificationCategoryEnum.LOW_VULNERABILITY) {
-                count = entry.getVulnerabilityList().size();
-                dataSet.put(VulnerabilityCache.VULNERABILITY_ID_SET, getVulnerabilityIdSet(entry));
-            }
+    private Collection<UserNotificationWrapper> createCateoryDataMap(final DigestTypeEnum digestType, final Collection<NotificationEntity> notificationList) {
+        final Map<String, Map<String, ProjectDataBuilder>> userProjectMap = createUserProjectMap(digestType, notificationList);
```
bamandel: Instead of having maps in maps would it be better to create a new class and add functionality there instead?

src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestNotificationProcessor.java
```diff
@@ -51,74 +58,115 @@
...
+            } catch (final NoResultException ex) {
+                logger.debug("user {} could not be found in the configuration", username);
+                logger.debug("Cause:", ex);
             }
+        });
+        return dataList;
+    }
+
+    private Map<String, Map<String, ProjectDataBuilder>> createUserProjectMap(final DigestTypeEnum digestType, final Collection<NotificationEntity> notificationList) {
+        final Map<String, Map<String, ProjectDataBuilder>> userProjectMap = new LinkedHashMap<>();
```
bamandel: another map in map where we could probably cut out some code through an extra class

PR #26, opened 2017-11-14
Add Basic Controller Functionality

gkillough:


src/main/java/com/blackducksoftware/integration/hub/alert/channel/email/EmailChannel.java
```diff
@@ -93,8 +104,16 @@ public String testMessage(final EmailConfigEntity emailConfigEntity) {
 
     @Override
     public void sendMessage(final EmailEvent emailEvent, final EmailConfigEntity emailConfigEntity) {
-        // TODO get the emails from the config
-        sendMessage("", emailEvent, emailConfigEntity);
+        final Long userId = emailEvent.getUserConfigId();
+        final HubUsersEntity entity = hubUsersRepository.findOne(userId);
+        final String username = entity.getUsername();
```
jrichardBD: We should check that the entity is not null. There is a chance that the configuration has been updated since the event was created.

PR #27, opened 2017-11-14
Changed the DB from in memory to embedded

bamandel:
This will allow an embedded DB as opposed to in-memory. At the moment the db file is stored in the Users directory and not in the project directory although this can probably be changed if desired.


src/main/resources/application.properties
```diff
@@ -17,7 +17,15 @@ alert.accumulator.cron=0 0/1 * 1/1 * *
 alert.digest.daily.cron=0 0 0 1/1 * ?
 alert.digest.realtime.cron=0 0/1 * 1/1 * *
 
-spring.datasource.url=jdbc:h2:mem:testdb
+spring.datasource.url=jdbc:h2:file:~/test
```
psantos1113: Let's change this to alertdb and see if we can change the path to the directory running the application. 

Also how did you test this?  The DataSourceConfig was needed because Spring batch had some conflicts with the JPA repository, specifically the transaction manager it was using.

src/main/resources/application.properties
```diff
@@ -17,7 +17,15 @@ alert.accumulator.cron=0 0/1 * 1/1 * *
 alert.digest.daily.cron=0 0 0 1/1 * ?
 alert.digest.realtime.cron=0 0/1 * 1/1 * *
```
gkillough: If we are going to use this file, we should remove any fields that need to be specified by the user through the UI.

bamandel: That's a good point. We won't need half this file anymore because of the UI. I'll delete

src/main/resources/application.properties
bamandel: I removed the datasource file as it mostly was configuring springs default in memory testdb. The thing that was creating this was the DataSource bean specifically. I can add the TransactionManager back but I would like to test your exact scenario first to see if we still have this conflict, or if just the in memory DB was the root cause of the issue.

bamandel:
Honestly if we do something like change db tables or need to remove data we can just remove the file from the data folder and restart the app. It'll regenerate the DB with the new values. I have not setup anything for rolling back tho

PR #28, opened 2017-11-17
Hub User Manager

gkillough:
Adding user manager functionality. This will later be used in conjunction with a HubUserPersistenceManager (or something along those lines) to update the database tables by checking for hub configuration changes on a schedule.


src/main/java/com/blackducksoftware/integration/hub/alert/datasource/relation/HubUserProjectVersionsRelationPK.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.Serializable;
+
+public class HubUserProjectVersionsRelationPK implements Serializable {
+    private static final long serialVersionUID = 8206061503103229279L;
+
+    public Long userConfigId;
+    public String projectName;
+    public String projectVersionName;
+
```
gkillough: This class functions as a primary key object for database operations. For serialization purposes, only a default constructor is recognized.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/relation/manager/HubUserManager.java
```diff
@@ -0,0 +1,240 @@
...
+                    channelRelation = clazz.newInstance();
+                } catch (InstantiationException | IllegalAccessException e) {
+                    logger.error("Unable to save Hub user channel config (hubUserId: {}, channelConfigId: {}) because {} could not be instantiated: {}.", id, channelId, clazz.getSimpleName(), e.getLocalizedMessage());
+                    return;
+                }
+                channelRelation.setUserConfigId(id);
+                channelRelation.setChannelConfigId(objectTransformer.stringToLong(channelId));
+                repository.save(channelRelation);
+            } else {
+                repository.delete(id);
```
gkillough: We want to delete the mapping from user to channel id if the channel id is null. A mapping should only be in the db if it is configured.

src/test/java/com/blackducksoftware/integration/hub/alert/digest/DigestNotificationProcessorIT.java
```diff
@@ -138,7 +137,6 @@ public void processNotificationDataWithSameEventKeyTestIT() {
         assertEquals(1, eventsCreated.size());
         final AbstractChannelEvent event = eventsCreated.get(0);
         assertTrue(event instanceof HipChatEvent);
-        assertEquals(hipChatUserRel.getUserConfidId(), event.getUserConfigId());
```
gkillough: We cannot predict the id in the tests. With auto-increment and no guaranteed test order, it is very likely to change.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/relation/manager/HubUserManager.java
```diff
@@ -0,0 +1,240 @@
...
+        return configId;
+    }
+
+    public void deleteConfig(final String id) {
+        deleteConfig(objectTransformer.stringToLong(id));
+    }
+
+    public void deleteConfig(final Long id) {
+        if (doesConfigExist(id)) {
+            hubUsersRepository.delete(id);
```
psantos1113: Perform the deletion in the opposite order of the save.  So the deletion of the user object should be the last thing that is deleted.

PR #29, opened 2017-11-21
Alert styling

bamandel:
There are still some small changes I want to make to styling, but here are some of the more recent changes i've made to our UI. I've moved a lot of repeated code into a parent class.

### DIFF
--- a/src/main/js/component/ConfigButtons.js
+++ b/src/main/js/component/ConfigButtons.js
@@ -18,7 +18,7 @@ export default class ConfigButtons extends React.Component {
             <div className={submitContainers}>
                 <input className={submitButtons} type={buttonType} value={buttonText} onClick={this.props.onClick}></input>
                 { includeTest == "true" &&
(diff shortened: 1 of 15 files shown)


src/main/js/component/ConfigButtons.js
```diff
@@ -0,0 +1,26 @@
...
+
+import { submitContainers, submitButtons } from '../../css/main.css';
+
+export default class ConfigButtons extends React.Component {
+    constructor(props) {
+        super(props);
+    }
+
+    render() {
+        const includeTest = this.props.includeTest || "false";
```
psantos1113: In the near future we need to spend time getting Redux working with this application.  There is a mechanism to set the default property values.  We don't want to set them in the render method.  I would prefer const {includeTest, buttonText, buttonType, buttonTestText, buttonTestType} = this.props;

bamandel:
Merged into James' Login branch

PR #30, opened 2017-11-21
Login

jrichardBD:


package.json
```diff
@@ -14,7 +14,9 @@
   "dependencies": {
     "react": "^15.3.2",
     "react-dom": "^15.3.2",
-    "react-router-dom": "^4.2.0",
```
jrichardBD: We may not need these router dependencies, since we dont actually use Router's

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/ConfigActions.java
```diff
@@ -63,26 +64,68 @@ public boolean doesConfigExist(final Long id) {
...
             }
             return Collections.emptyList();
         }
         final List<D> databaseEntities = repository.findAll();
         final List<R> restModels = objectTransformer.databaseEntitiesToConfigRestModels(databaseEntities, configRestModelClass);
-        return restModels;
+        return maskRestModels(restModels);
+    }
+
+    public R maskRestModel(final R restModel) {
```
jrichardBD: This gets overridden if needed by the actions to hide fields. We could make this abstract and force all actions to implement, ideas?

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/ConfigActions.java
```diff
@@ -63,26 +64,68 @@ public boolean doesConfigExist(final Long id) {
...
+    public D updateNewConfigWithSavedConfig(final D newConfig, final String id) {
+        if (StringUtils.isNotBlank(id)) {
+            final Long longId = objectTransformer.stringToLong(id);
+            final D savedConfig = repository.findOne(longId);
+            return updateNewConfigWithSavedConfig(newConfig, savedConfig);
+        }
+        return newConfig;
+    }
+
+    public D updateNewConfigWithSavedConfig(final D newConfig, final D savedConfig) {
```
jrichardBD: This gets overridden if needed by the actions to use saved fields if the newConfig has empty "hidden" fields like passwords. We could make this abstract and force all actions to implement, ideas?

src/main/java/com/blackducksoftware/integration/hub/alert/web/AuthenticationHandler.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+@EnableWebSecurity
+@Configuration
+public class AuthenticationHandler extends WebSecurityConfigurerAdapter {
+
+    @Override
+    protected void configure(final HttpSecurity http) throws Exception {
+        http.csrf().disable().authorizeRequests().anyRequest().permitAll().and().logout().logoutSuccessUrl("/");
+        // http.authorizeRequests().antMatchers("/configuration/**", "/api/**", "/verify").authenticated();
```
gkillough: Double check this line.

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/ConfigActions.java
gkillough: I think making it abstract works, but we should also consider how we are using inheritance. We have been abstracting a lot of functionality that is used in multiple places, but we should really decide whether the functionality makes sense as part of a class hierarchy, or as a separate component. Just food for thought.

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/ConfigActions.java
gkillough: After looking at this class with more context, I think our best bet is to abstract and force the implementation. In general, I don't like implementing something that will always be overridden. The same seems to apply to for the comment above.

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/LoginController.java
```diff
@@ -0,0 +1,147 @@
...
+            serverConfigBuilder.setAlwaysTrustServerCertificate(Boolean.valueOf(loginRestModel.getHubAlwaysTrustCertificate()));
+            serverConfigBuilder.setProxyHost(loginRestModel.getHubProxyHost());
+            serverConfigBuilder.setProxyPort(loginRestModel.getHubProxyPort());
+            serverConfigBuilder.setProxyUsername(loginRestModel.getHubProxyUsername());
+            serverConfigBuilder.setProxyPassword(loginRestModel.getHubProxyPassword());
+            validateHubConfiguration(serverConfigBuilder);
+            final RestConnection restConnection = createRestConnection(serverConfigBuilder);
+            restConnection.connect();
+            System.out.println("Connected");
+            // TODO check User's role
```
gkillough: Flagging TODO for your consideration before merging.

bamandel:
I have a couple things
- When logging in, it requires I specify a timeout which is hidden behind advanced config. I would pull that into the normal login section or add a default timeout value.
- After logging into http://localhost:8080/h2/ and entering a specific DB, you get security errors and a white screen displays. This will cause issues if we ever need to view the content of the DB.

jrichardBD:
1. The timeout should have a default value. The work to set the default should be done when we add the redux stuff
2. What is http://localhost:8080/h2/ ? Did you add that? Why do we have a UI for the database? Does it still give you security errors after logging in through the main login?

PR #32, opened 2017-11-29
Notification data purge

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/GlobalConfigEntity.java
```diff
@@ -67,11 +67,14 @@
     @Column(name = "alert_digest_daily_cron")
     private String dailyDigestCron;
 
+    @Column(name = "alert_purge_data_cron")
```
gkillough: Do we have a SQL script to initialize the global config table, and has it been updated?

src/main/java/com/blackducksoftware/integration/hub/alert/web/AuthenticationHandler.java
```diff
@@ -34,6 +34,8 @@
     @Override
     protected void configure(final HttpSecurity http) throws Exception {
         http.csrf().disable().authorizeRequests().anyRequest().permitAll().and().logout().logoutSuccessUrl("/");
+        // TODO: Make sure the H2 console is disabled by removing the following line of code.
```
gkillough: Is this being committed for debugging purposes?

psantos1113: This is for debugging purposes as we develop our initial version.  Before release this will need to be disabled.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/GlobalConfigEntity.java
psantos1113: We don't have an SQL script yet to initialize the tables.  I'm starting to look into that.

PR #35, opened 2017-12-04
Login action

bamandel:
Verifies that the user has the administrator role before they are allowed to log in.

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/LoginController.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/LoginController.java
@@ -22,46 +22,35 @@
  */
 package com.blackducksoftware.integration.hub.alert.web.controller;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Set;
-
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 
-import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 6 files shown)


src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/LoginController.java
```diff
@@ -22,46 +22,35 @@
...
 import com.blackducksoftware.integration.log.IntLogger;
 import com.blackducksoftware.integration.log.LogLevel;
 import com.blackducksoftware.integration.log.PrintStreamIntLogger;
-import com.blackducksoftware.integration.validator.AbstractValidator;
-import com.blackducksoftware.integration.validator.FieldEnum;
-import com.blackducksoftware.integration.validator.ValidationResult;
-import com.blackducksoftware.integration.validator.ValidationResults;
 
 @RestController
 public class LoginController {
```
gkillough: I think we should stick to only calling action methods here rather than implementing all of the functionality. I would like to see it abstracted up to the LoginActions class.

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/handler/LoginDataHandler.java
```diff
@@ -0,0 +1,83 @@
...
+import com.blackducksoftware.integration.hub.alert.web.ObjectTransformer;
+import com.blackducksoftware.integration.hub.alert.web.actions.LoginActions;
+import com.blackducksoftware.integration.hub.alert.web.model.LoginRestModel;
+import com.blackducksoftware.integration.hub.alert.web.model.ResponseBodyBuilder;
+import com.blackducksoftware.integration.hub.rest.exception.IntegrationRestException;
+import com.blackducksoftware.integration.log.IntLogger;
+import com.blackducksoftware.integration.log.LogLevel;
+import com.blackducksoftware.integration.log.PrintStreamIntLogger;
+
+public class LoginDataHandler extends ControllerHandler {
```
jamesrichard91: Could we rename this to LoginHandler?

PR #37, opened 2017-12-05
Slack restructure

bamandel:
Added the new slack functionality in addition to a bunch of small changes to organization and methods.


src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
```diff
@@ -52,17 +53,20 @@ public CommonDistributionRepository getCommonDistributionRepository() {
...
         }
-        logger.error("Global Config did not have the expected number of rows: expected 1, but found {}.", globalConfigs.size());
-        return null;
+
+        return globalConfigEntity;
     }
 
     public abstract void sendMessage(final E event, final C config);
 
-    public abstract String testMessage(final G globalConfig) throws IntegrationException;
```
gkillough: The reason we want to pass an entity here is because a test happens before anything gets saved. I'm thinking we should revisit this.

gkillough: Maybe this should take the distribution config. The test concept changes if we have a global config and a distribution config.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/email/EmailChannel.java
```diff
@@ -97,14 +97,11 @@ public void handleEvent(final EmailEvent emailEvent) {
...
-        if (globalConfig != null) {
-            final ProjectData data = new ProjectData(DigestTypeEnum.REAL_TIME, "Test Project", "Test Version", Collections.emptyMap());
-            final String emailAddress = globalConfig.getMailSmtpFrom();
-            sendMessage(Arrays.asList(emailAddress), new EmailEvent(data, null), globalConfig);
-            return "Attempted to send message with the given configuration.";
-        }
-        return null;
+    public String testMessage() {
+        final ProjectData data = new ProjectData(DigestTypeEnum.REAL_TIME, "Test Project", "Test Version", Collections.emptyMap());
+        final String emailAddress = getGlobalConfigEntity().getMailSmtpFrom();
```
gkillough: Again, the more I look at this, the more I think the testMessage will only relate to the distribution config. I don't know if it makes sense to send a test email to the same address it was sent from.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/global/GlobalSlackConfigEntity.java
```diff
@@ -36,36 +36,20 @@
 public class GlobalSlackConfigEntity extends DatabaseEntity {
     private static final long serialVersionUID = 4872590789715928839L;
 
-    @Column(name = "channel_name")
-    private String channelName;
-
-    @Column(name = "username")
-    private String username;
```
gkillough: I'm inclined to keep the username in the global config so that it is associated with the webhook.

E.g. "Hub Alert Bot" would be the username of the webhook being used.

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/SlackDistributionConfigController.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import com.blackducksoftware.integration.hub.alert.datasource.entity.distribution.SlackDistributionConfigEntity;
+import com.blackducksoftware.integration.hub.alert.web.actions.distribution.SlackDistributionConfigActions;
+import com.blackducksoftware.integration.hub.alert.web.model.distribution.SlackDistributionRestModel;
+
+@RestController
+public class SlackDistributionConfigController extends ConfigController<SlackDistributionRestModel> {
```
gkillough: Let's add the @PreAuthorize("hasRole('ADMIN')") annotation to the methods here.

src/main/java/com/blackducksoftware/integration/hub/alert/web/model/distribution/SlackDistributionRestModel.java
```diff
@@ -0,0 +1,71 @@
...
+    public void setChannelName(final String channelName) {
+        this.channelName = channelName;
+    }
+
+    public void setChannelUsername(final String channelUsername) {
+        this.channelUsername = channelUsername;
+    }
+
+    @Override
+    public String toString() {
```
gkillough: The superclass has this. We can remove it.

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/distribution/DistributionConfigActions.java
```diff
@@ -128,11 +133,15 @@ public String validateConfig(final R restModel) throws AlertFieldException {
         final List<D> allEntities = channelDistributionRepository.findAll();
         final List<R> constructedRestModels = new ArrayList<>();
         for (final D entity : allEntities) {
-            constructedRestModels.add(constructRestModel(entity));
+            try {
+                constructedRestModels.add(constructRestModel(entity));
+            } catch (final AlertException e) {
```
gkillough: This is a better approach for sure.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/email/EmailChannel.java
jamesrichard91: The only email address we have at this point is the one configured as the From address, so we must send the test email to that address OR we need to figure out if we want to add an additional field for the Test only OR should we use the email address of the Hub user configured.

src/main/java/com/blackducksoftware/integration/hub/alert/datasource/entity/global/GlobalSlackConfigEntity.java
jamesrichard91: +1

src/main/java/com/blackducksoftware/integration/hub/alert/web/model/distribution/SlackDistributionRestModel.java
jamesrichard91: +1

src/main/java/com/blackducksoftware/integration/hub/alert/channel/email/EmailChannel.java
gkillough: I like the last option because that will be consistent with testing the other channels.

bamandel:
I've updated the models so that they have the proper fields and removed redundant code. At the moment, I've left the testMessages the way we are so we can figure out exactly what we want to do with them.

PR #38, opened 2017-12-06
Freemarker & HipChat

gkillough:


src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -136,29 +135,24 @@ public String testMessage(final HipChatConfigEntity config) throws IntegrationEx
...
-                    htmlBuilder.append("<br />Type: " + data.getCategoryKey());
-                    htmlBuilder.append("<br />Number of Changes: " + data.getItemCount());
-                    for (final ItemData item : data.getItemList()) {
-                        final Map<String, Object> dataSet = item.getDataSet();
-                        htmlBuilder.append("<p>  Rule: " + dataSet.get(ItemTypeEnum.RULE.toString()));
-                        htmlBuilder.append(" | Component: " + dataSet.get(ItemTypeEnum.COMPONENT.toString()));
-                        htmlBuilder.append(" [" + dataSet.get(ItemTypeEnum.VERSION.toString()) + "]</p>");
-                    }
-                }
+        String html = "SOMETHING WENT WRONG";
```
gkillough: Would we rather throw a RuntimeException or send an error message to the channel? Obviously this message could be more in depth.

jamesrichard91: I think freemarkerTemplatingService. getResolvedTemplate() should throw an exception if something went wrong, otherwise we would need to have the same error handling here and in EmailMessagingService

PR #44, opened 2017-12-12
Configured Projects in RestModels

gkillough:
I will be adding tests before this is merged, but the logic should be fairly clear at this time.

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/distribution/DistributionConfigActions.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/distribution/DistributionConfigActions.java
@@ -89,6 +89,7 @@ public abstract class DistributionConfigActions<D extends DatabaseEntity, R exte
                     commonEntity.setDistributionConfigId(createdEntity.getId());
                     commonEntity = commonDistributionRepository.save(commonEntity);
                     saveConfiguredProjects(commonEntity, restModel);
+                    cleanUpStaleChannelConfigurations();
                     return createdEntity;
                 }
             } catch (final Exception e) {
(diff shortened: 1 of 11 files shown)


src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/distribution/DistributionConfigActions.java
```diff
@@ -129,6 +140,64 @@ public String validateConfig(final R restModel) throws AlertFieldException {
         return Collections.emptyList();
     }
 
+    protected void saveConfiguredProjects(final CommonDistributionConfigEntity commonEntity, final R restModel) {
+        if (Boolean.TRUE.equals(commonEntity.getFilterByProject())) {
+            final List<String> configuredProjectsFromUI = restModel.getConfiguredProjects();
+            if (configuredProjectsFromUI != null) {
+                // Remove the old DistributionProjectRelations for the distribution configuration:
```
bamandel: This comment and the one below it are describing something that can be explained through a method name. I'd recommend adding a remove and add project helper methods.

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/distribution/DistributionConfigActions.java
```diff
@@ -129,6 +140,64 @@ public String validateConfig(final R restModel) throws AlertFieldException {
...
+                final List<DistributionProjectRelation> distributionProjects = distributionProjectRepository.findByCommonDistributionConfigId(commonEntity.getId());
+                distributionProjectRepository.delete(distributionProjects);
+                // Add new DistributionProjectRelations (and ConfiguredProjectEntities if needed):
```
bamandel: This is the other section that can turn into another method

src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/distribution/DistributionConfigActions.java
```diff
@@ -129,6 +140,64 @@ public String validateConfig(final R restModel) throws AlertFieldException {
...
+        final List<String> configuredProjects = new ArrayList<>();
+        for (final DistributionProjectRelation relation : distributionProjects) {
+            final ConfiguredProjectEntity entity = configuredProjectsRepository.findOne(relation.getProjectId());
+            configuredProjects.add(entity.getProjectName());
+        }
+        return configuredProjects;
+    }
+
+    protected void cleanUpConfiguredProjects() {
+        final List<ConfiguredProjectEntity> configuredProjects = configuredProjectsRepository.findAll();
```
bamandel: I feel there is a query we can use that will get the projects needed automatically and maybe even delete them for us, all in one go.

bamandel: https://stackoverflow.com/questions/23723025/spring-data-delete-by-is-supported

PR #47, opened 2017-12-15
Distribution job config

jamesrichard91:


src/main/js/component/distribution/DistributionConfiguration.js
```diff
@@ -154,18 +141,114 @@ class DistributionConfiguration extends Component {
 				});
 			}
 		});
+        this.fetchDistributionJobs();
+    }
+
+    updateJobsTable() {
+        //TODO remove this and references to it
```
gkillough: Is this something you want to do before merging?

src/main/js/component/distribution/DistributionConfiguration.js
```diff
@@ -176,10 +259,49 @@ class DistributionConfiguration extends Component {
 
 	customJobConfigDeletionConfirm(next, dropRowKeys) {
 	  if (confirm("Are you sure you want to delete these Job configurations?")) {
+	  	console.log('Deleting the Job configs');
 	  	//TODO delete the Job configs from the backend
```
gkillough: This seems done?

PR #51, opened 2017-12-20
Test refactor

bamandel:


src/test/java/com/blackducksoftware/integration/hub/alert/web/actions/global/GlobalHubConfigActionsTest.java
```diff
@@ -224,35 +148,38 @@ public RestConnection answer(final InvocationOnMock invocation) throws Throwable
...
 
         Mockito.doAnswer(new Answer<GlobalHubConfigEntity>() {
             @Override
             public GlobalHubConfigEntity answer(final InvocationOnMock invocation) throws Throwable {
-                return mockUtils.createGlobalHubConfigEntity();
+                return mockUtils.createGlobalEntity();
             }
         }).when(mockedGlobalRepository).findOne(Mockito.anyLong());
 
         configActions.testConfig(partialRestModel);
```
bamandel: This test is broken because of our issues with sensitive fields

src/test/java/com/blackducksoftware/integration/hub/alert/TestProperties.java
```diff
@@ -0,0 +1,75 @@
...
+    private Properties properties;
+
+    public TestProperties() {
+        resourceLoader = new ResourceLoader();
+    }
+
+    public Properties getProperties() {
+        if (properties == null) {
+            try {
+                // TODO add additional locations to check for properties
```
gkillough: We should parameterize this method so that the location can be passed in, create a default version that uses that constant.

PR #52, opened 2017-12-21
Docker file support

psantos1113:


psantos1113:
NOTE that this will require that you specify the server.port value in you SPRING_APPLICATION_JSON environment variable or define the ALERT_SERVER_PORT environment variable to specify the port the server runs as.  The default value should be 8080

Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM openjdk:8-jre-alpine
+
+ARG VERSION
+
+LABEL com.blackducksoftware.integration.alert.vendor="Black Duck Software, Inc." \
```
gkillough: Should this be changed?

psantos1113: I will double check with Ari.  Our stuff on docker hub all follows this convention.

PR #58, opened 2017-12-29
Test refactor 2

bamandel:
Cleaned up tests and changed from a single large mock item to 4 individual ones. Separate files for Entity, GlobalEntity, RestModel, and GlobalRestModel.


src/test/java/com/blackducksoftware/integration/hub/alert/mock/entity/MockHipChatEntity.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private MockHipChatEntity(final Integer roomId, final Boolean notify, final String color, final Long id) {
+        this.roomId = roomId;
+        this.notify = notify;
+        this.color = color;
+        this.id = id;
+    }
+
+    @Test
+    public void tes() throws JSONException {
```
jamesrichard91: Can you give this test a real name?

src/test/java/com/blackducksoftware/integration/hub/alert/mock/entity/global/MockGlobalHubEntity.java
```diff
@@ -0,0 +1,91 @@
...
+    private MockGlobalHubEntity(final Integer hubTimeout, final String hubUsername, final String hubPassword, final Long id) {
+        super();
+        this.hubTimeout = hubTimeout;
+        this.hubUsername = hubUsername;
+        this.hubPassword = hubPassword;
+        this.id = id;
+    }
+
+    @Test
+    public void test() throws JSONException {
```
jamesrichard91: Can you give this test a real name?

src/test/java/com/blackducksoftware/integration/hub/alert/mock/entity/MockHipChatEntity.java
bamandel: These tests will be gone once I change from interface to an abstract class.

bamandel:
The main things that I've added are tests for processors and just the global controllers at the moment. Everything else is really a refactor of the old testing code I wrote.

src/main/java/com/blackducksoftware/integration/hub/alert/web/model/distribution/CommonDistributionConfigRestModel.java
```diff
@@ -24,10 +24,9 @@
 
 import java.util.List;
 
-import org.apache.commons.lang3.builder.RecursiveToStringStyle;
-import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
-
 import com.blackducksoftware.integration.hub.alert.web.model.ConfigRestModel;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
gkillough: Is there a son equivalent to this?

src/test/java/com/blackducksoftware/integration/hub/alert/processor/PolicyOverrideProcessorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import com.blackducksoftware.integration.hub.model.view.ComponentVersionView;
+import com.blackducksoftware.integration.hub.model.view.PolicyRuleView;
+import com.blackducksoftware.integration.hub.notification.processor.MapProcessorCache;
+import com.blackducksoftware.integration.test.TestLogger;
+
+public class PolicyOverrideProcessorTest {
+
+    @Test
+    public void testProcessRemoveEvent() throws HubIntegrationException, URISyntaxException {
+        // Test will break if ListProcessorCache is used because NotificationEvents does not have an equals method defined
```
gkillough: Why not add an equals method?

src/main/java/com/blackducksoftware/integration/hub/alert/web/model/distribution/CommonDistributionConfigRestModel.java
bamandel: I wasn't able to find a good replacement. If I were to actually use Gson on the object in its toString method, I would also need to change the expected json output for all our tests. If there are empty or null values in an object that has Gson ran on it, then the values aren't added to the new json string where our tests expect value: null

src/test/java/com/blackducksoftware/integration/hub/alert/processor/PolicyOverrideProcessorTest.java
bamandel: NotificationEvent is an object from another library. So we'd have to go through an entire release just to fix a test. We use the MapCache in alert so we should be using that in the test anyway, but maybe in the future we may want to add an equals method

gkillough: So then why is the comment necessary if we use a MapCache anyway?

bamandel: I wanted to worn people about a possible issue if it were changed. I think we should keep the comment, but maybe we can move it elsewhere

src/main/java/com/blackducksoftware/integration/hub/alert/web/model/distribution/CommonDistributionConfigRestModel.java
psantos1113: I don't want to introduce a different JSON serialization technology just to support tests.  I want to be consistent I would say go back to GSON. The tests for empty fields is handled by the entity tests where the fields are set to null.  The method: 
getEmptyEntityJson method in the test code should be in a base class and return the json string of "{}" which is an empty JSON object.

src/test/java/com/blackducksoftware/integration/hub/alert/mock/entity/MockHipChatEntity.java
```diff
@@ -0,0 +1,83 @@
...
+        final JsonObject json = new JsonObject();
+        json.addProperty("roomId", roomId);
+        json.addProperty("notify", notify);
+        json.addProperty("color", color);
+        json.addProperty("id", id);
+        return json.toString();
+    }
+
+    @Override
+    public String getEmptyEntityJson() {
```
psantos1113: These methods should be removed in classes like this.  a base class can return "{}" and empty json array for testing.

bamandel:
Added a couple more tests and resolved some issues with toString methods. **I made sensitive fields transient**. I did this because it worked best with Gson and also allows to possibly remove the need for sensitive field functionality. I verified that the slack integration still worked after this change (I would test hipchat which has transient fields but it's broken at the moment)

PR #60, opened 2018-01-04
Api Key Support

gkillough:


src/main/java/com/blackducksoftware/integration/hub/alert/web/actions/global/GlobalHubConfigActions.java
```diff
@@ -130,6 +130,9 @@ public String validateConfig(final GlobalHubConfigRestModel restModel) throws Al
         if (StringUtils.isNotBlank(restModel.getHubTimeout()) && !StringUtils.isNumeric(restModel.getHubTimeout())) {
             fieldErrors.put("hubTimeout", "Not an Integer.");
         }
+        if (StringUtils.isBlank(restModel.getHubApiKey())) {
```
gkillough: I don't think this is necessary.

PR #61, opened 2018-01-05
Jr audit changes

jamesrichard91:


src/main/java/com/blackducksoftware/integration/hub/alert/accumulator/RealTimeListener.java
```diff
@@ -51,8 +55,12 @@ public RealTimeListener(final Gson gson, final ChannelTemplateManager channelTem
...
     @Override
     public void receiveMessage(final String message) {
-        final RealTimeEvent event = getEvent(message);
-        final List<AbstractChannelEvent> events = notificationProcessor.processNotifications(DigestTypeEnum.REAL_TIME, event.getNotificationList());
-        channelTemplateManager.sendEvents(events);
+        try {
+            final RealTimeEvent event = getEvent(message);
+            final List<AbstractChannelEvent> events = notificationProcessor.processNotifications(DigestTypeEnum.REAL_TIME, event.getNotificationList());
+            channelTemplateManager.sendEvents(events);
+        } catch (final Exception e) {
```
gkillough: When does an exception get thrown here?

gkillough: Does it have to do with templating?

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
```diff
@@ -71,36 +71,45 @@ public CommonDistributionRepository getCommonDistributionRepository() {
...
-                        break;
+            try {
+                final AuditEntryEntity auditEntryEntity = getAuditEntryRepository().findOne(auditEntryId);
+                if (auditEntryEntity != null) {
+                    auditEntryEntity.setStatus(StatusEnum.FAILURE);
+                    auditEntryEntity.setErrorMessage(errorMessage);
+                    final String[] rootCause = ExceptionUtils.getRootCauseStackTrace(t);
+                    String exceptionStackTrace = "";
+                    for (final String line : rootCause) {
+                        if (exceptionStackTrace.length() + line.length() < 9999) {
```
gkillough: Is there a database constant for VARCHAR that we could use here?

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/handler/AuditEntryHandler.java
```diff
@@ -54,14 +56,14 @@ public AuditEntryRestModel get(final Long id) {
...
-            auditEntryEntity = auditEntryActions.resendNotification(id);
-            if (auditEntryEntity != null) {
-                return createResponse(HttpStatus.OK, id, "Attempting to resend notification...");
-            } else {
-                return createResponse(HttpStatus.BAD_REQUEST, id, "No audit entry with the provided id exists.");
-            }
+            auditEntries = auditEntryActions.resendNotification(id);
+
+            return createResponse(HttpStatus.OK, id, gson.toJson(auditEntries));
+
```
gkillough: "Whitespace" -ek

src/main/java/com/blackducksoftware/integration/hub/alert/accumulator/RealTimeListener.java
jamesrichard91: We use this as a catch all, because if an exception does get thrown the JmsListener will add the message back into the Queue and keep trying to send it

src/main/java/com/blackducksoftware/integration/hub/alert/channel/DistributionChannel.java
jamesrichard91: We set the limit in the AuditEntryEntity. We could use a static variable for the limit so we dont have to change it in both places

PR #66, opened 2018-01-08
Fix ProjectData cache key and update db entities with enum.

gkillough:
There are virtually no functional changes. I refactored to switch from Strings to enums in a few database objects.


src/main/java/com/blackducksoftware/integration/hub/alert/digest/model/ProjectDataFactory.java
```diff
@@ -50,17 +50,17 @@
     public Collection<ProjectData> createProjectDataCollection(final Collection<NotificationEntity> notifications, final DigestTypeEnum digestType) {
         final Map<String, ProjectDataBuilder> projectDataMap = new LinkedHashMap<>();
         for (final NotificationEntity entity : notifications) {
-            final String projectKey = entity.getEventKey();
+            final String projectKey = entity.getProjectName() + entity.getProjectVersion();
```
gkillough: This is the fix.

src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestRemovalProcessor.java
```diff
@@ -74,7 +74,7 @@ private String createCacheKey(final NotificationEntity entity) {
     }
 
     private boolean processPolicyNotifications(final Map<String, NotificationEntity> categoryMap, final NotificationEntity entity) {
-        final String notificationType = entity.getNotificationType();
+        final String notificationType = entity.getNotificationType().name();
         if (NotificationCategoryEnum.POLICY_VIOLATION.name().equals(notificationType)) {
```
bamandel: You can compare just enums and remove all string related expressions here

src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestRemovalProcessor.java
```diff
@@ -93,7 +93,7 @@ private boolean processPolicyNotifications(final Map<String, NotificationEntity>
     }
 
     private boolean processVulnerabilityNotifications(final String cacheKey, final Map<String, NotificationEntity> categoryMap, final NotificationEntity entity) {
-        final String notificationType = entity.getNotificationType();
+        final String notificationType = entity.getNotificationType().name();
```
bamandel: You can also try getting just the enum here as well but would require some additional changes to the maps below

src/test/java/com/blackducksoftware/integration/hub/alert/datasource/entity/NotificationEntityTest.java
```diff
@@ -72,7 +76,27 @@ public void assertEntityFieldsFull(final NotificationEntity entity) {
...
+    @Test
+    public void testEntity() throws JSONException {
+        // TODO figure out why the hash code keeps changing
+        final NotificationEntity configEntity = getMockUtil().createEntity();
+
+        assertEntityFieldsFull(configEntity);
+        assertEquals(Long.valueOf(getMockUtil().getId()), configEntity.getId());
+
+        // TODO figure out why the hash code keeps changing
+        // final int configHash = configEntity.hashCode();
```
bamandel: Looks like duplicate TODOs here

src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestRemovalProcessor.java
gkillough: That would require modifying the entity cache stuff. We don't want to do that for now.

src/test/java/com/blackducksoftware/integration/hub/alert/datasource/entity/NotificationEntityTest.java
gkillough: Good catch

src/main/java/com/blackducksoftware/integration/hub/alert/digest/DigestRemovalProcessor.java
gkillough: Actually, I think there's an easy way to do it.

PR #70, opened 2018-01-10
Vulnerability operations

jamesrichard91:


src/main/java/com/blackducksoftware/integration/hub/alert/accumulator/RealTimeListener.java
```diff
@@ -43,21 +46,24 @@
...
+        this.projectDataFactory = projectDataFactory;
+        this.eventManager = eventManager;
     }
 
     @JmsListener(destination = RealTimeEvent.TOPIC_NAME)
     @Override
     public void receiveMessage(final String message) {
         try {
             final RealTimeEvent event = getEvent(message);
-            final List<AbstractChannelEvent> events = notificationProcessor.processNotifications(DigestTypeEnum.REAL_TIME, event.getNotificationList());
```
gkillough: Is this used at all (now that it's deleted)?

gkillough: Or is it used in daily, and we just don't want to remove notifications from the real time digest?

jamesrichard91: Its used in daily, we dont want to remove deletions from real time. We can have a longer conversation about the processing if you'd like

PR #74, opened 2018-01-12
Slack format fix

psantos1113:


src/main/resources/email/templates/digest.ftl
```diff
@@ -46,28 +46,29 @@
...
-                                   <#elseif itemType == "UPDATED">
-                                     Updated (${item_value})
-                                   <#elseif itemType == "DELETED">
-                                     Deleted (${item_value})
-                                   <#elseif itemType == "PERSON">
-                                     By: ${item_value}
-                                   <#elseif itemType == "VERSION">
-                                     ${item_value}
-                                   <#else>
+                                   Component: ${item.dataSet.COMPONENT}
```
jamesrichard91: I added these changes to the email template, so the emails are always consistent in how they display component information

src/main/java/com/blackducksoftware/integration/hub/alert/channel/slack/SlackChannel.java
```diff
@@ -142,16 +144,35 @@ protected String createMessage(final ProjectData projectData) {
...
                         messageBuilder.append(System.lineSeparator());
                         final Map<String, Object> dataSet = item.getDataSet();
-                        messageBuilder.append("Rule: _" + dataSet.get(ItemTypeEnum.RULE.toString()));
-                        messageBuilder.append("_" + System.lineSeparator());
-                        messageBuilder.append("Component: _" + dataSet.get(ItemTypeEnum.COMPONENT.toString()));
-                        messageBuilder.append("_ [" + dataSet.get(ItemTypeEnum.VERSION.toString()) + "]");
+                        final String ruleKey = ItemTypeEnum.RULE.toString();
+                        if (dataSet.containsKey(ruleKey) && StringUtils.isNotBlank(dataSet.get(ruleKey).toString())) {
+                            messageBuilder.append("Rule: " + dataSet.get(ItemTypeEnum.RULE.toString()));
+                            messageBuilder.append(System.lineSeparator());
```
gkillough: This shouldn't cause an issue, but since this is being displayed through slack rather than something system dependent, we probably should just use a '\n'. I don't know how slack will format carriage returns (if alert is running on windows).

PR #81, opened 2018-01-22
Fix for okhttp leaked connection

bamandel:


src/main/java/com/blackducksoftware/integration/hub/alert/channel/ChannelRestFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import com.blackducksoftware.integration.exception.IntegrationException;
+import com.blackducksoftware.integration.hub.rest.RestConnection;
+import com.blackducksoftware.integration.hub.rest.exception.IntegrationRestException;
+
+import okhttp3.HttpUrl;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+public class ChannelRestFactory {
```
gkillough: How is this entire file new? I thought this already existed?

src/main/java/com/blackducksoftware/integration/hub/alert/channel/ChannelRestFactory.java
```diff
@@ -0,0 +1,61 @@
...
+        Response response = null;
+        try {
+            logger.info("Attempting to send message...");
+            response = restConnection.createResponse(request);
+            logger.info("Successfully sent a message!");
+            if (logger.isTraceEnabled()) {
+                logger.trace("Response: " + response.toString());
+            }
+        } catch (final IntegrationException e) {
+            throw new IntegrationRestException(400, "Failed to send message", e.getMessage(), e);
```
psantos1113: Don't throw a blanket HTTP 400 here.  That is not correct.  Check the status on the response above in the try section and throw a new IntegrationRestException with the correct HTTP status code from the response.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/ChannelRestFactory.java
bamandel: The other class is rest connection, this new file is just rest functionality

src/main/java/com/blackducksoftware/integration/hub/alert/channel/ChannelRestFactory.java
```diff
@@ -0,0 +1,96 @@
...
+import okhttp3.HttpUrl;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+public class ChannelRestFactory {
+    private final RestConnection restConnection;
+    private final Logger logger;
+
+    public ChannelRestFactory(final String url, final GlobalProperties globalProperties, final Logger logger) throws IntegrationRestException {
```
gkillough: We really shouldn't throw exceptions from a constructor.

PR #83, opened 2018-01-26
RestDistributionChannel

gkillough:
Create RestDistributionChannel to abstract repeated message sending functionality.


src/main/java/com/blackducksoftware/integration/hub/alert/channel/rest/RestDistributionChannel.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void sendMessage(final E event, final C config) {
+        try {
+            final RestConnection restConnection = channelRestConnectionFactory.createUnauthenticatedRestConnection(getApiUrl());
+            final ChannelRequestHelper channelRequestHelper = new ChannelRequestHelper(restConnection);
+
+            final Request request = createRequest(channelRequestHelper, config, event.getProjectData());
+            channelRequestHelper.sendMessageRequest(request, event.getTopic());
+            setAuditEntrySuccess(event.getAuditEntryId());
```
bamandel: After looking at the EmailChannel, it seems like we have some similarities here for all channels. We may want to abstract this out into the handleEvent method of the DistributionChannel class and maybe have sendMessage throw the exception that possibly occurs.

PR #85, opened 2018-01-30
Hipchat api key test

gkillough:
These are the back-end changes for testing the HipChat global config.

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
@@ -102,8 +102,10 @@ public class HipChatChannel extends RestDistributionChannel<HipChatEvent, Global
             final Request request = restConnection.createPostRequest(httpUrl, requestHeaders, body);
 
             final ChannelRequestHelper channelRequestHelper = new ChannelRequestHelper(restConnection);
-            try (Response response = channelRequestHelper.sendGenericRequest(request)) {
-                if (response.isSuccessful()) {
+            Response response = null;
+            try {
+                response = channelRequestHelper.sendGenericRequest(request);
+                if (response != null && response.isSuccessful()) {
                     return "API key is valid.";
                 }
                 return "Invalid API key: " + response.message();
--- a/src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
@@ -33,6 +33,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.jms.annotation.JmsListener;
 import org.springframework.stereotype.Component;
 
+import com.blackducksoftware.integration.exception.IntegrationException;
 import com.blackducksoftware.integration.hub.alert.AlertConstants;
 import com.blackducksoftware.integration.hub.alert.audit.repository.AuditEntryRepositoryWrapper;
 import com.blackducksoftware.integration.hub.alert.channel.ChannelFreemarkerTemplatingService;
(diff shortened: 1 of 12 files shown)


src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -73,6 +81,41 @@ public String getApiUrl() {
         return HIP_CHAT_API;
     }
 
+    @Override
+    public String testGlobalConfig(final GlobalHipChatConfigEntity entity) {
+        if (entity == null) {
```
bamandel: Maybe we should add a test method to the above RestDistributionChannel class? Theoretically, all rest channel classes will also have a their associated test method

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -73,6 +81,41 @@ public String getApiUrl() {
...
+            return "The provided entity was null.";
+        }
+        final RestConnection restConnection = channelRestConnectionFactory.createUnauthenticatedRestConnection(HIP_CHAT_API);
```
bamandel: For tests methods like this we can create our own event with appropriate ProjectData passed in and our test entity. Then pass those values to the original rest connection in the parent class and have everything done automatically.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/manager/DistributionChannelManager.java
```diff
@@ -69,8 +69,8 @@ public ObjectTransformer getObjectTransformer() {
         return objectTransformer;
     }
 
-    public ProjectData getTestMessageProjectData() {
-        return new ProjectData(DigestTypeEnum.REAL_TIME, "Hub Alert", "Test Message", Collections.emptyList(), Collections.emptyMap());
+    public String testGlobalConfig(final G globalConfigEntity) {
+        return getDistributionChannel().testGlobalConfig(globalConfigEntity);
```
bamandel: Create a new event here to pass along with our entity

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
gkillough: Slack doesn't. Each test will be specific to the rest api that is being tested. We can abstract it when it actually makes sense to abstract it.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/manager/DistributionChannelManager.java
gkillough: It's a test, so there shouldn't be an event. Especially because it's a global config.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
gkillough: That breaks the paradigm. Events may contain project data and configuration info, but they aren't meant to be a convenience class (aka a wrapper). Events should only ever be created when we are processing notifications.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -73,6 +81,41 @@ public String getApiUrl() {
...
+            final Map<String, String> queryParameters = new HashMap<>();
+            queryParameters.put("auth_test", "true");
+
+            final Map<String, String> requestHeaders = new HashMap<>();
+            requestHeaders.put("Authorization", "Bearer " + entity.getApiKey());
+            requestHeaders.put("Content-Type", "application/json");
+
+            final HttpUrl httpUrl = restConnection.createHttpUrl(urlSegments, queryParameters);
+            // The body = {"message":"test"} is required to avoid a BAD_REQUEST (OkHttp issue: #854)
+            final RequestBody body = restConnection.createJsonRequestBody("{\"message\":\"test\"}");
```
psantos1113: Can you use the createMessageRequest from ChannelRequestHelper to help encapsulate creating the request body and the post request.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/manager/DistributionChannelManager.java
```diff
@@ -80,6 +80,10 @@ public String sendTestMessage(final R restModel) throws AlertException {
         return "Attempting to send a test message...";
     }
 
+    public ProjectData getTestMessageProjectData() {
```
psantos1113: Is this method used anywhere? I cannot find it with changes in this pull request.

gkillough: I moved its location in the file. No functional change.

src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
gkillough: Yes. My mistake was adding this before I added that method in ChannelRequestHelper.

PR #88, opened 2018-01-31
Ui bug fixes

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -87,21 +87,22 @@ public String testGlobalConfig(final GlobalHipChatConfigEntity entity) {
...
-            final Request request = channelRequestHelper.createMessageRequest(httpUrl, requestHeaders, "{\"message\":\"test\"}");
+                // The {"message":"test"} is required to avoid a BAD_REQUEST (OkHttp issue: #854)
+                final ChannelRequestHelper channelRequestHelper = new ChannelRequestHelper(restConnection);
+                final HttpUrl httpUrl = restConnection.createHttpUrl(urlSegments, queryParameters);
+                final Request request = channelRequestHelper.createMessageRequest(httpUrl, requestHeaders, "{\"message\":\"test\"}");
+
+                Response response = null;
 
-            Response response = null;
-            try {
```
gkillough: Something I meant to fix, but missed was the null check on _response_. I don't think it can be null any more, but if that's not true, the 'return "Invalid API key: " + response.message()' needs to have a null check as well. I believe the former is true so we should remove it.

PR #90, opened 2018-02-02
Logout Fix

gkillough:
Changes suggested by Kevin K.


gkillough:
@psantos1113 The reason the names were changed here is because Spring has a default implementation of each of these endpoints, and that was being called. Since we are using the Hub for authentication, there's no reason we should take the extra dozen or so steps to reach parity with Spring's default implementation. The easiest way to solve this was to rename the endpoins because they are not easily overridden. The appropriate changes were made in the UI as well.

PR #91, opened 2018-02-05
Add '/api' prefix to REST endpoints

gkillough:


gkillough:
This should be reviewed after this PR: 
https://github.com/blackducksoftware/hub-alert/pull/90

src/main/java/com/blackducksoftware/integration/hub/alert/web/controller/BaseController.java
```diff
@@ -0,0 +1,31 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.blackducksoftware.integration.hub.alert.web.controller;
+
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RequestMapping("/api")
+public abstract class BaseController {
```
psantos1113: I have an idea here that I will run by you @gkillough might allow us to provide an API directory to allow for API discovery.  I'll reach out to you on this.

bamandel: I think Spring offers an easy way to specify all rest controllers with a certain prefix. That may be simpler than this parent class

PR #97, opened 2018-02-23
Web permissions fix

psantos1113:
Fixed an issue where all requests for the global configuration were permitted.  Remove the need for the preauthorize annotation and setup the security context in the AuthenticationHandler class.

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/alert/web/SSLAuthenticationHandler.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/alert/web/SSLAuthenticationHandler.java
@@ -36,6 +36,10 @@ public class SSLAuthenticationHandler extends WebSecurityConfigurerAdapter {
 
     @Override
     protected void configure(final HttpSecurity http) throws Exception {
-        http.requiresChannel().anyRequest().requiresSecure().and().csrf().disable().authorizeRequests().anyRequest().permitAll().and().logout().logoutSuccessUrl("/");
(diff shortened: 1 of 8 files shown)


src/main/java/com/blackducksoftware/integration/hub/alert/web/SSLAuthenticationHandler.java
```diff
@@ -36,6 +36,10 @@
 
     @Override
     protected void configure(final HttpSecurity http) throws Exception {
-        http.requiresChannel().anyRequest().requiresSecure().and().csrf().disable().authorizeRequests().anyRequest().permitAll().and().logout().logoutSuccessUrl("/");
+        final String[] allowedPaths = { "/", "/built/bundle.js", "index.html", "/api/configuration/global", "/api/login", "/api/logout" };
+        http.requiresChannel().anyRequest().requiresSecure().and().csrf().disable().authorizeRequests().antMatchers(allowedPaths).permitAll().and()
```
gkillough: We can do permitAll() and hasRole("ADMIN")?

psantos1113: yes the 'and' is essentially appending rules where the security context should be applied.  The first rule is that we requires SSL for any request.  Then for specific paths in the web app we permit all requests.  Then we add the rule to require the role of admin for any other request.  The permit all is applying to the paths specified in the antmatchers argument.

gkillough: That makes sense to me.

PR #98, opened 2018-02-26
Ps bug fixes

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/digest/model/ItemData.java
```diff
@@ -26,7 +26,6 @@
 import java.util.Map;
 
 public class ItemData extends DigestData {
-    // TODO make the key an ItemTypeEnum
```
bamandel: Did we never end up adding this enum? Or do we just not want to do this?

PR #99, opened 2018-03-07
Hip chat test fixes

psantos1113:
@foosbar take a look at the UI components especially the hipchatConfig.js file.  It fixes the issue with the fetching attribute in the state not getting set properly to enable the hipchat API key field.

The back end changes were to get the status message from the backend to the front end if the API key test fails or if the API key is null.

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
@@ -24,9 +24,7 @@
 package com.blackducksoftware.integration.hub.alert.channel.hipchat;
 
 import java.io.IOException;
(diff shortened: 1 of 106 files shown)


src/main/java/com/blackducksoftware/integration/hub/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -99,17 +99,16 @@ public String testGlobalConfig(final GlobalHipChatConfigEntity entity) {
 
                 // The {"message":"test"} is required to avoid a BAD_REQUEST (OkHttp issue: #854)
```
gkillough: If hub-common no longer uses OkHttp, we may be able to get rid of the hard-coded string in the request. Not a high priority, but worth noting.

PR #105, opened 2018-03-20
Dev/sonarcloud fixes

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/audit/repository/relation/AuditNotificationRelationPK.java
```diff
@@ -27,9 +27,25 @@
 
 @SuppressWarnings("serial")
 public class AuditNotificationRelationPK implements Serializable {
-    public Long auditEntryId;
-    public Long notificationId;
+    private Long auditEntryId;
```
gkillough: Did you test this? There seemed to be some weird behavior when these were private.

gkillough: It actually could have been because there was a constructor with the fields, but no setters. JPA didn't seem to like that.

psantos1113: it works correctly with the changes I have made with the set and get methods

PR #108, opened 2018-03-21
Dev/hub 4.5 auth fix

psantos1113:


src/main/js/store/actions/session.js
```diff
@@ -59,15 +59,12 @@ export function verifyLogin() {
         const csrfToken = getState().session.csrfToken;
```
foosbar: You can remove this line to get the csrf token since it's not being used in the request.

PR #114, opened 2018-03-23
Dev/audit refresh fix

psantos1113:


src/main/js/component/general/audit/Index.js
```diff
@@ -41,6 +42,7 @@ class Index extends Component {
             this.setState({ message: '' });
             this.setEntriesFromArray(nextProps.items);
         }
+        this.startAutoReload();
```
foosbar: This is problematic because this function will get called if any props change (either from redux or props passed in from HOC.  This differs from the distribution code because the fetch in done by redux.  In this case, you may want to wrap the startAutoReload in a condition to make sure fetching is false.

psantos1113: ok will do

PR #118, opened 2018-03-28
Dev/mail settings fix

psantos1113:
@foosbar please review the UI side.

@jamesrichard91 and @gkillough  please review the server side changes when you have a chance


src/main/js/component/channels/EmailConfiguration.js
```diff
@@ -29,20 +29,53 @@ class EmailConfiguration extends React.Component {
         if (nextProps.updateStatus === 'FETCHED' || nextProps.updateStatus === 'UPDATED') {
             this.setState({
                 mailSmtpHost: nextProps.mailSmtpHost || '',
-                mailSmtpFrom: nextProps.mailSmtpFrom || '',
-                mailSmtpAuth: nextProps.mailSmtpAuth || false,
                 mailSmtpUser: nextProps.mailSmtpUser || '',
                 mailSmtpPassword: nextProps.mailSmtpPassword || '',
                 mailSmtpPasswordIsSet: nextProps.mailSmtpPasswordIsSet || false,
                 mailSmtpPort: nextProps.mailSmtpPort || undefined,
```
foosbar: why undefined?  if the values don't exist they'd be undefined anyway.  A value like 0 would evaluate to false and the boolean values should already be true/false?

psantos1113: I think early on the values were strings that's probably why.  let me fix that.

PR #125, opened 2018-04-05
Dev/project data aggregation fix

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/audit/controller/AuditEntryActions.java
```diff
@@ -87,8 +90,23 @@ public AuditEntryActions(final AuditEntryRepositoryWrapper auditEntryRepository,
...
+
+    public AlertPagedRestModel<AuditEntryRestModel> get(final Integer pageNumber, final Integer pageSize) {
+        AlertPage<AuditEntryEntity> auditEntries;
+        logger.debug("Audit entry get. PageNumber: {} PageSize: {}", pageNumber, pageSize);
+        if (pageNumber != null && pageSize != null) {
+            final PageRequest pageRequest = new PageRequest(pageNumber, pageSize, new Sort(Sort.Direction.DESC, "timeLastSent"));
+            auditEntries = auditEntryRepository.findAll(pageRequest);
+        } else {
+            final List<AuditEntryEntity> contentList = auditEntryRepository.findAll();
+            auditEntries = new AlertPage<>(1, 1, contentList.size(), contentList);
```
gkillough: Based on the AlertPage constructor, I think the second parameter should be 0 (zero):
new AlertPage(final int totalPages, final int currentPage, final int pageSize, final List<D> contentList)

PR #130, opened 2018-04-20
Dev/timeout login page fix

psantos1113:
redirect to the login page if the session times out or if the server is restarted.


src/main/js/component/general/audit/Index.js
```diff
@@ -76,9 +77,14 @@ class Index extends Component {
         }).then((response) => {
             this.setState({ inProgress: false });
             if (!response.ok) {
-                return response.json().then((json) => {
-                    this.setState({ message: json.message });
-                });
+                switch(response.status) {
+                    case 401:
```
jamesrichard91: I could be wrong, but it looks like for 401 we do nothing, and for 403 we logout. Is that what is happening here?

src/main/js/store/actions/hipChatConfig.js
```diff
@@ -101,14 +103,19 @@ export function getConfig() {
...
+                if(response.ok) {
+                    response.json().then((body) => {
+                        if (body.length > 0) {
+                            dispatch(configFetched(body[0]));
+                        } else {
+                            dispatch(configFetched({}));
+                        }
+                    })
+                } else {
+                    dispatch(verifyLoginByStatus(response.status));
```
jamesrichard91: How do you handle non-authentication problems?

src/main/js/component/general/audit/Index.js
psantos1113: case 401:
case 403: 
together both 401 and 403 are treated the same.  That's how you group cases in switch statements.  We want the same behavior for both 401 and 403

PR #132, opened 2018-04-25
Fix the select all project selector in the project table

psantos1113:


src/main/js/component/general/distribution/ProjectConfiguration.js
```diff
@@ -21,25 +21,44 @@ class ProjectConfiguration extends Component {
...
+            const index = selectedArray.indexOf(row.name);
             if (index >= 0) {
-                selected.splice(index, 1); // if found, remove that element from selected array
+                selectedArray.splice(index, 1); // if found, remove that element from selected array
             }
         }
+    }
+
+    onRowSelectedAll(isSelected, rows) {
+        console.log("Selected rows ", rows);
```
jamesrichard91: Do we want to leave these logs?

psantos1113: No we do not thank you.  I missed that

PR #133, opened 2018-04-26
Add the project description to the models

psantos1113:
Display the description field in the project selection field.


src/main/js/component/general/distribution/ProjectConfiguration.js
```diff
@@ -11,10 +11,11 @@ function assignClassName(row, rowIdx) {
 }
 
 function assignDataFormat(cell, row) {
+    console.log("Cell ", cell);
```
jamesrichard91: Do we want to leave this log in?

psantos1113: No I will remove it

PR #135, opened 2018-05-03
Added more specific filtering

bamandel:


src/test/java/com/blackducksoftware/integration/hub/alert/digest/filter/NotificationPostProcessorTestIT.java
```diff
@@ -138,18 +122,4 @@ public void doFrequenciesMatchTest() {
         assertFalse(postProcessor.doFrequenciesMatch(configOther, projectDataOther));
     }
 
-    @Test
-    public void doNotificationTypesMatchWithNoneConfiguredTest() {
```
psantos1113: You will need to add a test for the filterMatchingNotificationTypes.  This method should still provide a valid test case

PR #136, opened 2018-05-03
Devf/about controller

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/alert/config/GlobalProperties.java
```diff
@@ -97,25 +99,34 @@
...
         }
     }
 
     public String getProductVersion() {
-        return productVersion;
+        if (aboutModel.isPresent()) {
+            return aboutModel.get().getVersion();
+        } else {
+            return PRODUCT_VERSION_UNKNOWN;
+        }
```
bamandel: I believe this sort of case can be handled using the Optional flatmap method.

PR #175, opened 2018-07-23
Audit table

jamesrichard91:
Fixing Audit table search bar and table sorting


src/main/java/com/blackducksoftware/integration/alert/web/audit/AuditEntryActions.java
```diff
@@ -120,6 +121,48 @@ public AuditEntryRestModel get(final Long id) {
         return null;
     }
 
+    public AlertPagedRestModel<AuditEntryRestModel> search(final Integer pageNumber, final Integer pageSize, final String searchTerm) {
```
bamandel: We use both AlertPagedRestModel and AlertPage objects in this class. AlertPage is only used in the get method above and is immediately converted into AlertPagedRestModel after it's created. We should remove AlertPage and just create AlertPagedRestModel's instead.

PR #179, opened 2018-07-26
Frequency refactor

psantos1113:
# Pull Request template

* Make the Frequency for real time and daily both pull from the data and perform the same processing.

*

*
*
*


src/main/java/com/blackducksoftware/integration/alert/workflow/scheduled/frequency/OnDemandTask.java
```diff
@@ -0,0 +1,73 @@
...
+        currentTime = currentTime.withZoneSameInstant(ZoneOffset.UTC);
+        final ZonedDateTime zonedStartDate = currentTime.minusSeconds(secondInterval);
+        final Date startDate = Date.from(zonedStartDate.toInstant());
+        final Date endDate = Date.from(currentTime.toInstant());
+        return new DateRange(startDate, endDate);
+    }
+
+    @Override
+    public DigestType getDigestType() {
+        return DigestType.REAL_TIME;
```
bamandel: Do we want to have the enum value match the class name? So instead of REAL_TIME we have ON_DEMAND or name the class RealTime?

PR #182, opened 2018-07-31
Rename hub to blackduck

jamesrichard91:
Initial attempt to rename Hub to Black Duck in Alert. The properties, repo, and libraries we depend on still need to be updated


src/main/resources/db/changelog/alert/changelog-3.0.0.xml
```diff
@@ -11,4 +11,9 @@
             <column name="CONTENT" type="CLOB"/>
         </createTable>
     </changeSet>
+    <changeSet author="jrichard" id="renameHubTable">
```
bamandel: I know that the other ID in this file is -1 but we generally try to create the ID's as the current time in millis. If you are making a bunch of changes at the same time, then you do something like 123456789-1 where the -1 is just a counter of how many things you've changed at that time.

PR #186, opened 2018-08-07
Write common notifications

psantos1113:
# Pull Request template

*

* Write the raw notification data into the database and update the audit page to render the raw notification

* Provider defines the notification types
* Provider starts an accumulator task
* accumulator task writes raw notification data into the database


src/main/java/com/blackducksoftware/integration/alert/common/descriptor/ProviderDescriptor.java
```diff
@@ -39,4 +41,6 @@ public ProviderDescriptor(final String name, final DescriptorConfig providerDesc
     public Provider getProvider() {
         return provider;
     }
+
+    public abstract Set<String> getNotificationTypes();
```
bamandel: should this call provider.getNotificationTypes instead of being abstract? Or will this have a different purpose?

src/main/java/com/blackducksoftware/integration/alert/provider/blackduck/descriptor/BlackDuckDescriptor.java
```diff
@@ -38,4 +40,8 @@ public BlackDuckDescriptor(final BlackDuckProviderDescriptorConfig providerDescr
         super(PROVIDER_NAME, providerDescriptorConfig, provider);
     }
 
+    @Override
+    public Set<String> getNotificationTypes() {
+        return getProvider().getNotificationTypes();
```
bamandel: We should have this method not be abstract and just call this line in the parent class

src/main/java/com/blackducksoftware/integration/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -126,25 +131,79 @@ public String testGlobalConfig(final HipChatGlobalConfigEntity entity) throws In
...
+            return eventContent.length() > MESSAGE_SIZE_LIMIT;
+        } else {
+            return false;
+        }
+    }
 
-            final Map<String, String> requestHeaders = new HashMap<>();
-            requestHeaders.put("Authorization", "Bearer " + globalConfig.getApiKey());
-            requestHeaders.put("Content-Type", "application/json");
+    private int calculateChunks(final String eventContent) {
```
jamesrichard91: Do we need this method for anything other than logging?

src/main/java/com/blackducksoftware/integration/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -126,25 +131,79 @@ public String testGlobalConfig(final HipChatGlobalConfigEntity entity) throws In
...
+        logger.info("Number of requests to submit: {}", requestCount);
+        return requestCount;
+    }
 
-            return channelRequestHelper.createPostMessageRequest(url, requestHeaders, jsonString);
+    private List<Request> createChunkedRequestList(final ChannelRequestHelper channelRequestHelper, final HipChatDistributionConfigEntity config, final HipChatGlobalConfigEntity globalConfig, final ChannelEvent event,
+            final int requestCount) throws IntegrationException {
+        final List<Request> requestList = new ArrayList<>(requestCount);
+        final String eventContent = event.getContent();
+        for (int index = 0; index < requestCount; index++) {
```
jamesrichard91: Couldn't this be a while loop? so that we dont need to calculate the chunks before hand. You can just keep going until the end > eventContent.length()

src/main/java/com/blackducksoftware/integration/alert/channel/hipchat/HipChatChannel.java
psantos1113: The return value of the method is used to be able to split the content into pieces.  We also use the return value to include in the message to denote the chunk number in the message we send to hipchat (1 of 10) for example.

psantos1113:
Reviewed with the team the changes have been made to address the issues.  merging

PR #196, opened 2018-08-28
Blackduck url configuration

psantos1113:
# Pull Request template

* Ability to set the blackduck URL to connect to via environment variable for fresh installs to work.

Adds the blackduck server URL to the database for the blackduck provider
* Allows the blackduck server URL to be set from an environment variable on startup
* Updated the deployment files to define the variables for the provider and scheduling components


src/main/js/component/providers/BlackDuckConfiguration.js
```diff
@@ -88,9 +89,14 @@ class BlackDuckConfiguration extends React.Component {
 
                 <form className="form-horizontal" onSubmit={this.handleSubmit}>
                     <div>
-                        <ReadOnlyField label="Url" name="blackDuckUrl" readOnly="true" value={this.props.blackDuckUrl}/>
+                        <TextInput label="Url"
```
bamandel: I'm worried that allowing people to change config while Alert is running will cause some issues.

psantos1113:
Changes were made to make the URL field read only until other changes are implemented.

PR #198, opened 2018-08-30
Provider distribution ui config

psantos1113:
# Pull Request template

*

* Update the UI to get the notification types from the provider dynamically

* Added a Distribution rest api for providers
* Create the field for the notification types
* Update the UI to get the notification types from the server for the selected provider
* Save the notification types selected per common distribution
* Fix a bug in the UI editing existing configurations


src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionRestApi.java
```diff
@@ -0,0 +1,52 @@
...
+public class BlackDuckDistributionRestApi extends RestApi {
+
+    @Autowired
+    public BlackDuckDistributionRestApi(final BlackDuckTypeConverter typeConverter, final BlackDuckRepositoryAccessor repositoryAccessor) {
+        super(typeConverter, repositoryAccessor);
+    }
+
+    @Override
+    public void validateConfig(final Config restModel, final Map<String, String> fieldErrors) {
+
```
jamesrichard91: Shouldn't there be configuration validation here?

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionRestApi.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public void testConfig(final DatabaseEntity entity) {
+
```
jamesrichard91: Shouldn't there be a way to test the config here?

src/main/js/component/general/distribution/JobAddModal.js
```diff
@@ -24,23 +25,30 @@ class JobAddModal extends Component {
...
 
+    componentDidMount() {
+        this.props.resetDistributionDescriptor();
+    }
+
     getCurrentJobConfig() {
         switch (this.state.values.typeValue) {
             case 'channel_email':
                 return (<GroupEmailJobConfiguration
+                    alertChannelName={this.state.values.typeValue}
```
jamesrichard91: These all have the same alertChannelName?

src/main/js/component/general/distribution/Index.js
```diff
@@ -84,18 +84,21 @@ class Index extends Component {
             const {distributionConfigId, distributionType} = currentRowSelected;
             if (distributionType === 'channel_email') {
                 return (<GroupEmailJobConfiguration
+                    alertChannelName={distributionType}
```
jamesrichard91: These all have the same alertChannelName?

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionRestApi.java
psantos1113: We could validate that project names exist or if the include all is selected. We never had distribution validation for Providers before.  I can add that.

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionRestApi.java
psantos1113: For testing hub project names what would we test here? This just creates a placeholder for us to perform distribution.  Perhaps based on your question here maybe Provider distributions shouldn't have a way to test since we aren't pushing a message to the provider.

src/main/js/component/general/distribution/JobAddModal.js
psantos1113: The property has to be named alertChannelName because the slack configuration that is read has a property channel_name.  The name of this property needed to be different in order to pass the correct channel name.  The this.state.values.typeValue is "channel_email", "channel_hipchat", or "channel_slack" based on what the user selected from the drop down. We want that value in alertChannelName in order to send a get request to get the right UI components for the distribution job.  In the future when we have a React component that renders the configuration screen based on the ui components from the server we won't need this switch statement.  We will just pass in the value for this.state.values.typeValue.

src/main/js/component/general/distribution/Index.js
psantos1113: Similar to the item about the distributionType is job specific and is "channel_email", "channel_hipchat", or "channel_slack" So they aren't the same.  It just looks that way because it's the same variable value.  In the future we won't have a switch statement here.

PR #199, opened 2018-09-04
Digest type rename

psantos1113:
# Pull Request template

*

* Rename the DigestType enum to FrequencyEnum since it is only the frequency to run

* . Rename the enum class and update the classes that use it
* Rename the column in the UI to Frequency Type from Digest Type
*


src/main/java/com/synopsys/integration/alert/web/controller/UIComponentController.java
```diff
@@ -103,7 +103,7 @@ public UIComponent getDistributionUIComponent(@RequestParam(value = "providerNam
             final UIComponent providerUIComponent = providerUIConfig.generateUIComponent();
             final List<ConfigField> combinedFields = new ArrayList<>();
             final ConfigField name = new TextInputConfigField("name", "Name", true, false);
-            final ConfigField frequency = new SelectConfigField("frequency", "Digest type", true, false, Arrays.stream(DigestType.values()).map(type -> type.getDisplayName()).collect(Collectors.toList()));
+            final ConfigField frequency = new SelectConfigField("frequency", "Digest type", true, false, Arrays.stream(FrequencyType.values()).map(type -> type.getDisplayName()).collect(Collectors.toList()));
```
bamandel: Should Digest type here be changed to Frequency?

psantos1113: yes it should

psantos1113:
Tested both a clean install of the database as an existing database before the refactor.  Both worked correctly.  Merging this pull request.

PR #200, opened 2018-09-10
Sync blackduck data

jamesrichard91:
# Pull Request template

* Syncing the BlackDuck data with the Alert DB every minute

* Sync User emails
* Sync UserGroups
* Sync relations between users and groups
* Sync Projects
* Sync relations between project and users, projects and groups(the users in the groups)


src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/EmailSyncTask.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public DatabaseEntity createEntity(final String data) {
+        return new BlackDuckUserEntity(data, false);
+    }
+
+    @Override
+    public void addRelations(final Map<String, ? extends HubView> currentDataMap, final List<? extends DatabaseEntity> storedEntities, final HubService hubService) throws IOException, IntegrationException {
+
```
gkillough: Is this just empty because we don't have any email-only configurations yet?

src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/GroupSyncTask.java
```diff
@@ -0,0 +1,149 @@
...
+        final Set<GroupData> storedGroups = blackDuckGroupEntities.stream().map(blackDuckGroupEntity -> new GroupData(blackDuckGroupEntity.getName(), blackDuckGroupEntity.getActive(), blackDuckGroupEntity.getHref()))
+                                                .collect(Collectors.toSet());
+        return storedGroups;
+    }
+
+    @Override
+    public List<BlackDuckGroupEntity> getEntitiesToRemove(final List<? extends DatabaseEntity> storedEntities, final Set<GroupData> dataToRemove) {
+        final List<BlackDuckGroupEntity> blackDuckGroupEntities = (List<BlackDuckGroupEntity>) storedEntities;
+        final List<BlackDuckGroupEntity> blackDuckGroupsToRemove = blackDuckGroupEntities.stream()
+                                                                       .filter(blackDuckGroupEntity -> {
```
gkillough: When it comes to streams, do we want lambdas with more than one line (i.e. block-lambdas) to have their own method?

src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/EmailSyncTask.java
jamesrichard91: This is empty, because the Email syncing does not add relationships. The Group syncing and Project syncing do.  I can remove this from the object contract and leave an empty default method in the parent if you would prefer that

PR #203, opened 2018-09-13
Refine json extractor

gkillough:

* Before this, we used the JsonExtractor to deserialize objects every time we needed to access a single field.
* These changes give us a JsonFieldAccessor that is created by deserializing something once, then allowing us to get values by field in constant time.
* These changes also reduce the code footprint in TopicCollector and JsonExtractor in exchange for a model that can be reused throughout the processing of a single notification.


gkillough:
I still need to add a few test cases, but all of the existing tests pass.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckTopicCollector.java
```diff
@@ -37,15 +35,4 @@
     public BlackDuckTopicCollector(final JsonExtractor jsonExtractor, final BlackDuckDescriptor blackDuckDescriptor, final List<TopicFormatter> topicFormatterList) {
```
gkillough: I'm not sure if we really need this class any more.

PR #205, opened 2018-09-14
Ps collector selection

psantos1113:
# Pull Request template

*

* process notifications per distribution job in parallel

*
*
*


src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckTopicCollectorFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.beans.factory.ObjectFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.workflow.processor.TopicCollector;
+import com.synopsys.integration.alert.provider.blackduck.collector.BlackDuckPolicyTopicCollector;
+import com.synopsys.integration.alert.provider.blackduck.collector.BlackDuckVulnerabilityTopicCollector;
+
+@Component
+public class BlackDuckTopicCollectorFactory {
```
gkillough: I almost feel like we should just have TopicCollectorFactory that has a method that takes in a NotificationType and returns a TopicCollector. It's constructor could a Collection of TopicCollectors and search for the one that applies to the notification provided.

src/main/java/com/synopsys/integration/alert/workflow/filter/NotificationFilter.java
```diff
@@ -102,6 +102,34 @@ public NotificationFilter(final JsonExtractor jsonExtractor, final List<Provider
...
+        notificationsByType.forEach((type, groupedNotifications) -> {
+            final Set<StringHierarchicalField> filterableFields = getFilterableFieldsByNotificationType(type, providerContentTypes);
+            final Predicate<NotificationContent> filterForNotificationType = createJobFilter(filterableFields, jobConfiguration);
+
+            final List<NotificationContent> matchingNotifications = applyFilter(groupedNotifications, filterForNotificationType);
+            filteredNotifications.addAll(matchingNotifications);
+        });
+
+        return filteredNotifications
+                   .parallelStream()
```
gkillough: I don't think parallelStream() buys us anything with sorted(). I implemented it this way too, but thinking about it now doesn't make a lot of sense.

src/main/java/com/synopsys/integration/alert/workflow/processor/JobProcessor.java
```diff
@@ -0,0 +1,126 @@
...
+            notificationsForJob.parallelStream()
+                .filter(notificationContent -> collectorMap.containsKey(notificationContent.getNotificationType()))
+                .forEach(notificationContent -> collectorMap.get(notificationContent.getNotificationType()).insert(notificationContent));
+
+            return collectorMap.values().parallelStream().flatMap(collector -> collector.collect(formatType).stream()).collect(Collectors.toList());
+        }
+    }
+
+    //TODO notificationFilter also has this should it be made common?
+    private <T> List<T> applyFilter(final Collection<T> notificationList, final Predicate<T> filter) {
```
gkillough: We might not need this method if we just operated on Streams rather than Collections.

psantos1113:
Merging into the notification_formatting branch so that the comments provided can be worked on there.

PR #207, opened 2018-09-19
New hipchat template

gkillough:


src/main/java/com/synopsys/integration/alert/channel/hipchat/HipChatChannel.java
```diff
@@ -224,14 +224,12 @@ private String createHtmlMessage(final String contentTitle, final String content
...
             }
-
             final ChannelFreemarkerTemplatingService freemarkerTemplatingService = new ChannelFreemarkerTemplatingService(templateDirectoryPath);
 
             final HashMap<String, Object> model = new HashMap<>();
-            model.put("content", content);
-            model.put("contentTitle", contentTitle);
+            model.put("content", messageContent);
 
-            return freemarkerTemplatingService.getResolvedTemplate(model, "audit.ftl");
```
gkillough: Should I delete the old notification.ftl?

gkillough: Obviously not the audit one.

psantos1113: yes you can delete the old notification.ftl since it is no longer used.  The audit.ftl file was originally to display the raw content.  That may not be needed anymore either.

PR #213, opened 2018-09-21
Alert project owner

jamesrichard91:

* Allow the User to send Alerts only to the Project Owner if configured.


src/main/java/com/synopsys/integration/alert/channel/event/ChannelEventFactory.java
```diff
@@ -157,30 +159,36 @@ public EmailChannelEvent createEmailChannelTestEvent(final Config restModel) {
...
-                                               .stream()
-                                               .map(userProjectRelation -> blackDuckUserRepositoryAccessor.readEntity(userProjectRelation.getBlackDuckUserId()))
-                                               .filter(userEntity -> userEntity.isPresent())
-                                               .map(userEntity -> ((BlackDuckUserEntity) userEntity.get()).getEmailAddress())
-                                               .collect(Collectors.toSet());
+        final Set<String> emailAddresses;
+        if (projectOwnerOnly && StringUtils.isNotBlank(blackDuckProjectEntity.getProjectOwnerEmail())) {
+            emailAddresses = new HashSet<>();
+            emailAddresses.add(blackDuckProjectEntity.getProjectOwnerEmail());
+        } else {
```
jamesrichard91: If they configured to send alerts to the Project owner only and there is none, should we not send anything? Or should we send it to the Users assigned to the project?
Right now the code is set to send it to the Users assigned to the project if there is no project owner

src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/ProjectSyncTask.java
```diff
@@ -108,18 +108,30 @@ public void run() {
...
+    public Map<BlackDuckProject, ProjectView> getCurrentData(final List<ProjectView> projectViews, HubService hubService) {
+        final Map<BlackDuckProject, ProjectView> projectMap = new ConcurrentHashMap<>();
+        projectViews
+            .parallelStream()
+            .forEach(projectView -> {
+                String projectOwnerEmail = null;
+                if (StringUtils.isNotBlank(projectView.projectOwner)) {
+                    try {
+                        UserView projectOwner = hubService.getResponse(projectView.projectOwner, UserView.class);
+                        projectOwnerEmail = projectOwner.email;
```
jamesrichard91: The 'projectOwner' field is a link we have to follow to get the user information

src/main/java/com/synopsys/integration/alert/channel/event/ChannelEventFactory.java
gkillough: Right, I meant somewhere in the configuration, so the test event would work.

PR #214, opened 2018-09-26
Revisit @Transactional

gkillough:
- I have cut down the number of classes in src/main/java that use @Transactional to 12. 
- I have eliminated the use of javax transactions in favor of spring for consistency.
- There is still some work to be done to ensure that private methods are being utilized correctly in transactions (methods annotated with @Transactional must be public or they will be ignored).

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/ChannelTemplateManager.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/ChannelTemplateManager.java
@@ -26,11 +26,10 @@ package com.synopsys.integration.alert.channel;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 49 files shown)


src/main/java/com/synopsys/integration/alert/channel/ChannelTemplateManager.java
```diff
@@ -43,7 +42,6 @@
 import com.synopsys.integration.alert.database.audit.AuditNotificationRepository;
 import com.synopsys.integration.alert.database.audit.relation.AuditNotificationRelation;
 
-@Transactional
```
gkillough: Removing the annotation from the class level should increase efficiency of methods that do not interact with the DB.

PR #215, opened 2018-09-28
Openshift protoform support

psantos1113:
# Pull Request template

* Changes to the code to remove chown and chmod from the dockerfile to support openshift
* changes to support upgrading from 2.x to 3.0

*
*
*


src/main/js/component/general/audit/Details.js
```diff
@@ -54,7 +54,9 @@ class Details extends Component {
         if (this.state.values.errorStackTrace) {
             errorStackTrace = <TextArea inputClass="stackTraceContainer" label="Stack Trace" readOnly name="errorStackTrace" value={this.state.values.errorStackTrace}/>;
         }
-        const jsonPrettyPrintContent = JSON.stringify(JSON.parse(this.state.values.content), null, 2);
+        const defaultContent = Object.assign({}, {'warning': 'Content in an Unknown Format'});
+        const jsonContent = this.state.values.content || JSON.stringify(defaultContent);
```
jamesrichard91: For the default content, we use JSON.stringify then we parse it and stringify it again? That seems sill. We should have an if to correctly display the real value if available, otherwise show the default

PR #216, opened 2018-10-02
Category item add content

psantos1113:
# Pull Request template

* Add the policy overrider name to the collection of items
* Add the severity of the vulnerability to the collection of items

*
*
*

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
@@ -150,9 +150,7 @@ public abstract class MessageContentCollector {
         final List<LinkableItem> list = new ArrayList<>();
         if (null != items) {
             for (final LinkableItem item : items) {
-                if (null != item) {
-                    list.add(item);
-                }
+                list.add(item);
             }
         }
         return list;
--- a/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
@@ -148,9 +148,11 @@ public abstract class MessageContentCollector {
 
     protected final List<LinkableItem> createLinkableItemList(final LinkableItem... items) {
         final List<LinkableItem> list = new ArrayList<>();
-        if (items != null) {
+        if (null != items) {
             for (final LinkableItem item : items) {
-                list.add(item);
+                if (null != item) {
+                    list.add(item);
+                }
             }
         }
         return list;
@@ -182,7 +184,7 @@ public abstract class MessageContentCollector {
         for (int index = 0; index < count; index++) {
             final LinkableItem topicItem = topicItems.get(index);
 
-            LinkableItem subTopic;
+            final LinkableItem subTopic;
             if (!subTopicItems.isEmpty()) {
                 subTopic = subTopicItems.get(index);
             } else {
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderContentTypes.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderContentTypes.java
@@ -39,13 +39,41 @@ import com.synopsys.integration.blackduck.api.generated.enumeration.Notification
 import com.synopsys.integration.blackduck.notification.content.VulnerabilitySourceQualifiedId;
 
 public class BlackDuckProviderContentTypes {
+    //common fields
+    public static final String JSON_FIELD_CONTENT = "content";
+    public static final String JSON_FIELD_PROJECT_NAME = "projectName";
+    public static final String JSON_FIELD_PROJECT_VERSION_NAME = "projectVersionName";
+    public static final String JSON_FIELD_PROJECT_VERSION = "projectVersion";
+    public static final String JSON_FIELD_COMPONENT_NAME = "componentName";
+    public static final String JSON_FIELD_COMPONENT = "component";
+    public static final String JSON_FIELD_COMPONENT_VERSION_NAME = "componentVersionName";
+    public static final String JSON_FIELD_COMPONENT_VERSION = "componentVersion";
+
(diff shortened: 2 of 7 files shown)


src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
```diff
@@ -148,9 +148,11 @@ protected void addItem(final List<CategoryItem> categoryItems, final CategoryIte
 
     protected final List<LinkableItem> createLinkableItemList(final LinkableItem... items) {
         final List<LinkableItem> list = new ArrayList<>();
-        if (items != null) {
+        if (null != items) {
             for (final LinkableItem item : items) {
-                list.add(item);
+                if (null != item) {
```
gkillough: Why would there be null items in this list? Are we explicitly adding something that can be null?

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderContentTypes.java
```diff
@@ -68,6 +72,8 @@
             new StringHierarchicalField(Arrays.asList("content"), "component", FieldContentIdentifier.CATEGORY_ITEM, LABEL_COMPONENT_NAME + HierarchicalField.LABEL_URL_SUFFIX),
             new StringHierarchicalField(Arrays.asList("content"), "componentVersionName", FieldContentIdentifier.CATEGORY_ITEM, LABEL_COMPONENT_VERSION_NAME),
             new StringHierarchicalField(Arrays.asList("content"), "componentVersion", FieldContentIdentifier.CATEGORY_ITEM, LABEL_COMPONENT_VERSION_NAME + HierarchicalField.LABEL_URL_SUFFIX),
+            new StringHierarchicalField(Arrays.asList("content"), "firstName", FieldContentIdentifier.CATEGORY_ITEM, LABEL_POLICY_OVERRIDE_FIRST_NAME),
+            new StringHierarchicalField(Arrays.asList("content"), "lastName", FieldContentIdentifier.CATEGORY_ITEM, LABEL_POLICY_OVERRIDE_LAST_NAME),
```
jamesrichard91: Why do we have the static strings above and we aren't using them here?

src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
psantos1113: The first and last names for policy are only there for policy override notifications.  So if it isn't present send a null value to avoid branching code in the policy collector.  The same is true in the edge case of losing connectivity to BlackDuck in order to get the severity of the vulnerabilities.  There I would need more complicated branching code to do the same call.  Let me look at the code another time to see if I can refactor into something clean.  I don't like checking for null either.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckVulnerabilityMessageContentCollector.java
```diff
@@ -71,35 +82,72 @@ protected void addCategoryItems(final List<CategoryItem> categoryItems, final Js
...
-            addVulnerabilityItems(categoryItems, notificationContent.getId(), ItemOperation.UPDATE, BlackDuckProviderContentTypes.LABEL_VULNERABILITY_NEW, componentItem, componentVersionItem, updatedVulnerabilityList);
-
             final List<VulnerabilitySourceQualifiedId> deletedVulnerabilityList = getFieldValueObjectsByLabel(jsonFieldAccessor, vulnerabilityFields, BlackDuckProviderContentTypes.LABEL_VULNERABILITY_DELETED, targetClass);
-            addVulnerabilityItems(categoryItems, notificationContent.getId(), ItemOperation.DELETE, BlackDuckProviderContentTypes.LABEL_VULNERABILITY_NEW, componentItem, componentVersionItem, deletedVulnerabilityList);
-        } catch (final AlertException e) {
-            logger.error("Mishandled the expected type of a notification field", e);
+            final Optional<BlackduckRestConnection> restConnectionOptional = blackDuckProperties.createRestConnectionAndLogErrors(logger);
+            // TODO: this seems overly complicated to get a service connection to BlackDuck.  Need to implement a component that manages the connections and ask that class for a service.
```
jamesrichard91: The problem with creating a Component to create the services is, how/when do we close the BlackDuckRestConnection? At least this way we know we aren't leaving open connections

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderContentTypes.java
psantos1113: I'll refactor what is reused.

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderContentTypes.java
```diff
@@ -39,15 +39,47 @@
...
+    public static final String JSON_FIELD_UPDATED_VULNERABILITY_IDS = "updatedVulnerabilityIds";
+    public static final String JSON_FIELD_DELETED_VULNERABILITY_IDS = "deletedVulnerabilityIds";
+
+    // labels
     public static final String LABEL_COMPONENT_NAME = "Component";
     public static final String LABEL_COMPONENT_VERSION_NAME = "Component Version";
     public static final String LABEL_POLICY_NAME = "Policy";
     public static final String LABEL_PROJECT_NAME = "Project";
     public static final String LABEL_PROJECT_VERSION_NAME = "Project Version";
+    public static final String LABEL_POLICY_OVERRIDE_FIRST_NAME = JSON_FIELD_FIRST_NAME;
```
gkillough: Keep in mind, the label is the thing that shows up in the channel message, so case might matter for consistency:
e.g.
firstName: Bob
vs
Policy Overrider: Bob

psantos1113: This concern is currently addressed in the policy collector object

PR #227, opened 2018-10-29
Ps post 3 cleanup

psantos1113:

* SonarCloud fixes and removal of unused classes


src/main/java/com/synopsys/integration/alert/workflow/startup/StartupManager.java
```diff
@@ -133,8 +133,12 @@ public void startup() {
     }
 
     public void checkEncryptionProperties() {
-        alertProperties.getAlertEncryptionPassword().orElseThrow(() -> new IllegalArgumentException("Encryption password not configured"));
-        alertProperties.getAlertEncryptionGlobalSalt().orElseThrow(() -> new IllegalArgumentException("Encryption salt not configured"));
+        if (!alertProperties.getAlertEncryptionPassword().isPresent()) {
```
gkillough: I think we actually want to prefer the original implementation here. As optional becomes ubiquitous, best practices become more apparent:
https://dzone.com/articles/optional-ispresent-is-bad-for-you

PR #232, opened 2018-11-13
Switch to json path

jamesrichard91:


src/main/java/com/synopsys/integration/alert/common/field/JsonField.java
```diff
@@ -0,0 +1,102 @@
...
+    private final JsonPath configNameMapping;
+    private final JsonPath jsonPath;
+
+    private final String fieldName;
+    private final String label;
+
+    public static <T> JsonField<T> createObjectField(final JsonPath jsonPath, final String fieldName, final FieldContentIdentifier contentIdentifier, final String label, final TypeRef<?> typeRef) {
+        return new JsonField<T>(typeRef, jsonPath, fieldName, contentIdentifier, label);
+    }
+
```
jamesrichard91: Because we set the configuration of JsonPath to always return lists, the TypeRef must always be a List of the type you want

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderContentTypes.java
```diff
@@ -165,6 +184,10 @@
 
     public static final List<ProviderContentType> ALL = Arrays.asList(BOM_EDIT, LICENSE_LIMIT, POLICY_OVERRIDE, RULE_VIOLATION, RULE_VIOLATION_CLEARED, VULNERABILITY);
 
+    private static JsonPath createJsonPath(final String pattern, final String... fields) {
```
gkillough: I believe this should be a public static method on JsonField because that is the thing that demands JsonPath (and providers other than Black Duck will likely need access to the it).

psantos1113: I agree with Gavin here.  This method should be in JsonField

jamesrichard91: Moved to JsonField

PR #235, opened 2018-11-26
Test Config Modal for Email and HipChat

gkillough:
Allows users to send a test message to a specified destination.


src/main/js/component/common/ChannelTestModal.js
```diff
@@ -0,0 +1,72 @@
...
+        this.setState({destination: target.value});
+    }
+
+    handleSendTestMessage(event) {
+        event.preventDefault();
+        event.stopPropagation();
+        this.setState({
+            spinning: true
+        });
+        this.props.sendTestMessage(this.state.destination);
```
jamesrichard91: Couldn't we set spinning to false again after the sendTestMessage() is complete?

src/main/java/com/synopsys/integration/alert/channel/slack/descriptor/SlackDistributionDescriptorActionApi.java
```diff
@@ -64,7 +64,7 @@ public void validateConfig(final Config restModel, final Map<String, String> fie
     }
 
     @Override
-    public void testConfig(final Config restModel) throws IntegrationException {
+    public void testConfig(final Config restModel, final String ignoredDestination) throws IntegrationException {
```
jamesrichard91: Since we have ignoredDestination in several places, would it make sense that those classes have a different descriptor? Or should the classes that use the destination have a different descriptor?

src/main/js/component/common/ChannelTestModal.js
gkillough: Yes, but the problem is that sendTestMessage does a dispatch asynchronously, so as soon as the method is called, the spinner is hidden. This results in the spinner not appearing at all.

src/main/java/com/synopsys/integration/alert/channel/slack/descriptor/SlackDistributionDescriptorActionApi.java
gkillough: Yes, I discussed this with @psantos1113 last week, but we did not come up with a solution.

psantos1113: Since it is specific to a channel should we have the ChannelDescriptor class define the abstract method or DescriptorActionApi create a sub-class that is ChannelDescriptorActionApi and the Global and Distribution Action API classes extend that rather than just DescriptorActionApi?

PR #236, opened 2018-11-26
Ps system setup

psantos1113:

* Add the ability to configure alert the first time
* Display a configuration form if the system isn't installed
* Added tests to SystemMessages that were missing


src/main/java/com/synopsys/integration/alert/web/controller/SystemController.java
```diff
@@ -49,4 +54,15 @@ public SystemController(final SystemHandler handler) {
     public ResponseEntity<String> getSystemMessages(@RequestParam(value = "startDate", required = false) final String startDate, @RequestParam(value = "endDate", required = false) final String endDate) {
         return handler.getSystemMessages(startDate, endDate);
     }
+
+    @GetMapping(value = "/system/setup/initial")
```
psantos1113: Because we want to add a Setup menu item to display the form the initial configuration needs to be accessible without requiring login.  Therefore the /initial sub path will have permissions to be accessed without login.  When Setup becomes a menu item then /system/setup will be the URL path for that.  Both should be rendered by the same UI Component in the future.  Just wanted to provide some context on the path.

src/main/java/com/synopsys/integration/alert/web/controller/handler/SystemHandler.java
```diff
@@ -74,4 +78,33 @@ public SystemHandler(final ContentConverter contentConverter, final SystemAction
...
+            return new ResponseEntity<>(headers, HttpStatus.FOUND);
+        } else {
+            return new ResponseEntity<>(getContentConverter().getJsonString(actions.getCurrentSystemSetup()), HttpStatus.OK);
+        }
+    }
+
+    public ResponseEntity<String> saveRequiredInformation(final SystemSetupModel requiredSystemConfiguration) {
+        final ResponseEntity<String> response;
+        if (actions.isSystemInitialized()) {
+            response = new ResponseEntity<>(HttpStatus.METHOD_NOT_ALLOWED);
```
bamandel: Gavin and I were just talking about something similar to this. Should we respond with METHOD_NOT_ALLOWED, when the type of method is actually allowed, but just not all criteria is met?

src/main/java/com/synopsys/integration/alert/workflow/startup/install/SystemInitializer.java
```diff
@@ -35,27 +39,74 @@
...
+        Integer blackDuckConnectionTimeout = null;
+        String blackDuckApiToken = null;
+        if (blackDuckConfigEntity.isPresent()) {
+            final GlobalBlackDuckConfigEntity blackDuckEntity = blackDuckConfigEntity.get();
+            blackDuckUrl = blackDuckEntity.getBlackDuckUrl();
+            blackDuckConnectionTimeout = blackDuckEntity.getBlackDuckTimeout();
+            blackDuckApiToken = blackDuckEntity.getBlackDuckApiKey();
+        }
+
+        return new RequiredSystemConfiguration(blackDuckUrl,
```
bamandel: Converting something from an entity to another object seems like a good job for the TypeConverter. If we end up keeping this when multiple providers are added, we should aim for that change.

src/main/java/com/synopsys/integration/alert/database/system/SystemStatusUtility.java
```diff
@@ -67,6 +65,12 @@ public Date getStartupTime() {
         return getSystemStatus().getStartupTime();
     }
 
+    private Date createCurrentDateTimestamp() {
```
jamesrichard91: Would this code make sense in DateRange? I think we get the current time using the same code elsewhere in Alert

src/main/java/com/synopsys/integration/alert/web/actions/SystemActions.java
```diff
@@ -74,6 +85,40 @@ public SystemActions(final SystemStatusUtility systemStatusUtility, final System
...
+        return new SystemMessageModel(systemMessage.getSeverity(), createdAt, systemMessage.getContent(), systemMessage.getType());
+    }
+
+    public SystemSetupModel getCurrentSystemSetup() {
+        final RequiredSystemConfiguration systemConfiguration = systemInitializer.getCurrentSystemSetup();
+        return new SystemSetupModel(systemConfiguration.getBlackDuckProviderUrl(),
+            systemConfiguration.getBlackDuckConnectionTimeout(),
+            systemConfiguration.getBlackDuckApiToken(),
+            StringUtils.isNotBlank(systemConfiguration.getBlackDuckApiToken()),
+            systemConfiguration.getGlobalEncryptionPassword(),
```
jamesrichard91: I don't think we want to return the BlackDuckApiToken, GlobalEncryptionPassword, GlobalEncryptionSalt, or the ProxyPassword. We only want to return if they are set since this is an unauthenticated endpoint that will be getting this information.

jamesrichard91: We never want to return any kind of sensitive information to the UI authenticated or not.

src/main/js/store/reducers/system.js
```diff
@@ -20,6 +40,50 @@ const config = (state = initialState, action) => {
...
             });
+        case SYSTEM_SETUP_FETCHING:
+            return Object.assign({}, state, {
+                fetchingSetup: true,
+                updateStatus: 'FETCHING'
+            });
+        case SYSTEM_SETUP_FETCH_REDIRECTED:
+            return Object.assign({}, state, {
+                fetchingSetup: false,
+                setupRedirect: true
```
jamesrichard91: In all cases that don't set updateStatus to a value we should explicitly set it to an empty string. Otherwise it will carry over in the state and that has caused issues for us

PR #237, opened 2018-11-27
Add css class to field wrapper to correct error spacing

gkillough:


src/main/css/field.scss
```diff
@@ -29,7 +29,7 @@ small {
 
 .fieldError {
   color: #a94442;
-  padding: 4px 0;
+  padding: 4px 0px 0px;
```
gkillough: https://www.w3schools.com/css/css_padding.asp#midcontentadcontainer

jamesrichard91: Should we add the last '0px' to cover all of the padding?

gkillough: If the padding property has three values:
- padding: 25px 50px 75px;
- top padding is 25px
- right and left paddings are 50px
- bottom padding is 75px

jamesrichard91: Well then what you have is perfect

PR #242, opened 2018-12-07
Ps policy duplication fix

psantos1113:

* Fix the construction of the policy rule messages to avoid creating N x M categories which creates incorrect policy messages. 

This affects the RULE_VIOLATION and RULE_VIOLATION_CLEARED
N = # of policies (policyInfos attribute in JSON)
M = # of components (componentVersionStatuses in JSON)

For each policy and component a new category message would be created even if the component doesn't violate the policy.  

The fix is to map the components to the policies they are affected by and then create the category items from that.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
@@ -26,17 +26,22 @@ package com.synopsys.integration.alert.provider.blackduck.collector;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.SortedSet;
+import java.util.TreeMap;
 import java.util.TreeSet;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
+import com.jayway.jsonpath.TypeRef;
 import com.synopsys.integration.alert.common.enumeration.ItemOperation;
+import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.field.JsonField;
 import com.synopsys.integration.alert.common.model.CategoryItem;
 import com.synopsys.integration.alert.common.model.CategoryKey;
@@ -48,6 +53,8 @@ import com.synopsys.integration.alert.provider.blackduck.BlackDuckProviderConten
 import com.synopsys.integration.alert.workflow.filter.field.JsonExtractor;
 import com.synopsys.integration.alert.workflow.filter.field.JsonFieldAccessor;
 import com.synopsys.integration.blackduck.api.generated.enumeration.NotificationType;
+import com.synopsys.integration.blackduck.notification.content.ComponentVersionStatus;
+import com.synopsys.integration.blackduck.notification.content.PolicyInfo;
 
 @Component
 @Scope("prototype")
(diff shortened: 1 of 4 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
```diff
@@ -62,6 +70,63 @@ public BlackDuckPolicyMessageContentCollector(final JsonExtractor jsonExtractor,
 
     @Override
     protected void addCategoryItems(final List<CategoryItem> categoryItems, final JsonFieldAccessor jsonFieldAccessor, final List<JsonField<?>> notificationFields, final NotificationContent notificationContent) {
+        if (NotificationType.POLICY_OVERRIDE.name().equals(notificationContent.getNotificationType())) {
```
gkillough: Should we split this into two Collectors? PolicyOverride and PolicyRule

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
```diff
@@ -62,6 +70,63 @@ public BlackDuckPolicyMessageContentCollector(final JsonExtractor jsonExtractor,
...
+    private void processPolicy(final List<CategoryItem> categoryItems, final JsonFieldAccessor jsonFieldAccessor, final List<JsonField<?>> notificationFields, final NotificationContent notificationContent) {
+        final List<JsonField<PolicyInfo>> policyFields = getFieldsOfType(notificationFields, new TypeRef<PolicyInfo>() {});
+        final List<JsonField<ComponentVersionStatus>> componentFields = getFieldsOfType(notificationFields, new TypeRef<ComponentVersionStatus>() {});
+        try {
+            final List<PolicyInfo> policyItems = getFieldValueObjectsByLabel(jsonFieldAccessor, policyFields, BlackDuckProviderContentTypes.LABEL_POLICY_INFO_LIST);
+            final List<ComponentVersionStatus> componentVersionStatuses = getFieldValueObjectsByLabel(jsonFieldAccessor, componentFields, BlackDuckProviderContentTypes.LABEL_COMPONENT_VERSION_STATUS);
+            final Map<String, SortedSet<LinkableItem>> policyItemMap = mapPolicyToComponents(componentVersionStatuses);
+
+            final ItemOperation operation = getOperationFromNotification(notificationContent);
+            if (operation == null) {
```
gkillough: We should do this before the json parsing to save time if we do end up "short-circuiting".

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
```diff
@@ -72,47 +137,48 @@ protected void addCategoryItems(final List<CategoryItem> categoryItems, final Js
...
     }
 
-    private void addApplicableItems(final List<CategoryItem> categoryItems, final Long notificationId, final LinkableItem policyItem, final String policyUrl, final ItemOperation operation,
-        final Optional<LinkableItem> nameItem, final SortedSet<LinkableItem> applicableItems) {
+    private void addApplicableItems(final List<CategoryItem> categoryItems, final Long notificationId, final LinkableItem policyItem, final String policyUrl, final ItemOperation operation, final SortedSet<LinkableItem> applicableItems) {
+        final List<String> keyItems = applicableItems.stream().map(LinkableItem::getValue).collect(Collectors.toList());
+        String[] keyArray = new String[keyItems.size() + 1];
+        keyArray = keyItems.toArray(keyArray);
+        keyArray[keyArray.length - 1] = policyUrl;
+        final CategoryKey categoryKey = CategoryKey.from(CATEGORY_TYPE, keyArray);
```
gkillough: Would it be better to just add a new static initializer that takes a List<String>?

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
psantos1113: We should.  Given that we are trying to release the next version I would want to limit the changes in this version and enhance master with changes to create a separate collector for policy overrides.  Limit the number of changes in this branch.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyMessageContentCollector.java
```diff
@@ -62,6 +70,63 @@ public BlackDuckPolicyMessageContentCollector(final JsonExtractor jsonExtractor,
...
+
+            final ItemOperation operation = getOperationFromNotification(notificationContent);
+            if (operation == null) {
+                return;
+            }
+            for (final PolicyInfo policyItem : policyItems) {
+                final String policyUrl = policyItem.policy;
+                final String policyName = policyItem.policyName;
+                final LinkableItem policyLinkableItem = new LinkableItem(BlackDuckProviderContentTypes.LABEL_POLICY_NAME, policyName, policyUrl);
+                final SortedSet<LinkableItem> applicableItems = policyItemMap.get(policyUrl);
```
gkillough: Map.get() could return null (if the json we parsed is bad for some reason). Let's make sure we handle this.

PR #245, opened 2018-12-11
Audit table update

jamesrichard91:


src/main/java/com/synopsys/integration/alert/channel/event/DistributionEvent.java
```diff
@@ -23,12 +23,17 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
 import com.synopsys.integration.alert.common.event.ContentEvent;
 import com.synopsys.integration.alert.common.model.AggregateMessageContent;
 
 public abstract class DistributionEvent extends ContentEvent {
     private final Long commonDistributionConfigId;
-    private Long auditEntryId;
+    private Map<Long, Long> notificationIdToAuditId;
```
jamesrichard91: We have a map now because AggregateMessageContent can include multiple Notifications now, and we still want an audit for each of the notifications, otherwise when you resend a notification that was sent in a AggregateMessageContent with other notifications the 'Last Sent' time gets updated for all the notifications that were in that initial AggregateMessageContent

src/main/java/com/synopsys/integration/alert/web/model/NotificationContentConverter.java
```diff
@@ -49,8 +56,8 @@ public Config getConfigFromJson(final String json) {
     @Override
     public DatabaseEntity populateEntityFromConfig(final Config restModel) {
         final NotificationConfig notificationConfig = (NotificationConfig) restModel;
-        final Date createdAt = getContentConverter().getValue(notificationConfig.getCreatedAt(), Date.class);
```
jamesrichard91: This way of converting String to Date throws exceptions. The method is apparently never called which is why we didn't catch it sooner

src/main/js/component/common/RefreshTableCellFormatter.js
```diff
@@ -1,16 +1,17 @@
...
 
-const RefreshTableCellFormatter = ({ handleButtonClicked, currentRowSelected }) => (
+const RefreshTableCellFormatter = ({handleButtonClicked, currentRowSelected}) => (
     <button
         className="btn btn-link editJobButton"
         type="button"
         title="Refresh"
-        onClick={() => {
+        onClick={(e) => {
+            e.stopPropagation();
```
jamesrichard91: This is needed to stop the modal from popping up when you try to resend the notification

PR #246, opened 2018-12-13
Ps ldap authentication

psantos1113:
# Pull Request template

* Implement Ldap authentication
*
*


src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -0,0 +1,130 @@
...
+    private LdapAuthenticationProvider authenticationProvider;
+    private LdapProperties currentConfiguration;
+
+    @Autowired
+    public LdapManager(final LdapProperties ldapProperties) {
+        updateConfiguration(ldapProperties);
+    }
+
+    public boolean isLdapEnabled() {
+        return currentConfiguration == null ? false : currentConfiguration.isEnabled();
```
gkillough: I think this is one of those cases where the ternary op hurts readability. If the default is a boolean literal, we should probably do the more conventional && operation: 
`return currentConfiguration != null && currentConfiguration.isEnabled();`

src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public void updateContext() {
+        final LdapProperties configuration = getCurrentConfiguration();
+        final LdapContextSource ldapContextSource = new LdapContextSource();
+        ldapContextSource.setUrl(configuration.getServer());
+        ldapContextSource.setUserDn(configuration.getManagerDN());
+        ldapContextSource.setPassword(configuration.getManagerPassword());
+        ldapContextSource.setReferral(configuration.getLdapReferral());
+        ldapContextSource.setAuthenticationStrategy(createAuthenticationStrategy(configuration));
```
gkillough: What is the behavior when this is null?

PR #247, opened 2018-12-14
Java 11 and blackduck-common 40 (SNAPSHOT)

gkillough:


src/main/java/com/synopsys/integration/alert/channel/rest/ChannelRestConnectionFactory.java
```diff
@@ -51,72 +50,33 @@ public ChannelRestConnectionFactory(final AlertProperties alertProperties) {
...
-    private URL getUrlFromString(final String apiUrl) {
-        URL url = null;
-        try {
-            url = new URL(apiUrl);
-        } catch (final MalformedURLException e) {
-            logger.error("Problem generating the URL: " + apiUrl, e);
-        }
-        return url;
+    public RestConnection createRestConnection(final IntLogger intLogger, final int timeout) {
+        final Optional<Boolean> alertTrustCertificate = alertProperties.getAlertTrustCertificate();
```
bamandel: Instead of allowing this to be set to true, should we just default it to false since our stance is they should import their certs?

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProperties.java
```diff
@@ -114,11 +114,12 @@ public Integer getBlackDuckTimeout() {
         return Optional.empty();
     }
 
-    public HubServicesFactory createBlackDuckServicesFactory(final BlackduckRestConnection restConnection, final IntLogger logger) {
-        return new HubServicesFactory(HubServicesFactory.createDefaultGson(), HubServicesFactory.createDefaultJsonParser(), restConnection, logger);
+    public BlackDuckServicesFactory createBlackDuckServicesFactory(final BlackDuckRestConnection restConnection, final IntLogger logger) {
+        final Gson gson = BlackDuckServicesFactory.createDefaultGson();
```
bamandel: Should we create a defaultGson or autowire our version?

jamesrichard91: We should update the Gson that we create as a Bean in Application if this is the Gson we want to use, and then autowire it in here to be used

src/main/java/com/synopsys/integration/alert/channel/rest/ChannelRestConnectionFactory.java
psantos1113: I would leave this here.

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProperties.java
psantos1113: agreed we should autowire it in.

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckProviderDescriptorActionApi.java
```diff
@@ -74,9 +71,9 @@ public void validateConfig(final Config config, final Map<String, String> fieldE
         }
         if (StringUtils.isNotBlank(blackDuckConfig.getBlackDuckApiKey())) {
             if (blackDuckConfig.getBlackDuckApiKey().length() < 64) {
-                fieldErrors.put("blackDuckApiKey", "Not enough characters to be a Hub API Key.");
+                fieldErrors.put("blackDuckApiKey", "Not enough characters to be a Black Duck API Key.");
```
psantos1113: We should just state that the API token is invalid for both cases.  We might be giving information that's too specific.

src/test/java/com/synopsys/integration/alert/channel/rest/RestDistributionChannelTest.java
```diff
@@ -44,16 +44,16 @@
 import com.synopsys.integration.rest.request.Request;
 
 public class RestDistributionChannelTest extends ChannelTest {
-    @Test
+    // TODO figure out what this was supposed to be testing:
```
psantos1113: This was test was to increase the test coverage of RestDistributionChannel to 100%. In hindsight testing the other channels like HipChat and Slack which extend RestDistributionChannel should be modified to get us to 100% code coverage for this class as well.  This test we may be able to delete.

PR #254, opened 2019-01-03
Descriptor UI Controller

gkillough:


src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/DescriptorMetadata.java
```diff
@@ -26,24 +26,31 @@
...
+    private final String name;
+    private final DescriptorType type;
+    private final ConfigContextEnum context;
     private final String fontAwesomeIcon;
     private final boolean automaticallyGenerateUI;
     private List<ConfigField> fields;
 
-    public UIComponent(final String label, final String urlName, final String descriptorName, final String fontAwesomeIcon, final List<ConfigField> fields) {
-        this(label, urlName, descriptorName, fontAwesomeIcon, true, fields);
+    public DescriptorMetadata(final String label, final String urlName, final String name, final DescriptorType type, final ConfigContextEnum context, final String fontAwesomeIcon, final List<ConfigField> fields) {
```
jamesrichard91: DescriptorMetadata seems too generic, would you be against DescriptorFieldMetadata?

src/main/java/com/synopsys/integration/alert/web/controller/metadata/DescriptorController.java
```diff
@@ -0,0 +1,100 @@
...
+    @Autowired
+    public DescriptorController(final Collection<Descriptor> descriptors) {
+        this.descriptors = descriptors;
+    }
+
+    @GetMapping(DESCRIPTORS_PATH)
+    public Set<DescriptorMetadata> getDescriptors(@RequestParam(required = false) final String name, @RequestParam(required = false) final String type, @RequestParam(required = false) final String context) {
+        Set<Descriptor> filteredDescriptors = filter(descriptors, Descriptor::hasUIConfigs);
+        if (name != null) {
+            filteredDescriptors = filter(filteredDescriptors, descriptor -> descriptor.getName().equalsIgnoreCase(name));
```
jamesrichard91: Would it be better to gather the predicates and then filter the list  rather than streaming over the list for every filter?

src/main/java/com/synopsys/integration/alert/web/controller/metadata/DescriptorController.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        final ConfigContextEnum contextEnum = EnumUtils.getEnumIgnoreCase(ConfigContextEnum.class, context);
+        if (contextEnum != null) {
+            filteredDescriptors = filter(filteredDescriptors, descriptor -> descriptor.hasUIConfigForType(contextEnum));
+        }
+        return generateUIComponents(filteredDescriptors, contextEnum);
+    }
+
+    private Set<Descriptor> filter(final Collection<Descriptor> descriptors, final Predicate<Descriptor> predicate) {
```
jamesrichard91: Should this take a collection of predicates instead?

src/main/java/com/synopsys/integration/alert/web/controller/metadata/DescriptorController.java
gkillough: Good idea

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionUIConfig.java
```diff
@@ -32,15 +32,17 @@
...
+import com.synopsys.integration.alert.common.enumeration.DescriptorType;
 
 @Component
 public class EmailDistributionUIConfig extends UIConfig {
 
     @Override
-    public UIComponent generateUIComponent() {
-        return new UIComponent("Email", "email", EmailGroupChannel.COMPONENT_NAME, "envelope", setupFields());
+    public DescriptorMetadata generateDescriptorMetadata() {
+        return new DescriptorMetadata("Email", "email", EmailGroupChannel.COMPONENT_NAME, DescriptorType.CHANNEL, ConfigContextEnum.DISTRIBUTION, "envelope", setupFields());
```
bamandel: A lot of this info is in the Descriptor that references this. Is there a way we can populate that data when this is called through the descriptor?

src/main/java/com/synopsys/integration/alert/web/controller/UIComponentController.java
```diff
@@ -72,48 +72,48 @@ public UIComponentController(final DescriptorMap descriptorMap) {
...
+            return descriptorMap.getDescriptorMetadataList(contextType);
         } else {
             return descriptorMap.getAllUIComponents();
         }
     }
 
     // TODO split out provider and distribution endpoints
     @GetMapping("/distribution")
-    public UIComponent getDistributionUIComponent(@RequestParam(value = "providerName", required = true) final String providerName, @RequestParam(value = "channelName", required = true) final String channelName) {
+    public DescriptorMetadata getDistributionUIComponent(@RequestParam(value = "providerName", required = true) final String providerName, @RequestParam(value = "channelName", required = true) final String channelName) {
```
bamandel: I believe there will be a time where you hit this endpoint and only know the channel name and not the provider name. So we might want to support that scenario.

gkillough: I don't know if we will use this class going forward. We might only want to expose metadata in terms of descriptors.

src/main/java/com/synopsys/integration/alert/web/controller/metadata/DescriptorController.java
psantos1113: Are you suggesting to create a list of predicates pass it into this method and before the return statement create a single predicate object that is an and() of all the predicates in the list?

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionUIConfig.java
psantos1113: you'd end up with a circular dependency since the descriptor requires this object.

src/main/java/com/synopsys/integration/alert/web/controller/UIComponentController.java
```diff
@@ -72,48 +74,48 @@ public UIComponentController(final DescriptorMap descriptorMap) {
                     final ConfigContextEnum contextType = EnumUtils.getEnum(ConfigContextEnum.class, context);
                     final UIConfig uiConfig = descriptor.getUIConfig(contextType);
                     if (uiConfig != null) {
-                        return Arrays.asList(uiConfig.generateUIComponent());
+                        return Arrays.asList(uiConfig.generateDescriptorMetadata());
```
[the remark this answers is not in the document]
gkillough: This class will be deleted soon, so I'm not gonna worry about it.

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionUIConfig.java
gkillough: We could create this from the descriptor and throw away the UIConfig classes.

PR #255, opened 2019-01-03
Audit endpoints

jamesrichard91:


src/main/java/com/synopsys/integration/alert/web/audit/AuditEntryActions.java
```diff
@@ -113,8 +113,31 @@ public AuditEntryModel get(final Long id) {
         return null;
     }
 
-    public AlertPagedModel<AuditEntryModel> resendNotification(final Long notificationdId, final Long commonConfigId) throws IntegrationException {
-        final NotificationContent notificationContent = notificationManager.findById(notificationdId).orElseThrow(() -> new AlertNotificationPurgedException("No notification with this id exists."));
+    public JobAuditModel getAuditInfoForJob(final Long jobId) {
+        if (jobId != null) {
```
gkillough: What defines jobId? @bamandel may have changed this concept.

bamandel: JobId will be the configurationId

src/main/java/com/synopsys/integration/alert/web/audit/AuditEntryActions.java
```diff
@@ -113,8 +113,31 @@ public AuditEntryModel get(final Long id) {
...
+                    timeLastSent = notificationContentConverter.getContentConverter().getStringValue(auditEntryEntity.getTimeLastSent());
+                }
+                String status = null;
+                if (null != auditEntryEntity.getStatus()) {
+                    status = auditEntryEntity.getStatus();
+                }
+                return new JobAuditModel(timeCreated, timeLastSent, status);
+            }
+        }
+        return null;
```
psantos1113: If we're returning null here should this method return an Optional and return an Optional.empty().  The controller would need to change to return an Optional<JobAuditModel> but what I have initially researched is that the controller returning an Optional.empty will return a 404 that the resource cannot be found which is valid here given that the jobId is a path variable indicating a specific resource.

psantos1113: Would need to test the Optional returned from the controller to be sure it returns a 404

PR #256, opened 2019-01-03
Config endpoint

bamandel:
# Pull Request template

* Tests to verify Config endpoints work

* Added some constructors to models
* Fixed an issue with boolean logic
* Added tests


src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -27,52 +27,64 @@
...
     }
 
-    public List<FieldModel> getConfigs(final @RequestParam ConfigContextEnum context, @RequestParam(required = false) final String descriptorName) {
+    @GetMapping("/{context}")
+    public List<FieldModel> getConfigs(final @PathVariable ConfigContextEnum context, @RequestParam(required = false) final String descriptorName) {
         return controllerHandler.getConfigs(context, descriptorName);
     }
 
+    @GetMapping
     public FieldModel getConfig(final Long id) {
```
psantos1113: I think we want this to be a path variable since it is a specific resource.  Right now from looking at the test the id is a query parameter.  We want it to be a path variable so that the path is /api/configuration/{id} to get a single resource

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -27,52 +27,64 @@
...
     }
 
-    public ResponseEntity<String> validateConfig(@RequestBody(required = false) final FieldModel restModel) {
+    @PostMapping("/validate")
+    public ResponseEntity<String> validateConfig(@RequestBody(required = true) final FieldModel restModel) {
         return controllerHandler.validateConfig(restModel);
     }
 
+    @DeleteMapping
     public ResponseEntity<String> deleteConfig(final Long id) {
```
psantos1113: The id here should be a path variable like the comment above on the get method.

src/test/java/com/synopsys/integration/alert/util/DatabaseConfiguredFieldTest.java
```diff
@@ -8,58 +8,29 @@
...
-import com.synopsys.integration.alert.common.descriptor.DescriptorMap;
 import com.synopsys.integration.alert.common.enumeration.ConfigContextEnum;
 import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
 import com.synopsys.integration.alert.database.api.configuration.ConfigurationAccessor.ConfigurationModel;
 import com.synopsys.integration.alert.database.api.configuration.ConfigurationFieldModel;
 
 public abstract class DatabaseConfiguredFieldTest extends AlertIntegrationTest {
 
-    private final List<Descriptor> descriptors;
+    private final List<Descriptor> descriptors = new ArrayList<>();
```
psantos1113: If the size of the list isn't known or can grow to a large arbitrary size think about using LinkedList to avoid the expense of array copies if the list hits the threshold and need to be resized.  If we know the size of the list or can calculate it then ArrayList is good.  This is just a general comment.  The number of descriptors we currently have is less than the default size of an ArrayList.

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -27,52 +27,64 @@
...
 
-    // TODO Change the endpoint to have configuration/context/type/descriptorName/id
-
     @Autowired
     public ConfigController(final ConfigControllerHandler controllerHandler) {
         this.controllerHandler = controllerHandler;
     }
 
-    public List<FieldModel> getConfigs(final @RequestParam ConfigContextEnum context, @RequestParam(required = false) final String descriptorName) {
+    @GetMapping("/{context}")
```
gkillough: Anything that does not uniquely identify a resource by itself should be a RequestParam rather than a PathVariable.

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -27,52 +27,64 @@
...
     }
 
+    @GetMapping
     public FieldModel getConfig(final Long id) {
         return controllerHandler.getConfig(id);
     }
 
-    public ResponseEntity<String> postConfig(@RequestBody(required = false) final FieldModel restModel, @RequestParam final ConfigContextEnum context) {
+    @PostMapping
+    public ResponseEntity<String> postConfig(@RequestBody(required = true) final FieldModel restModel, @RequestParam final ConfigContextEnum context) {
```
gkillough: Be careful with enums in RequestParams, if the string provided is invalid, Spring will throw an exception and return a 400.

src/main/java/com/synopsys/integration/alert/web/model/FieldValueModel.java
```diff
@@ -48,8 +53,8 @@ public void setValues(final Collection<String> values) {
         return values.stream().findFirst();
     }
 
-    public void setValue(String value) {
-        this.values = Arrays.asList(value);
+    public void setValue(final String value) {
+        values = Arrays.asList(value);
```
gkillough: Let's use Set.of(value) for the singleton case. I know you didn't add the Arrays.asList(), but as these cases come to light, we should convert them to the Java 11 standard (let's also use Set.of() for the constructor.

src/main/java/com/synopsys/integration/alert/web/model/FieldModel.java
```diff
@@ -35,6 +36,16 @@
     private final String descriptorName;
     private final String context;
 
+    public FieldModel() {
+        this("", "", new HashMap<>());
```
gkillough: What is this case for? I'm a little wary of initializing a String that represents an enum to anything other than a valid value or null.

bamandel: Blank constructor is for Gson serialization. I made these private and inserted null instead of values.

PR #261, opened 2019-01-09
Updating the Email channel

jamesrichard91:


src/main/java/com/synopsys/integration/alert/channel/email/EmailAddressHandler.java
```diff
@@ -0,0 +1,122 @@
...
+        this.blackDuckProjectRepositoryAccessor = blackDuckProjectRepositoryAccessor;
+        this.userProjectRelationRepositoryAccessor = userProjectRelationRepositoryAccessor;
+        this.blackDuckUserRepositoryAccessor = blackDuckUserRepositoryAccessor;
+    }
+
+    public FieldAccessor updateEmailAddresses(final String provider, final AggregateMessageContent content, final FieldAccessor originalAccessor) {
+        if (StringUtils.isBlank(provider)) {
+            return originalAccessor;
+        }
+        if (BlackDuckProvider.COMPONENT_NAME.equals(provider)) {
```
bamandel: Might be easier to comprehend in the future if we separate provider specific content out into different classes or at least separate methods. Or better yet, we can add the option for email functionality to providers then get the provider and call their own functionality.

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionDescriptorActionApi.java
```diff
@@ -100,7 +101,7 @@ public TestConfigModel createTestConfigModel(final FieldModel fieldModel, final
             blackDuckProjectEntities
                 .stream()
                 .forEach(project -> {
-                    final Set<String> emailsForProject = emailGroupChannel.getEmailAddressesForProject(project, projectOwnerOnly);
+                    final Set<String> emailsForProject = emailAddressHandler.getBlackDuckEmailAddressesForProject(project, projectOwnerOnly);
```
bamandel: This call will need to be specific to provider in the future, correct? If we require a provider to define its own email address functionality, then we can call the provider method here as well.

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionDescriptorActionApi.java
```diff
@@ -72,56 +73,66 @@ public void validateChannelConfig(final FieldAccessor fieldAccessor, final Map<S
...
-                                           .collect(Collectors.toSet());
+        if (providerName.isPresent() && BlackDuckProvider.COMPONENT_NAME.equals(providerName.get())) {
+            Set<BlackDuckProjectEntity> blackDuckProjectEntities = null;
+            if (filterByProject.isPresent() && BooleanUtils.toBoolean(filterByProject.get())) {
+                final Set<String> configuredProjects = fieldModel.getField(BlackDuckDescriptor.KEY_CONFIGURED_PROJECT).getValues().stream().collect(Collectors.toSet());
+                final Optional<String> projectNamePattern = fieldModel.getField(BlackDuckDescriptor.KEY_PROJECT_NAME_PATTERN).getValue();
+                blackDuckProjectEntities = blackDuckProjectRepositoryAccessor.readEntities()
+                                               .stream()
+                                               .map(databaseEntity -> (BlackDuckProjectEntity) databaseEntity)
+                                               .filter(databaseEntity -> doesProjectNameMatchThePattern(databaseEntity.getName(), projectNamePattern.orElse(""))
```
gkillough: I think we should abstract this filter into its own method. It feels like there is way too much happening in one stream operation. Once abstracted, we could call it like this: `this::isProjectConfigured`

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionDescriptorActionApi.java
```diff
@@ -72,56 +73,66 @@ public void validateChannelConfig(final FieldAccessor fieldAccessor, final Map<S
...
-                    if (emailsForProject.isEmpty()) {
-                        projectsWithoutEmails.add(project.getName());
+            }
+            if (null != blackDuckProjectEntities) {
+                final Optional<String> projectOwnerOnlyOptional = fieldModel.getField(EmailDescriptor.KEY_PROJECT_OWNER_ONLY).getValue();
+                final Boolean projectOwnerOnly = Boolean.parseBoolean(projectOwnerOnlyOptional.orElse("false"));
+                final Set<String> projectsWithoutEmails = new HashSet<>();
+                blackDuckProjectEntities
+                    .stream()
+                    .forEach(project -> {
```
gkillough: I think we want to avoid using `Collection.forEach` for non-trivial operations. As a side note, I don't think `.stream().forEach()` buys us anything over a `.forEach()` without a stream unless we are going from an "ordered" collection and the inner logic doesn't care about order.

PR #267, opened 2019-01-15
Config field validation

psantos1113:

* Changes to implement validation per field defined in the ConfigField objects.
*
*

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
+++ b/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
@@ -91,13 +91,11 @@ public class ConfigField extends Stringable {
     Collection<String> validate(final FieldValueModel fieldToValidate, final FieldModel fieldModel, final List<BiFunction<FieldValueModel, FieldModel, Collection<String>>> validationFunctions) {
         final Collection<String> errors = new LinkedList<>();
         errors.addAll(validateRequiredField(fieldToValidate));
-        final boolean performValidation = errors.isEmpty() && (!fieldToValidate.isSet() || fieldToValidate.hasValues());
+        final boolean performValidation = errors.isEmpty() && fieldToValidate.hasValues();
         if (performValidation) {
-            if (fieldToValidate.hasValues()) {
-                for (final BiFunction<FieldValueModel, FieldModel, Collection<String>> validation : validationFunctions) {
-                    if (null != validation) {
-                        errors.addAll(validation.apply(fieldToValidate, fieldModel));
-                    }
+            for (final BiFunction<FieldValueModel, FieldModel, Collection<String>> validation : validationFunctions) {
+                if (null != validation) {
+                    errors.addAll(validation.apply(fieldToValidate, fieldModel));
                 }
             }
         }


src/main/java/com/synopsys/integration/alert/common/descriptor/config/context/DescriptorActionApi.java
```diff
@@ -41,13 +46,44 @@
...
 
-    public abstract void validateConfig(final FieldModel fieldModel, final Map<String, String> fieldErrors);
+    public static final String REQUIRED_FIELD_MISSING = "Required field missing";
+
+    public void validateConfig(final Collection<ConfigField> descriptorFields, final FieldModel fieldModel, final Map<String, String> fieldErrors) {
+        for (final ConfigField field : descriptorFields) {
+            final String fieldKey = field.getKey();
+            final Optional<FieldValueModel> optionalField = fieldModel.getField(fieldKey);
+            if (field.isRequired()) {
+                if (optionalField.isEmpty()) {
```
gkillough: For this outer if-statement, I would be more comfortable doing an `.isPresent()` check and before a `.get()` in case the code ever changes. We should also minimize the number of times we do the get by storing the value from the Optional in a local variable.

src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -34,27 +41,66 @@
...
+        }
+        return validate(fieldToValidate, fieldModel, validationFunctions);
+    }
+
+    Collection<String> validate(final FieldValueModel fieldToValidate, final FieldModel fieldModel, final List<BiFunction<FieldValueModel, FieldModel, Collection<String>>> validationFunctions) {
+        final Collection<String> errors = new LinkedList<>();
+        final boolean performValidation = !fieldToValidate.isSet() || fieldToValidate.hasValues();
+        if (performValidation) {
+            if (fieldToValidate.hasValues()) {
+                for (final BiFunction<FieldValueModel, FieldModel, Collection<String>> validation : validationFunctions) {
```
gkillough: How does the default validation make it into this? Or do we not want the default validation if someone providers alternative(s)?

src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -34,27 +41,66 @@
...
+        this(key, label, type, required, sensitive, group, "", null);
+    }
+
+    public ConfigField(final String key, final String label, final String type, final boolean required, final boolean sensitive, final FieldGroup group, final BiFunction<FieldValueModel, FieldModel, Collection<String>> validationFunction) {
+        this(key, label, type, required, sensitive, group, "", validationFunction);
     }
 
     public ConfigField(final String key, final String label, final String type, final boolean required, final boolean sensitive, final String subGroup) {
-        this(key, label, type, required, sensitive, FieldGroup.DEFAULT, subGroup);
+        this(key, label, type, required, sensitive, FieldGroup.DEFAULT, subGroup, null);
```
gkillough: I would feel better about a storing a private constant with a lambda to pass here instead of null. If we are already going to be applying validation to fields, we could do something like this: 
`private static final BiFunction<FieldValueModel, FieldModel, Collection<String>> NO_VALIDATION = (x, y) -> List.of();`
I'm not 100% clear on how we want to do validation with default options, but something along these lines seems safer, though it will add more operations.

src/main/java/com/synopsys/integration/alert/common/descriptor/config/context/DescriptorActionApi.java
```diff
@@ -41,13 +46,44 @@
 
 public abstract class DescriptorActionApi {
 
-    public abstract void validateConfig(final FieldModel fieldModel, final Map<String, String> fieldErrors);
+    public static final String REQUIRED_FIELD_MISSING = "Required field missing";
+
+    public void validateConfig(final Collection<ConfigField> descriptorFields, final FieldModel fieldModel, final Map<String, String> fieldErrors) {
+        for (final ConfigField field : descriptorFields) {
```
gkillough: Since we know what field this is, I think we can get a little more explicit about which required field is missing. Then we don't have to do extra validation operations in the various UIConfig classes with `.StringUtils.isBlank(...)`. Or am I missing something?

src/main/java/com/synopsys/integration/alert/component/scheduling/SchedulingUIConfig.java
```diff
@@ -46,9 +50,33 @@ public SchedulingUIConfig() {
 
     @Override
     public List<ConfigField> createFields() {
-        final ConfigField digestHour = SelectConfigField.createRequired(KEY_DAILY_DIGEST_HOUR_OF_DAY, "Daily digest hour of day", Arrays.asList("1", "2", "3", "4", "5", "6", "7", "8", "10", "11", "12"));
-        final ConfigField purgeFrequency = SelectConfigField.createRequired(KEY_PURGE_DATA_FREQUENCY_DAYS, "Purge data frequency in days", Arrays.asList("1", "2", "3"));
+        final ConfigField digestHour = SelectConfigField.createRequired(KEY_DAILY_DIGEST_HOUR_OF_DAY, "Daily digest hour of day", Arrays.asList("1", "2", "3", "4", "5", "6", "7", "8", "10", "11", "12"), this::validateDigestHourOfDay);
```
gkillough: Might as well change this to `List.of(...)` while we're in here.

src/main/java/com/synopsys/integration/alert/common/descriptor/config/context/DescriptorActionApi.java
```diff
@@ -41,13 +46,45 @@
...
+                        }
+                    } else {
+                        if (!fieldValueModel.isSet()) {
+                            fieldErrors.put(fieldKey, REQUIRED_FIELD_MISSING);
+                        }
+                    }
+                } else {
+                    fieldErrors.put(fieldKey, REQUIRED_FIELD_MISSING);
+                }
+            }
```
bamandel: I believe we can pass the optional FieldValueModel to the ConfigField validate method and have an automatic required test there instead of here. This method I imagine would be more focused on building the error map from the results.

src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -23,38 +23,81 @@
...
+
+    public ConfigField(final String key, final String label, final String type, final boolean required, final boolean sensitive, final BiFunction<FieldValueModel, FieldModel, Collection<String>> validationFunction) {
+        this(key, label, type, required, sensitive, FieldGroup.DEFAULT, "", validationFunction);
+    }
+
+    public Collection<String> validate(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
+        return validate(fieldToValidate, fieldModel, List.of(validationFunction));
+    }
+
+    Collection<String> validate(final FieldValueModel fieldToValidate, final FieldModel fieldModel, final List<BiFunction<FieldValueModel, FieldModel, Collection<String>>> validationFunctions) {
```
bamandel: Should this be public/private/protected? Or did we intentionally leave it blank?

src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -23,38 +23,81 @@
...
+        this(key, label, type, required, sensitive, FieldGroup.DEFAULT, "", NO_VALIDATION);
+    }
+
+    public ConfigField(final String key, final String label, final String type, final boolean required, final boolean sensitive, final BiFunction<FieldValueModel, FieldModel, Collection<String>> validationFunction) {
+        this(key, label, type, required, sensitive, FieldGroup.DEFAULT, "", validationFunction);
+    }
+
+    public Collection<String> validate(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
+        return validate(fieldToValidate, fieldModel, List.of(validationFunction));
+    }
```
bamandel: We should have an abstract method that is ran after we determine if the field should be required and after we run the passed function. This would help force a contract with child classes

src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
```diff
@@ -44,9 +48,18 @@
...
+            Collectors.toList()), this::validateNotificationTypes);
         final ConfigField formatField = SelectConfigField.createRequired(KEY_FORMAT_TYPE, "Format", provider.getSupportedFormatTypes().stream().map(FormatType::name).collect(Collectors.toList()));
 
         return Arrays.asList(notificationTypesField, formatField);
     }
+
+    private Collection<String> validateNotificationTypes(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
+        final Collection<String> notificationTypes = Optional.ofNullable(fieldToValidate.getValues()).orElse(List.of());
+        if (notificationTypes == null || notificationTypes.size() <= 0) {
+            return List.of("Must have at least one notification type.");
```
bamandel: Shouldn't this be covered by the required check?

src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/CommonDistributionUIConfig.java
```diff
@@ -23,33 +23,77 @@
...
+                                                       .anyMatch(configurationFieldModel -> jobName.equals(configurationFieldModel.getFieldValue().get()));
+                if (foundDuplicateName) {
+                    errors.add("A distribution configuration with this name already exists.");
+                }
+            } catch (final AlertDatabaseConstraintException e) {
+                logger.error("Could not retrieve distributions of {}", jobName);
+            }
+
+        } else {
+            errors.add("Name cannot be blank.");
```
bamandel: This part I think should be handled by the required check.

src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/CommonDistributionUIConfig.java
```diff
@@ -23,33 +23,77 @@
...
+        final String jobName = fieldToValidate.getValue().orElse(null);
+        if (StringUtils.isNotBlank(jobName)) {
+            try {
+                final List<ConfigurationModel> configurations = configurationAccessor.getConfigurationsByDescriptorName(descriptorName);
+                final Boolean foundDuplicateName = configurations.stream()
+                                                       .map(configurationModel -> configurationModel.getField(CommonDistributionUIConfig.KEY_NAME).orElse(null))
+                                                       .filter(configurationFieldModel -> (null != configurationFieldModel) && configurationFieldModel.getFieldValue().isPresent())
+                                                       .anyMatch(configurationFieldModel -> jobName.equals(configurationFieldModel.getFieldValue().get()));
+                if (foundDuplicateName) {
+                    errors.add("A distribution configuration with this name already exists.");
```
bamandel: Does this only check duplicates between descriptors? Is this the check we want to do? Or do we want to check all configurations of a specific context type? I don't think we want to check a single descriptor.

PR #269, opened 2019-01-17
Ps apply ldap settings

psantos1113:

* Implement the changes to use the settings configuration for LDAP rather than LDAP properties.
* Added logging to have visibility if environment variables for alert configuration data is set or not


src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -38,64 +42,80 @@
...
-    public LdapProperties getCurrentConfiguration() {
-        return currentConfiguration;
+        return enabled;
     }
 
-    public LdapAuthenticationProvider getAuthenticationProvider() throws AlertLDAPConfigurationException {
-        if (authenticationProvider == null) {
-            updateConfiguration(currentConfiguration);
+    public ConfigurationModel getCurrentConfiguration() throws AlertDatabaseConstraintException, AlertLDAPConfigurationException {
+        final List<ConfigurationModel> configurations = configurationAccessor.getConfigurationsByDescriptorName(SettingsDescriptor.SETTINGS_COMPONENT);
```
gkillough: This could become: 
```
return configurationAccessor.getConfigurationsByDescriptorName(SettingsDescriptor.SETTINGS_COMPONENT)
.stream()
.findFirst()
.orElseThrow(() -> new AlertLDAPConfigurationException("more than 1 settings configuration present"));
```

src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -108,39 +128,49 @@ private DirContextAuthenticationStrategy createAuthenticationStrategy(final Ldap
...
         return authenticator;
     }
 
-    private LdapAuthoritiesPopulator createAuthoritiesPopulator() {
-        final DefaultLdapAuthoritiesPopulator authoritiesPopulator = new DefaultLdapAuthoritiesPopulator(contextSource, currentConfiguration.getGroupSearchBase());
-        authoritiesPopulator.setGroupSearchFilter(currentConfiguration.getGroupSearchFilter());
-        authoritiesPopulator.setGroupRoleAttribute(currentConfiguration.getGroupRoleAttribute());
-        authoritiesPopulator.setRolePrefix(currentConfiguration.getRolePrefix());
+    private String[] createSetFromCSV(final String commaSeparatedString) {
+        return StringUtils.split(commaSeparatedString, ",");
```
jamesrichard91: This is a little confusing since the method name implies it is creating a set not an array

src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -38,64 +42,80 @@
...
-        final LdapContextSource ldapContextSource = new LdapContextSource();
-
         try {
-            if (StringUtils.isNotBlank(configuration.getServer())) {
-                ldapContextSource.setUrl(configuration.getServer());
-                ldapContextSource.setUserDn(configuration.getManagerDN());
-                ldapContextSource.setPassword(configuration.getManagerPassword());
-                ldapContextSource.setReferral(configuration.getLdapReferral());
-                ldapContextSource.setAuthenticationStrategy(createAuthenticationStrategy(configuration));
+            if (isLdapEnabled()) {
```
gkillough: We might want to consider making this short-circuit since nothing will get executed if this  returns false. I don't know if we are trying to minimize the number of return paths, but nesting a lot of code can cause readability issues when unnecessary.

src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -117,11 +126,15 @@ private String convertKeyToPropery(final String descriptorName, final String key
...
+        String found = "No";
         String value = System.getProperty(propertyKey);
         if (StringUtils.isBlank(value)) {
-            logger.debug("Not found in system env, checking Spring env");
             value = environment.getProperty(propertyKey);
+            if (environment.containsProperty(propertyKey)) {
+                found = "Yes";
+            }
         }
+        logger.info("  {}: {}", propertyKey, found);
```
jamesrichard91: Do we need to keep this logging?

src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -38,64 +42,80 @@
...
     }
 
-    private DirContextAuthenticationStrategy createAuthenticationStrategy(final LdapProperties configuration) {
-        final String authenticationType = configuration.getAuthenticationType();
+    private String getFieldValueOrEmpty(final ConfigurationModel configurationModel, final String fieldKey) {
+        return configurationModel.getField(fieldKey).flatMap(field -> field.getFieldValue()).orElse("");
+    }
+
+    private DirContextAuthenticationStrategy createAuthenticationStrategy(final ConfigurationModel configuration) {
+        final String authenticationType = getFieldValueOrEmpty(configuration, SettingsDescriptor.KEY_LDAP_AUTHENTICATION_TYPE);
```
bamandel: A couple time we're using getFieldValueOrEmpty and then checking if the variable is blank later on. Since we would normally get an optional (And no blank value should be stored in the values table), should we use the optional to check for isPresent instead?

src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -67,9 +69,16 @@ public AlertStartupInitializer(final DescriptorMap descriptorMap, final Environm
...
         final Set<String> descriptorNames = descriptorMap.getDescriptorMap().keySet();
+        logger.info("** --------------------------------- **");
+        logger.info("Initializing descriptors with environment variables...");
         for (final String descriptorName : descriptorNames) {
+            logger.info("---------------------------------");
+            logger.info("Descriptor: {}", descriptorName);
+            logger.info("---------------------------------");
             final Map<String, String> newConfiguration = new HashMap<>();
-            final List<DefinedFieldModel> fieldsForDescriptor = descriptorAccessor.getFieldsForDescriptor(descriptorName, ConfigContextEnum.GLOBAL);
+            final List<DefinedFieldModel> fieldsForDescriptor = new ArrayList<>(descriptorAccessor.getFieldsForDescriptor(descriptorName, ConfigContextEnum.GLOBAL));
```
bamandel: Can we create a TreeSet here instead and pass the collection through? If need be, we can modify the DefinedFieldModel to compare itself better (Compare on the key instead of everything)

src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -117,11 +126,15 @@ private String convertKeyToPropery(final String descriptorName, final String key
...
         String value = System.getProperty(propertyKey);
         if (StringUtils.isBlank(value)) {
-            logger.debug("Not found in system env, checking Spring env");
             value = environment.getProperty(propertyKey);
+            if (environment.containsProperty(propertyKey)) {
+                found = "Yes";
+            }
         }
+        logger.info("  {}: {}", propertyKey, found);
         return value;
```
gkillough: Should this return `Optional.ofNullable(value);`?

psantos1113: I will change that

src/main/java/com/synopsys/integration/alert/web/security/authentication/ldap/LdapManager.java
```diff
@@ -86,7 +83,9 @@ public LdapAuthenticationProvider getAuthenticationProvider() throws AlertLDAPCo
 
     public void updateContext() throws AlertLDAPConfigurationException {
         try {
-            if (isLdapEnabled()) {
+            if (!isLdapEnabled()) {
```
gkillough: If we're gonna use the else statement, we can go back to the old way. I just wanted to avoid nesting when there was no other code that executed otherwise. Sorry for the confusion.

PR #273, opened 2019-01-23
Ps apply proxy settings

psantos1113:
# Pull Request template

* Read the settings descriptor configuration for proxy settings
* Create a component to encapsulate the management of the proxy settings.
*


src/main/java/com/synopsys/integration/alert/common/ProxyManager.java
```diff
@@ -0,0 +1,115 @@
...
+public class ProxyManager {
+    private final BaseConfigurationAccessor configurationAccessor;
+
+    @Autowired
+    public ProxyManager(final BaseConfigurationAccessor configurationAccessor) {
+        this.configurationAccessor = configurationAccessor;
+    }
+
+    private ConfigurationModel getSettingsConfiguration() throws AlertRuntimeException, AlertDatabaseConstraintException {
+        return configurationAccessor.getConfigurationByDescriptorNameAndContext(SettingsDescriptor.SETTINGS_COMPONENT, ConfigContextEnum.GLOBAL)
```
bamandel: Since this method is accessing the DB every time we call a field, we're hitting the DB every time we need to access data from the same configuration. Why don't we have this be public or a new object that extends Configuration? Then we can add additional methods in ProxyManager to access the fields of that newly defined Configuration model.

src/main/java/com/synopsys/integration/alert/provider/polaris/PolarisProperties.java
```diff
@@ -94,7 +100,12 @@ public AccessTokenRestConnection createRestConnection(final IntLogger intLogger)
 
     public AccessTokenRestConnection createRestConnection(final IntLogger intLogger, final String baseUrl, final String accessToken, final Integer timeout) {
         final Boolean alwaysTrustCertificate = alertProperties.getAlertTrustCertificate().orElse(Boolean.FALSE);
-        final ProxyInfo proxyInfo = alertProperties.createProxyInfo();
+        ProxyInfo proxyInfo = ProxyInfo.NO_PROXY_INFO;
+        try {
+            proxyInfo = proxyManager.createProxyInfo();
+        } catch (AlertDatabaseConstraintException | AlertRuntimeException ex) {
+            logger.error("Error creating proxy information", ex);
+        }
```
bamandel: Since this seems to be in multiple locations, maybe we can create a restconnection wrapper for everything that automatically adds our Proxy stuff to that connection.

PR #274, opened 2019-01-23
Handler removal

bamandel:
# Pull Request template

*

* Handler's seemed like an unnecessary step in our web workflow. I have removed them.

* Moved handler handling to controllers
* Created a ResponseFactory class that builds responses based off of status


src/main/java/com/synopsys/integration/alert/web/controller/ResponseFactory.java
```diff
@@ -0,0 +1,83 @@
...
+        final String responseBody = new ResponseBodyBuilder(id, message).build();
+        return new ResponseEntity<>(responseBody, status);
+    }
+
+    public ResponseEntity<String> createResponse(final HttpStatus status, final Long id, final String message) {
+        return createResponse(status, String.valueOf(id), message);
+    }
+
+    public ResponseEntity<String> createResponse(final HttpStatus status, final String message) {
+        return createResponse(status, -1L, message);
```
psantos1113: Can we change the response body builder to not assign -1 for the id.  Add a check for a missing ID field and then have the ResponseBodyBuilder only add the id  if it is not the missing ID value.

PR #282, opened 2019-01-28
Bm group endpoints

bamandel:
# Pull Request template

*

* Adding new endpoints for saving jobs

* Pulled out common code in ConfigActions
* Added common fields to all appropriate descriptors


src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
```diff
@@ -54,9 +53,15 @@ public ProviderDistributionUIConfig(final String label, final String urlName, fi
             Collectors.toList()), this::validateNotificationTypes);
         final ConfigField formatField = SelectConfigField.createRequired(KEY_FORMAT_TYPE, "Format", provider.getSupportedFormatTypes().stream().map(FormatType::name).collect(Collectors.toList()));
 
-        return Arrays.asList(notificationTypesField, formatField);
+        final List<ConfigField> configFields = List.of(notificationTypesField, formatField);
+        final List<ConfigField> providerDistributionFields = createProviderDistributionFields();
+        providerDistributionFields.addAll(configFields);
```
gkillough: We have to be careful using `.addAll(...)` on lists we don't create. It could very easily be implemented as an unmodifiable list (e.g. `List.of(...)`).
An alternative could be:
```java
Streams.concat(providerDistributionFields, configFields).collect(Collectors.toList())
```

src/main/java/com/synopsys/integration/alert/web/config/FieldModelProcessor.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    public ConfigurationModel getSavedEntity(final Long id) throws AlertException {
+        if (null != id) {
+            final Optional<ConfigurationModel> configuration = configurationAccessor.getConfigurationById(id);
+            if (configuration.isPresent()) {
+                return configuration.get();
+            }
+        }
+        return null;
```
gkillough: If this method is supposed to be public, it should return an Optional.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
```diff
@@ -0,0 +1,210 @@
...
+    public ResponseEntity<String> getJob(UUID id) {
+        Optional<GroupedFieldModel> optionalModel;
+        try {
+            optionalModel = jobConfigActions.getJobById(id);
+        } catch (final AlertException e) {
+            logger.error(e.getMessage(), e);
+            return responseFactory.createResponse(HttpStatus.INTERNAL_SERVER_ERROR, "There was an issue retrieving data from the database for ID: " + id);
+        }
+
+        if (optionalModel.isEmpty()) {
```
gkillough: I think we should swap the `.isEmpty()` and `.isPresent()` lines.

src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
jamesrichard91: We should not be modifying the list that is provided as a parameter if we are returning the modified list. What we could do is return a new list and add the things we want to the new list

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -96,11 +96,11 @@ public ConfigController(final ConfigActions configActions, final ContentConverte
             return responseFactory.createNotFoundResponse("Configuration not found for the specified id");
         }
 
-        if (optionalModel.isPresent()) {
-            return new ResponseEntity<>(contentConverter.getJsonString(optionalModel.get()), HttpStatus.OK);
+        if (optionalModel.isEmpty()) {
```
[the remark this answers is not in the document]
jamesrichard91: Sonarqube does not yet recognize isEmpty() so we should favor isPresent() when possible

src/main/java/com/synopsys/integration/alert/web/config/FieldModelProcessor.java
```diff
@@ -0,0 +1,206 @@
...
+    public String testFieldModel(FieldModel fieldModel, String destination) throws IntegrationException {
+        final Optional<DescriptorActionApi> descriptorActionApi = retrieveDescriptorActionApi(fieldModel);
+        if (descriptorActionApi.isPresent()) {
+            final DescriptorActionApi descriptorApi = descriptorActionApi.get();
+            final TestConfigModel testConfig = descriptorApi.createTestConfigModel(fieldModel, destination);
+            // TODO look to see if FieldAccessor can be modified to allow mutable maps. FieldAccessor and destination can then be passed here
+            descriptorApi.testConfig(testConfig);
+            return "Successfully sent test message.";
+        } else {
+            return "Descriptor action api did not exist: " + fieldModel.getDescriptorName();
```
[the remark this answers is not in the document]
jamesrichard91: There is similar code elsewhere to do this. I think it makes sense to have this. It may not make sense to a User but if our testing is good they should never see this and it make development a lot easier

src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -33,40 +33,57 @@
...
         final ConfigField frequency = SelectConfigField.createRequired(KEY_FREQUENCY, "Frequency", Arrays.stream(FrequencyType.values()).map(type -> type.getDisplayName()).collect(Collectors.toList()), this::validateJobName);
-        final ConfigField channelName = SelectConfigField.createRequired(KEY_CHANNEL_NAME, "Channel Type", channelDescriptors);
-        final ConfigField providerName = SelectConfigField.createRequired(KEY_PROVIDER_NAME, "Provider Type", providerDescriptors);
+        final ConfigField channelName = SelectConfigField.createRequired(KEY_CHANNEL_NAME, "Channel Type", getDescriptorNames(DescriptorType.CHANNEL));
+        final ConfigField providerName = SelectConfigField.createRequired(KEY_PROVIDER_NAME, "Provider Type", getDescriptorNames(DescriptorType.PROVIDER));
 
-        return List.of(name, channelName, frequency, providerName);
+        final List<ConfigField> configFields = List.of(name, channelName, frequency, providerName);
+        final List<ConfigField> channelDistributionFields = createChannelDistributionFields();
+        channelDistributionFields.addAll(configFields);
```
psantos1113: Same comment here with the addAll as the ProviderUIConfig.  Don't use addAll here since you don't know if createChannelDistributionFields may return an immutable list.

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionDescriptorActionApi.java
```diff
@@ -23,20 +23,12 @@
...
 @Component
 public class BlackDuckDistributionDescriptorActionApi extends ProviderDistributionDescriptorActionApi {
-    private final ContentConverter contentConverter;
-
-    @Autowired
-    public BlackDuckDistributionDescriptorActionApi(final ContentConverter contentConverter) {
-        this.contentConverter = contentConverter;
-    }
 
     // TODO Add the Delete/update/save overrides here to customize what Blackduck does in each scenario(Modifying projects)
```
psantos1113: Is this still relevant?

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -96,11 +96,11 @@ public ConfigController(final ConfigActions configActions, final ContentConverte
...
+            return responseFactory.createNotFoundResponse("Configuration not found for the specified id");
         }
 
-        return responseFactory.createNotFoundResponse("Configuration not found for the specified id");
+        return new ResponseEntity<>(contentConverter.getJsonString(optionalModel.get()), HttpStatus.OK);
```
psantos1113: use the ResponseFactory here.  responseFactory.createOkContentResponse

src/main/java/com/synopsys/integration/alert/web/config/FieldModelProcessor.java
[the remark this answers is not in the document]
psantos1113: I would maybe change the message to "Failed to send test message:  Failed to find a component to process the request."
maybe change component to descriptor in the message. 
In general I don't like the idea of showing text in the UI that references descriptor or action api since those are server concepts in the product.  Might be leaking information about the internals.  A log message would be ok,  but not a message that shows up in the UI.  If it is implemented correctly it is a message they will never see.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    public GroupedFieldModel saveJob(GroupedFieldModel groupedFieldModel) throws AlertFieldException, AlertDatabaseConstraintException {
+        validateJob(groupedFieldModel);
+        Set<String> descriptorNames = new HashSet<>();
+        Set<ConfigurationFieldModel> configurationFieldModels = new HashSet<>();
+        for (FieldModel fieldModel : groupedFieldModel.getFieldModels()) {
+            descriptorNames.add(fieldModel.getDescriptorName());
+            final Collection<ConfigurationFieldModel> configurationFieldModelMap = fieldModelProcessor.saveFieldModel(fieldModel).values();
```
psantos1113: consider renaming the variable configurationFieldModelMap.  It isn't a map it's a collection. configurationFieldModels would work.

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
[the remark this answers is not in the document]
bamandel: I had forgotten. Will change it

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionDescriptorActionApi.java
bamandel: I was imagining we needed to delete some things on save/delete in projects and what not... but I now think that it is no longer necessary. Will remove

PR #283, opened 2019-01-29
Ps add env override flag

psantos1113:
# Pull Request template

* Remove the logging of the properties to occur in one location
* Add a checkbox to the settings page to determine if the environment variables should override or not
* Check the settings configuration and environment variable for the override flag first.
* List descriptor properties for settings first then the other descriptors in alphabetical order
* Update the copyright to 2019


src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -67,60 +66,112 @@ public AlertStartupInitializer(final DescriptorMap descriptorMap, final Environm
...
 
-    public void initializeConfigs(final boolean overwriteCurrentConfig) throws IllegalArgumentException, SecurityException, AlertDatabaseConstraintException {
-        final Set<String> descriptorNames = descriptorMap.getDescriptorMap().keySet();
+    public void initializeConfigs() throws IllegalArgumentException, SecurityException {
         logger.info("** --------------------------------- **");
         logger.info("Initializing descriptors with environment variables...");
+        boolean overwriteCurrentConfig = isEnvironmentOverrideEnabled();
+        logger.info("Environment variables override configuration: {}", overwriteCurrentConfig);
+        initializeConfiguration(List.of(SettingsDescriptor.SETTINGS_COMPONENT), overwriteCurrentConfig);
+        List<String> descriptorNames = descriptorMap.getDescriptorMap().keySet().stream().filter(key -> !key.equals(SettingsDescriptor.SETTINGS_COMPONENT)).sorted().collect(Collectors.toList());
```
gkillough: Does order matter when initializing configs? I suspect it might, but I'm not sure `.sorted()` will do it in a meaningful way unless we have specified how to compare them. If order doesn't matter, let's remove the sort to improve performance, otherwise let's make sure the sort is doing something meaningful.

src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -67,60 +66,112 @@ public AlertStartupInitializer(final DescriptorMap descriptorMap, final Environm
...
+        }
+        return environmentOverride;
+    }
+
+    private Optional<ConfigurationModel> findSettingsConfiguration() throws AlertDatabaseConstraintException {
+        final List<ConfigurationModel> settingsConfigurationModels = fieldConfigurationAccessor.getConfigurationByDescriptorNameAndContext(SettingsDescriptor.SETTINGS_COMPONENT, ConfigContextEnum.GLOBAL);
+        if (settingsConfigurationModels.size() != 1) {
+            return Optional.empty();
+        }
+        return Optional.of(settingsConfigurationModels.get(0));
```
gkillough: We might be able to replace lines 103-106 with: 
```java
return settingsConfigurationModels.stream.findFirst();
```

src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
psantos1113: The sorting is to ensure that the descriptors are processed an logged in alphabetical order.  Sorted will use an objects natural ordering which in the case of a string it will be done lexicographically.  The only descriptor out of order is the settings descriptor.  That takes precedence to determine if environment variables should override.

We used to log all of the environment variables in alphabetical order.  Logging that data became redundant.  We started logging the environment variables per descriptor to determine if environment variables were set or not.  Then it became duplicate data in the logs.

Processing the descriptors and writing the data into the database does not matter because the data written happens before and channels or providers begin their tasks.  The ordering is there to make it easier for someone looking through the logs when debugging an issue.  That was the only reason for the sorting of the descriptor names.

PR #284, opened 2019-01-29
Fixing conversion

jamesrichard91:
This changes the way we handle the configuration from the UI going to the backend, to ensure it is correctly populated before we try to use it to test


src/main/java/com/synopsys/integration/alert/common/ConfigurationFieldModelConverter.java
```diff
@@ -116,4 +139,13 @@ public final FieldAccessor convertToFieldAccessor(final FieldModel fieldModel) t
         }
         return Optional.of(configurationModel);
     }
+
+    private boolean areValuesEmptyOrBlank(final Collection<String> values) {
+        if (null == values || values.isEmpty()) {
+            return true;
+        }
+        final Optional<String> first = values.stream().filter(StringUtils::isNotBlank).findFirst();
```
gkillough: You might want to try `return values.stream().allMatch(StringUtils::isBlank);` (or `.anyMatch(...)` depending on what you're testing).

src/main/java/com/synopsys/integration/alert/web/model/configuration/FieldModel.java
```diff
@@ -72,10 +70,6 @@ public String getContext() {
         return getField(key).flatMap(fieldValueModel -> fieldValueModel.getValue());
     }
 
-    public Collection<String> getFieldValues(final String key) {
```
gkillough: Why are you removing this?

jamesrichard91: It isn't used anywhere so I removed it. I can put it back in case we use it in the future I guess

PR #293, opened 2019-02-05
Bm sonarcloud fixes

bamandel:
# Pull Request template

* Address Sonarcloud issues

* Fixes all over the place

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/hipchat/descriptor/HipChatDistributionUIConfig.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/hipchat/descriptor/HipChatDistributionUIConfig.java
@@ -24,10 +24,8 @@
 package com.synopsys.integration.alert.channel.hipchat.descriptor;
 
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
 
-import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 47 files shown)


src/main/java/com/synopsys/integration/alert/channel/hipchat/descriptor/HipChatDistributionUIConfig.java
```diff
@@ -38,31 +36,20 @@
...
+    public HipChatDistributionUIConfig(final BaseConfigurationAccessor configurationAccessor, final BaseDescriptorAccessor descriptorAccessor) {
         super(HipChatDescriptor.HIP_CHAT_LABEL, HipChatDescriptor.HIP_CHAT_URL, HipChatDescriptor.HIP_CHAT_ICON, configurationAccessor, descriptorAccessor);
     }
 
     @Override
     public List<ConfigField> createChannelDistributionFields() {
-        final ConfigField roomId = NumberConfigField.createRequired(HipChatDescriptor.KEY_ROOM_ID, "Room Id", this::validateRoomID);
+        final ConfigField roomId = NumberConfigField.createRequired(HipChatDescriptor.KEY_ROOM_ID, "Room Id");
         final ConfigField notify = CheckboxConfigField.create(HipChatDescriptor.KEY_NOTIFY, "Notify");
         final ConfigField color = SelectConfigField.create(HipChatDescriptor.KEY_COLOR, "Color", Arrays.asList("Yellow", "Green", "Red", "Purple", "Gray", "Random"));
```
gkillough: Might as well change this to `List.of(...)` while we're at it.

src/main/java/com/synopsys/integration/alert/common/ProxyManager.java
```diff
@@ -56,14 +56,14 @@ public ProxyManager(final BaseConfigurationAccessor configurationAccessor) {
...
                        .stream()
                        .findFirst();
-        } catch (AlertDatabaseConstraintException ex) {
+        } catch (final AlertDatabaseConstraintException ex) {
             logger.error("Could not find the settings configuration for proxy data", ex);
         }
         return Optional.empty();
     }
 
-    public ProxyInfo createProxyInfo() throws IllegalArgumentException {
```
gkillough: SonarQube might complain about declaring unchecked exceptions, but we often do it to avoid accidentally introducing bugs when using certain methods. I think this is one of those cases since the exception is bubbling up from `integration-rest` and it is declared as part of the method signature.

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionUIConfig.java
```diff
@@ -79,10 +79,9 @@ public BlackDuckDistributionUIConfig(final BlackDuckProvider provider) {
         final String projectNamePattern = fieldModel.getField(BlackDuckDescriptor.KEY_PROJECT_NAME_PATTERN).flatMap(field -> field.getValue()).orElse(null);
         final boolean trueTextPresent = Boolean.TRUE.toString().equalsIgnoreCase(filterByProject);
         final boolean falseTextPresent = Boolean.FALSE.toString().equalsIgnoreCase(filterByProject);
-        if (!trueTextPresent && !falseTextPresent) {
-            if ((null == configuredProjects || configuredProjects.isEmpty()) && StringUtils.isBlank(projectNamePattern)) {
-                return List.of("You must select at least one project.");
-            }
+        final boolean missingProject = (null == configuredProjects || configuredProjects.isEmpty()) && StringUtils.isBlank(projectNamePattern);
+        if (!trueTextPresent && !falseTextPresent && missingProject) {
```
gkillough: Negating both the true and false case looks like a bug. I know you didn't necessarily write the original logic, but we should be using `final boolean shouldFilter = Boolean.parseBoolean(filterByProject)` to get the value, then using that for our logic. Anything else kills readability.

PR #294, opened 2019-02-05
Password Reset

gkillough:


src/main/java/com/synopsys/integration/alert/channel/email/EmailMessagingService.java
```diff
@@ -54,6 +54,8 @@
 import freemarker.template.TemplateException;
 
 public class EmailMessagingService {
+    private static final String SIMPLE_EMAIL_ADDRESS_PATTERN = ".{1,64}@.+";
```
bamandel: There should be some already existing validators we can use to verify Email addresses.

src/main/java/com/synopsys/integration/alert/component/settings/PasswordResetService.java
```diff
@@ -0,0 +1,101 @@
...
+import com.synopsys.integration.alert.common.enumeration.EmailPropertyKeys;
+import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
+import com.synopsys.integration.alert.common.exception.AlertException;
+import com.synopsys.integration.alert.database.api.configuration.model.ConfigurationModel;
+import com.synopsys.integration.alert.database.api.user.UserAccessor;
+import com.synopsys.integration.alert.database.api.user.UserModel;
+
+@Component
+public class PasswordResetService {
+    private static final int TEMP_PASSWORD_LENGTH = 50;
```
bamandel: 50 seems a bit long. Do we think 10 would suffice as well? I know sometimes I don't have the opportunity to copy and paste when changing passwords like this and have to type in the new value manually.

src/main/java/com/synopsys/integration/alert/component/settings/PasswordResetService.java
```diff
@@ -0,0 +1,101 @@
...
+        final EmailTarget passwordResetEmail = new EmailTarget(userModel.getEmailAddress(), TEMPLATE_NAME, templateFields, Map.of());
+        handleSendAndUpdateDatabase(passwordResetEmail, emailProperties, username, tempPassword);
+    }
+
+    private void handleSendAndUpdateDatabase(final EmailTarget passwordResetEmail, final EmailProperties emailProperties, final String username, final String tempPassword) throws AlertException {
+        try {
+            final EmailMessagingService emailService = new EmailMessagingService(alertProperties.getAlertTemplatesDir(), emailProperties);
+            emailService.sendEmailMessage(passwordResetEmail);
+            // Only change the password if there isn't an issue with sending the email
+            userAccessor.changeUserPassword(username, tempPassword);
```
bamandel: I think this should be pulled out of this method and just called itself after you send the email.

src/main/java/com/synopsys/integration/alert/database/api/user/UserAccessor.java
```diff
@@ -121,7 +121,19 @@ public boolean changeUserPassword(final String username, final String newPasswor
         final Optional<UserEntity> entity = userRepository.findByUserName(username);
         if (entity.isPresent()) {
             final UserEntity oldEntity = entity.get();
-            final UserEntity updatedEntity = new UserEntity(oldEntity.getUserName(), defaultPasswordEncoder.encode(newPassword));
+            final UserEntity updatedEntity = new UserEntity(oldEntity.getUserName(), defaultPasswordEncoder.encode(newPassword), oldEntity.getEmailAddress());
+            updatedEntity.setId(oldEntity.getId());
+            return userRepository.save(updatedEntity) != null;
+        } else {
```
bamandel: else is super necessary here

src/main/java/com/synopsys/integration/alert/web/controller/AuthenticationController.java
```diff
@@ -91,4 +99,22 @@ public LoginController(final LoginActions loginActions, final ResponseFactory re
...
         }
     }
+
+    @PostMapping(value = "/resetPassword")
+    public ResponseEntity<String> resetPassword() {
+        return responseFactory.createBadRequestResponse(null, "Password Reset Error: A username must be specified");
+    }
+
+    @PostMapping(value = "/resetPassword/{username}")
+    public ResponseEntity<String> resetPassword(@PathVariable final String username) {
```
bamandel: Why have 2 endpoints defined here if the one without the username path variable just return a bad request? That could be the very first check in the correct method

src/main/java/com/synopsys/integration/alert/web/controller/AuthenticationController.java
```diff
@@ -91,4 +99,22 @@ public LoginController(final LoginActions loginActions, final ResponseFactory re
...
+        final String errorPrefix = "Password Reset Error: ";
+        try {
+            passwordResetService.resetPassword(username);
+            return responseFactory.createOkResponse(null, "Password reset email sent");
```
bamandel: you can call createOkContentResponse which doesn't require any ID

src/main/java/com/synopsys/integration/alert/channel/email/EmailMessagingService.java
jamesrichard91: `  try {
                final InternetAddress emailAddr = new InternetAddress(testEmailAddress);
                emailAddr.validate();
            } catch (final AddressException ex) {
                throw new AlertException(String.format("%s is not a valid email address. %s", testEmailAddress, ex.getMessage()));
            }`

src/main/java/com/synopsys/integration/alert/component/settings/PasswordResetService.java
gkillough: Luckily we allow users to copy and paste the password since we are the application developers ;)

src/main/java/com/synopsys/integration/alert/component/settings/PasswordResetService.java
gkillough: After the catch blocks?

src/main/java/com/synopsys/integration/alert/component/settings/PasswordResetService.java
gkillough: Since our code is open source, and we do not currently limit login attempts, the ability to brute force a login is significantly stymied by increasing the number of characters in the temp password.
36^10 ≈ 3.65 x 10^15
36^50 ≈ 6.53 x 10^77

src/main/java/com/synopsys/integration/alert/web/controller/AuthenticationController.java
gkillough: If we don't define the route without the path variable, it returns a 404

PR #295, opened 2019-02-05
Updating the error messaging around the project name pattern

jamesrichard91:


src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionDescriptorActionApi.java
```diff
@@ -75,16 +75,22 @@ public TestConfigModel createTestConfigModel(final String configId, final FieldA
             if (filterByProject.isPresent() && BooleanUtils.toBoolean(filterByProject.get())) {
                 final Optional<ConfigurationFieldModel> projectField = fieldAccessor.getField(BlackDuckDescriptor.KEY_CONFIGURED_PROJECT);
                 final Set<String> configuredProjects = projectField.map(ConfigurationFieldModel::getFieldValues).orElse(Set.of()).stream().collect(Collectors.toSet());
-                final Optional<String> projectNamePattern = fieldAccessor.getString(BlackDuckDescriptor.KEY_PROJECT_NAME_PATTERN);
-                blackDuckProjectEntities = blackDuckProjectRepositoryAccessor.readEntities()
+                final String projectNamePattern = fieldAccessor.getString(BlackDuckDescriptor.KEY_PROJECT_NAME_PATTERN).orElse("");
+                final List<BlackDuckProjectEntity> blackDuckProjects = blackDuckProjectRepositoryAccessor.readEntities();
+                final boolean noProjectsMatchPattern = blackDuckProjects.stream().noneMatch(databaseEntity -> databaseEntity.getName().matches(projectNamePattern));
```
gkillough: Since we can ensure that `projectNamePattern` is not null, we could switch the lambda to: 
```java
databaseEntity -> projectNamePattern.matches(databaseEntity.getName())
```
This way we don't have to chain methods and we don't have to worry about `databaseEntity.getName()` returning `null`. I know it won't (because we control it) by it removes the concern of a maintainer seeing a method in the chain that returns String (which is notoriously nullable) and having to ensure that it will never, in fact, be null.

PR #296, opened 2019-02-05
Bm project desc fix

bamandel:
# Pull Request template

* Fixed bug with long project descriptions

* add a length limit of 250
* deleted all upgrade code


src/main/java/com/synopsys/integration/alert/database/provider/blackduck/data/BlackDuckProjectRepositoryAccessor.java
```diff
@@ -48,8 +50,24 @@ public BlackDuckProjectEntity findByName(final String name) {
         return blackDuckProjectRepository.findByName(name);
     }
 
+    @Override
+    public BlackDuckProjectEntity saveEntity(final BlackDuckProjectEntity blackDuckProjectEntity) {
```
psantos1113: Use StringUtils.abbreviate method.  It take a max length and appends ... to the end already. 
`final description = StringUtils.abbreviate(blackDuckProjectEntity.getDescription(), MAX_DESCRIPTION_LENGTH);`
you won't need the trimDescription method or the if check

PR #303, opened 2019-02-08
Updating  the ldap Referral field and Ldap validations

jamesrichard91:


src/main/java/com/synopsys/integration/alert/component/settings/SettingsUIConfig.java
```diff
@@ -166,6 +166,30 @@ private boolean validateFieldExists(final FieldModel fieldModel, final String fi
         return List.of();
     }
 
+    private Collection<String> validateLDAPAuthenticationType(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
+        if (isLDAPEnabled(fieldModel)) {
+            final String authenticationType = fieldToValidate.getValue().map(String::trim).orElse("");
+            final boolean doesNotMatchKnownAuthentication = "simple".equals(authenticationType) || "none".equals(authenticationType) || "digest".equals(authenticationType);
```
gkillough: Shouldn't this be called doesMatchKnownAuthentication?

src/main/java/com/synopsys/integration/alert/component/settings/SettingsUIConfig.java
```diff
@@ -166,6 +166,30 @@ private boolean validateFieldExists(final FieldModel fieldModel, final String fi
...
+            }
+        }
+
+        return List.of();
+    }
+
+    private Collection<String> validateLDAPReferral(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
+        if (isLDAPEnabled(fieldModel)) {
+            final String referral = fieldToValidate.getValue().map(String::trim).orElse("");
+            final boolean doesNotMatchKnownReferral = "ignore".equals(referral) || "follow".equals(referral) || "throw".equals(referral);
```
bamandel: This sort of validation should be done for all SelectConfigFields. We want to make sure the value passed to us is contained in the possible values list. We can probably move something very similar to this over to the SelectConfigField class.

src/main/java/com/synopsys/integration/alert/component/settings/SettingsDescriptor.java
```diff
@@ -76,6 +76,8 @@
     public static final String FIELD_ERROR_LDAP_SERVER_MISSING = "LDAP Server missing";
     public static final String FIELD_ERROR_LDAP_USERNAME_MISSING = "LDAP Manager DN is missing";
     public static final String FIELD_ERROR_LDAP_PASSWORD_MISSING = "LDAP Manager DN password is missing";
+    public static final String FIELD_ERROR_LDAP_AUTHENTICATION_TYPE_INVALID = "Invalid LDAP Authentication type specified.";
+    public static final String FIELD_ERROR_LDAP_REFERRAL_INVALID = "Invalid LDAP Referral specified";
```
bamandel: If we move the validation to the SelectConfigField then we should be able to generate these errors from the data we already have.

PR #307, opened 2019-02-11
Fixing validation

jamesrichard91:


src/main/java/com/synopsys/integration/alert/common/descriptor/config/context/ChannelDistributionDescriptorActionApi.java
```diff
@@ -86,4 +98,39 @@ public FieldModel deleteConfig(final FieldModel fieldModel) {
...
+        final String jobNameError = validateJobNameUnique(fieldModel);
+        if (StringUtils.isNotBlank(jobNameError)) {
+            return Map.of(ChannelDistributionUIConfig.KEY_NAME, jobNameError);
+        }
+        return Map.of();
+    }
+
+    private String validateJobNameUnique(final FieldModel fieldModel) {
+        final String descriptorName = fieldModel.getDescriptorName();
+        return fieldModel.getField(ChannelDistributionUIConfig.KEY_NAME).map(jobNameField -> {
```
gkillough: This is a pretty heavily loaded lambda. Can we break it up a little more for readability? Maybe an additional method or two.

src/test/java/com/synopsys/integration/alert/channel/ChannelDescriptorTest.java
```diff
@@ -130,8 +130,8 @@ public FieldModel createInvalidDistributionFieldModel() {
     }
 
     public Map<String, String> createValidCommonDistributionFieldMap() {
-        return Map.of(ChannelDistributionUIConfig.KEY_NAME, "name", ChannelDistributionUIConfig.KEY_FREQUENCY, FrequencyType.REAL_TIME.name(), ChannelDistributionUIConfig.KEY_CHANNEL_NAME, "channelName",
-            ChannelDistributionUIConfig.KEY_PROVIDER_NAME, "providerName");
+        return Map.of(ChannelDistributionUIConfig.KEY_NAME, "name", ChannelDistributionUIConfig.KEY_FREQUENCY, FrequencyType.REAL_TIME.name(), ChannelDistributionUIConfig.KEY_CHANNEL_NAME, "channel_slack",
```
gkillough: Let's use the constants here so that if they ever change the tests won't break.

PR #308, opened 2019-02-11
Temporary fix for job running

bamandel:
We are storing some enums in the DB as their pretty printed coutnerparts. We should not do this and store the actual Enum values. For now, we parse this stored value and put it into memory as expected.


bamandel:
Closing as the original source of the issue is being fixed elsewhere.

PR #314, opened 2019-02-12
Attempt to fix audit once and for all

gkillough:


src/main/java/com/synopsys/integration/alert/database/entity/repository/NotificationContentRepository.java
```diff
@@ -45,20 +45,40 @@
...
+
+    @Query(value =
+               "SELECT RAW_NOTIFICATION_CONTENT.ID, RAW_NOTIFICATION_CONTENT.CREATED_AT, RAW_NOTIFICATION_CONTENT.PROVIDER, RAW_NOTIFICATION_CONTENT.PROVIDER_CREATION_TIME, RAW_NOTIFICATION_CONTENT.NOTIFICATION_TYPE, RAW_NOTIFICATION_CONTENT.CONTENT "
+                   + "FROM ALERT.RAW_NOTIFICATION_CONTENT "
+                   + "JOIN ALERT.AUDIT_NOTIFICATION_RELATION ON AUDIT_NOTIFICATION_RELATION.NOTIFICATION_ID = RAW_NOTIFICATION_CONTENT.ID "
+                   + "JOIN ALERT.AUDIT_ENTRIES ON AUDIT_ENTRIES.ID = AUDIT_NOTIFICATION_RELATION.AUDIT_ENTRY_ID "
+                   + "JOIN ALERT.CONFIG_GROUPS ON CONFIG_GROUPS.JOB_ID = AUDIT_ENTRIES.COMMON_CONFIG_ID "
+                   + "JOIN ALERT.DESCRIPTOR_CONFIGS ON DESCRIPTOR_CONFIGS.ID = CONFIG_GROUPS.CONFIG_ID "
+                   + "JOIN ALERT.FIELD_VALUES ON FIELD_VALUES.CONFIG_ID = DESCRIPTOR_CONFIGS.ID "
+                   + "JOIN ALERT.DEFINED_FIELDS ON DEFINED_FIELDS.ID = FIELD_VALUES.FIELD_ID "
```
jamesrichard91: Do we want the JOIN to be LEFT JOIN like the old query? or does LEFT JOIN not work in the raw query?

src/main/java/com/synopsys/integration/alert/database/entity/repository/NotificationContentRepository.java
```diff
@@ -45,20 +45,40 @@
...
+                   + "WHERE RAW_NOTIFICATION_CONTENT.PROVIDER ILIKE %:searchTerm% "
+                   + "OR RAW_NOTIFICATION_CONTENT.NOTIFICATION_TYPE ILIKE %:searchTerm% "
```
jamesrichard91: Do we want to wrap the values in LOWER(...) ? or does that not work in the raw sql?

jamesrichard91:
We should comment out the configGroupEntity field in AuditEntryEntity in this PR so we can test if this query will work

gkillough:
Going to take a different approach.

PR #315, opened 2019-02-12
Started fix for broken sensitive tests

bamandel:
Fixed potential issue with sensitive fields when testing

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckProviderUIConfig.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckProviderUIConfig.java
@@ -54,12 +54,14 @@ public class BlackDuckProviderUIConfig extends UIConfig {
     }
 
     private Collection<String> validateAPIToken(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
-        final String apiKey = fieldToValidate.getValue().orElse("");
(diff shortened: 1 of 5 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckProviderUIConfig.java
```diff
@@ -54,12 +54,14 @@ public BlackDuckProviderUIConfig() {
     }
 
     private Collection<String> validateAPIToken(final FieldValueModel fieldToValidate, final FieldModel fieldModel) {
-        final String apiKey = fieldToValidate.getValue().orElse("");
-        if (StringUtils.isBlank(apiKey)) {
-            return List.of(ConfigField.REQUIRED_FIELD_MISSING);
-        } else {
-            if (apiKey.length() < 64 || apiKey.length() > 256) {
-                return List.of("Invalid Black Duck API Token.");
+        if (fieldToValidate.hasValues() || !fieldToValidate.isSet()) {
```
jamesrichard91: Why do we need this if? and is it doing what you expect? the condition seems very confusing

psantos1113: if there are values we want to validate if the list is empty and isSet is false we want to validate.  When isSet=true because it is saved the values list is empty which is ok.

PR #319, opened 2019-02-14
Audit job changes

jamesrichard91:


src/main/java/com/synopsys/integration/alert/database/audit/AuditEntryEntity.java
```diff
@@ -73,9 +72,9 @@
     @OneToMany(mappedBy = "auditEntryEntity")
     private List<AuditNotificationRelation> auditNotificationRelations;
 
-    @ManyToOne
-    @JoinColumn(name = "common_config_id", referencedColumnName = "job_id", insertable = false, updatable = false)
-    private ConfigGroupEntity configGroupEntity;
+    @OneToMany
+    @JoinColumn(name = "audit_id", referencedColumnName = "id", insertable = false, updatable = false)
```
gkillough: Since this is OneToMany, do we need the JoinColumn? I was under the impression we only needed mappedBy.

PR #320, opened 2019-02-14
Remove all empty fields from fieldmodel when validating

bamandel:
Removed all empty/null fields from fieldmodel when validating


bamandel:
Do we want to trim the map for all endpoints or only when necessary (Like in validation)?

src/main/java/com/synopsys/integration/alert/web/config/FieldModelProcessor.java
```diff
@@ -101,16 +101,34 @@ public FieldModel performUpdateAction(final FieldModel fieldModel) {
...
+            final Map<String, ConfigField> configFields = retrieveUIConfigFields(trimmedFieldModel.getContext(), trimmedFieldModel.getDescriptorName())
                                                               .stream()
                                                               .collect(Collectors.toMap(ConfigField::getKey, Function.identity()));
-            descriptorActionApi.get().validateConfig(configFields, fieldModel, fieldErrors);
+            descriptorActionApi.get().validateConfig(configFields, trimmedFieldModel, fieldErrors);
         }
         return fieldErrors;
     }
 
+    public FieldModel trimFieldModelValues(final FieldModel fieldModel) {
```
psantos1113: Why would we only want to do this for validation.  When creating a configuration or updating wouldn't we want to save just the necessary values.  I think we would want this trim method in the save and update methods as well.  But it would need to be tested.

PR #325, opened 2019-02-15
Address Code Smells

gkillough:


src/main/java/com/synopsys/integration/alert/channel/email/EmailMessagingService.java
```diff
@@ -200,8 +200,8 @@ private void sendAuthenticated(final EmailProperties emailProperties, final Mess
     }
 
     private String generateContentId(final String value) {
-        final String cid = value.replaceAll("[^A-Za-z0-9]", "bd").trim() + "@blackducksoftware.com";
-        return cid;
+        // TODO should this be changed to '@synopsys.com'?
+        return value.replaceAll("[^A-Za-z0-9]", "bd").trim() + "@blackducksoftware.com";
```
psantos1113: I believe this is just a unique identifier in the content of the email message so changing to synopsys.com should not cause any issues here.

src/main/java/com/synopsys/integration/alert/web/model/configuration/FieldValueModel.java
```diff
@@ -43,8 +44,9 @@ public FieldValueModel(final Collection<String> values, final boolean isSet) {
 
     public Collection<String> getValues() {
         if (null != values) {
-            return values.stream().filter(singleValue -> null != singleValue).collect(Collectors.toList());
+            return values.stream().filter(Objects::nonNull).collect(Collectors.toList());
         }
+        // FIXME should this return Set.of(); ?
```
psantos1113: If isSet is true we expect values to be null right now.  The code that is in place for the UI may work correctly with Set.of() with the latest changes that were put in to the UI.

PR #327, opened 2019-02-20
Fixing issue with category item lookup.

jamesrichard91:


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyViolationCollector.java
```diff
@@ -85,7 +85,7 @@ protected void addCategoryItems(final List<CategoryItem> categoryItems, final Js
                 addApplicableItems(categoryItems, notificationContent.getId(), policyLinkableItem, policyUrl, operation, applicableItems);
             }
         }
-        
+
```
gkillough: :(

PR #333, opened 2019-03-01
Field help text

jamesrichard91:


src/main/js/providers/BlackDuckConfiguration.js
```diff
@@ -120,6 +122,7 @@ class BlackDuckConfiguration
                         <NumberInput
                             id={KEY_BLACKDUCK_TIMEOUT}
                             label="Timeout"
+                            description="The timeout in seconds for all connections to the Black Duck server. Default= 300"
```
gkillough: Since this is help text, I would go with "Default: 300."

src/main/js/distribution/job/HipChatJobConfiguration.js
```diff
@@ -99,6 +99,7 @@ class HipChatJobConfiguration extends Component {
                 <TextInput
                     id={KEY_ROOM_ID}
                     label="Room Id"
+                    description="The ID of the room to receive Alerts."
```
gkillough: Might want to change this to say, "The API ID of rht room to receive alerts."

It's officially called API ID in HipChat docs.

PR #334, opened 2019-03-04
This should not schedule the Polaris task until it has a valid configuration

jamesrichard91:


src/main/java/com/synopsys/integration/alert/provider/polaris/PolarisProvider.java
```diff
@@ -45,19 +47,24 @@
...
     }
 
     @Override
     public void initialize() {
         logger.info("Initializing Polaris provider...");
         taskManager.registerTask(projectSyncTask);
-        taskManager.scheduleCronTask(BlackDuckAccumulator.DEFAULT_CRON_EXPRESSION, projectSyncTask.getTaskName());
+        final Optional<AccessTokenPolarisHttpClient> polarisHttpClient = polarisProperties.createPolarisHttpClientSafely(logger);
+        polarisHttpClient.ifPresent(client -> {
+            taskManager.scheduleCronTask(BlackDuckAccumulator.DEFAULT_CRON_EXPRESSION, projectSyncTask.getTaskName());
```
bamandel: Since this sort of functionality seems to be appearing multiple times, maybe we want to build it into the provider or make a new parent Provider called ScheduledProvider (Or something similar). The provider will have all the same methods available, but also only start the job if a config exists.

src/main/java/com/synopsys/integration/alert/provider/polaris/descriptor/PolarisGlobalDescriptorActionApi.java
```diff
@@ -73,4 +80,14 @@ public void testConfig(final TestConfigModel testConfig) throws IntegrationExcep
...
     }
+
+    @Override
+    public FieldModel saveConfig(final FieldModel fieldModel) {
+        final Optional<AccessTokenPolarisHttpClient> polarisHttpClient = polarisProperties.createPolarisHttpClientSafely(logger);
+        final Optional<String> nextRunTime = taskManager.getNextRunTime(BlackDuckAccumulator.TASK_NAME);
+        if (polarisHttpClient.isPresent() && nextRunTime.isEmpty()) {
+            taskManager.scheduleCronTask(PolarisProjectSyncTask.DEFAULT_CRON_EXPRESSION, PolarisProjectSyncTask.TASK_NAME);
+        }
+        return super.saveConfig(fieldModel);
```
bamandel: Similar to what I said previously, this seems to be something that should be abstracted out now. Anytime we have a scheduled provider, we will need to perform this funcitonality. So perhaps a parent class or a utility class to perform this would help a bunch.

src/main/java/com/synopsys/integration/alert/provider/polaris/tasks/PolarisProjectSyncTask.java
```diff
@@ -63,6 +64,7 @@
 @Component
 public class PolarisProjectSyncTask extends ScheduledTask {
     public static final String TASK_NAME = "polaris-project-sync-task";
+    public static final String DEFAULT_CRON_EXPRESSION = BlackDuckAccumulator.DEFAULT_CRON_EXPRESSION;
```
gkillough: Can we put the constant DEFAULT_CRON_EXPRESSION in ScheduledTask or something so that we don't have providers referencing each other?

PR #338, opened 2019-03-12
Working on settings and validation

jamesrichard91:


src/main/java/com/synopsys/integration/alert/common/ConfigurationFieldModelConverter.java
```diff
@@ -62,15 +63,23 @@ public final FieldAccessor convertToFieldAccessor(final FieldModel fieldModel) t
         return convertToConfigurationFieldModelMap(fieldModel);
     }
 
-    public final Optional<ConfigurationFieldModel> convertFromDefinedFieldModel(final DefinedFieldModel definedFieldModel, final String value) {
+    public final Optional<ConfigurationFieldModel> convertFromDefinedFieldModel(final DefinedFieldModel definedFieldModel, final String value, final boolean isSet) {
         final Optional<ConfigurationFieldModel> configurationModel = createEmptyModel(definedFieldModel);
-        configurationModel.ifPresent(model -> model.setFieldValue(value));
+        final boolean actualSetValue = isSet || StringUtils.isNotBlank(value);
+        configurationModel.ifPresent(model -> {
```
gkillough: We could abstract the ifPresent functionality into a method to reuse here and bellow.

src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -71,28 +71,28 @@ public AlertStartupInitializer(final DescriptorMap descriptorMap, final Environm
...
-            Optional<ConfigurationModel> settingsConfiguration = findSettingsConfiguration();
+            final Optional<ConfigurationModel> settingsConfiguration = findSettingsConfiguration();
             final String fieldKey = SettingsDescriptor.KEY_STARTUP_ENVIRONMENT_VARIABLE_OVERRIDE;
-            boolean overwriteSavedInDB = settingsConfiguration
-                                             .flatMap(configurationModel -> configurationModel.getField(fieldKey))
-                                             .flatMap(field -> field.getFieldValue())
-                                             .map(value -> Boolean.valueOf(value)).orElse(Boolean.FALSE);
+            final boolean overwriteSavedInDB = settingsConfiguration
+                                                   .flatMap(configurationModel -> configurationModel.getField(fieldKey))
+                                                   .flatMap(field -> field.getFieldValue())
```
gkillough: This could become a method reference.

src/main/java/com/synopsys/integration/alert/common/ConfigurationFieldModelConverter.java
jamesrichard91: I dont think we would get value from that since the body of this ifPresent uses different method calls than the one below.

PR #341, opened 2019-03-18
Merging sync changes to master

jamesrichard91:
This should ensure all the provider data that gets synced in the database is encapsulated in a single transaction to prevent read access during the sync process

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ProviderDataAccessor.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ProviderDataAccessor.java
@@ -25,6 +25,7 @@ package com.synopsys.integration.alert.common.persistence.accessor;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
@@ -50,4 +51,6 @@ public interface ProviderDataAccessor {
     List<ProviderUserModel> getAllUsers(final String providerName);
 
     List<ProviderUserModel> deleteAndSaveAllUsers(final String providerName, final Collection<ProviderUserModel> usersToDelete, final Collection<ProviderUserModel> usersToAdd);
+
+    void updateProjectAndUserData(String providerName, final Map<ProviderProject, Set<String>> projectToUserData);
 }
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
@@ -177,7 +177,7 @@ public class DefaultProviderDataAccessor implements ProviderDataAccessor {
     @Override
     public void updateProjectAndUserData(final String providerName, final Map<ProviderProject, Set<String>> projectToUserData) {
         updateProjectDB(providerName, projectToUserData.keySet());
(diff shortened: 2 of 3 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ProviderDataAccessor.java
```diff
@@ -50,4 +51,6 @@
     List<ProviderUserModel> getAllUsers(final String providerName);
 
     List<ProviderUserModel> deleteAndSaveAllUsers(final String providerName, final Collection<ProviderUserModel> usersToDelete, final Collection<ProviderUserModel> usersToAdd);
+
+    void updateProjectAndUserData(String providerName, final Map<ProviderProject, Set<String>> projectToUserData);
```
gkillough: Can we rename the `projectToUserData` variable to `projectToUserEmailMap` or something that clarifies what the Set of Strings is?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -169,4 +174,67 @@ private ProviderUserEntity convertToUserEntity(final String providerName, final
         return new ProviderUserEntity(providerUserModel.getEmailAddress(), providerUserModel.getOptOut(), providerName);
     }
 
+    @Override
+    public void updateProjectAndUserData(final String providerName, final Map<ProviderProject, Set<String>> projectToUserData) {
+        updateProjectDB(providerName, projectToUserData.keySet());
+        final Set<String> userData = projectToUserData.values().stream().flatMap(userDataSet -> userDataSet.stream()).collect(Collectors.toSet());
```
gkillough: Could the `flatMap` operation use a method reference like `Collection::stream`?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -169,4 +174,67 @@ private ProviderUserEntity convertToUserEntity(final String providerName, final
...
+        updateUserDB(providerName, userData);
+        updateUserProjectRelations(projectToUserData);
+    }
+
+    private List<ProviderProject> updateProjectDB(final String providerName, final Set<ProviderProject> currentProjects) {
+        logger.info("{} projects", currentProjects.size());
+        return deleteAndSaveAllProjects(providerName, currentProjects);
```
gkillough: Considering Polaris and Black Duck will both depend on this functionality eventually, we should probably track the deltas like we do with users.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ProviderDataAccessor.java
jamesrichard91: We could but I was thinking that if we change that to a User object in the future, because it becomes more than just email addresses, then we wouldn't have to change the method name too.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
jamesrichard91: Created a ticket to update this

PR #343, opened 2019-03-18
Project sync update

jamesrichard91:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -90,9 +90,10 @@ public ProviderProject saveProject(final String providerName, final ProviderProj
     }
 
     @Override
-    public List<ProviderProject> deleteAndSaveAllProjects(final String providerName, final Collection<ProviderProject> providerProjects) {
-        providerProjectRepository.deleteAllByProvider(providerName);
-        final Iterable<ProviderProjectEntity> providerProjectEntities = providerProjects
+    public List<ProviderProject> deleteAndSaveAllProjects(final String providerName, final Collection<ProviderProject> providerProjectsToRemove, final Collection<ProviderProject> providerProjectsToAdd) {
```
bamandel: Should this be split into 2 separate methods? We can still have this method, but having the 2 functionalities split out as well makes more sense to me.

jamesrichard91: Then would you also want the same done for Users, which does the same?

bamandel: If Users also has a method that does multiple things like this, then yes

gkillough: We should also remove "all" from the method name since it is no longer touching every single project.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -183,8 +194,28 @@ public void updateProjectAndUserData(final String providerName, final Map<Provid
...
 
     private List<ProviderProject> updateProjectDB(final String providerName, final Set<ProviderProject> currentProjects) {
-        logger.info("{} projects", currentProjects.size());
-        return deleteAndSaveAllProjects(providerName, currentProjects);
+        final Set<ProviderProject> projectsToAdd = new HashSet<>();
+        final Set<ProviderProject> projectsToRemove = new HashSet<>();
+        final List<ProviderProject> storedProjects = findByProviderName(providerName);
+
+        projectsToRemove.addAll(storedProjects);
+        projectsToRemove.removeIf(project -> currentProjects
```
gkillough: We might want to create a HashSet of href to project to improve efficiency here since we're doing a lot of work inside of a transaction so cycles matter. We can then use `.containsKey(href)` on each element.

psantos1113: If you had a HashSet<String> linkSet; the remove if would just change to removeIf(linkSet::contains)

PR #344, opened 2019-03-18
Environment override default

psantos1113:
Create a stored procedure to set the default for the environment variable override to true

### DIFF
--- a/src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-global-migration.xml
+++ b/src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-global-migration.xml
@@ -261,7 +261,7 @@
             <column name="CONTEXT_ID" valueComputed="GET_ID_FOR_CONFIG_CONTEXT('GLOBAL')"/>
         </insert>
         <sql dbms="h2" stripComments="true">
-            CALL INSERT_GLOBAL_FIELD_VALUE('component_settings','settings.startup.environment.variable.override', 'true');
+            CALL SET_LATEST_CONFIG_GLOBAL_FIELD_VALUE('settings.startup.environment.variable.override', 'true');
         </sql>
     </changeSet>
 
--- a/src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-global-migration.xml
+++ b/src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-global-migration.xml
@@ -261,7 +261,7 @@
             <column name="CONTEXT_ID" valueComputed="GET_ID_FOR_CONFIG_CONTEXT('GLOBAL')"/>
         </insert>
         <sql dbms="h2" stripComments="true">
-            CALL INSERT_FIELD_VALUE('component_settings','settings.startup.environment.variable.override', 'true');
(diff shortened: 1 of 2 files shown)


src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-stored-procedures.xml
```diff
@@ -364,4 +364,26 @@
...
+    <!-- Create job migration procedure -->
+    <changeSet author="psantos" id="1545000000000-4">
+        <createProcedure
+                dbms="h2"
+                relativeToChangelogFile="true"
+                schemaName="ALERT"
+                procedureName="INSERT_FIELD_VALUE"
+        >
+            CREATE ALIAS INSERT_GLOBAL_FIELD_VALUE AS $$
+                java.lang.Void insertGlobalFieldValue(final java.sql.Connection connection, final java.lang.String descriptor_name, final java.lang.String fieldKey, final java.lang.String fieldValue) throws java.sql.SQLException {
```
gkillough: This is minor, but you're using both snake_case and camelCase for the method parameters.

src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-stored-procedures.xml
```diff
@@ -364,4 +364,26 @@
...
+                    if (fieldValue != null &amp;&amp; !fieldValue.isBlank()) {
+                        try (final java.sql.Statement insertIntoFieldValues = connection.createStatement()) {
+                            final java.lang.String insertStatement = java.lang.String.format("INSERT INTO ALERT.FIELD_VALUES (CONFIG_ID, FIELD_ID, FIELD_VALUE) VALUES (GET_ID_FOR_REGISTERED_DESCRIPTOR_NAME('%s'), GET_FIELD_ID_FOR_SOURCE_KEY('%s'), '%s')",
```
gkillough: I think `GET_ID_FOR_REGISTERED_DESCRIPTOR_NAME(...)` returns the ID from the REGISTERED_DESCRIPTORS table, so this won't be the CONFIG_ID.

src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-global-migration.xml
```diff
@@ -252,4 +252,17 @@
         </sql>
     </changeSet>
 
-</databaseChangeLog>
\ No newline at end of file
+    <!-- Global Scheduling Config -->
+    <changeSet author="psantos" id="1548361428926-5">
```
gkillough: Just to confirm: a config for this descriptor doesn't exist before this, right?

psantos1113: Correct I tried to put the stored procedure call in the changelog-configuration-initialization.xml file first but no configuration exists for the settings component at the time that xml file is processed.  So I needed to do the same thing that was done with the other default configurations.

src/main/resources/db/changelog/alert/4.0.0/changelog-configuration-stored-procedures.xml
psantos1113: I'll fix it

PR #345, opened 2019-03-19
Bm auto provider ui

bamandel:
I made a much larger PR than I anticipated due to adding descriptions to each field in Java. If you would like to skip all those files, you can search for /PolarisGlobalUIConfig.java and my actual changes look to start from there. I apologize for not making 2 PR's.

Other than that, I have made an auto generated global Provider UI! Using the fields we defined in our java code, our javascript will automatically create those fields in the order you've specified. This is only setup to work with Polaris at the moment but we can expand it in the future to work for most Global descriptors.


jamesrichard91:
We may want to move the text of the field descriptions into static strings to make it easier to localize later

src/main/js/channels/SlackConfiguration.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 const SlackConfiguration = () => (
     <div>
         <h1>
-            <span className="fa fa-slack" />
```
jamesrichard91: I don't think we can remove the "fa-" from this class name

src/main/js/component/common/ConfigurationLabel.js
```diff
@@ -47,9 +48,7 @@ class ConfigurationLabel extends Component {
         return (
             <div className="d-inline-flex col-sm-4">
                 <h1>
-                    {fontAwesomeIcon &&
-                    <span className={iconClasses} />
-                    }
+                    <FontAwesome name={fontAwesomeIcon} fixedWidth />
```
jamesrichard91: Do we want to wrap this in
`
{fontAwesomeIcon &&
...
} 
`
like we had before, so it will only display if the icon is set

bamandel: I was worried I was removing a feature with this removal. At the moment, is there any page we use that doesn't use fontawesome? At the moment, it's almost a requirement for each descriptor.

bamandel:
> We may want to move the text of the field descriptions into static strings to make it easier to localize later

I can do this in another PR

src/main/js/component/common/ConfigurationLabel.js
jamesrichard91: If that is the case, then we should update the proptypes to make this property required

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -51,16 +51,19 @@
...
     }
 
     @Override
     public List<ConfigField> createFields() {
-        final ConfigField name = TextInputConfigField.createRequired(KEY_NAME, "Name");
-        final ConfigField frequency = SelectConfigField.createRequired(KEY_FREQUENCY, "Frequency", Arrays.stream(FrequencyType.values()).map(FrequencyType::name).collect(Collectors.toList()));
-        final ConfigField channelName = SelectConfigField.createRequired(KEY_CHANNEL_NAME, "Channel Type", getDescriptorNames(DescriptorType.CHANNEL));
-        final ConfigField providerName = SelectConfigField.createRequired(KEY_PROVIDER_NAME, "Provider Type", getDescriptorNames(DescriptorType.PROVIDER));
+        final ConfigField name = TextInputConfigField.createRequired(KEY_NAME, "Name", "The name of the distribution job. Must be unique.");
+        final ConfigField frequency = SelectConfigField.createRequired(KEY_FREQUENCY, "Frequency", "Select how frequent this job should check for notifications to send.",
```
gkillough: The description should say "frequently".

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionUIConfig.java
```diff
@@ -53,12 +53,14 @@ public BlackDuckDistributionUIConfig(final BlackDuckProvider provider) {
 
     @Override
     public List<ConfigField> createProviderDistributionFields() {
-        final ConfigField filterByProject = CheckboxConfigField.createRequired(CommonDistributionConfiguration.KEY_FILTER_BY_PROJECT, "Filter by project");
-        final ConfigField projectNamePattern = TextInputConfigField.create(CommonDistributionConfiguration.KEY_PROJECT_NAME_PATTERN, "Project name pattern", this::validateProjectNamePattern);
+        final ConfigField filterByProject = CheckboxConfigField.createRequired(CommonDistributionConfiguration.KEY_FILTER_BY_PROJECT, "Filter by project",
+            "If true, all projects will be included. Any notifications matching the configured notification types will be processed.");
```
gkillough: I think this should say "If false, all projects will be included..."

src/main/js/providers/FieldsPanel.js
```diff
@@ -0,0 +1,75 @@
...
+    fieldKeys: PropTypes.array.isRequired,
+    descriptorFields: PropTypes.array.isRequired,
+    currentConfig: PropTypes.object.isRequired,
+    updateStatus: PropTypes.string,
+    fieldErrors: PropTypes.object
+};
+
+// Mapping redux state -> react props
+const mapStateToProps = state => ({
+    updateStatus: state.provider.updateStatus,
```
psantos1113: We aren't going to be able to rely on the provider state in redux.  We should pass the updateStatus and the fieldErrors as required properties and remove the need for redux state here.  The changing of the required fields by the parent class will cause componentDidUpdate to get called.   We want to separate this component from specific redux state since it will eventually relate to channel configuration and distribution job configuration

src/main/js/providers/ProviderConfiguration.js
```diff
@@ -0,0 +1,115 @@
...
+                    {errorMessage}
+                </div>}
+
+                {actionMessage && <div className="alert alert-success">
+                    {actionMessage}
+                </div>}
+
+                <form className="form-horizontal" onSubmit={this.handleSubmit}>
+                    <div>
+                        <FieldsPanel currentConfig={this.state.currentConfig} fieldKeys={this.state.currentKeys} descriptorFields={fields} />
```
psantos1113: Pass in the Provider updateStatus and the fieldErrors into the FieldsPanel

src/main/js/store/actions/provider.js
```diff
@@ -0,0 +1,157 @@
+import { CONFIG_FETCHED, CONFIG_FETCHING, CONFIG_TEST_FAILED, CONFIG_TEST_SUCCESS, CONFIG_TESTING, CONFIG_UPDATE_ERROR, CONFIG_UPDATED, CONFIG_UPDATING } from 'store/actions/types';
```
psantos1113: We should rename this from provider.js this is really just configuration methods across providers and channels. Perhaps configuration.js

src/main/js/store/reducers/index.js
```diff
@@ -14,12 +14,14 @@ import schedulingConfig from 'store/reducers/schedulingConfig';
...
 import system from 'store/reducers/system';
 import refresh from 'store/reducers/refresh';
+import provider from 'store/reducers/provider';
 
 
 const rootReducer = combineReducers({
     about,
     audit,
     blackduck,
+    provider,
```
psantos1113: rename to configuration

src/main/js/store/reducers/provider.js
```diff
@@ -0,0 +1,107 @@
+import { CONFIG_FETCHED, CONFIG_FETCHING, CONFIG_TEST_FAILED, CONFIG_TEST_SUCCESS, CONFIG_TESTING, CONFIG_UPDATE_ERROR, CONFIG_UPDATED, CONFIG_UPDATING, SERIALIZE } from 'store/actions/types';
```
psantos1113: rename to configuration

src/main/js/util/fieldMapping.js
```diff
@@ -0,0 +1,59 @@
...
+import TextArea from 'field/input/TextArea';
+import PasswordInput from 'field/input/PasswordInput';
+import NumberInput from 'field/input/NumberInput';
+import CheckboxInput from 'field/input/CheckboxInput';
+import ReadOnlyField from 'field/ReadOnlyField';
+import * as FieldModelUtilities from 'util/fieldModelUtilities';
+
+export function getField(fieldType, props) {
+    switch (fieldType) {
+        case 'Select':
```
psantos1113: You could look into a static map containing references to functions that you can lookup by field type.
You can look to see if that makes the code more readable than the switch statement.

```
export function buildTextInput(props) {
    return <TextInput {...props} />
}

export static const builders = { 
    'TextInput': buildTextInput,
     ...
}

export function getField(fieldType, props) {
    const builder = builders[fieldType];
    return builder(props);
}
```

src/main/java/com/synopsys/integration/alert/provider/blackduck/descriptor/BlackDuckDistributionUIConfig.java
jamesrichard91: So, this is a bit complicated. The UI shows "Include all projects" for this field, which is the opposite of filtering by projects.  We have handled this is in the font end for several versions.  At some point we will need to sync up the front end and the back end so  they are consistent. I'm in favor of updating the backend since that would not require a major version change and it would have the least impact to customers. But I don't think we should do that in this PR

PR #347, opened 2019-03-20
Fix the initialization based on environment variables to populate if …

psantos1113:
Fix the initialization based on environment variables to add fields even if the environment variable for override is false.  It will add new fields if it is missing.  If the environment override is specified only the environment variables defined will be used.  This will allow for clearing fields from configuration.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
+++ b/src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
@@ -26,6 +26,7 @@ package com.synopsys.integration.alert.workflow.startup;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -146,12 +154,10 @@ private void updateConfigurationFields(final String descriptorName, final boolea
...
         if (!configurationModels.isEmpty()) {
             if (!foundConfigurationModels.isEmpty()) {
-                if (overwriteCurrentConfig) {
-                    final ConfigurationModel configurationModel = foundConfigurationModels.get(0);
-                    logger.info("  Overwriting configuration values with environment for descriptor.");
-                    final Collection<ConfigurationFieldModel> updatedFields = updateAction(descriptorName, configurationModels);
-                    fieldConfigurationAccessor.updateConfiguration(configurationModel.getConfigurationId(), updatedFields);
-                }
+                final ConfigurationModel foundModel = foundConfigurationModels.get(0);
+                logger.info("  Overwriting configuration values with environment for descriptor.");
```
jamesrichard91: Why do these logs have a space at the beginning?

psantos1113: indentation to nest them so it's a little easier to read the server startup logs.

PR #352, opened 2019-03-26
fix add a check for the roles to make sure the user can use alert

psantos1113:
Add a check to make sure the user contains a role that alert recognizes during the login.  This authorization check is used to determine if the user can login to the application.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/actions/LoginActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/actions/LoginActions.java
@@ -33,11 +33,13 @@ import org.springframework.security.authentication.BadCredentialsException;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.authentication.dao.DaoAuthenticationProvider;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/alert/web/actions/LoginActions.java
```diff
@@ -89,6 +93,15 @@ private Authentication performLdapAuthentication(final LoginConfig loginConfig)
         return result;
     }
 
+    private boolean isAuthorized(final Authentication authentication) {
+        EnumSet<UserRole> allowedRoles = EnumSet.allOf(UserRole.class);
+        return authentication.getAuthorities().stream()
+                   .map(authority -> authority.getAuthority())
```
gkillough: This could be a method reference.

src/main/java/com/synopsys/integration/alert/web/actions/LoginActions.java
```diff
@@ -89,6 +93,15 @@ private Authentication performLdapAuthentication(final LoginConfig loginConfig)
...
+                   .filter(role -> role.startsWith("ROLE_"))
```
gkillough: Do we have a constant for `"ROLE_"`?

PR #353, opened 2019-03-26
Bm related fields

bamandel:
Removed custom validation for related required fields and added a built in functionality into our validation.

To use this, you need to list keys associated with a field that the field requires. Only if a user puts data into that field (Or marks a checkbox as true), do we validate the other fields have values as well.


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/FieldValueModel.java
```diff
@@ -74,6 +73,14 @@ public void setIsSet(final boolean set) {
...
-        return values != null && !values.isEmpty();
+        return !getValues().isEmpty();
+    }
+
+    public boolean containsNoData() {
+        return !hasValues() && !isSet();
+    }
+
+    private void cleanValues() {
+        values = getValues().stream().filter(Objects::nonNull).filter(StringUtils::isNotBlank).collect(Collectors.toSet());
```
gkillough: I think `Objects::nonNull` is redundant because isNotBlank checks null as well.

src/test/java/com/synopsys/integration/alert/provider/polaris/descriptor/PolarisGlobalDescriptorActionApiTest.java
```diff
@@ -61,8 +62,8 @@ public void validateConfigWhenValidTest() throws Exception {
                                                             .stream()
                                                             .collect(Collectors.toMap(ConfigField::getKey, Function.identity()));
         actionApi.validateConfig(configFieldMap, fieldModel, fieldErrors);
-        assertEquals(ConfigField.REQUIRED_FIELD_MISSING, fieldErrors.get(PolarisDescriptor.KEY_POLARIS_ACCESS_TOKEN));
-        assertEquals(ConfigField.REQUIRED_FIELD_MISSING, fieldErrors.get(PolarisDescriptor.KEY_POLARIS_TIMEOUT));
+        assertNull(fieldErrors.get(PolarisDescriptor.KEY_POLARIS_ACCESS_TOKEN), "Apitoken should be populated with valid token");
```
gkillough: There should be a space between "Api" and "token". We might also want to use "API Token" instead of "Api token".

src/test/java/com/synopsys/integration/alert/provider/polaris/descriptor/PolarisGlobalDescriptorActionApiTest.java
```diff
@@ -61,8 +62,8 @@ public void validateConfigWhenValidTest() throws Exception {
...
+        assertNull(fieldErrors.get(PolarisDescriptor.KEY_POLARIS_TIMEOUT), "timeout should be populated with valid timeout");
```
gkillough: Capitalize the first letter.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionApi.java
```diff
@@ -49,21 +51,45 @@
...
+    public Boolean shouldValidateField(final FieldValueModel fieldValueModel, final String type) {
+        final Boolean isValueTrue = fieldValueModel.getValue().map(Boolean::parseBoolean).orElse(false);
+        final Boolean isCheckbox = FieldType.CHECKBOX_INPUT.getFieldTypeName().equals(type);
+        return (isValueTrue && isCheckbox) || (!fieldValueModel.containsNoData() && !isCheckbox);
+    }
+
+    public void validateRelatedFields(final ConfigField field, final FieldModel fieldModel, final Map<String, String> fieldErrors) {
+        final String key = field.getKey();
+        final Optional<FieldValueModel> optionalFieldValue = fieldModel.getField(key);
+        if (optionalFieldValue.isEmpty() || (optionalFieldValue.isPresent() && optionalFieldValue.get().containsNoData())) {
```
psantos1113: we are making sure that related fields are set here and not calling validate on the related fields.  Does it make sense to change the method name to validateAnyRelatedFieldsMissing?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -166,6 +170,22 @@ public void setSubGroup(final String subGroup) {
...
 
+    public Set<String> getRequiredRelatedFields() {
+        return requiredRelatedFields;
+    }
+
+    public void setRequiredRelatedFields(final Set<String> requiredRelatedFields) {
+        this.requiredRelatedFields = requiredRelatedFields;
+    }
+
+    public void requireField(final String configFieldKey) {
```
psantos1113: you could have this follow the builder pattern where it returns the this pointer and rather than void it returns ConfigField.  This way the requiring of the fields can be done as an inline operation rather than saving a local variable and calling the local variable requireField.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -166,6 +170,22 @@ public void setSubGroup(final String subGroup) {
...
+        requiredRelatedFields.add(configFieldKey);
+    }
+
+    public void requireFields(final Collection<String> configFieldKeys) {
```
psantos1113: See the comment above

PR #354, opened 2019-03-27
Made our tasks async

bamandel:
Noticed our tasks weren't running asynchronously causing things to get backed up. Now all tasks are async.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
@@ -60,8 +60,10 @@ public abstract class ScheduledTask implements Runnable {
     }
 
     @Override
-    public void run() {
+    public final void run() {
+        logger.info("### {} Task Started...", taskName);
         runTask();
+        logger.info("### {} Task Finished", taskName);
     }
 
     @Async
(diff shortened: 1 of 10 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
```diff
@@ -58,6 +59,14 @@ public String getTaskName() {
         return taskName;
     }
 
+    @Override
+    public void run() {
```
gkillough: Make this final and do the logging here!

src/main/java/com/synopsys/integration/alert/Application.java
```diff
@@ -114,7 +116,9 @@ public SimpleJobLauncher jobLauncher() {
 
     @Bean
     public TaskScheduler taskScheduler() {
-        return new ThreadPoolTaskScheduler();
+        final ThreadPoolTaskScheduler threadPoolTaskScheduler = new ThreadPoolTaskScheduler();
+        threadPoolTaskScheduler.setPoolSize(Runtime.getRuntime().availableProcessors());
```
gkillough: Is this restricted by the security manager? If so, we should do some sort of check and set it to a default otherwise.

PR #355, opened 2019-03-28
Added a validation check for disallowed related fields

bamandel:
Similar to the related fields functionality, you can now set fields that are not allowed if other fields are selected.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionApi.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionApi.java
@@ -61,32 +61,14 @@ public abstract class DescriptorActionApi {
             if (!fieldErrors.containsKey(key) && optionalFieldValue.isPresent()) {
                 // field is present now validate the field
                 final FieldValueModel fieldValueModel = optionalFieldValue.get();
-                if (shouldValidateField(fieldValueModel, field.getType())) {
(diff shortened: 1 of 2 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionApi.java
```diff
@@ -132,4 +114,40 @@ public FieldModel saveConfig(final FieldModel fieldModel) {
...
+    private Boolean hasValueOrChecked(final FieldValueModel fieldValueModel, final String type) {
+        final Boolean isValueTrue = fieldValueModel.getValue().map(Boolean::parseBoolean).orElse(false);
+        final Boolean isCheckbox = FieldType.CHECKBOX_INPUT.getFieldTypeName().equals(type);
+        return (isValueTrue && isCheckbox) || (!fieldValueModel.containsNoData() && !isCheckbox);
+    }
+
+    private void validateAnyRelatedFieldsMissing(final ConfigField field, final FieldModel fieldModel, final Map<String, String> fieldErrors) {
+        final String key = field.getKey();
+        final Optional<FieldValueModel> optionalFieldValue = fieldModel.getField(key);
+        if (optionalFieldValue.isEmpty() || (optionalFieldValue.isPresent() && optionalFieldValue.get().containsNoData())) {
```
gkillough: You might be able to map the Optional to `containsNoData`. Then you don't need the is present check separate.

PR #358, opened 2019-03-29
Removed isSet functionality from DB objects

bamandel:
Changed some names and made sure that we delete empty models from the DB the proper way.


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ConfigurationFieldModel.java
```diff
@@ -89,12 +85,8 @@ public void setFieldValues(final Collection<String> values) {
         return Collections.emptySet();
     }
 
-    public void setSet(final boolean set) {
-        isSet = set;
-    }
-
     public boolean isSet() {
-        return isSet;
+        return fieldValues != null;
```
gkillough: Is checking if the collection is empty an appropriate check that should be included?

PR #361, opened 2019-04-01
Pretty Print Polaris Issue Types

gkillough:


src/main/java/com/synopsys/integration/alert/provider/polaris/PolarisApiHelper.java
```diff
@@ -124,22 +126,26 @@ public PolarisApiHelper(final IssueService issueService, final RoleAssignmentsSe
...
+            final IssueResourcesSingle populatedIssueResouce = issueService.getIssueForProjectBranchAndIssueKeyWithDefaultIncluded(projectId, branchId, issueKey);
+            final Optional<IssueTypeV0Resource> optionalIssueTypeResource = issueService.getIssueTypeFromPopulatedIssueResources(populatedIssueResouce);
+            optionalIssueTypeResource.ifPresent(issueTypeResource -> {
+                final String issueType = issueTypeResource.getAttributes().getName();
+                if (!issueTypeCounts.containsKey(issueType)) {
+                    issueTypeCounts.put(issueType, 0);
+                }
+                final Integer tempCount = issueTypeCounts.get(issueType);
+                issueTypeCounts.put(issueType, tempCount + 1);
+            });
```
bamandel: When do we want to use ifPresent vs an if statement? ifPresent is nice if we have to reference the object multiple times. But here we just reference the object once and only need to call a single get. So... what do we prefer?

PR #365, opened 2019-04-02
fix create AggregateMessageContent without sub topic when necessary

psantos1113:
Added code to make the sub topic linkable item optional.  The AggregateMessageContent object has a constructor
that does not take a sub topic as a parameter.  Therefore use that constructor when a sub topic is not present.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
@@ -180,22 +180,32 @@ public abstract class MessageContentCollector {
         for (int index = 0; index < count; index++) {
             final LinkableItem topicItem = topicItems.get(index);
 
-            final LinkableItem subTopic;
+            final Optional<LinkableItem> subTopic;
             if (!subTopicItems.isEmpty()) {
-                subTopic = subTopicItems.get(index);
+                subTopic = Optional.ofNullable(subTopicItems.get(index));
             } else {
-                subTopic = new LinkableItem(null, null);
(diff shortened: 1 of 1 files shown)


src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollector.java
```diff
@@ -180,24 +180,34 @@ private JsonFieldAccessor createJsonAccessor(final List<JsonField<?>> notificati
...
+                subTopic = Optional.ofNullable(subTopicItems.get(index));
             } else {
-                subTopic = new LinkableItem(null, null);
+                subTopic = Optional.empty();
             }
 
-            final AggregateMessageContent foundContent = findTopicContent(topicItem.getName(), topicItem.getValue(), subTopic.getName(), subTopic.getValue());
+            final String subTopicName = subTopic.map(LinkableItem::getName).orElse(null);
+            final String subTopicValue = subTopic.map(LinkableItem::getValue).orElse(null);
+            final AggregateMessageContent foundContent = findTopicContent(topicItem.getName(), topicItem.getValue(), subTopicName, subTopicValue);
```
gkillough: We might want to reexamine how `findTopicContent(...)` works with null values.

psantos1113: We were creating a linkable item with null values and findTopicContent was working for the other notification types before.  I wanted to preserve null values to pass into findTopicContent in order to minimize the size and scope of the changes to avoid breaking something else inadvertently.  I agree when I saw this code with the null values I questioned it.  I think we should revisit it in the future.

PR #370, opened 2019-04-04
Temporary fix for License notifications

bamandel:
License notifications do not have groups associated with them. Because of this, we need to check if the message is about licenses then add all the emails associated with the job.

This is meant to be temporary until we rework the way providers handle the Email channel.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/DefaultEmailHandler.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/DefaultEmailHandler.java
@@ -23,6 +23,7 @@
 package com.synopsys.integration.alert.provider;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/alert/provider/DefaultEmailHandler.java
```diff
@@ -93,4 +101,32 @@ public FieldAccessor updateFieldAccessor(final AggregateMessageContent content,
         }
         return Set.of();
     }
+
+    // FIXME temporary fix for license notifications before we rewrite the way emails are handled in our workflow
+    private Set<String> licenseNotificationCheck(final AggregateMessageContent content, final FieldAccessor fieldAccessor, final boolean projectOwnerOnly) {
+        final Optional<LinkableItem> subTopic = content.getSubTopic();
+        if (subTopic.isEmpty()) {
+            final Boolean filterByProject = fieldAccessor.getBoolean(CommonDistributionConfiguration.KEY_FILTER_BY_PROJECT).orElse(false);
```
gkillough: This isn't for this pull request, but in some major version, we should have filterByProject default to true for email. Otherwise, a mistake could generate lots of unwanted emails.

src/main/js/component/audit/Index.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { BootstrapTable, ButtonGroup, TableHeaderColumn } from 'react-bootstrap-table';
-import { getAuditData } from 'store/actions/audit';
+import { getAuditData, resendNotification } from 'store/actions/audit';
```
psantos1113: Why was this added?  Was it missing from Index.js beforehand and causing an error in the javascript console?

jamesrichard91: Yes it was missing before

bamandel: This was because of the issue with My and Gavin's save actions. So I added it as I noticed it was broken when trying to test this feature.

PR #373, opened 2019-04-09
Bm fix policy message for BD

bamandel:
We now group message content based off of policy violations and component name. Component version and the policies will be grouped if there are similar results from the BlackDuckProvider. New messages can look like this:

Type: ADD
Component: GNU Compiler Collection
Component Version: [5.4.0][6.0.1]

Policy: [No GNU compiler collection][No dynamic]

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyCollector.java
@@ -24,7 +24,10 @@ package com.synopsys.integration.alert.provider.blackduck.collector;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/PolicyComponentMapping.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.synopsys.integration.alert.provider.blackduck.collector;
+
+import java.util.Set;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class PolicyComponentMapping extends AlertSerializableModel {
+    // Do not delete this member. This is used for checking equals and filtering.
+    private final String componentName;
```
bamandel: I'm not sure the best way to handle this. It's technically unused as we don't reference it directly. But it is used in our equals method. Is there a better way to do this?

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyCollector.java
```diff
@@ -44,12 +47,19 @@ public BlackDuckPolicyCollector(final JsonExtractor jsonExtractor, final List<Me
         super(jsonExtractor, messageContentProcessorList, contentTypes);
     }
 
-    protected void addApplicableItems(final List<CategoryItem> categoryItems, final Long notificationId, final LinkableItem policyItem, final String policyUrl, final ItemOperation operation, final SortedSet<LinkableItem> applicableItems) {
-        final List<String> keyItems = Stream.concat(applicableItems.stream().map(LinkableItem::getValue), Stream.of(policyUrl)).collect(Collectors.toList());
+    protected void addApplicableItems(final List<CategoryItem> categoryItems, final Long notificationId, final Set<LinkableItem> policyItems, final ItemOperation operation, final Set<LinkableItem> applicableItems) {
+        final List<String> keyItems = Stream.concat(
+            applicableItems.stream().map(LinkableItem::getValue),
```
gkillough: Is the "value" of the LinkableItem a url? If so, we name the collection in a way that makes it obvious; if not, why are we collecting a group of urls and "unknown" items?

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyLinkableItem.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+import com.synopsys.integration.alert.common.message.model.LinkableItem;
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+import com.synopsys.integration.alert.provider.blackduck.BlackDuckProviderContentTypes;
+
+public class BlackDuckPolicyLinkableItem extends AlertSerializableModel {
+    private final SortedSet<LinkableItem> linkableItems;
```
gkillough: Do we know what these LinkableItems are here? We could be more explicit if so (e.g. policyRuleLinks or something like that).

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyViolationCollector.java
```diff
@@ -72,18 +76,22 @@ protected void addCategoryItems(final List<CategoryItem> categoryItems, final Js
         final List<JsonField<PolicyInfo>> policyFields = getFieldsOfType(notificationFields, new TypeRef<PolicyInfo>() {});
         final List<JsonField<ComponentVersionStatus>> componentFields = getFieldsOfType(notificationFields, new TypeRef<ComponentVersionStatus>() {});
 
-        final List<PolicyInfo> policyItems = getFieldValueObjectsByLabel(jsonFieldAccessor, policyFields, BlackDuckProviderContentTypes.LABEL_POLICY_INFO_LIST);
+        final Map<String, PolicyInfo> policyItems = getFieldValueObjectsByLabel(jsonFieldAccessor, policyFields, BlackDuckProviderContentTypes.LABEL_POLICY_INFO_LIST).stream()
+                                                        .collect(Collectors.toMap(PolicyInfo::getPolicy, Function.identity()));
```
gkillough: Instead of creating a `Map` from the results here, why not just iterate over them in a loop?

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/PolicyComponentMapping.java
gkillough: I think we need some feedback from @ekerwin

PR #378, opened 2019-04-11
Saml fallback

psantos1113:
This branch was originally intended to fall back to our regular authentication scheme if SAML failed.  It couldn't be done as of now. 

However the authentication methods of the LoginActions were encapsulated into a AuthenticationProvider.  This makes it fit with Spring Security conventions and cleans up the code in case we need other authentication schemes from our login screen.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
+++ b/src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
@@ -1,0 +1,93 @@
+package com.synopsys.integration.alert.web.security.authentication;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.authentication.dao.DaoAuthenticationProvider;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.ldap.authentication.LdapAuthenticationProvider;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.exception.AlertLDAPConfigurationException;
+import com.synopsys.integration.alert.common.rest.model.UserModel;
+import com.synopsys.integration.alert.database.user.UserRole;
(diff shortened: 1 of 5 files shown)


src/main/java/com/synopsys/integration/alert/web/security/authentication/saml/AlertSAMLEntryPoint.java
```diff
@@ -54,6 +51,8 @@ public void commence(final HttpServletRequest request, final HttpServletResponse
             super.commence(request, response, e);
             return;
         }
-        throw new ServletException(e);
+        if (e instanceof InsufficientAuthenticationException) {
```
gkillough: Do we want to do anything with the exception if this returns false?

psantos1113: I do not believe so the ServletException was a way to force our UI to display the login screen but the log would be spammed with the exception string.  This is handled in a filter chain so we don't want the exception to bubble up.  It was bubbled up to force the Login Screen to show up because SAML wasn't enabled.  This is a better solution and I would want to make this single file change for 4.1.0 after this is review.

src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private Authentication performDatabaseAuthentication(final Authentication pendingAuthentication) {
+        logger.info("Attempting database authentication...");
+        return alertDatabaseAuthProvider.authenticate(pendingAuthentication);
+    }
+
+    private Authentication performLdapAuthentication(final Authentication pendingAuthentication) {
+        logger.info("Checking ldap based authentication...");
+        Authentication result;
```
bamandel: We could set result = pendingAuthentication right here then only add the success one if it never throws the exception

src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
```diff
@@ -0,0 +1,115 @@
...
+        return result;
+    }
+
+    private boolean isAuthorized(final Authentication authentication) {
+        final EnumSet<UserRole> allowedRoles = EnumSet.allOf(UserRole.class);
+        return authentication.getAuthorities().stream()
+                   .map(GrantedAuthority::getAuthority)
+                   .filter(role -> role.startsWith(UserModel.ROLE_PREFIX))
+                   .map(role -> StringUtils.substringAfter(role, UserModel.ROLE_PREFIX))
+                   .anyMatch(roleName -> allowedRoles.contains(UserRole.valueOf(roleName)));
```
bamandel: We COULD add one more map step here instead of calling the valueOf method inside. We would convert it to the following instead.
`.map(UserRole::valueOf)
.anyMatch(allowedRoles::contains)`

psantos1113: sounds good I will make that change

PR #379, opened 2019-04-11
JAVA: Updating the previous issue counts from the DB since we were setting them to 0 previously so all issues looked new everytime

jamesrichard91:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultPolarisIssueAccessor.java
```diff
@@ -68,6 +68,25 @@ public DefaultPolarisIssueAccessor(final PolarisIssueRepository polarisIssueRepo
                    .collect(Collectors.toList());
     }
 
+    @Override
+    public Optional<PolarisIssueModel> getProjectIssuesByIssueType(final String projectHref, final String issueType) throws AlertDatabaseConstraintException {
```
gkillough: "Issues" should become "Issue" in the method name.

PR #380, opened 2019-04-11
Bm auto test modals

bamandel:
Added a new item to descriptor metadata that will autogenerate the test modal for global configurations. All you pass this new field is the label for the destination field that you want displayed. The modal should work exactly as before.

In the future, I would like to expand this feature to accept multiple field definitions so we can have a mini testing form, but that is not necessary at the moment.


src/main/js/dynamic/GlobalConfiguration.js
```diff
@@ -10,21 +10,27 @@ import * as FieldModelUtilities from 'util/fieldModelUtilities';
 import * as DescriptorUtilities from 'util/descriptorUtilities';
 import * as FieldMapping from 'util/fieldMapping';
 import StatusMessage from 'field/StatusMessage';
+import ChannelTestModal from "../component/common/ChannelTestModal";
```
jamesrichard91: Make sure the import path is not relative and is in single quotes.

PR #382, opened 2019-04-16
Disable saml

psantos1113:
Remove saml support


psantos1113:
The DB changes to define the fields were not removed.  The database changes to define the fields are ok to leave since the fields will be used in the future and this is a last minute disable.  This reserves the configuration fields for when it is enabled again.

PR #381, opened 2019-04-15
Bm component auto ui

bamandel:
This converts BOTH scheduling and settings to auto-generated UI's. The majority of changes here are in regards to updating the Java UIConfigs to match what was previously in the UI and adding the select functionality to auto-generation.

*Important* Scheduling is not quite complete in this PR. It does not have the countdown featue it used to have. This will be resolved someway in a separate PR.

PR #383, opened 2019-04-16
Created a refresh redux action for updating only config.

bamandel:
Added a refresh function that lets us keep the page always up to date with the DB. This function is called whenever a user changes the data in the DB.


bamandel:
> This looks fine for the changes. I am assuming the UI components that use these changes are coming in another pull request. Correct?

This change was actually in regards to the other PR I had https://github.com/blackducksoftware/blackduck-alert/pull/381. The read only fields weren't being updated until I refreshed the page. Now data is always going to be refreshed.

PR #384, opened 2019-04-17
Countdown field

bamandel:
To finish the scheduling page, I needed to add a countdown timer to match our accumulator. I did that and fixed the actual countdown value. Apparently we were counting down from the BD accumulator which ran after 60 seconds but the ondemand task runs every 10 seconds.

The only issue with this change is the countdown field takes a bit longer than all the other fields to display a value for some reason. That can be fixed some time in the future if necessary.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/component/scheduling/SchedulingDescriptorActionApi.java
+++ b/src/main/java/com/synopsys/integration/alert/component/scheduling/SchedulingDescriptorActionApi.java
@@ -23,7 +23,6 @@
 package com.synopsys.integration.alert.component.scheduling;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import org.springframework.beans.factory.annotation.Autowired;
@@ -33,13 +32,12 @@ import com.synopsys.integration.alert.common.descriptor.action.NoTestActionApi;
 import com.synopsys.integration.alert.common.rest.model.FieldModel;
 import com.synopsys.integration.alert.common.rest.model.FieldValueModel;
 import com.synopsys.integration.alert.common.workflow.task.TaskManager;
-import com.synopsys.integration.alert.provider.blackduck.tasks.BlackDuckAccumulator;
 import com.synopsys.integration.alert.workflow.scheduled.PurgeTask;
 import com.synopsys.integration.alert.workflow.scheduled.frequency.DailyTask;
+import com.synopsys.integration.alert.workflow.scheduled.frequency.OnDemandTask;
 
 @Component
 public class SchedulingDescriptorActionApi extends NoTestActionApi {
-
     private final TaskManager taskManager;
 
     @Autowired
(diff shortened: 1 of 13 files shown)


src/main/java/com/synopsys/integration/alert/component/scheduling/SchedulingDescriptorActionApi.java
```diff
@@ -49,8 +47,8 @@ public SchedulingDescriptorActionApi(final TaskManager taskManager) {
 
     @Override
     public FieldModel readConfig(final FieldModel fieldModel) {
-        final Optional<Long> accumulatorNextRun = taskManager.getDifferenceToNextRun(BlackDuckAccumulator.TASK_NAME, TimeUnit.SECONDS);
-        fieldModel.putField(SchedulingDescriptor.KEY_ACCUMULATOR_NEXT_RUN, new FieldValueModel(List.of(accumulatorNextRun.map(String::valueOf).orElse("")), true));
+        final String accumulatorNextRun = taskManager.getDifferenceToNextRun(OnDemandTask.TASK_NAME, TimeUnit.SECONDS).map(String::valueOf).orElse("");
```
jamesrichard91: This won't work since the OnDemandTask does not do the accumulation from the provider(s).  We should have a countdown per provider accumulator.

bamandel: Ah, I see what I missed. I will change that to BD and Polaris. Do we want a timer for the onDemand task or no?

jamesrichard91: We do not want a timer for the on demand task. It runs frequently enough that it should fine to not display it

src/main/java/com/synopsys/integration/alert/component/scheduling/SchedulingDescriptorActionApi.java
```diff
@@ -49,8 +47,8 @@ public SchedulingDescriptorActionApi(final TaskManager taskManager) {
...
+        fieldModel.putField(SchedulingDescriptor.KEY_ACCUMULATOR_NEXT_RUN, new FieldValueModel(List.of(accumulatorNextRun), true));
```
psantos1113: We may want to iterate through the providers and have keys inserted with a prefix and the provider name i.e. 
accumulator.next.run.blackduck, accumulator.next.run.polaris

Something like that.

src/main/java/com/synopsys/integration/alert/component/scheduling/SchedulingUIConfig.java
```diff
@@ -50,13 +55,15 @@ public SchedulingUIConfig() {
 
     @Override
     public List<ConfigField> createFields() {
+        final Long countdownMax = TimeUnit.MILLISECONDS.toSeconds(OnDemandTask.DEFAULT_INTERVAL_MILLISECONDS);
+        final ConfigField accumulatorNextRun = CountdownConfigField.create(SchedulingDescriptor.KEY_ACCUMULATOR_NEXT_RUN, LABEL_ACCUMULATOR_NEXT_RUN, ACCUMULATOR_NEXT_RUN_DESCRIPTION, countdownMax);
```
psantos1113: Iterate through the providers and create a countdown field per provider accumulator.
The scheduled tasks that are related to the accumulators should implement an interface.  i.e. Accumulator which is in alert-common.  If the provider has an accumulator task then we ask the accumulator for it's name or task name to include when creating the CountdownConfigField.

bamandel: I think these are both good ideas, however.... We don't intend on keeping this page around for a super long time. The idea is to add the task manager component to the UI once that has all been implemented. Things will be done dynamically there.

If we add more providers before that happens, then I think we should make these changes. Otherwise, I think we can keep it this way for now.

PR #386, opened 2019-04-18
Decoupling channels from descriptors

bamandel:
Instead of looping through descriptors to register our channels, I now just loop through the channels. The channels are gone from all descriptors.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/channel/DistributionChannel.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/channel/DistributionChannel.java
@@ -64,7 +64,7 @@ public abstract class DistributionChannel extends MessageReceiver {
                 logger.error("There was an error sending the message.", ex);
             }
         } else {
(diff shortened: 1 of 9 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/DistributionChannel.java
```diff
@@ -57,14 +57,14 @@ public AlertProperties getAlertProperties() {
...
-        if (event.getDestination().equals(getDistributionType())) {
+        if (event.getDestination().equals(getDestinationName())) {
             try {
                 sendAuditedMessage(event);
             } catch (final IntegrationException ex) {
                 logger.error("There was an error sending the message.", ex);
             }
         } else {
-            logger.warn("Received an event of type '{}', but this channel is for type '{}'.", event.getDestination(), getDistributionType());
+            logger.warn("Received an event of type '{}', but this channel is for type '{}'.", event.getDestination(), getDestinationName());
```
gkillough: Does this log message convey the correct thing with the new variable?

PR #388, opened 2019-04-19
Add UI link to polaris notifications

gkillough:


src/main/java/com/synopsys/integration/alert/provider/polaris/tasks/PolarisProjectSyncTask.java
```diff
@@ -150,10 +151,10 @@ private ProviderProject convertToProviderProject(final PolarisApiHelper polarisA
         return polarisApiHelper.createIssueModelsForProject(projectId, projectModel.getName(), branchIds);
     }
 
-    private Set<AlertNotificationWrapper> createNotificationContentForProject(final ProviderProject project, final Set<PolarisIssueModel> projectIssues) {
+    private Set<AlertNotificationWrapper> createNotificationContentForProject(final String projectLink, final ProviderProject project, final Set<PolarisIssueModel> projectIssues) {
```
gkillough: Only the notifications need the UI link, the database should still use the API link.

PR #389, opened 2019-04-24
Saml implementation

psantos1113:
Implement the changes to enable SAML support in Alert.  Work around an issue where ActiveMQ sets a security provider that conflicts with the SAML library and causes an issue starting up the web server.


src/main/java/com/synopsys/integration/alert/component/settings/SettingsDescriptorActionApi.java
```diff
@@ -132,17 +132,22 @@ private void saveEncryptionProperties(final FieldModel fieldModel) {
...
+        final Optional<FieldValueModel> metadataBaseURLFieldValueOptional = fieldModel.getFieldValueModel(SettingsDescriptor.KEY_SAML_ENTITY_BASE_URL);
+        if (metadataURLFieldValueOptional.isPresent() && metadataEntityFieldValueOptional.isPresent() && metadataBaseURLFieldValueOptional.isPresent()) {
             final FieldValueModel metadataURLFieldValue = metadataURLFieldValueOptional.get();
+            final FieldValueModel metadataEntityFieldValue = metadataEntityFieldValueOptional.get();
+            final FieldValueModel metadataBaseUrValueModel = metadataBaseURLFieldValueOptional.get();
             final String metadataURL = metadataURLFieldValue.getValue().orElse("");
-            samlManager.updateSAMLConfiguration(samlEnabled, metadataURL);
-        }*/
+            final String entityId = metadataEntityFieldValue.getValue().orElse("");
+            final String baseUrl = metadataBaseUrValueModel.getValue().orElse("");
```
bamandel: Can we combine these three calls into the initial gets above?

build.gradle
```diff
@@ -1,7 +1,7 @@
 buildscript {
     ext {
         // if you change this version remember to change the version in the plugins closure
-        springBootVersion = '2.1.1.RELEASE'
+        springBootVersion = '2.1.4.RELEASE'
```
gkillough: Do we also need to change this in the plugins closure?

bamandel: If this is the case and they should always be the same, could we create some constant to refer to for this value?

psantos1113: I have already tried before to make a constant for this.  Unfortunately so far I have not been able to.  The buildscript part is evaluated differently than the plugins portion.

src/main/java/com/synopsys/integration/alert/component/settings/SettingsDescriptorActionApi.java
psantos1113: I will be writing more tests around the classes in a separate PR I will clean up this code then

PR #390, opened 2019-04-25
Fix issue type comparison in DB.

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultPolarisIssueAccessor.java
```diff
@@ -81,7 +81,7 @@ public DefaultPolarisIssueAccessor(final PolarisIssueRepository polarisIssueRepo
         if (optionalProjectId.isPresent()) {
             return polarisIssueRepository.findByProjectId(optionalProjectId.get())
                        .stream()
-                       .filter(entity -> issueType == entity.getIssueType())
+                       .filter(entity -> issueType.equals(entity.getIssueType()))
```
gkillough: This was the actual bug.

PR #391, opened 2019-04-25
Bm event handler

bamandel:
Long PR as DescriptorActionApis were coupled with a lot of things. I have completely removed DescriptorActionApis and replaced them with EventListeners and TestActions. All channel descriptors should now be UI based and perform very little functionality (Validation functionality is the only processing that occurs now).


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/FieldModel.java
```diff
@@ -78,6 +78,10 @@ public void putField(final String key, final FieldValueModel field) {
         keyToValues.put(key, field);
     }
 
+    public void removeField(final String key) {
+        keyToValues.remove(key);
+    }
```
bamandel: Since events now have the opportunity modify the FieldModel content, we can no longer pass unmodifiable items to keyToValues. I was not able to find a better alternative for using events to modify FieldModels.

PR #393, opened 2019-04-26
bm-api-action merging into bm-event-handler

bamandel:
Created a class called ConfigurationAction which contains all actions that can be done in regards to configuration. You can add testing functionality or config actions hoooks to this object which will automatically be picked up by ConfigActions.

No more Events... for now...


src/main/java/com/synopsys/integration/alert/web/config/FieldModelProcessor.java
```diff
@@ -56,72 +56,94 @@
...
-        return fireEvent(fieldModel, ConfigurationEventType.CONFIG_UPDATE_AFTER);
+        return retrieveApiAction(fieldModel)
+                   .map(apiAction -> apiAction.afterUpdateAction(fieldModel))
+                   .orElse(fieldModel);
     }
 
-    private FieldModel fireEvent(final FieldModel fieldModel, final ConfigurationEventType configurationEventType) {
-        configurationEventPublisher.publishConfigurationEvent(fieldModel, configurationEventType);
-        return fieldModel;
+    public Optional<ApiAction> retrieveApiAction(final FieldModel fieldModel) {
```
psantos1113: Since these retrieve methods seem to be used internally in the FieldModelProcessor and no where else make the methods private.  Or if you want to leave them available to sub-classes of FieldModelProcessor make them protected.

PR #395, opened 2019-04-29
SUMMARY Format

gkillough:
Any feedback about reducing the code footprint would be appreciated.


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private AggregateMessageContent summarize(final AggregateMessageContent message) {
+        final SortedSet<CategoryItem> originalCategoryItems = message.getCategoryItems();
+        if (null == originalCategoryItems) {
+            return message;
+        }
+
+        final Map<ItemOperation, SortedSet<CategoryItem>> itemsByOperation = sortByOperation(originalCategoryItems);
+
+        final SortedSet<CategoryItem> summarizedCategoryItems = new ConcurrentSkipListSet<>();
```
jamesrichard91: Why did you choose ConcurrentSkipListSet over TreeSet? We use TreeSet everywhere else as the type of SortedSet

PR #397, opened 2019-04-30
Encryption secrets file

psantos1113:
Look for the ALERT_ENCRYPTION_PASSWORD and ALERT_ENCRYPTION_SALT in the /run/secrets directory that docker will mount.


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/FilePersistenceUtil.java
```diff
@@ -78,6 +92,10 @@ public void delete(final String fileName) throws IOException {
     }
 
     private File createFilePath(final String fileName) {
```
jamesrichard91: Should these methods be named createFile instead?

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/FilePersistenceUtil.java
```diff
@@ -78,6 +92,10 @@ public void delete(final String fileName) throws IOException {
...
-        return new File(parentDataDirectory, fileName);
+        return createFilePath(parentDataDirectory, fileName);
+    }
+
+    private File createFilePath(final File parent, final String fileName) {
```
jamesrichard91: same as above

PR #399, opened 2019-05-02
Add collapsing to digest items and fix Slack

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/MessageContentCollapser.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.enumeration.ItemOperation;
+import com.synopsys.integration.alert.common.message.model.AggregateMessageContent;
+import com.synopsys.integration.alert.common.message.model.CategoryItem;
+import com.synopsys.integration.alert.common.message.model.CategoryKey;
+
+@Component
+public class MessageContentCollapser {
```
gkillough: This class is a one-to-one copy of the previous DigestMessageContentProcessor.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/AggregateMessageContent.java
```diff
@@ -22,37 +22,52 @@
...
         super(name, value, url);
-        this.subTopic = subTopic;
+        if (null != subTopic) {
+            this.subTopics = Set.of(subTopic);
+        } else {
+            this.subTopics = Set.of();
+        }
         this.categoryItems = categoryItems;
-        this.messageContentKey = MessageContentKey.from(name, value, subTopic.getName(), subTopic.getValue());
+        this.messageContentKey = MessageContentKey.from(name, value, subTopics);
```
bamandel: Can this constructor call this then wrap SubTopic when passing it through?

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
```diff
@@ -91,7 +102,12 @@ public void setSummarizable(final boolean summarizable) {
 
     @Override
     public int compareTo(final LinkableItem otherItem) {
-        return CompareToBuilder.reflectionCompare(this, otherItem);
+        if (collapsible && !otherItem.collapsible) {
+            return 1;
```
bamandel: Do we want to use the constant here or is 1 enough?

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageContentKey.java
```diff
@@ -44,7 +43,32 @@ public static MessageContentKey from(final String topicName, final String topicV
...
 
+    public static MessageContentKey from(final String topicName, final String topicValue, final Set<LinkableItem> subTopics) {
+        if (null == subTopics || subTopics.isEmpty()) {
+            return from(topicName, topicValue);
+        }
+
+        final StringBuilder builder = new StringBuilder();
+        builder.append(topicName);
+        builder.append(SEPARATOR);
+        builder.append(topicValue);
```
bamandel: These could be collapsed into String.format("%s%s%s", topicName, SEPARATOR, topicValue); and then appended

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageContentKey.java
```diff
@@ -44,7 +43,32 @@ public static MessageContentKey from(final String topicName, final String topicV
...
+
+        for (final LinkableItem subTopic : subTopics) {
+            builder.append(SEPARATOR);
+            builder.append(subTopic.getName());
+            builder.append(SEPARATOR);
+            builder.append(subTopic.getValue());
```
bamandel: These can probably be collapsed as well.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
```diff
@@ -23,91 +23,119 @@
...
-        for (final AggregateMessageContent topic : messages) {
-            final Map<CategoryKey, CategoryItem> categoryDataCache = new LinkedHashMap<>();
-            topic.getCategoryItems().forEach(item -> processOperation(categoryDataCache, item));
-
-            final Optional<AggregateMessageContent> collapsedContent = rebuildTopic(topic, categoryDataCache.values());
-            if (collapsedContent.isPresent()) {
-                collapsedTopicList.add(collapsedContent.get());
-            }
+        final List<AggregateMessageContent> collapsedMessages = messageContentCollapser.collapse(messages);
+        final Map<LinkableItem, List<AggregateMessageContent>> topicInfoToTopics = groupByTopic(collapsedMessages);
```
bamandel: Could this be a list of objects? It doesn't look like we use the map functionality at all here and just loop through each item.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
```diff
@@ -23,91 +23,119 @@
...
+        final List<String> keyParts = new ArrayList<>();
+        keyParts.add(categoryItem.getOperation().name());
+        for (final LinkableItem item : categoryItem.getItems()) {
+            if (!item.isCollapsible()) {
+                keyParts.add(item.getName());
+                keyParts.add(item.getValue());
+            }
         }
+
+        return CategoryKey.from("digest", keyParts);
```
bamandel: Do we have or want a constant for digest?

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
jamesrichard91: 1 should be the expected return type

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageContentKey.java
jamesrichard91: I prefer it the way it is and not collapsed

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageContentKey.java
jamesrichard91: same as above

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
jamesrichard91: I think the map is appropriate here

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
gkillough: I'm not sure. This string could be anything, my choice to use "digest" was mostly just because that was the class this was in. It will only be used internally. Does that still make it worth it?

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
```diff
@@ -91,7 +102,12 @@ public void setSummarizable(final boolean summarizable) {
...
+        }
+
+        final List<String> excludedFields = List.of("collapsible", "countable", "isNumericValue", "summarizable");
```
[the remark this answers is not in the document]
gkillough: Maybe. I'm not sure this is the best way to handle reflection equals. Any alternatives?

PR #400, opened 2019-05-03
Message Content Groups

gkillough:
One new class, 43 files affected...


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageContentGroup.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class MessageContentGroup extends AlertSerializableModel {
+    private final List<AggregateMessageContent> subContent;
+
+    private LinkableItem commonTopic;
+
+    public static MessageContentGroup singleton(final AggregateMessageContent message) {
```
jamesrichard91: is this named singleton because it creating an actual Singleton? Or because this group will be for 1 message?
If it is for a single message then I think we should change the name to singleMessage or newSingleMessage

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
```diff
@@ -51,41 +53,35 @@ public DigestMessageContentProcessor(final MessageContentCollapser messageConten
...
+            final LinkableItem topic = topicAndSubTopic.getLeft();
+            final SortedSet<CategoryItem> combinedCategoryItems = gatherCategoryItems(groupedMessageEntry.getValue());
 
-            final Set<LinkableItem> newSubTopics = gatherSubTopics(groupedMessages);
-            final SortedSet<CategoryItem> newCategoryItems = gatherCategoryItems(groupedMessages);
-
-            final AggregateMessageContent newMessage = new AggregateMessageContent(topicInfo.getName(), topicInfo.getValue(), topicInfo.getUrl().orElse(null), newSubTopics, newCategoryItems);
-            digestMessages.add(newMessage);
+            final AggregateMessageContent newMessage = new AggregateMessageContent(topic.getName(), topic.getValue(), topic.getUrl().orElse(null), topicAndSubTopic.getRight(), combinedCategoryItems);
+            messageGroups.computeIfAbsent(topic.getValue(), ignored -> new MessageContentGroup()).add(newMessage);
```
bamandel: I mentioned it on the last PR and I think it's even more true with the most recent changes, We should create an object for the map being used here. Create a list of these objects and iterate through them. Add some methods to this object that does the above functionality. It will be easier to understand imo.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
```diff
@@ -51,41 +53,35 @@ public DigestMessageContentProcessor(final MessageContentCollapser messageConten
...
         }
 
-        return digestMessages;
+        return new ArrayList<>(messageGroups.values());
     }
 
-    private Map<LinkableItem, List<AggregateMessageContent>> groupByTopic(final List<AggregateMessageContent> messages) {
-        final Map<LinkableItem, List<AggregateMessageContent>> groupedMessages = new LinkedHashMap<>();
+    private Map<Pair<LinkableItem, LinkableItem>, List<AggregateMessageContent>> groupByTopicAndSubTopicValue(final List<AggregateMessageContent> messages) {
```
bamandel: The map mentioned above is even used here which means we would get repeated use from the object.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageContentGroup.java
gkillough: For what it's worth, Java calls Collections with a single element singleton in their APIs. If that doesn't sway you I can change it. Singleton is a native java concept and distinct from the Singleton Pattern.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/DigestMessageContentProcessor.java
jamesrichard91: This could get confusing with the Objects you are using here. We may want to encapsulate the Pair of LinkableItem's in an object we define

PR #402, opened 2019-05-06
Fix License Limit Link

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
```diff
@@ -52,7 +53,10 @@ public String getValue() {
     }
 
     public Optional<String> getUrl() {
-        return Optional.ofNullable(url);
+        if (StringUtils.isNotBlank(url)) {
```
gkillough: License Limit notifications return an empty String for some of their fields. 
`Optional.ofNullable(url)` will end up wrapping an empty String in these cases.

PR #403, opened 2019-05-07
Additional roles poc

psantos1113:
Changes to store the permissions in a matrix format in the database.

Added new roles ALERT_USER and ALERT_JOB_MANAGER

Configured all the roles to have permissions associated with them in the database.  Created an AuthorizationManager component that loads the permissions for a user into a cache and removes them when the user logs out.

A new pull request will be created to use the AuthorizationManager in the application.  The scope of this pull request is to get the components and permission structure in place


alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/UserRole.java
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.enumeration;
+
+public enum UserRole {
+    ALERT_ADMIN,
+    ALERT_JOB_MANAGER,
+    ALERT_USER;
+
+    public static final String ALERT_ADMIN_TEXT = "ALERT_ADMIN";
```
jamesrichard91: Why do we need these Strings? Are these not the same as the enum.toString()?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
```diff
@@ -0,0 +1,137 @@
...
+            userRoleRepository.saveAll(roleRelations);
+        }
+    }
+
+    private Optional<String> getRoleName(final Long roleId) {
+        return roleRepository.findById(roleId).map(RoleEntity::getRoleName);
+    }
+
+    @Override
+    public PermissionMatrixModel mergePermissionsForRoles(final Collection<String> roleNames) {
```
jamesrichard91: I dont understand this method name. What is it merging?

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/PermissionKeys.java
```diff
@@ -0,0 +1,46 @@
...
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.enumeration;
+
+public enum PermissionKeys {
+    CONFIG_PROVIDER_BLACKDUCK("config.provider.blackduck"),
+    CONFIG_PROVIDER_POLARIS("config.provider.polaris"),
+    CONFIG_CHANNEL_EMAIL("config.channel.email"),
+    CONFIG_CHANNEL_HIPCHAT("config.channel.hipchat"),
+    CONFIG_CHANNEL_SLACK("config.channel.slack"),
```
bamandel: Are we planning on adding permissions specifically for each channel/provider? I'm not sure how this will be implemented, but is there a was we could make this dynamic?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
```diff
@@ -0,0 +1,137 @@
...
+        return userRoles;
+    }
+
+    @Override
+    public void updateUserRoles(final Long userId, final Collection<UserRoleModel> roles) {
+        if (null != userId) {
+            userRoleRepository.deleteAllByUserId(userId);
+        }
+
+        if (null != roles) {
```
bamandel: Is there a specific reason we do 2 different operations depending on the data passed us? This could cause some confusing functionality. You accidentally only pass a UserId and a null roles and now we've randomly deleted all userRoles in an update method. I would prefer we check for null in a single if statement then do all our steps.

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/UserRole.java
psantos1113: In the integration tests the @WithMockUser annotation must use a constant.  I can remove these string literals in here and refactor the tests to reference a constant class just for the tests to clean this up.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
psantos1113: it merges the permissions for all the roles specified into one PermissionMatrixModel.  It's a superset of the permissions.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/UserModel.java
```diff
@@ -67,17 +69,22 @@ public String getEmailAddress() {
...
+        return roles.stream()
+                   .map(UserRoleModel::getName)
+                   .map(role -> ROLE_PREFIX + role)
+                   .map(SimpleGrantedAuthority::new)
+                   .collect(Collectors.toList());
     }
 
     public boolean hasRole(final String role) {
-        return roles.contains(role);
+        //Make this performs better.
```
gkillough: If this is the solution for now, add a TODO so we don't lose track of it. Also: "performs" -> "perform".

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/PermissionKeys.java
psantos1113: With permissions the UI needs to be changed to look at these permissions keys and the backend as well when displaying descriptors or enabling or disabling functionality in the UI.  Therefore you have to a have a fixed set of permission keys.  I suppose a descriptor could provide permission keys that it cares about to be a bit more dynamic, but audit right now isn't a descriptor and we want to have permission to perform actions with the audit screen.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
psantos1113: Need to make a change there to make it better.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    private Optional<String> getRoleName(final Long roleId) {
+        return roleRepository.findById(roleId).map(RoleEntity::getRoleName);
+    }
+
+    @Override
+    public PermissionMatrixModel mergePermissionsForRoles(final Collection<String> roleNames) {
+        final List<RoleEntity> roles = roleRepository.findRoleEntitiesByRoleName(roleNames);
```
bamandel: Can we change this method name to findRoleEntitiesByRoleNames? Just to clarify the collection aspect of it.

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/PermissionKeys.java
bamandel: Sounds like a possible future improvement then.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
gkillough: getPermissionsForRoleNames ?

src/main/resources/db/changelog/alert/4.2.0/changelog-authentication.xml
```diff
@@ -0,0 +1,538 @@
...
+                baseTableName="PERMISSION_MATRIX"
+                baseColumnNames="ACCESS_OPERATION_ID"
+                constraintName="FK_PERMISSION_OPERATION"
+                referencedTableSchemaName="ALERT"
+                referencedTableName="ACCESS_OPERATIONS"
+                referencedColumnNames="ID"
+                onDelete="CASCADE"
+        />
+    </changeSet>
+    <changeSet author="psantos" id="1556732237461-5">
```
gkillough: I think some stored procedures could help readability here rather than 4 lines per insert with only one value changing. It might also help to split out table creation and table initialization.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthorizationUtility.java
psantos1113: I can change it I had readPermissionsForRoles but it takes the permissions for all the roles and creates a superset for the roles specified.  I didn't think read was the best name given it was putting the permissions together.  I'm not sold on the name.  I'm happy to change it to something clearer.

src/main/resources/db/changelog/alert/4.2.0/changelog-authentication.xml
psantos1113: I will look into this.  I was thinking about it beforehand.  The matrix is slightly different between roles.

PR #404, opened 2019-05-07
Decouple providers and descriptors

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/provider/Provider.java
```diff
@@ -43,11 +45,14 @@ public String getName() {
 
     public abstract void destroy();
 
-    public abstract Set<ProviderContentType> getProviderContentTypes();
+    public abstract Set<MessageContentCollector> createTopicCollectors();
 
-    public abstract Set<FormatType> getSupportedFormatTypes();
+    public ProviderContent getProviderContent() {
+        return providerContent;
+    }
```
bamandel: Should we add a helper method called getSupportedFormatTypes here to add some convenience?

gkillough: I'd rather keep the Provider API small. Turns out we only retrieve this object like 3 places.

PR #406, opened 2019-05-08
Bm dynamic distribution

bamandel:
Deleted old Distribution job code and replaced it with the new dynamic generation. If you now add any new channel or provider or component, things should all automatically be added to the UI. yay!

The biggest issue with these changes is a slightly slower response time in the UI. But nothing unbearable.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -50,48 +46,61 @@
...
-        return Set.of();
+    public String getChannelName() {
+        return channelName;
+    }
+
+    private Collection<LabelValueSelectOption> getProviderLabelValues() {
+        return descriptorMap.getDescriptorByType(DescriptorType.PROVIDER).stream()
+                   .map(descriptor -> descriptor.getMetaData(ConfigContextEnum.DISTRIBUTION))
+                   .filter(Optional::isPresent)
+                   .map(Optional::get)
```
gkillough: The lines 
```java
.filter(Optional::isPresent)
.map(Optional::get)
```
should become
```java
.flatMap(Optional::stream)
```

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -50,48 +46,61 @@
...
+                   .map(descriptorMetadata -> new LabelValueSelectOption(descriptorMetadata.getLabel(), descriptorMetadata.getName(), descriptorMetadata.getFontAwesomeIcon()))
+                   .collect(Collectors.toSet());
+    }
+
+    private Collection<LabelValueSelectOption> getChannelLabelValues() {
+        return descriptorMap.getDescriptorByType(DescriptorType.CHANNEL).stream()
+                   .map(descriptor -> descriptor.getUIConfig(ConfigContextEnum.DISTRIBUTION))
+                   .filter(Optional::isPresent)
+                   .map(Optional::get)
```
gkillough: Ditto

src/main/java/com/synopsys/integration/alert/channel/email/descriptor/EmailDistributionUIConfig.java
```diff
@@ -26,34 +26,33 @@
...
+        super(EmailChannel.COMPONENT_NAME, EmailDescriptor.EMAIL_LABEL, EmailDescriptor.EMAIL_URL, EmailDescriptor.EMAIL_ICON, descriptorMap);
     }
 
     @Override
     public List<ConfigField> createChannelDistributionFields() {
-        final ConfigField emailAddresses = TextInputConfigField.create(EmailDescriptor.KEY_EMAIL_ADDRESSES, LABEL_EMAIL_ADDRESSES, EMAIL_ADDRESSES_DESCRIPTION);
         final ConfigField subjectLine = TextInputConfigField.create(EmailDescriptor.KEY_SUBJECT_LINE, LABEL_SUBJECT_LINE, EMAIL_SUBJECT_LINE_DESCRIPTION);
         final ConfigField projectOwnerOnly = CheckboxConfigField.create(EmailDescriptor.KEY_PROJECT_OWNER_ONLY, LABEL_PROJECT_OWNER_ONLY, EMAIL_PROJECT_OWNER_ONLY_DESCRIPTION);
-        return Arrays.asList(emailAddresses, subjectLine, projectOwnerOnly);
+        return Arrays.asList(subjectLine, projectOwnerOnly);
```
gkillough:
```java
List.of(...)
```

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -50,48 +46,61 @@
...
     private static final String LABEL_PROVIDER_NAME = "Provider Type";
 
     private static final String DESCRIPTION_NAME = "The name of the distribution job. Must be unique.";
     private static final String DESCRIPTION_FREQUENCY = "Select how frequently this job should check for notifications to send.";
     private static final String DESCRIPTION_CHANNEL_NAME = "Select the channel. Notifications generated through Alert will be sent through this channel.";
     private static final String DESCRIPTION_PROVIDER_NAME = "Select the provider. Only notifications for that provider will be processed in this distribution job.";
 
-    private final Logger logger = LoggerFactory.getLogger(ChannelDistributionUIConfig.class);
-    private final DescriptorAccessor descriptorAccessor;
+    private final DescriptorMap descriptorMap;
```
gkillough: I thought we were trying to get rid of DescriptorMap?

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -0,0 +1,315 @@
...
+        const jsonBody = this.buildJsonBody();
+        if (jobId) {
+            Object.assign(jsonBody, { jobId });
+            this.props.updateDistributionJob(jsonBody);
+        } else {
+            this.props.saveDistributionJob(jsonBody);
+        }
+    }
+
+    // FIXME this function is probably not necessary
```
gkillough: FYI

src/main/js/distribution/Index.js
```diff
@@ -102,52 +99,37 @@ class Index extends Component {
...
+            <DistributionConfiguration
+                projects={this.state.projects}
+                handleCancel={this.cancelRowSelect}
+                onModalClose={() => {
+                    this.props.fetchDistributionJobs();
+                    this.cancelRowSelect();
+                    onModalClose();
+                }}
+                onSave={this.saveBtn}
+                columns={columns}
```
jamesrichard91: columns doesn't appear to be one of the properties on DistributionConfiguration. This should be removed until it is supported in DistributionConfiguration

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -0,0 +1,315 @@
...
+        if (selectedChannelOption && prevChannelName !== selectedChannelOption) {
+            const newChannel = this.props.descriptors.find(descriptor => descriptor.name === selectedChannelOption && descriptor.context === DescriptorUtilities.CONTEXT_TYPE.DISTRIBUTION);
+            const channelKeys = FieldMapping.retrieveKeys(newChannel.fields);
+            const emptyChannelConfig = FieldModelUtilities.createEmptyFieldModel(channelKeys, newChannel.context, newChannel.name);
+            const updatedChannelConfig = {};
+            Object.assign(updatedChannelConfig, FieldModelUtilities.updateFieldModelSingleValue(emptyChannelConfig, KEY_CHANNEL_NAME, selectedChannelOption));
+            const name = FieldModelUtilities.getFieldModelSingleValue(channelConfig, KEY_NAME);
+            const frequency = FieldModelUtilities.getFieldModelSingleValue(channelConfig, KEY_FREQUENCY);
+            const provider = FieldModelUtilities.getFieldModelSingleValue(channelConfig, KEY_PROVIDER_NAME);
+            Object.assign(updatedChannelConfig, FieldModelUtilities.updateFieldModelSingleValue(updatedChannelConfig, KEY_NAME, name));
```
jamesrichard91: Do these Object.assign(...)'s need to be here? You are ignoring the object they return. I would expect something like 
const thing = Object.assign(...)

bamandel: The first argument passed is the one that gets updated. with all the other objects that are passed. So I'm updating the updatedChannelConfig object

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
bamandel: Oh wow, I didn't think of that as an alternative.

src/main/js/util/fieldMapping.js
```diff
@@ -8,46 +9,75 @@ import CheckboxInput from 'field/input/CheckboxInput';
...
+function buildNumberInput(items) {
+    const trimmedValue = trimValue(items);
+    return <NumberInput {...trimmedValue} />;
 }
 
-export function buildCheckboxInput(items) {
+function buildCheckboxInput(items) {
     const { value } = items;
     const checkedValue = value.toString().toLowerCase() === 'true';
     Object.assign(items, { isChecked: checkedValue });
```
jamesrichard91: Why are we ignoring the return type when doing Object.assign?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
bamandel: We were trying to get rid of descriptorMap and replace it with the DB calls, but for the auto-generated UI we actually needed a bit more data from the descriptor such as name, label and fontawesome icon. If we stored this data in the DB, then that'd be it. But unfortunately we do and I don't think that we should.

src/main/js/dynamic/DistributionConfiguration.js
jamesrichard91: That may be a side effect of the Object.assign() but it is bad practice to rely on that. You should be using the returned object as the new "complete" object rather than expecting the target to be updated. Unless Object.assign specifically documents that this is alright to do.

src/main/js/dynamic/DistributionConfiguration.js
bamandel: I will probably clean up this and other redundant code in a later PR. There is already a lot going on here.

PR #414, opened 2019-05-10
JS code cleanup

bamandel:


src/main/js/field/input/NumberInput.js
```diff
@@ -24,13 +24,13 @@ NumberInput.propTypes = {
     readOnly: PropTypes.bool,
     inputClass: PropTypes.string,
     name: PropTypes.string,
-    value: PropTypes.number,
+    value: PropTypes.string,
```
gkillough: Will this affect the way the field gets displayed?

bamandel: From what I've seen so far, no.

src/main/js/util/fieldMapping.js
```diff
@@ -13,18 +13,21 @@ import DescriptorOption from 'component/common/DescriptorOption';
 
 function trimValue(items) {
     const { value } = items;
-    const trimmedValue = (Array.isArray(value)) ? value[0] : value;
+    return (Array.isArray(value)) ? value[0] : value;
```
jamesrichard91: What is the point of this? The value is never trimmed, it just returns the first value if its an array? Otherwise return the value?

src/main/js/util/fieldMapping.js
```diff
@@ -13,18 +13,21 @@ import DescriptorOption from 'component/common/DescriptorOption';
...
+}
 
+function insertTrimmedValue(items) {
+    const trimmedValue = trimValue(items);
     return Object.assign(items, { value: trimmedValue });
```
jamesrichard91: If I understand the trimValue method, then this is adding the same value into items that was already there?

src/main/js/util/fieldMapping.js
bamandel: The name trim can probably be changed, but the intent of this function is to handle the arrays we use on the backend. All our fields expect a certain type of data to be passed to it, but the backend always returns an array of strings. If a field uses this function, it will confirm we have an array of strings then grab the first one (assuming there's only one item). If it's not an array, then it's assumed we have the exact value we need and we return it.

src/main/js/util/fieldMapping.js
bamandel: What happens here is we get the single item from an array or we get the single value already passed to us. Then we insert it into the old items object and overwrite the old value with this new, potentially changed, value.

src/main/js/util/fieldMapping.js
jamesrichard91: I think a name change would fix my confusion here. Maybe, extractSingleValue?

src/main/js/util/fieldMapping.js
jamesrichard91: I think we may want to rename this to reduce confusion. overwriteSingleValue?

PR #415, opened 2019-05-13
Ps authorization checks

psantos1113:
Changes to the database to update the permissions.   Added stored procedures to help with the assignment of the roles and permissions.

Updated the controllers to use the AuthorizationManager to check permissions.


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/AuthorizationUtil.java
```diff
@@ -29,6 +29,8 @@
 import com.synopsys.integration.alert.common.persistence.model.UserRoleModel;
 
 public interface AuthorizationUtil {
+    Set<UserRoleModel> createRoleModels();
```
psantos1113: I'm thinking this might need a different name.

src/main/java/com/synopsys/integration/alert/web/config/ConfigController.java
```diff
@@ -88,6 +94,9 @@ public ConfigController(final ConfigActions configActions, final ContentConverte
     @GetMapping("/{id}")
     public ResponseEntity<String> getConfig(@PathVariable final Long id) {
         final Optional<FieldModel> optionalModel;
+        if (!authorizationManager.hasAnyConfigReadPermissions()) {
```
jamesrichard91: Dont we also need to know the context and descriptor name in order to properly handle permissions here? So if they dont have permission to the global email configuration but they do have permission to the global Black Duck configuration, then they would have access to retrieve both as long as they know the id's.
This will be relevant when User's can define their own roles/permission matrix

src/main/java/com/synopsys/integration/alert/web/controller/AboutController.java
```diff
@@ -41,12 +42,14 @@
     private final AboutActions aboutActions;
     private final ResponseFactory responseFactory;
     private final ContentConverter contentConverter;
+    private final AuthorizationManager authorizationManager;
```
jamesrichard91: This class does not appear to use AuthorizationManager

src/main/java/com/synopsys/integration/alert/web/security/authorization/AuthorizationManager.java
```diff
@@ -0,0 +1,142 @@
...
+import com.synopsys.integration.alert.common.enumeration.ConfigContextEnum;
+import com.synopsys.integration.alert.common.enumeration.PermissionKeys;
+import com.synopsys.integration.alert.common.persistence.accessor.AuthorizationUtil;
+import com.synopsys.integration.alert.common.persistence.model.PermissionMatrixModel;
+import com.synopsys.integration.alert.common.persistence.model.UserModel;
+import com.synopsys.integration.alert.common.persistence.model.UserRoleModel;
+
+@Component
+public class AuthorizationManager {
+
```
gkillough: :(

src/main/java/com/synopsys/integration/alert/web/controller/AboutController.java
psantos1113: I just removed the member variable.

PR #418, opened 2019-05-15
Updating real time handling

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/event/EventListenerConfigurer.java
```diff
@@ -34,31 +32,28 @@
...
     }
 
     @Override
     public void configureJmsListeners(final JmsListenerEndpointRegistrar registrar) {
         logger.info("Registering JMS Listeners");
-        channelDescriptorList.forEach(descriptor -> {
-            final MessageListener channelListener = descriptor.getChannelListener();
-            if (channelListener != null) {
-                final String listenerId = createListenerId(descriptor.getName());
+        alertEventListeners.forEach(listener -> {
```
gkillough: Seeing this now, we may want a traditional for-loop here since `.forEach(...)` doesn't buy us anything.

src/main/java/com/synopsys/integration/alert/workflow/processor/NotificationReceiver.java
```diff
@@ -0,0 +1,84 @@
...
+    public NotificationReceiver(final Gson gson, final NotificationManager notificationManager, final NotificationProcessor notificationProcessor, final EventManager eventManager) {
+        super(gson, NotificationEvent.class);
+        this.notificationManager = notificationManager;
+        this.notificationProcessor = notificationProcessor;
+        this.eventManager = eventManager;
+    }
+
+    @Override
+    public void handleEvent(final NotificationEvent event) {
+        if (event.getDestination().equals(NotificationEvent.NOTIFICATION_EVENT_TYPE)) {
```
gkillough: I prefer seeing the constant call `.equals()` so we don't have to worry about NPEs.
```java
NotificationEvent.NOTIFICATION_EVENT_TYPE.equals(event.getDestination())
```

alert-common/src/main/java/com/synopsys/integration/alert/common/event/EventManager.java
```diff
@@ -20,30 +20,29 @@
...
-public class ChannelTemplateManager {
+public class EventManager {
     private final JmsTemplate jmsTemplate;
-    private final DefaultAuditUtility auditUtility;
+    private final AuditUtility auditUtility;
     private final ContentConverter contentConverter;
 
     @Autowired
-    public ChannelTemplateManager(final ContentConverter contentConverter, final DefaultAuditUtility auditUtility, final JmsTemplate jmsTemplate) {
+    public EventManager(final ContentConverter contentConverter, @Lazy final AuditUtility auditUtility, final JmsTemplate jmsTemplate) {
```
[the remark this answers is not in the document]
jamesrichard91: EventManager -->  AuditUtility (DefaultAuditUtility) --> DefaultNotificationManager --> EventManager

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
```diff
@@ -57,17 +61,27 @@
...
-    public AlertNotificationWrapper saveNotification(final AlertNotificationWrapper notification) {
-        return notificationContentRepository.save((NotificationContent) notification);
+    public List<AlertNotificationWrapper> saveAllNotifications(final Collection<AlertNotificationWrapper> notifications) {
+        final List<AlertNotificationWrapper> notificationContents = notifications.stream()
+                                                                        .map(notification -> notificationContentRepository.save((NotificationContent) notification))
+                                                                        .collect(Collectors.toList());
+        final List<Long> notificationIds = notificationContents.stream()
+                                               .map(AlertNotificationWrapper::getId)
+                                               .collect(Collectors.toList());
+        eventManager.sendEvent(new NotificationEvent(notificationIds));
```
bamandel: I feel that this functionality doesn't exactly fit here. We should do our save then if people want, the user can send the event. This would also help remove the circular dependency caused by Autowiring the EventManager.

src/main/java/com/synopsys/integration/alert/workflow/processor/NotificationProcessor.java
```diff
@@ -52,12 +55,18 @@ public NotificationProcessor(final MessageContentAggregator messageContentAggreg
     }
 
     public List<DistributionEvent> processNotifications(final FrequencyType frequencyType, final List<AlertNotificationWrapper> notificationList) {
-        final Map<CommonDistributionConfiguration, List<AggregateMessageContent>> messageContentList = messageContentAggregator.processNotifications(frequencyType, notificationList);
-        return notificationToEventConverter.convertToEvents(messageContentList);
+        logger.info("Notifications to Process: {}", notificationList.size());
+        if (notificationList.isEmpty()) {
+            return List.of();
+        } else {
```
bamandel: Removing else would make everything soooo much prettier.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
jamesrichard91: The problem is, if we don't do it here then anywhere we call saveAllNotifications needs access to the EventManager AND they would have this same duplicated code
`
 final List<Long> notificationIds = notificationContents.stream()
                                               .map(AlertNotificationWrapper::getId)
        eventManager.sendEvent(new NotificationEvent(notificationIds));
`

PR #420, opened 2019-05-17
Feat(Java): Added caching to prevent requesting the issue type information for subtools we have already seen.

jamesrichard91:


src/main/java/com/synopsys/integration/alert/provider/polaris/PolarisApiHelper.java
```diff
@@ -138,24 +138,35 @@ public PolarisApiHelper(final IssueService issueService, final RoleAssignmentSer
...
+            } else {
+                final String issueKey = queryIssue.getAttributes().getIssueKey();
+                final IssueResourcesSingle populatedIssueResource = issueService.getIssueForProjectBranchAndIssueKeyWithDefaultIncluded(projectId, branchId, issueKey);
+                final String issueType = issueService
+                                             .getIssueTypeFromPopulatedIssueResources(populatedIssueResource)
+                                             .map(IssueTypeV0Resource::getAttributes)
+                                             .map(IssueTypeV0Attributes::getName)
+                                             .orElse(subTool);
+                issueTypeMap.put(subTool, issueType);
+                updateIssueTypeCounts(issueTypeCounts, issueType);
```
gkillough: We would only have to call this once if we moved the issueTypes variable before the if-statement.

jamesrichard91: I updated the code, so we only update the counts in one place

PR #422, opened 2019-05-17
Separating distribution event handling out of EventManager into ChannelEventManager.

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/event/EventManager.java
```diff
@@ -23,28 +23,22 @@
...
 
 @Component
 public class EventManager {
     private final JmsTemplate jmsTemplate;
-    private final AuditUtility auditUtility;
     private final ContentConverter contentConverter;
 
     @Autowired
-    public EventManager(final ContentConverter contentConverter, @Lazy final AuditUtility auditUtility, final JmsTemplate jmsTemplate) {
+    public EventManager(final ContentConverter contentConverter, final JmsTemplate jmsTemplate) {
```
gkillough: Where are we using this vanilla version of the class? Can it be abstract? Does it need to be a component still?

jamesrichard91: This is current used for NotificationEvent's but it could be used in the future for all non-channel events

PR #423, opened 2019-05-17
Improve query performance

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
```diff
@@ -98,9 +99,11 @@ public DefaultNotificationManager(final NotificationContentRepository notificati
     public Page<AlertNotificationWrapper> findAllWithSearch(final String searchTerm, final PageRequest pageRequest, final boolean onlyShowSentNotifications) {
         final String lcSearchTerm = searchTerm.toLowerCase(Locale.ENGLISH);
         if (onlyShowSentNotifications) {
-            return notificationContentRepository.findMatchingSentNotification(lcSearchTerm, pageRequest);
+            final Page<NotificationContent> matchingSentNotification = notificationContentRepository.findMatchingSentNotification(lcSearchTerm, pageRequest);
+            return safelyConvertToGenericPage(matchingSentNotification);
```
psantos1113: can we just call the return at the end since in both the if and else we call safelyConvertToGenericPage.  Move the Page<NotificationContent> outside the if else statement.

PR #424, opened 2019-05-20
Ps ui authorization

psantos1113:
update the permission keys so that the keys can be generated in code.  Add more logic to restrict the controllers.  Update the UI to display components based on the permissions of the descriptor.


src/main/js/util/fieldMapping.js
```diff
@@ -86,6 +102,10 @@ function buildCounterField(items, field) {
     return <CounterField {...items} />;
 }
 
+function convertStringtoBoolean(value) {
```
gkillough: Capitalize the 't' in 'to'.

PR #429, opened 2019-05-23
Ps icon refactor

psantos1113:
- Update the Font Awesome icon usage to the FontAwesomeIcon React component to make it consistent. 
- Change the icon paths for slack and jira to have the icon set name before the icon name i.e. 'fab/sack' and 'fab/jira' 
- Create a utility to construct the array that FontAwesomeIcon needs to load branded icons correctly.
- Created the alert-icon CSS class to add space between the FontAwesomeIcons and text elements.

### DIFF
--- a/src/main/js/distribution/ProjectConfiguration.js
+++ b/src/main/js/distribution/ProjectConfiguration.js
@@ -3,13 +3,19 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { BootstrapTable, TableHeaderColumn } from 'react-bootstrap-table';
 import { getProjects } from 'store/actions/projects';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 function assignClassName(row, rowIdx) {
     return 'tableRow';
 }
 
 function assignDataFormat(cell, row) {
-    const cellContent = (row.missing) ? <span className="missingBlackDuckData"><span className="fa fa-exclamation-triangle fa-fw" aria-hidden="true" />{cell}</span> : cell;
+    const cellContent = (row.missing) ? <span className="missingBlackDuckData">
+        <span className="fa-layers fa-fw">
+            <FontAwesomeIcon icon="exclamation-triangle" className="alert-icon" size="lg" />{cell}
+        </span> </span>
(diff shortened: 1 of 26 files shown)


src/main/js/distribution/ProjectConfiguration.js
```diff
@@ -3,13 +3,19 @@ import PropTypes from 'prop-types';
...
 }
 
 function assignDataFormat(cell, row) {
-    const cellContent = (row.missing) ? <span className="missingBlackDuckData"><span className="fa fa-exclamation-triangle fa-fw" aria-hidden="true" />{cell}</span> : cell;
+    const cellContent = (row.missing) ? <span className="missingBlackDuckData">
+        <span className="fa-layers fa-fw">
+            <FontAwesomeIcon icon="exclamation-triangle" className="alert-icon" size="lg" />{cell}
+        </span> </span>
+        : cell;
+
```
bamandel: Can we format this differently? It's a bit hard to read at the moment. I'd recommend:
`const cellContent = (row.missing) ?
    <span className="missingBlackDuckData">
        <span className="fa-layers fa-fw">
            <FontAwesomeIcon icon="exclamation-triangle" className="alert-icon" size="lg" />{cell}
        </span>
     </span>
        : cell;`

Or something similar. It's a very large ternary operator, so even a variable for the HTML might be a better option.

bamandel: That did not format at all. So I think another variable for the HTML might make this easier to read.

src/main/js/util/iconUtility.js
```diff
@@ -0,0 +1,19 @@
...
+    if (iconPathData) {
+        path = ['fas'];
+        if (iconPathData.includes('/')) {
+            path = iconPathData.split('/');
+        } else {
+            path.push(iconPathData);
+        }
+    }
+    return path;
+}
```
bamandel: Another option is to create our own fontawesome component that wraps the original and adds this functionality when displaying the field.

psantos1113: I tried that last week and was concerned about the number of properties our wrapper would need to support.  I have another idea.

PR #430, opened 2019-05-28
Bm startup refactor

bamandel:
Removed a lot of code from StartupManager in favor of our new StartupComponent classes which allows people to insert code for startup. These components can have an Order annotation to denote what order the items should be loaded into the list.


src/main/java/com/synopsys/integration/alert/workflow/startup/component/StartupComponent.java
```diff
@@ -0,0 +1,39 @@
...
+public abstract class StartupComponent {
+    private final Logger logger = LoggerFactory.getLogger(StartupComponent.class);
+
+    public void initializeComponent() {
+        final String runningStartupComponentLog = String.format("Running startup component: %s", getClass().getSimpleName());
+        logger.info(runningStartupComponentLog);
+        run();
+    }
+
+    public abstract void run();
```
psantos1113: can you make this protected since we want initializeComponent to be the public interface.  Also change the method from run to initialize.  Run makes me think of it as a thread which it is not at the moment.

src/main/java/com/synopsys/integration/alert/workflow/startup/component/CronJobsStartupComponent.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private final ConfigurationAccessor configurationAccessor;
+    private final TaskManager taskManager;
+    private final PurgeTask purgeTask;
+    private final DailyTask dailyTask;
+    private final PhoneHomeTask phoneHomeTask;
+    private final UpdateNotifierTask updateNotifierTask;
+
+    @Autowired
+    public CronJobsStartupComponent(final ConfigurationAccessor configurationAccessor, final TaskManager taskManager, final PurgeTask purgeTask, final DailyTask dailyTask, final PhoneHomeTask phoneHomeTask,
```
gkillough: I think we should have a method on ScheduledTask along the lines of `initializeOnStartup()` or something. Then we can just autowire all of the ScheduledTasks and call that method; if they get scheduled some time after startup, the method could just do nothing.

psantos1113: Since there is no order on this task then yes that would make sense to get all the tasks.  Filter the list of tasks to only the list where the initializeOnStartup returns true and then schedule them.

PR #431, opened 2019-05-28
Ps jira distribution fields

psantos1113:
Add the distribution fields for the Jira Channel. 
Fix some permission issues.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDescriptor.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDescriptor.java
@@ -34,9 +34,9 @@ public class JiraDescriptor extends ChannelDescriptor {
     public static final String KEY_JIRA_USERNAME = "jira.user.name";
     public static final String KEY_JIRA_ACCESS_TOKEN = "jira.access.token";
 
-    public static final String KEY_ADD_COMMENTS = "jira.add.comments";
-    public static final String KEY_ISSUE_CREATOR = "jira.issue.creator";
-    public static final String KEY_JIRA_PROJECT_NAME = "jira.project.name";
+    public static final String KEY_ADD_COMMENTS = "channel.jira.add.comments";
+    public static final String KEY_ISSUE_CREATOR = "channel.jira.issue.creator";
+    public static final String KEY_JIRA_PROJECT_NAME = "channel.jira.project.name";
 
     public static final String JIRA_LABEL = "Jira Cloud";
     public static final String JIRA_URL = "jira";
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDescriptor.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDescriptor.java
@@ -34,9 +34,9 @@ public class JiraDescriptor extends ChannelDescriptor {
     public static final String KEY_JIRA_USERNAME = "jira.cloud.user.name";
     public static final String KEY_JIRA_ACCESS_TOKEN = "jira.cloud.access.token";
 
-    public static final String KEY_ADD_COMMENTS = "channel.cloud.jira.add.comments";
-    public static final String KEY_ISSUE_CREATOR = "channel.cloud.jira.issue.creator";
-    public static final String KEY_JIRA_PROJECT_NAME = "channel.cloud.jira.project.name";
+    public static final String KEY_ADD_COMMENTS = "channel.jira.cloud.add.comments";
+    public static final String KEY_ISSUE_CREATOR = "channel.jira.cloud.issue.creator";
+    public static final String KEY_JIRA_PROJECT_NAME = "channel.jira.cloud.project.name";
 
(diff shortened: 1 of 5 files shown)


src/main/resources/db/changelog/alert/5.0.0/changelog-configuration-initialization.xml
```diff
@@ -22,4 +22,16 @@
             CALL DEFINE_FIELD('jira.access.token', TRUE, 'channel_jira', 'GLOBAL');
         </sql>
     </changeSet>
+    <!-- Register Distribution Jira Cloud Fields. -->
+    <changeSet author="psantos" id="1558524516692-3">
+        <sql dbms="h2" stripComments="true">
+            CALL DEFINE_FIELD('jira.add.comments', FALSE, 'channel_jira', 'DISTRIBUTION');
```
psantos1113: For these distribution specific fields do we want channel. as the prefix?  i.e. channel.jira.add.comments
For the HipChat and Slack distribution fields we do that.  So to be consistent we can do that here as well. We also have the FIELD_CONTEXTS table that creates the mapping between the field ID and the Context.

gkillough: I think we should add the prefix for consistency.

bamandel: I would like this as we could have a Jira provider in the future and the prefix would help avoid conflicts.

psantos1113: ok no problem.  I prefer the consistency as well but wanted to make sure I wasn't being too particular.

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDescriptor.java
```diff
@@ -30,9 +30,13 @@
...
-    public static final String KEY_JIRA_URL = "jira.url";
-    public static final String KEY_JIRA_USERNAME = "jira.user.name";
-    public static final String KEY_JIRA_ACCESS_TOKEN = "jira.access.token";
+    public static final String KEY_JIRA_URL = "jira.cloud.url";
+    public static final String KEY_JIRA_USERNAME = "jira.cloud.user.name";
+    public static final String KEY_JIRA_ACCESS_TOKEN = "jira.cloud.access.token";
+
+    public static final String KEY_ADD_COMMENTS = "channel.cloud.jira.add.comments";
+    public static final String KEY_ISSUE_CREATOR = "channel.cloud.jira.issue.creator";
+    public static final String KEY_JIRA_PROJECT_NAME = "channel.cloud.jira.project.name";
```
jamesrichard91: Why are these field cloud.jira when the other fields have jira.cloud?

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDistributionUIConfig.java
```diff
@@ -30,18 +30,32 @@
...
 
 @Component
 public class JiraDistributionUIConfig extends ChannelDistributionUIConfig {
+
+    public static final String LABEL_ADD_COMMENTS = "Add Comments";
+    public static final String LABEL_ISSUE_CREATOR = "Issue Creator";
+    public static final String LABEL_JIRA_PROJECT = "Jira Project";
+
+    public static final String DESCRIPTION_ADD_COMMENTS = "If true, this will add comments to the Jira ticket with data describing the latest change to the ticket due to the job processing.";
+    public static final String DESCRIPTION_ISSUE_CREATOR = "The user name of the Jira Cloud user to assign to the issue creator field of the Jira ticket.";
```
jamesrichard91: The user name of the Jira Cloud user to assign _as_ the issue creator field of the Jira ticket.

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDistributionUIConfig.java
```diff
@@ -30,18 +30,32 @@
...
 import com.synopsys.integration.alert.common.descriptor.config.ui.ChannelDistributionUIConfig;
 
 @Component
 public class JiraDistributionUIConfig extends ChannelDistributionUIConfig {
+
+    public static final String LABEL_ADD_COMMENTS = "Add Comments";
+    public static final String LABEL_ISSUE_CREATOR = "Issue Creator";
+    public static final String LABEL_JIRA_PROJECT = "Jira Project";
+
+    public static final String DESCRIPTION_ADD_COMMENTS = "If true, this will add comments to the Jira ticket with data describing the latest change to the ticket due to the job processing.";
```
jamesrichard91: Can we change the end of this to
...with data describing the latest change.

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDescriptor.java
psantos1113: It will be fixed

PR #434, opened 2019-05-30
Projects now associated with current user

bamandel:
If a user sets up Alert and they are not associated with a project that is set in a distribution job, we will automatically add the user to the project.

Also updated to next version of blackduck common.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckProjectSyncTask.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckProjectSyncTask.java
@@ -22,6 +22,7 @@
  */
 package com.synopsys.integration.alert.provider.blackduck.tasks;
 
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckProjectSyncTask.java
```diff
@@ -135,4 +144,32 @@ public void runTask() {
...
+
+    private void updateBlackDuckProjectPermissions(final Set<String> configuredProjects, final List<ProjectView> projectViews, final ProjectUsersService projectUsersService, final BlackDuckService blackDuckService)
+        throws IntegrationException {
+        final UserView currentUser = blackDuckService.getResponse(ApiDiscovery.CURRENT_USER_LINK_RESPONSE);
+        final Set<ProjectView> matchingProjects = projectViews.parallelStream()
+                                                      .filter(projectView -> configuredProjects.contains(projectView.getName()))
+                                                      .collect(Collectors.toSet());
+        for (final ProjectView projectView : matchingProjects) {
+            logger.info("Adding user to Project {}", projectView.getName());
+            projectUsersService.addUserToProject(projectView, currentUser);
```
jamesrichard91: If the User is already assigned to the Project(s), this method will generate a log every time. We may want to wait until that is addressed in the library OR implement it ourselves.

I am in favor of using this method, fixing the library and releasing it, and update to the newer version.

bamandel: I had seen that issue when I originally implemented it. I considered doing a check before but I thought doing a check beforehand just to ignore a log was not beneficial.

PR #437, opened 2019-06-04
Fixed links to BlackDuck

bamandel:
Created a more appropriate link for BlackDuck policy and vulnerability components and their versions. Added a new Parent class that uses BlackDuck end points to properly update which URL should be hit.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
@@ -1,0 +1,96 @@
+/**
+ * blackduck-alert
+ *
+ * Copyright (c) 2019 Synopsys, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.provider.blackduck.collector;
+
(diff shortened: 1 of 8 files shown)


psantos1113:
Just refactor the getProjectQueryLink to getProjectComponentLink since it relates to components in the BOM to avoid confusion.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public Optional<String> getProjectQueryLink(final String projectLink, final String componentName) {
+        return Optional.of(String.format("%s?q=componentName:%s", projectLink, componentName));
+    }
+
+    public Optional<String> getProjectLink(final String projectVersionUrl, final String link) {
+        if (blackDuckService.isPresent() && bucketService.isPresent()) {
+            try {
+                final ProjectVersionView projectVersionView = blackDuckService.get().getResponse(projectVersionUrl, ProjectVersionView.class);
```
jamesrichard91: We should check the Bucket before we try to request the ProjectVersionView.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyViolationCollector.java
```diff
@@ -137,17 +148,18 @@ private PolicyComponentMapping createPolicyComponentMapping(final ComponentVersi
...
-
         final String componentVersionName = componentVersionStatus.getComponentVersionName();
         if (StringUtils.isNotBlank(componentVersionName)) {
-            blackDuckPolicyLinkableItem.addComponentVersionItem(componentVersionName, componentVersionStatus.getComponentVersion());
+            blackDuckPolicyLinkableItem.addComponentVersionItem(componentVersionName, projectVersionWithComponentLink);
+        }
+
+        final String componentName = componentVersionStatus.getComponentName();
+        if (StringUtils.isNotBlank(componentName)) {
+            final String componentLink = (StringUtils.isBlank(componentVersionName)) ? projectVersionWithComponentLink : componentVersionStatus.getComponent();
```
jamesrichard91: I don't think this inline condition is correct. The link should always be the same. If a the component version is missing then the link should be applied to the component name, if the component version is present then the link should be applied to the version

bamandel: I can make this clearer by changing componentVersionStatus.getComponent() to null. The idea is if we have a component name, but the component version is blank, we add the link to the name. If we don't want a link we add null. The value I used instead of null HAPPENS to be null if there is no version, so I am not sure why I used that instead of just null.

PR #438, opened 2019-06-05
Added Global jira test functionality

bamandel:
Added local dependency for jira common and implemented the Jira Cloud global test functionality.

This won't be merged until later so we can update the dependency to the appropriate version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,6 +80,7 @@ dependencies {
     implementation project(':alert-database')
 
     implementation 'com.synopsys.integration:polaris-common:0.12.0'
+    implementation 'com.synopsys.integration:int-jira-common:0.0.1-SNAPSHOT'
 
     // Spring
     implementation 'org.springframework:spring-webmvc'
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ dependencies {
     implementation project(':alert-database')
 
     implementation 'com.synopsys.integration:polaris-common:0.12.0'
-    implementation 'com.synopsys.integration:int-jira-common:0.0.1-SNAPSHOT'
+    implementation 'com.synopsys.integration:int-jira-common:0.0.2'
 
     // Spring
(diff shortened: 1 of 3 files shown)


src/main/java/com/synopsys/integration/alert/channel/jira/actions/JiraGlobalTestAction.java
```diff
@@ -24,18 +24,58 @@
...
+    }
 
     @Override
     public void testConfig(final TestConfigModel testConfig) throws IntegrationException {
-        logger.info("Test config invoked {}", testConfig);
+        final FieldAccessor fieldAccessor = testConfig.getFieldAccessor();
+        final JiraServerConfigBuilder jiraServerConfigBuilder = new JiraServerConfigBuilder();
+        final String jiraUrl = fieldAccessor.getString(JiraDescriptor.KEY_JIRA_URL).orElse(null);
+        final String accessToken = fieldAccessor.getString(JiraDescriptor.KEY_JIRA_ACCESS_TOKEN).orElse(null);
+        final String username = fieldAccessor.getString(JiraDescriptor.KEY_JIRA_USERNAME).orElse(null);
```
jamesrichard91: Is this supposed to be a Username or an email address?

bamandel: I was not sure what to do about this. The username in Jira is the users email... from what I could tell. Our UI specifies this field as the username since it's what they use to log in. Should we change it all to email?

build.gradle
```diff
@@ -80,6 +80,7 @@ dependencies {
     implementation project(':alert-database')
 
     implementation 'com.synopsys.integration:polaris-common:0.12.0'
+    implementation 'com.synopsys.integration:int-jira-common:0.0.1-SNAPSHOT'
```
psantos1113: use 0.0.2 of int-jira-common.  That has been released.

PR #446, opened 2019-06-13
Jira channel

bamandel:
This has MOST of the functionality we wanted for the Jira channel. There are a few cases where we don't have the information we want to and a fixme should be there explaining what we want or need.

There are also Freemarker template service changes here which I originally planned on using but ended up not.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ dependencies {
     implementation project(':alert-database')
 
     implementation 'com.synopsys.integration:polaris-common:0.12.0'
-    implementation 'com.synopsys.integration:int-jira-common:0.0.2'
+    implementation 'com.synopsys.integration:int-jira-common:0.0.3'
 
     // Spring
     implementation 'org.springframework:spring-webmvc'
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ dependencies {
     implementation project(':alert-database')
 
     implementation 'com.synopsys.integration:polaris-common:0.12.0'
-    implementation 'com.synopsys.integration:int-jira-common:0.0.3'
+    implementation 'com.synopsys.integration:int-jira-common:0.0.4'
 
     // Spring
     implementation 'org.springframework:spring-webmvc'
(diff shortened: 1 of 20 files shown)


build.gradle
```diff
@@ -80,7 +80,7 @@ dependencies {
     implementation project(':alert-database')
 
     implementation 'com.synopsys.integration:polaris-common:0.12.0'
-    implementation 'com.synopsys.integration:int-jira-common:0.0.2'
```
gkillough: 0.0.4 is now available

src/main/java/com/synopsys/integration/alert/channel/ChannelFreemarkerTemplatingService.java
```diff
@@ -30,39 +30,40 @@
...
-    private final Configuration configuration;
 
-    public ChannelFreemarkerTemplatingService(final String templateDirectory) throws IOException {
-        this.templateDirectory = templateDirectory;
-        this.configuration = createFreemarkerConfig();
+    public String getTemplatePath(final String templatesDirectory, final String channelDirectory) {
+        if (StringUtils.isNotBlank(templatesDirectory)) {
+            return String.format("%s/%s", templatesDirectory, channelDirectory);
+        }
+        return String.format("%s/src/main/resources/%s/templates", System.getProperties().getProperty("user.dir"), channelDirectory);
```
gkillough: For production the templates are exported to a different location to be accessible in the docker container. See _distributions_ in the build.gradle.

psantos1113: We should refactor this to use AlertProperties.getAlertTemplatesDir() at this point.  This is already used to get the templates. We should also refactor the build.gradle to pass the property in via the command line.  Also update the application properties file used for the tests to set this alert.templates.dir as well.  This way we can remove line 50.

bamandel: When you say update build.gradle, will I only be updating the runServer command to add this property value? I'm not aware of anywhere else I'd need to make changes for this.

PR #448, opened 2019-06-13
Remove BlackDuckVerifier from SystemValidator

bamandel:
To add to what is already done in your branch, this additional functionality allows different providers to insert their own validation to the System validator.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckValidator.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckValidator.java
@@ -1,0 +1,93 @@
+package com.synopsys.integration.alert.provider.blackduck;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Optional;
+
+import org.apache.commons.lang3.BooleanUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.AlertProperties;
+import com.synopsys.integration.alert.common.enumeration.SystemMessageSeverity;
+import com.synopsys.integration.alert.common.enumeration.SystemMessageType;
+import com.synopsys.integration.alert.common.exception.AlertRuntimeException;
+import com.synopsys.integration.alert.common.provider.ProviderValidator;
+import com.synopsys.integration.alert.database.system.SystemMessageUtility;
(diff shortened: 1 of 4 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckValidator.java
```diff
@@ -0,0 +1,93 @@
...
+    public boolean validate() {
+        logger.info("Validating BlackDuck Provider...");
+        systemMessageUtility.removeSystemMessagesByType(SystemMessageType.BLACKDUCK_PROVIDER_CONNECTIVITY);
+        systemMessageUtility.removeSystemMessagesByType(SystemMessageType.BLACKDUCK_PROVIDER_URL_MISSING);
+        systemMessageUtility.removeSystemMessagesByType(SystemMessageType.BLACKDUCK_PROVIDER_LOCALHOST);
+        try {
+
+            final Optional<String> blackDuckUrlOptional = blackDuckProperties.getBlackDuckUrl();
+            if (blackDuckUrlOptional.isEmpty()) {
+                logger.error("  -> BlackDuck Provider Invalid; cause: Black Duck URL missing...");
```
gkillough: I think "Black Duck" should always be printed as two words.

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckValidator.java
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                } else {
+                    final String message = "The BlackDuck configuration is not valid.";
+                    connectivityWarning(systemMessageUtility, message);
+                }
+            }
+        } catch (final MalformedURLException | IntegrationException | AlertRuntimeException ex) {
+            logger.error("  -> BlackDuck Provider Invalid; cause: {}", ex.getMessage());
+            logger.debug("  -> BlackDuck Provider Stack Trace: ", ex);
+            systemMessageUtility.addSystemMessage("BlackDuck Provider invalid: " + ex.getMessage(), SystemMessageSeverity.WARNING, SystemMessageType.BLACKDUCK_PROVIDER_CONNECTIVITY);
```
gkillough: Do we want to show the exception message in system messages? I think, given the limited screen real estate, we should omit it since we already divulge that information when a "Test Connection" is performed.

psantos1113: In this version of Alert we don't need to show the exception message anymore.  With the 3.x versions we were required to have a valid Black Duck connection to even log in. That error message would show the certificate errors for example.  If the provider is invalid they can now test the connection so we don't need that detail anymore in the UI.

PR #451, opened 2019-06-14
fix(java): Add a contentKey to the provider message content.

psantos1113:


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -114,6 +114,12 @@ public ComponentItem build() throws AlertException {
                 key = componentKey;
             } else {
                 final List<String> keyParts = new LinkedList<>(componentKeyParts);
+                keyParts.add(componentName);
+                keyParts.add(componentValue);
+                if (StringUtils.isNotBlank(subComponentName)) {
```
gkillough: We need to check name and value. I would argue value is the most important piece here.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ProviderMessageContent.java
```diff
@@ -33,18 +34,31 @@
...
-    private ProviderMessageContent(LinkableItem provider, LinkableItem topic, LinkableItem subTopic, Set<ComponentItem> componentItems) {
+    private ProviderMessageContent(LinkableItem provider, LinkableItem topic, LinkableItem subTopic, String contentKey, Set<ComponentItem> componentItems) {
         this.provider = provider;
         this.topic = topic;
         this.subTopic = subTopic;
+        this.contentKey = contentKey;
         this.componentItems = componentItems;
     }
 
+    public static final String generateContentKey(String providerName, String topicName, String topicValue, String subTopicName, String subTopicValue) {
```
gkillough: Do we want a ComponentKey object that has a string key and a "pretty print" method?

psantos1113: That might be better then.  We have a MessageContentKey class we could follow the same pattern here to provide that pretty print.

PR #452, opened 2019-06-17
Add ComponentKey

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor2/MessageContentCollapser.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        return collapsedMessages;
+    }
+
+    private BiFunction<Map<ComponentKey, ComponentItem>, ComponentItem, Void> createAddFunction() {
+        return (categoryDataCache, categoryItem) -> {
+            categoryDataCache.put(categoryItem.getComponentKey(), categoryItem);
+            return null;
```
jamesrichard91: Does the BiFunction require a return?

psantos1113: the return type is void so returning null fulfills that part of the contract.

gkillough: Exactly. Also, this is mostly just copied/pasted from the old logic.

PR #457, opened 2019-06-18
Removing the Reset Password link from the LoginPage

jamesrichard91:


src/main/js/LoginPage.js
```diff
@@ -88,16 +76,6 @@ class LoginPage extends Component {
                                     </div>
                                 </div>
                             </div>
-                            <div>
-                                <ResetPasswordModal
```
bamandel: If the ResetPasswordModal is no longer being used here, should this components file be removed as well?

jamesrichard91: I want to leave the backend parts and the Component in case we support more than one database user in the future and we need to bring this feature back.

PR #458, opened 2019-06-19
Ps component priority

psantos1113:
Changes to implement a comparator for ComponentItems to handle sorting the data without need custom component item objects.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
@@ -34,24 +34,23 @@ import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
 
+import com.synopsys.integration.alert.common.enumeration.ComponentItemPriority;
 import com.synopsys.integration.alert.common.enumeration.ItemOperation;
 import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 import com.synopsys.integration.builder.Buildable;
 
-public class ComponentItem extends AlertSerializableModel implements Buildable, Comparable<ComponentItem> {
+public class ComponentItem extends AlertSerializableModel implements Buildable {
     private final LinkableItem component;
     private final LinkableItem subComponent;
     private final Set<LinkableItem> componentAttributes;
-    private final Integer priority;
+    private final ComponentItemPriority priority;
     private final String category;
     private final ComponentKey componentKey;
     private final ItemOperation operation;
     private final Long notificationId;
-    private final transient Comparator<ComponentItem> comparator;
 
-    private ComponentItem(LinkableItem component, LinkableItem subComponent, Set<LinkableItem> componentAttributes, Integer priority, String category, ComponentKey componentKey, ItemOperation operation, Long notificationId,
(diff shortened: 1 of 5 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -126,7 +141,16 @@ public ComponentItem build() throws AlertException {
 
             final String additionalDataString = ComponentKey.generateAdditionalDataString(componentAttributes);
             ComponentKey key = new ComponentKey(category, componentName, componentValue, subComponentName, subComponentValue, additionalDataString);
-            return new ComponentItem(component, subComponent, componentAttributes, category, key, operation, notificationId);
+            Comparator<ComponentItem> comparator = createComparator();
+            return new ComponentItem(component, subComponent, componentAttributes, priority, category, key, operation, notificationId, comparator);
+        }
+
+        private Comparator<ComponentItem> createComparator() {
+            return Comparator.comparing(ComponentItem::getComponent)
```
gkillough: Couldn't we compare the ComponentKey before the priority?

psantos1113: that certainly makes it easier.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -38,23 +39,28 @@
 import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 import com.synopsys.integration.builder.Buildable;
 
-public class ComponentItem extends AlertSerializableModel implements Buildable {
+public class ComponentItem extends AlertSerializableModel implements Buildable, Comparable<ComponentItem> {
     private final LinkableItem component;
     private final LinkableItem subComponent;
     private final Set<LinkableItem> componentAttributes;
+    private final Integer priority;
```
gkillough: This could be an enum:
HIGH,
MEDIUM,
LOW,
STANDARD

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -96,13 +105,19 @@ public Long getNotificationId() {
...
         for (final LinkableItem item : componentAttributes) {
             final String name = item.getName();
-            map.computeIfAbsent(name, ignored -> new ArrayList<>()).add(item);
+            map.computeIfAbsent(name, ignored -> new LinkedList<>()).add(item);
         }
         return map;
     }
 
+    @Override
+    public final int compareTo(final ComponentItem otherItem) {
```
gkillough: Rather than being Comparable, this could be a `createComparator` method that would only be used for processing.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentKey.java
```diff
@@ -97,4 +100,11 @@ public boolean equals(final Object otherObject) {
         return false;
     }
 
+    @Override
+    public int compareTo(final ComponentKey other) {
```
gkillough: Should either `compareTo` or `equals` reference the other so that we do not have duplication?

PR #462, opened 2019-06-20
Ps license support

psantos1113:
Add the license information to the messages.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
@@ -25,17 +25,21 @@ package com.synopsys.integration.alert.provider.blackduck.collector;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.synopsys.integration.alert.common.message.model.LinkableItem;
 import com.synopsys.integration.alert.common.provider.ProviderContentType;
 import com.synopsys.integration.alert.common.workflow.MessageContentCollector;
 import com.synopsys.integration.alert.common.workflow.filter.field.JsonExtractor;
 import com.synopsys.integration.alert.common.workflow.processor.MessageContentProcessor;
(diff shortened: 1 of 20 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
```diff
@@ -85,6 +89,28 @@ public BlackDuckCollector(final JsonExtractor jsonExtractor, final List<MessageC
...
+                return Optional.empty();
+            }
+        }
+        return Optional.empty();
+    }
+
+    public List<LinkableItem> getLicenseLinkableItems(final VersionBomComponentView bomComponentView) {
+        return bomComponentView.getLicenses().stream().map(licenseView -> {
+            LinkableItem item = new LinkableItem(BlackDuckContent.LABEL_COMPONENT_LICENSE, licenseView.getLicenseDisplay(), licenseView.getLicense());
+            item.setSummarizable(true);
```
gkillough: Is license something relevant to a summary view?

psantos1113: no it actually isn't I will remove that. We don't summarize by license here.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyOverrideCollector.java
```diff
@@ -91,7 +93,7 @@ public BlackDuckPolicyOverrideCollector(final JsonExtractor jsonExtractor, final
             ComponentItemPriority priority = mapSeverityToPriority(policyItem.getSeverity());
             Set<LinkableItem> attributeSet = new LinkedHashSet<>();
             nameItem.ifPresent(attributeSet::add);
-            attributeSet.addAll(createPolicyLinkableItems(policyItem));
+            attributeSet.addAll(createPolicyLinkableItems(policyItem, bomComponentVersionUrl.orElse("")));
```
gkillough: I think down the line, this bomComponentVersionUrl trickles to the bucket. How does it handle empty strings? If it does an `isBlank` check, I would be more comfortable with null here so that we don't potentially have the side affect of an empty string as a key in the bucket if it ever changes.

PR #463, opened 2019-06-21
Updating how we check for newer versions of Alert.

jamesrichard91:


build.gradle
```diff
@@ -249,7 +249,9 @@ task createAboutText(type: Task) {
     def projectUrlText = readmeContentArray.subList(projectUrlStart, projectUrlEnd).join(" ")
     def gitUrl = projectUrlText.substring(projectUrlText.indexOf("https"))
     gitUrl = gitUrl.substring(0, gitUrl.indexOf("releases"))
-    final def aboutJson = JsonOutput.toJson([version: version, projectUrl: gitUrl, description: description])
+
+    def time = new Date().format('yyy-MM-dd\'T\'HH:mm:ss.SSSSSS\'Z\'');
```
gkillough: We should use `UpdateChecker. DATE_FORMAT` here.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/UpdateChecker.java
```diff
@@ -112,31 +128,83 @@ private boolean isComprisedOfNumericTokens(final String version) {
...
-            final String[] firstVersionTokens = StringUtils.split(firstVersion, VERSION_SEPARATOR);
-            final String[] secondVersionTokens = StringUtils.split(secondVersion, VERSION_SEPARATOR);
+    private Comparator<DockerTagModel> tagOrder() {
+        return (firstTag, secondTag) -> compareVersionAndDates(firstTag.getName(), firstTag.getLastUpdated(), secondTag.getName(), secondTag.getLastUpdated(), this::compareDateStrings);
+    }
 
-            for (int i = 0; i < firstVersionTokens.length && i < secondVersionTokens.length; i++) {
-                final int firstToken = Integer.parseInt(firstVersionTokens[i]);
-                final int secondToken = Integer.parseInt(secondVersionTokens[i]);
+    private int compareVersionAndDates(final String firstVersion, final String firstDate, final String secondVersion, final String secondDate, final BiFunction<String, String, Integer> dateComparison) {
```
gkillough: Why do we need to pass a method reference here? Do we have more than one date comparison method? These methods are both in the same class.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/model/UpdateModel.java
```diff
@@ -22,36 +22,38 @@
...
     }
 
-    public String getLatestAvailableVersion() {
-        return latestAvailableVersion;
+    public String getCurrentCreatedDate() {
+        return currentCreatedDate;
+    }
+
+    public DockerTagModel getLatestAvailableTag() {
+        return latestAvailableTag;
```
gkillough: Is this the newest tag on Docker? Or is this what we determined to be the latest tag?

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/UpdateChecker.java
```diff
@@ -112,31 +128,83 @@ private boolean isComprisedOfNumericTokens(final String version) {
                    .allMatch(NumberUtils::isParsable);
     }
 
-    private Comparator<String> versionOrder() {
-        return (firstVersion, secondVersion) -> {
-            final String[] firstVersionTokens = StringUtils.split(firstVersion, VERSION_SEPARATOR);
-            final String[] secondVersionTokens = StringUtils.split(secondVersion, VERSION_SEPARATOR);
+    private Comparator<DockerTagModel> tagOrder() {
```
bamandel: A separate SemanticVersion model class (Or perhaps an already built utility) would probably make this all easier to follow. It would separate out version comparisons from this task.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/UpdateNotifierTask.java
```diff
@@ -53,14 +53,14 @@ public UpdateNotifierTask(final TaskScheduler taskScheduler, final UpdateChecker
...
+        if (updateModel.getUpdatable()) {
             addSystemMessage(updateModel);
             updateEmailService.sendUpdateEmail(updateModel);
         }
     }
 
     private void addSystemMessage(final UpdateModel updateModel) {
-        final String message = String.format("There is a new version of %s available: %s", AlertConstants.ALERT_APPLICATION_NAME, updateModel.getLatestAvailableVersion());
+        final String message = String.format("There is a new version of %s available: %s", AlertConstants.ALERT_APPLICATION_NAME, updateModel.getLatestAvailableTag().getName());
         systemMessageUtility.removeSystemMessagesByType(SystemMessageType.UPDATE_AVAILABLE);
```
bamandel: If we're only using the name here, we might just want to pass the name to the method and not the entire object.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/model/UpdateModel.java
jamesrichard91: This is the latest tag that we determined from Docker

build.gradle
jamesrichard91: I don't believe we can reference the class from the gradle file. If you know how to do that please let me know

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/UpdateChecker.java
```diff
@@ -67,10 +72,28 @@ public UpdateModel getUpdateModel() {
...
+
+        final DockerTagModel latestAvailableVersion = getLatestAvailableTag(dockerTagRetriever, isProduction);
         final String repositoryUrl = dockerTagRetriever.getRepositoryUrl();
-        return new UpdateModel(currentVersion, optionalLatestAvailableVersion.orElse(null), repositoryUrl);
+
+        return getUpdateModel(currentVersion, alertCreated, latestAvailableVersion, repositoryUrl);
+    }
+
+    public UpdateModel getUpdateModel(final String currentVersion, final String alertCreated, final DockerTagModel latestAvailableVersion, final String repositoryUrl) {
+        int comparison = compareVersions(currentVersion, latestAvailableVersion.getName());
```
psantos1113: why not use the comparator from getTagOrder here? appears to be the same comparison

build.gradle
gkillough: Oh, you're right. I think you can only do it with buildscript dependencies.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/model/UpdateModel.java
gkillough: I'm still not clear on this. If we uploaded 5.0.0 one day, then uploaded 4.2.0.1 the next day, which of the two would be considered the latest tag?

PR #464, opened 2019-06-21
Provider Message Content

gkillough:
This will also be included in Code Review.


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageContentCollector.java
```diff
@@ -97,7 +112,18 @@ public void insert(final AlertNotificationWrapper notification) {
...
+    private Optional<ProviderMessageContent> buildMessageContent(ProviderMessageContent.Builder builder) {
+        try {
+            return Optional.of(builder.build());
+        } catch (AlertException ex) {
+            logger.error("Error building message content continuing on", ex);
+            return Optional.empty();
+        }
+    }
+
+    //protected abstract void addCategoryItems(SortedSet<CategoryItem> categoryItems, final JsonFieldAccessor jsonFieldAccessor, final List<JsonField<?>> notificationFields, final AlertNotificationWrapper notificationContent);
```
bamandel: Comment here

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageContentCollector.java
```diff
@@ -191,45 +205,41 @@ private JsonFieldAccessor createJsonAccessor(final List<JsonField<?>> notificati
...
-            if (foundContent != null) {
-                aggregateMessageContentsForNotifications.add(foundContent);
+            if (null != foundContent) {
+                buildersForNotifications.add(foundContent);
             } else {
-                final SortedSet<CategoryItem> categoryItems = new TreeSet<>();
-                aggregateMessageContentsForNotifications.add(createAggregateMessageContent(topicItem, subTopic, categoryItems));
+                ProviderMessageContent.Builder builder = new ProviderMessageContent.Builder();
+                builder
+                    .applyProvider(providerItem.getValue(), providerItem.getUrl().orElse(null))
```
bamandel: Can this line be used on line 233 instead and you add these items immediately after creating the builder? The formatting is weird here with a single item on 233

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ContentKey.java
```diff
@@ -0,0 +1,89 @@
...
+package com.synopsys.integration.alert.common.message.model;
+
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class ContentKey extends AlertSerializableModel {
+    private static final String KEY_SEPARATOR = "_";
```
jamesrichard91: This is the same KEY_SEPARATOR as the one in ComponentKeys, could we have them share the same separator?

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ContentKey.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private String generateContentKey(String providerName, String topicName, String topicValue, String subTopicName, String subTopicValue) {
+        final List<String> keyParts;
+        if (StringUtils.isNotBlank(subTopicName)) {
+            keyParts = List.of(providerName, topicName, topicValue, subTopicName, subTopicValue);
+        } else {
+            keyParts = List.of(providerName, topicName, topicValue);
+        }
+        return String.join(KEY_SEPARATOR, keyParts);
```
jamesrichard91: We use StringUtils.join(..) in ComponentKeys. We should make them consistent if possible.

gkillough: Will do. These are two classes written on two different branches by two different people haha. Those branches were combined to make this branch.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
```diff
@@ -70,11 +87,11 @@ public BlackDuckCollector(final JsonExtractor jsonExtractor, final List<MessageC
     }
 
     public Optional<String> getProjectLink(final String projectVersionUrl, final String link) {
-        if (blackDuckService.isPresent() && bucketService.isPresent()) {
+        if (null != blackDuckService && null != bucketService) {
```
jamesrichard91: We could update the constructor to set a boolean field called hasValidConnection, if either of the services are missing we could set that field to false and just check for that here.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
```diff
@@ -84,12 +101,34 @@ public BlackDuckCollector(final JsonExtractor jsonExtractor, final List<MessageC
         return Optional.empty();
     }
 
-    public Optional<BlackDuckBucketService> getBucketService() {
-        return bucketService;
+    public Optional<VersionBomComponentView> getBomComponentView(final String bomComponentUrl) {
+        if (getBucketService().isPresent()) {
```
jamesrichard91: If we added the field hasValidConnection, we could check for that here

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckPolicyCollector.java
```diff
@@ -22,50 +22,95 @@
...
 public abstract class BlackDuckPolicyCollector extends BlackDuckCollector {
-    public static final String CATEGORY_TYPE = "policy";
+    public static final String CATEGORY_TYPE = "Policy";
+    private final Map<String, ComponentItemPriority> priorityMap = new HashMap<>();
+    private Logger logger = LoggerFactory.getLogger(getClass());
 
     public BlackDuckPolicyCollector(final JsonExtractor jsonExtractor, final List<MessageContentProcessor> messageContentProcessorList, final Collection<ProviderContentType> contentTypes, final BlackDuckProperties blackDuckProperties) {
         super(jsonExtractor, messageContentProcessorList, contentTypes, blackDuckProperties);
+
+        priorityMap.put("blocker", ComponentItemPriority.HIGH);
```
jamesrichard91: What ends up being displayed in the end? Blocker or High? The customers are going to want to see the provider specified value, not our interpretation of it. And the ComponentItemPriority needs to be expanded to support more granularity. Having Major policy violations show up above Blockers will not go over well

gkillough: The priority is only used for sorting.

gkillough: Severity is added as its own linkable item.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ContentKey.java
```diff
@@ -58,7 +58,7 @@ private String generateContentKey(String providerName, String topicName, String
         } else {
             keyParts = List.of(providerName, topicName, topicValue);
         }
-        return String.join(KEY_SEPARATOR, keyParts);
+        return StringUtils.join(KEY_SEPARATOR, keyParts);
```
jamesrichard91: Are you sure the arguments are in the correct order?

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
```diff
@@ -68,6 +69,7 @@ public BlackDuckCollector(final JsonExtractor jsonExtractor, final List<MessageC
         bucketService = blackDuckServicesFactory
                             .map(BlackDuckServicesFactory::createBlackDuckBucketService)
                             .orElse(null);
+        hasValidConnection = null != blackDuckService && null != bucketService;
```
jamesrichard91: Should this condition be an OR not an AND?

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/ComponentItemPriority.java
```diff
@@ -23,10 +23,11 @@
 package com.synopsys.integration.alert.common.enumeration;
 
 public enum ComponentItemPriority {
-    CRITICAL,
+    HIGHEST,
     HIGH,
     MEDIUM,
     LOW,
+    LOWEST,
     STANDARD;
```
gkillough: STANDARD is lower than LOWEST?

jamesrichard91: I would say so. I don't think STANDARD is correctly named. Should we change it to UNDEFINED?

PR #468, opened 2019-06-25
Jira App

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssuePropertyHelper.java
```diff
@@ -0,0 +1,90 @@
...
+        return findIssues(category, bomComponentUri, null);
+    }
+
+    public Optional<IssueSearchResponseModel> findIssues(String category, String bomComponentUri, String policyName) throws IntegrationException {
+        final StringBuilder jqlBuilder = new StringBuilder();
+        if (StringUtils.isNotBlank(category)) {
+            jqlBuilder.append(createPropertySearchString(JiraConstants.JIRA_ISSUE_PROPERTY_OBJECT_KEY_CATEGORY, category));
+            jqlBuilder.append(StringUtils.SPACE);
+        }
+        if (StringUtils.isNotBlank(bomComponentUri)) {
```
jamesrichard91: This seems to be very Black Duck specific and how would it handle License Limit notifications for example?

src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssuePropertyHelper.java
```diff
@@ -0,0 +1,90 @@
...
+
+        final String jql = jqlBuilder.toString();
+        if (!jql.isBlank()) {
+            final IssueSearchResponseModel issueSearchResponseModel = issueSearchService.queryForIssues(jql);
+            return Optional.of(issueSearchResponseModel);
+        }
+        return Optional.empty();
+    }
+
+    public void addPropertiesToIssue(String issueKey, String category, String bomComponentUri) throws IntegrationException {
```
bamandel: I think We were going to use one of the content keys as the property key now to still be unique and also independent of providers

src/main/resources/jira/atlassian-connect.json
```diff
@@ -0,0 +1,48 @@
...
+        "name": {
+          "value": "Alert Issue Property"
+        },
+        "entityType": "issue",
+        "keyConfigurations": [
+          {
+            "propertyKey": "com-synopsys-integration-alert",
+            "extractions": [
+              {
+                "objectName": "bomComponentUri",
```
bamandel: If we can find an alternative to bomComponentUri, that would be the best.

PR #469, opened 2019-06-26
Added save functionality to the Custom Jira endpoint button.

bamandel:
Added saving functionality to the configure button to make the user experience better. New flow is like this.

Add normal text fields.
Save.
Error as you didn't press button to configure plugin (Normal fields are still saved).
Properly install custom plugin by pressing button.
Automatic save of button field occurs on success.
User does not need to press save again after pressing button (They can if they'd like).

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
@@ -22,7 +22,10 @@
  */
 package com.synopsys.integration.alert.channel.jira.web;
 
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
```diff
@@ -125,4 +130,26 @@ private String getAppropriateAccessToken(final FieldValueModel fieldAccessToken)
...
+                                                                              .stream()
+                                                                              .filter(configurationFieldModel -> !JiraDescriptor.KEY_JIRA_CONFIGURE_PLUGIN.equals(configurationFieldModel.getFieldKey()))
+                                                                              .collect(Collectors.toList());
+                configurationFields.add(configurePluginField);
+                configurationAccessor.updateConfiguration(configurationId, configurationFields);
+            } else {
+                configurationAccessor.createConfiguration(JiraChannel.COMPONENT_NAME, ConfigContextEnum.GLOBAL, List.of(configurePluginField));
+            }
+        } catch (final AlertDatabaseConstraintException e) {
+            logger.error("Was unable to retrieve Global Jira Cloud data from the Database.");
```
gkillough: Let's drop the "was".

PR #470, opened 2019-06-26
Jira Channel Enhancements

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -0,0 +1,253 @@
...
+        final PageOfProjectsResponseModel projectsResponseModel = projectService.getProjectsByName(projectName);
+        final ProjectComponent projectComponent = projectsResponseModel.getProjects()
+                                                      .stream()
+                                                      .findFirst()
+                                                      .orElseThrow(() -> new AlertException("Expected to be passed an existing project name."));
+
+        final String projectId = projectComponent.getId();
+        final String projectKey = projectComponent.getKey();
+        final String issueType = fieldAccessor.getString(JiraDescriptor.KEY_ISSUE_TYPE).orElse(JiraDistributionUIConfig.DEFAULT_ISSUE_TYPE);
+        final boolean matchingIssueTypeFound = issueTypeService.getAllIssueTypes()
```
jamesrichard91: It looks like 'matchingIssueTypeFound' is never used. If we can't find the Issue Type shouldn't we throw an exception the way we do for the Jira Project and the Provider?

src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
```diff
@@ -55,7 +55,7 @@
     private static final Logger logger = LoggerFactory.getLogger(JiraCustomEndpoint.class);
 
     // FIXME whenever we change the atlassian-connect.json file, we need to update this URL. This is very error prone and we'll want to find a better way to do this. Perhaps GH-pages
-    public static final String JIRA_PLUGIN_URL = "https://raw.githubusercontent.com/blackducksoftware/blackduck-alert/810dd02/src/main/resources/jira/atlassian-connect.json";
+    public static final String JIRA_PLUGIN_URL = "https://raw.githubusercontent.com/blackducksoftware/blackduck-alert/gk_jira_channel_enhancements/src/main/resources/jira/atlassian-connect.json";
```
jamesrichard91: Do we need this URL? Can we have Alert host the file, instead of hosting it on github?

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
gkillough: Yes, that was unused before this PR. The purpose of this PR is to add appropriate abstraction to what existed and begin working on the issue remapping functionality. I'll go ahead and take care of this anyway, but that's why it hadn't been touched.

src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
gkillough: This will eventually be hosted, this is just a workaround while we are developing.

PR #471, opened 2019-06-27
Data backup

jamesrichard91:
This seems to work well in the docker-compose deployment. I don't really have a way to test the other deployment mechanisms.


psantos1113:
I tested Kubernetes.

PR #476, opened 2019-07-01
Bm breakup FieldModelProcessor

bamandel:
FieldModelProcessor methods have been broken out into two other classes. DescriptorProcessor and ConfigurationFieldModelConverter.

Also did some small TODOs as the main intent was to finish the one above.


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/ConfigurationFieldModelConverter.java
```diff
@@ -106,4 +132,20 @@ public final FieldAccessor convertToFieldAccessor(final FieldModel fieldModel) t
         return Optional.of(configurationModel);
     }
 
+    private void populateAndSecureFields(final ConfigurationFieldModel fieldModel, final Map<String, FieldValueModel> fields) {
+        final String key = fieldModel.getFieldKey();
+        Collection<String> values = Collections.emptyList();
```
gkillough: Why use Collection here over list? Why not use `List.of()` for initialization?

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -150,7 +149,7 @@ public JobFieldModel updateJob(final UUID id, final JobFieldModel jobFieldModel)
         final Set<FieldModel> jobFieldModels = jobFieldModel.getFieldModels();
         for (final FieldModel fieldModel : jobFieldModels) {
             final FieldModel beforeUpdateEventFieldModel = fieldModelProcessor.performBeforeUpdateAction(fieldModel);
-            final Long fieldModelId = contentConverter.getLongValue(beforeUpdateEventFieldModel.getId());
+            final Long fieldModelId = Long.parseLong(beforeUpdateEventFieldModel.getId());
```
gkillough: Does this method differ in null handling than the one it is replacing?

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/ConfigurationFieldModelConverter.java
```diff
@@ -95,6 +99,28 @@ public final FieldAccessor convertToFieldAccessor(final FieldModel fieldModel) t
...
+            fields.put(key, fieldValueModel);
+        }
+        return fields;
+    }
+
+    public FieldModel convertToFieldModel(final ConfigurationModel configurationModel) throws AlertDatabaseConstraintException {
+        final Long configId = configurationModel.getConfigurationId();
+        final String descriptorName = getDescriptorName(configurationModel);
+        final Map<String, FieldValueModel> fields = new HashMap<>();
+        for (final ConfigurationFieldModel fieldModel : configurationModel.getCopyOfFieldList()) {
```
psantos1113: Why couldn't you use the convertToFieldValuesMap method here on the copy of field list and then iterate through the map and use the .setValues (null) method call if the isSensitive field returns true?  That way there is one method to handle converting the configuration model to a map of FieldValueModels.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
psantos1113: This will throw a NumberFormatException if the Id is null or the empty string.  We would return null in the past if the id was null or the empty string.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/ConfigurationFieldModelConverter.java
bamandel: It's a good idea, but we actually need the ConfigurationFieldModel to tell us if a field is sensitive or not. Using that flag, we empty the values.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
bamandel: In the past we used contentConverter which does what Paulo stated. But these are ID's that are all generated and maintained by us. At no point should we have our UI hit this endpoint without ID's specified. With the nul, we would fail silently if we messed something up in the UI. Now, we'll get NumberFormatException's if we mess it up. Perhaps there is a better way to do this, but I felt this was a better approach to the issue at the moment.

PR #478, opened 2019-07-02
Fix SonarCloud issues

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -96,12 +96,13 @@ public String createOrUpdateIssues(final FieldAccessor fieldAccessor, final Mess
...
-        issueTypeService.getAllIssueTypes()
-            .stream()
-            .map(IssueTypeResponseModel::getName)
-            .filter(name -> name.equals(issueType))
-            .findAny()
-            .orElseThrow(() -> new AlertException(String.format("The issue type '%s' could not be found", issueType)));
+        boolean isValidIssueType = issueTypeService.getAllIssueTypes()
+                                       .stream()
+                                       .map(IssueTypeResponseModel::getName)
+                                       .anyMatch(name -> name.equals(issueType));
```
bamandel: I believe you can do issueType::equals here

PR #479, opened 2019-07-03
Ps update email enhancement

psantos1113:
Added the Alert server URL to the template.  Needed to create a method in AlertProperties to get the server URL.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/workflow/startup/StartupManager.java
+++ b/src/main/java/com/synopsys/integration/alert/workflow/startup/StartupManager.java
@@ -138,20 +138,21 @@ public class StartupManager {
 
         logger.info("----------------------------------------");
         logger.info("Alert Configuration: ");
-        logger.info("Logging level:           {}", alertProperties.getLoggingLevel().orElse(""));
+        logger.info("Alert Server URL:          {}", alertProperties.getServerUrl().orElse(""));
+        logger.info("Logging level:             {}", alertProperties.getLoggingLevel().orElse(""));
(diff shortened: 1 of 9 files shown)


src/main/java/com/synopsys/integration/alert/workflow/startup/StartupManager.java
```diff
@@ -138,21 +136,13 @@ public void logConfiguration() {
...
         logger.info("Alert Configuration: ");
-        logger.info("Logging level:           {}", alertProperties.getLoggingLevel().orElse(""));
+        logger.info("Alert Server URL:          {}", alertProperties.getServerUrl().orElse(""));
+        logger.info("Logging level:             {}", alertProperties.getLoggingLevel().orElse(""));
         logger.info("Alert Proxy Host:          {}", proxyHost.orElse(""));
         logger.info("Alert Proxy Port:          {}", proxyPort.orElse(""));
         logger.info("Alert Proxy Authenticated: {}", authenticatedProxy);
         logger.info("Alert Proxy User:          {}", proxyUsername.orElse(""));
         logger.info("");
-        logger.info("BlackDuck URL:                 {}", blackDuckProperties.getBlackDuckUrl().orElse(""));
```
jamesrichard91: We aren't logging any of the Black Duck information anymore?

psantos1113: BlackDuck isn't required any more.  We could perhaps iterate through the providers and log their fields.  Or loop through descriptors and log their configuration other than sensitive data.

PR #480, opened 2019-07-03
fix(deployment): Add the SAML environment variables to the files.

psantos1113:
Add the environment variables for SAML.


deployment/blackduck-alert.env
```diff
@@ -52,6 +52,13 @@ ALERT_COMPONENT_SETTINGS_SETTINGS_LDAP_USER_DN_PATTERNS=
 ALERT_COMPONENT_SETTINGS_SETTINGS_LDAP_USER_SEARCH_BASE=
 ALERT_COMPONENT_SETTINGS_SETTINGS_LDAP_USER_SEARCH_FILTER=
 
+# -- SAML
```
gkillough: Might as well delete HipChat while you're in here :)

psantos1113: I will remove the fields from the master branch when merging to master. 4.2.0 technically still has HipChat support.

PR #484, opened 2019-07-03
Changed configure button functionality

bamandel:
Jira configure button no longer saves itself and no longer does fancy validation. It is simply a button to install a plugin on the target server. (These changes do not validate that the target server has the plugin installed).

Changed labels and descriptions for Jira fields and removed the popup fields from the Jira configure button. Also modified testing buttons to first verify whether or not the server has our plugin installed.


src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
```diff
@@ -35,22 +35,16 @@
 @Component
 public class JiraGlobalUIConfig extends UIConfig {
     public static final String LABEL_URL = "Url";
-    public static final String LABEL_USER_NAME = "Username";
-    public static final String LABEL_ACCESS_TOKEN = "Access Token";
+    public static final String LABEL_USER_NAME = "Admin Username";
+    public static final String LABEL_ACCESS_TOKEN = "Admin Api Token";
```
jamesrichard91: Can we also update the variable name?

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
```diff
@@ -35,22 +35,16 @@
...
     public static final String LABEL_CONFIGURE_PLUGIN = "Configure Jira Cloud plugin";
-    public static final String LABEL_ADMIN_USERNAME = "Admin Username";
-    public static final String LABEL_ADMIN_ACCESS_TOKEN = "Admin Access Token";
 
     public static final String DESCRIPTION_URL = "The URL of the Jira Cloud server.";
-    public static final String DESCRIPTION_USER_NAME = "The email address used to log into the Jira Cloud server that has generated the access token.";
-    public static final String DESCRIPTION_ACCESS_TOKEN = "The access token used to send API requests to the Jira Cloud server.";
+    public static final String DESCRIPTION_USER_NAME = "The email address of the admin used to log into the Jira Cloud server that has generated the api token.";
+    public static final String DESCRIPTION_ACCESS_TOKEN = "The admin api token used to send API requests to the Jira Cloud server.";
```
jamesrichard91: Can we also update the variable name?

src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
```diff
@@ -79,18 +74,12 @@ public JiraCustomEndpoint(final CustomEndpointManager customEndpointManager, fin
...
             final JiraAppService jiraAppService = jiraServicesCloudFactory.createJiraAppService();
-            final FieldValueModel adminUsername = fieldValueModels.get(JiraGlobalUIConfig.KEY_JIRA_ADMIN_USERNAME);
-            final FieldValueModel adminAccessToken = fieldValueModels.get(JiraGlobalUIConfig.KEY_JIRA_ADMIN_ACCESS_TOKEN);
-            if (null == adminUsername || null == adminAccessToken) {
-                throw new IntegrationException("Both Username and Access Token are required to install plugin.");
-            }
-            final String username = adminUsername.getValue().orElseThrow(() -> new IntegrationException("Expected to be passed an admin username but did not receive any."));
-            final String accessToken = adminAccessToken.getValue().orElseThrow(() -> new IntegrationException("Expected to be passed an admin username but did not receive any."));
+            final String username = jiraProperties.getUsername();
+            final String accessToken = jiraProperties.getAccessToken();
```
jamesrichard91: What error is shown when these are not provided? Shouldn't we still check that these have values?

psantos1113: I would expect that if we are missing either of these we would get a 401 from line 79.  We should probably check the error code and if it is unauthorized return that rather than a generic 400 bad request.  For all other status codes then it makes sense to return a 400 Bad Request

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
gkillough: Let's also be consistent with the string "api". Should it be "API", "Api", or "api"?

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
gkillough: This string has both "api" and "API" in it.

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
```diff
@@ -35,22 +35,16 @@
 @Component
 public class JiraGlobalUIConfig extends UIConfig {
     public static final String LABEL_URL = "Url";
-    public static final String LABEL_USER_NAME = "Username";
-    public static final String LABEL_ACCESS_TOKEN = "Access Token";
+    public static final String LABEL_USER_NAME = "Admin Username";
```
gkillough: Should this be "Admin User Email"?

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
jamesrichard91: It should be API

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
jamesrichard91: Any references to API should be in all caps since it is an acronym.

src/main/java/com/synopsys/integration/alert/channel/jira/web/JiraCustomEndpoint.java
bamandel: Paulo is correct, the following call will now handle this error. Additionally, we'd throw an exception in our JiraProperties when trying to connect to the instance.

src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraGlobalUIConfig.java
bamandel: The description does specify that it should be an email format. Doesn't jira still refer to this value as a username though?

PR #496, opened 2019-07-10
Ps fix message order

psantos1113:
Sort the aggregate message content before formatting by the creation of the notification from the provider.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageContentCollector.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageContentCollector.java
@@ -71,7 +71,7 @@ public abstract class MessageContentCollector {
         try {
             final List<JsonField<?>> notificationFields = getFieldsForNotificationType(notification.getNotificationType());
             final JsonFieldAccessor jsonFieldAccessor = createJsonAccessor(notificationFields, notification.getContent());
-            final List<AggregateMessageContent> contents = getContentsOrCreateIfDoesNotExist(jsonFieldAccessor, notificationFields);
+            final List<AggregateMessageContent> contents = getContentsOrCreateIfDoesNotExist(jsonFieldAccessor, notificationFields, notification);
             for (final AggregateMessageContent content : contents) {
                 addCategoryItems(content.getCategoryItems(), jsonFieldAccessor, notificationFields, notification);
                 addContent(content);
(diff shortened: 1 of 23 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageContentCollector.java
```diff
@@ -208,17 +208,17 @@ private JsonFieldAccessor createJsonAccessor(final List<JsonField<?>> notificati
...
                 final Set<CategoryItem> categoryItems = new LinkedHashSet<>();
-                aggregateMessageContentsForNotifications.add(createAggregateMessageContent(topicItem, subTopic, categoryItems));
+                aggregateMessageContentsForNotifications.add(createAggregateMessageContent(topicItem, subTopic, categoryItems, notification));
             }
         }
         return aggregateMessageContentsForNotifications;
     }
 
-    private AggregateMessageContent createAggregateMessageContent(final LinkableItem topicItem, final Optional<LinkableItem> subTopic, final Set<CategoryItem> categoryItems) {
+    private AggregateMessageContent createAggregateMessageContent(final LinkableItem topicItem, final Optional<LinkableItem> subTopic, final Set<CategoryItem> categoryItems, AlertNotificationWrapper notification) {
```
jamesrichard91: Shouldn't we just be passing in the creation date if that is the only field we are using from the notification?

src/main/java/com/synopsys/integration/alert/workflow/processor/MessageContentAggregator.java
```diff
@@ -50,6 +53,7 @@
 
 @Component
 public class MessageContentAggregator {
+    private static final Logger logger = LoggerFactory.getLogger(MessageContentAggregator.class);
```
bamandel: It looks like we needed a logger then never actually used it.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageContentCollector.java
psantos1113: We used to pass just the notificationID in another method.  Then I added the date for the provider creation.  Rather than just passing each individual field I pass the actual notification in case in the future we need to grab other fields from the notification.  The dependency is on data within AlertNotificationWrapper so pass the object and allow the method to extract what it needs from the object rather than passing each individual field.  That's why I opted to use the object instead of just the date.

src/main/java/com/synopsys/integration/alert/workflow/processor/MessageContentAggregator.java
psantos1113: It was used earlier for debugging.  I will remove it.

PR #499, opened 2019-07-11
New resource loader to use anytime we access resources.

bamandel:


alert-common/src/main/java/com/synopsys/integration/alert/common/AlertResourceLoader.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public InputStream getInputStreamResource(String path) throws IOException {
+        return getResource(path).getInputStream();
+    }
+
+    public Resource getResource(String path) {
+        return new ClassPathResource(path);
+    }
+
+    public String getTemplatePath(String channelDirectory) {
```
gkillough: Do we want the variable to be named `channelDirectory`? I know where this code is coming from, but nothing about the class or method name suggests that it is channel specific.

bamandel: I thought of that as well but couldn't think of a time when we would have a template without a channel associated.

PR #501, opened 2019-07-11
Ps fix summary

psantos1113:
Refactor the combination of items in the summary processor.  Add the type to the CategoryKey to be consistent with 5.0.0 keys that are created.


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -94,9 +94,11 @@ private AggregateMessageContent summarize(final AggregateMessageContent message)
 
     private LinkedHashSet<CategoryItem> createSummarizedCategoryItems(final ItemOperation operation, final Set<CategoryItem> categoryItemsForOperation) {
         final List<CategoryItem> summarizedCategoryItems = new LinkedList<>();
+        final Map<CategoryKey, Map<String, List<LinkableItem>>> itemsByCategory = collectCategoryItemData(categoryItemsForOperation);
```
gkillough: This feels a little hacky, but since this is 4.2.0, I don't think it is worth adding extra models.

psantos1113: I had the same thought.  I wanted to limit the number of changes.  In master I am re-testing.  I think we already fixed the issue in master.

PR #509, opened 2019-07-17
Custom Message Endpoint

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/CustomMessageConfigModel.java
```diff
@@ -28,38 +28,48 @@
 
 import com.synopsys.integration.alert.common.persistence.accessor.FieldAccessor;
 
-public class TestConfigModel extends AlertSerializableModel {
+public class CustomMessageConfigModel extends JobFieldModel {
```
bamandel: If global test actions are also being converted to use CustomMessageConfigModel, we should not extend JobFieldModel. Can we get this functionality while still extending AlertSerializableModel?

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/CustomMessageConfigModel.java
```diff
@@ -28,38 +28,48 @@
...
     private final String destination;
     private final FieldAccessor fieldAccessor;
-    private String configId;
```
bamandel: configId was useful here because more than just jobs can be tested. Now that tests use this model, we have to reference this value through jobId which will be confusing.

PR #512, opened 2019-07-19
Fix: Make summary format consistent with what it was in 4.2.0

gkillough:


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckVulnerabilityCollector.java
```diff
@@ -183,6 +183,7 @@ private LinkableItem getSeverity(final String vulnerabilityUrl) {
         }
 
         severityItem.setSummarizable(true);
+        severityItem.setPartOfKey(true);
```
gkillough: Now that the `LinkableItem` name is the same for all vulnerabilities (rather than "High Vulnerabilities" etc) this is necessary. I does not appear to have an impact on other format types.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -123,20 +120,41 @@ private ProviderMessageContent summarize(final ProviderMessageContent message) {
...
+    private String createSummaryKey(ComponentKeys componentKeys, ComponentAttributeMap itemsOfSameName) {
+        StringBuilder summaryKeyBuilder = new StringBuilder();
+        summaryKeyBuilder.append(componentKeys.getShallowKey());
+        for (List<LinkableItem> namedItems : itemsOfSameName.values()) {
+            if (namedItems.size() == 1) {
+                final LinkableItem item = namedItems
+                                              .stream()
+                                              .findAny()
+                                              .orElseThrow();
+                if (!item.isCollapsible() && item.isPartOfKey()) {
```
gkillough: Severity? Yes
Vulnerabilities? No

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -123,20 +120,41 @@ private ProviderMessageContent summarize(final ProviderMessageContent message) {
...
 
+    private String createSummaryKey(ComponentKeys componentKeys, ComponentAttributeMap itemsOfSameName) {
+        StringBuilder summaryKeyBuilder = new StringBuilder();
+        summaryKeyBuilder.append(componentKeys.getShallowKey());
+        for (List<LinkableItem> namedItems : itemsOfSameName.values()) {
+            if (namedItems.size() == 1) {
+                final LinkableItem item = namedItems
+                                              .stream()
+                                              .findAny()
+                                              .orElseThrow();
```
gkillough: What are your thoughts on the basic .orElseThrow() when we've already verified that there is exactly one item?

PR #515, opened 2019-07-19
Fix(startup): If the overwrite is set then change the values for the …

psantos1113:
…configuration.

Implement the 5.0.0 fix for Startup Initializer to preserve settings on reboot.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
+++ b/src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
@@ -153,10 +153,12 @@ public class AlertStartupInitializer {
         throws AlertDatabaseConstraintException {
         if (!configurationModels.isEmpty()) {
             if (!foundConfigurationModels.isEmpty()) {
-                final ConfigurationModel foundModel = foundConfigurationModels.get(0);
-                logger.info("  Overwriting configuration values with environment for descriptor.");
-                final Collection<ConfigurationFieldModel> updatedFields = updateAction(descriptorName, configurationModels, foundModel, overwriteCurrentConfig);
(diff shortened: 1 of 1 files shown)


src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -153,10 +153,12 @@ private void updateConfigurationFields(final String descriptorName, final boolea
         throws AlertDatabaseConstraintException {
         if (!configurationModels.isEmpty()) {
             if (!foundConfigurationModels.isEmpty()) {
-                final ConfigurationModel foundModel = foundConfigurationModels.get(0);
-                logger.info("  Overwriting configuration values with environment for descriptor.");
-                final Collection<ConfigurationFieldModel> updatedFields = updateAction(descriptorName, configurationModels, foundModel, overwriteCurrentConfig);
-                fieldConfigurationAccessor.updateConfiguration(foundModel.getConfigurationId(), updatedFields);
+                if (overwriteCurrentConfig) {
+                    final ConfigurationModel foundModel = foundConfigurationModels.get(0);
```
gkillough: We should really be using .stream().findFirst().orElseThrow(...)

gkillough: Especially considering this is on an unconstrained `Set`.

gkillough: We should get rid of the second if-statement and replace it with 
```java 
Optional<ConfigurationModel> optionalFoundModel = foundConfigurationModels
    .stream()
    .findFirst()
    .filter(overwriteCurrentConfig);
```

PR #516, opened 2019-07-19
Fix(startup): Update the initializer based on PR feedback.

psantos1113:
update from PR 515


src/main/java/com/synopsys/integration/alert/workflow/startup/AlertStartupInitializer.java
```diff
@@ -153,12 +154,14 @@ private void updateConfigurationFields(final String descriptorName, final boolea
         throws AlertDatabaseConstraintException {
         if (!configurationModels.isEmpty()) {
             if (!foundConfigurationModels.isEmpty()) {
```
gkillough: We no longer need this line. Instead of the `.orElseThrow()` on line 161, we should get the optional and use that for the if-statement.

PR #520, opened 2019-07-25
Allow email distribution jobs to include manually added email addresses

gkillough:


src/main/js/field/ProviderDataSelectField.js
```diff
@@ -41,9 +41,9 @@ class ProviderDataSelectField extends Component {
         console.log(`Post Fetch: ${this.state.providerData}`);
     }
 
-    providerDataError(errorMessage) {
+    providerDataError(fetched, errorMessage) {
```
jamesrichard91: Is fetched ever going to be false?

gkillough: When the constructor is called.

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -196,6 +196,7 @@ class DistributionConfiguration extends Component {
                     fieldErrors={this.props.fieldErrors}
                     self={this}
                     stateName="providerConfig"
+                    provider={currentProvider.name}
```
bamandel: why are we passing provider here? Is this left over from some previous old setup?

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -251,6 +252,7 @@ class DistributionConfiguration extends Component {
                                 fieldErrors={this.props.fieldErrors}
                                 self={this}
                                 stateName="channelConfig"
+                                provider={selectedProvider}
```
bamandel: why are we passing provider here? Is this left over from some previous old setup?

src/main/js/field/ProviderDataSelectField.js
```diff
@@ -0,0 +1,162 @@
...
+                isSearchable={searchable}
+                removeSelected={removeSelected}
+                options={options}
+                placeholder={placeholder}
+                value={selectValue}
+                isMulti={multiSelect}
+                closeMenuOnSelect={!multiSelect}
+                components={components}
+                isDisabled={readOnly}
+                noOptionsMessage={this.noOptionsMessage}
```
bamandel: Instead of having a separate function you can do something like this (() => this.state.errorMessage)

src/main/js/field/ProviderDataSelectField.js
```diff
@@ -0,0 +1,162 @@
...
+        });
+    }
+
+    fetchProviderData(endpoint, providerName) {
+        const resolvedEndpoint = endpoint.replace('{provider}', providerName);
+        const requestUrl = `/alert/api${resolvedEndpoint}`;
+        return fetch(requestUrl, {
+            credentials: 'same-origin'
+        })
+            .then((response) => {
```
bamandel: It looks like your save actions are trying to format the JS. I think we generally want the .then on the same line as the  closing parenthesis

src/main/js/field/ProviderDataSelectField.js
```diff
@@ -0,0 +1,162 @@
...
+        </div>);
+        return (
+            <LabeledField field={field} {...this.props} />
+        );
+    }
+}
+
+ProviderDataSelectField.propTypes = {
+    id: PropTypes.string,
+    providerDataEndpoint: PropTypes.string,
```
bamandel: you could make this prop required as it seems vital to this components functionality.

src/main/js/store/actions/types.js
```diff
@@ -25,6 +25,11 @@ export const PROJECTS_FETCHING = 'PROJECTS_FETCHING';
 export const PROJECTS_FETCHED = 'PROJECTS_FETCHED';
 export const PROJECTS_FETCH_ERROR = 'PROJECTS_FETCH_ERROR';
 
+// User email Actions
+export const USER_EMAIL_FETCHING = 'USER_EMAIL_FETCHING';
+export const USER_EMAIL_FETCHED = 'USER_EMAIL_FETCHED';
+export const USER_EMAIL_FETCH_ERROR = 'USER_EMAIL_FETCH_ERROR';
```
bamandel: Are these still neccessary? It doesn't look like anything here is part of redux.

PR #523, opened 2019-07-29
Exposing the missing transition errors

jamesrichard91:


jamesrichard91:
Screen Shot 2019-07-29 at 3 50 53 PM
Screen Shot 2019-07-29 at 3 51 11 PM

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -124,43 +125,54 @@ public String createOrUpdateIssues(final FieldAccessor fieldAccessor, final Mess
...
+                    } else {
+                        logger.warn("Expected to find an existing issue with key '{}' but none existed.", trackingKey);
+                    }
                 }
+            } catch (JiraMissingTransitionException e) {
+                missingTransitions.add(e.getTransition());
             }
         }
+        if (!missingTransitions.isEmpty()) {
+            final String joinedMissingTransitions = missingTransitions.stream().collect(Collectors.joining(","));
```
bamandel: Are we sure we want to loop through all transitions instead of failing on the first one? They'll still get an error for each one as they're fixing them.

psantos1113: For a user it would be better to know all the problems at once than to fix test and have it fail again.  If we know all the issues it is better to report them to the user once so they can address them all.  It would be similar to the compiler telling you the first compiler error out of 10 it finds.  You would need to compile 10 times to find and fix them all.

gkillough: On a UI related note, should we join with `", "` rather than `","` (add a space)?

PR #525, opened 2019-08-01
Ps fix vulnerability summary

psantos1113:
Fix the summary to contain totals and remove the remediation information from the results.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
@@ -29,7 +29,6 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -110,6 +109,8 @@ public class SummaryMessageContentProcessor extends MessageContentProcessor {
     }
 
     private LinkedHashSet<ComponentItem> summarizeCategoryComponentCounts(Set<ComponentItem> componentItems) {
+        // Organize the items by category such as policy and vulnerability.
+        // Then have the item priority with the set of components.
         Map<String, Map<ComponentItemPriority, Set<ComponentItem>>> itemsByCategory = new LinkedHashMap<>();
 
         for (final ComponentItem item : componentItems) {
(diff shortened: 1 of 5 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -105,60 +109,78 @@ private ProviderMessageContent summarize(final ProviderMessageContent message) {
...
+            for (Map.Entry<ComponentItemPriority, Set<ComponentItem>> priorityEntries : entry.getValue().entrySet()) {
+                try {
+                    Set<ComponentItem> items = priorityEntries.getValue();
 
-    private Map<String, ComponentAttributeMap> collectComponentItemData(Set<ComponentItem> componentItemsForOperation) {
-        Map<String, ComponentAttributeMap> itemsByCategory = new LinkedHashMap<>();
-        for (final ComponentItem categoryItem : componentItemsForOperation) {
-            ComponentAttributeMap itemsOfSameName = categoryItem.getItemsOfSameName();
-            String summaryKey = createSummaryKey(categoryItem.getComponentKeys(), itemsOfSameName);
+                    List<LinkableItem> attributeSet = items.stream()
```
gkillough: This variable is not a set, so we should call it something like `summarizableAttributes`.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -105,60 +109,78 @@ private ProviderMessageContent summarize(final ProviderMessageContent message) {
...
+                                                          .map(ComponentItem::getComponentAttributes)
+                                                          .flatMap(Set::stream)
+                                                          .filter(LinkableItem::isSummarizable)
+                                                          .collect(Collectors.toList());
+                    if (!attributeSet.isEmpty()) {
+                        Set<Long> notificationIds = items.stream()
+                                                        .map(ComponentItem::getNotificationIds)
+                                                        .flatMap(Set::stream)
+                                                        .collect(Collectors.toSet());
+                        Optional<ComponentItem> componentItem = items.stream().findFirst();
```
gkillough: Since we only use this variable once, we could immediately map this to Operation and get the value orElse some default (UPDATE?).

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -105,60 +109,78 @@ private ProviderMessageContent summarize(final ProviderMessageContent message) {
...
-                if (!summarizedLinkableItems.isEmpty()) {
-                    ComponentItem newComponentItem = messageOperationCombiner.createNewComponentItem(componentItem, summarizedLinkableItems);
-                    summarizedCategoryItems.add(newComponentItem);
-                }
-            } catch (AlertException e) {
-                // If this happens, it means there is a bug in the Collector logic.
-                throw new AlertRuntimeException(e);
-            }
+    private LinkedHashSet<ComponentItem> summarizeCategoryComponentCounts(Set<ComponentItem> componentItems) {
+        Map<String, Map<ComponentItemPriority, Set<ComponentItem>>> itemsByCategory = new LinkedHashMap<>();
```
bamandel: Objects like this are very tough to follow. Would we be able to convert this into an object instead?

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/SummaryMessageContentProcessor.java
```diff
@@ -105,60 +109,78 @@ private ProviderMessageContent summarize(final ProviderMessageContent message) {
...
+        for (final LinkableItem item : attributes) {
+            final String name = item.getName();
+            attributeMap.computeIfAbsent(name, ignored -> new LinkedList<>()).add(item);
+        }
+        return createSummarizedLinkableItems(attributeMap);
     }
 
-    private SortedSet<LinkableItem> createSummarizedLinkableItems(ComponentAttributeMap componentAttributeMap) {
-        final SortedSet<LinkableItem> summarizedLinkableItems = new TreeSet<>();
+    private Collection<LinkableItem> createSummarizedLinkableItems(ComponentAttributeMap componentAttributeMap) {
```
bamandel: Don't we want the return type to be List here and use Collection for passed args?

PR #527, opened 2019-08-02
Ps single select error

psantos1113:
Add the clearable option for the select fields.  
Always create an array for selected values.  Just push the single value when a single value is selected.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/SelectConfigField.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/SelectConfigField.java
@@ -37,67 +37,79 @@ public class SelectConfigField extends ConfigField {
     private boolean searchable;
     private boolean multiSelect;
     private boolean removeSelected;
+    private boolean clearable;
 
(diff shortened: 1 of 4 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/SelectConfigField.java
```diff
@@ -36,92 +36,124 @@
     private Collection<LabelValueSelectOption> options;
     private boolean searchable;
     private boolean multiSelect;
+    private boolean removeSelected;
```
gkillough: Will there ever be a case where we don't want to do this?

psantos1113: Right now I don't see a reason not to always do this for our product.

src/main/js/field/input/DynamicSelect.js
```diff
@@ -4,19 +4,20 @@ import Select from 'react-select';
 import LabeledField from 'field/LabeledField';
 
 function noOptionsMessage() {
-    return null;
+    return 'No options available...';
```
bamandel: This function can be removed and the noOptions prop can be passed {() => 'No options available...'} instead

PR #530, opened 2019-08-06
Ps saml auth fix

psantos1113:
After system setup redirect to the correct login page by reloading the application.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
+++ b/src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
@@ -25,6 +25,7 @@ package com.synopsys.integration.alert.web.security.authentication;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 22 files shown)


src/main/java/com/synopsys/integration/alert/web/security/authentication/AlertAuthenticationProvider.java
```diff
@@ -107,7 +108,9 @@ private boolean isAuthorized(final Authentication authentication) {
                    .map(GrantedAuthority::getAuthority)
                    .filter(role -> role.startsWith(UserModel.ROLE_PREFIX))
                    .map(role -> StringUtils.substringAfter(role, UserModel.ROLE_PREFIX))
-                   .map(UserRole::valueOf)
+                   .map(UserRole::findUserRole)
+                   .filter(Optional::isPresent)
```
gkillough: This should be `.flatMap(Optional::stream)`.

src/main/js/store/actions/system.js
```diff
@@ -120,9 +120,10 @@ export function getLatestMessages() {
         fetch(LATEST_MESSAGES_URL)
             .then((response) => {
                 if (response.ok) {
-                    response.json().then((body) => {
-                        dispatch(latestSystemMessagesFetched(body));
-                    });
+                    response.json()
```
gkillough: Most of the changes in here are just formatting. You might want to double check your JS save actions.

bamandel: The SaveActions in Intellij are intended for Java and often mess up our JS files because of it. I usually disable the save actions plugin when working in JS land.

src/main/js/App.js
```diff
@@ -17,8 +17,17 @@ IconUtility.loadIconData();
...
 class App extends Component {
     componentDidMount() {
-        this.props.verifyLogin();
         this.props.getSettings();
+        this.props.verifyLogin();
+    }
+
+    componentDidUpdate(prevProps) {
+        if (this.props.systemInitialized && !prevProps.systemInitialized
+            && !this.props.logoutPerformed && !this.props.loggedIn) {
```
bamandel: could be cleaner to call the props via `const {systemInitializer, logourPerformed, loggedIn} = this.props;`

PR #532, opened 2019-08-08
Update Permissions Matrix to use relations rather than hard-coded Strings

gkillough:


src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
```diff
@@ -60,49 +64,49 @@
...
     public ResponseEntity<String> getJobs() {
-        if (!authorizationManager.anyReadPermission(REQUIRED_PERMISSION_KEYS)) {
-            return responseFactory.createForbiddenResponse();
-        }
-        final List<JobFieldModel> models = new LinkedList<>();
         try {
+            final Set<String> descriptorNames = descriptorAccessor.getRegisteredDescriptors()
+                                                    .stream()
+                                                    .map(RegisteredDescriptorModel::getName)
+                                                    .filter(name -> name.startsWith("channel") || name.startsWith("provider"))
```
bamandel: This is a very risky check. Going forward, we probably want to remove these sort of checks wherever we see them and use some other method.

bamandel: To add to this, Eric right now is talking to Paulo about calling his msteams channel something that doesn't much the assumed state above. If we have new people adding content, we'll want to find a better solution to this (Like checking the actual descriptor type).

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
```diff
@@ -117,10 +121,8 @@ public JobConfigController(final JobConfigActions jobConfigActions, final Respon
 
         if (optionalModel.isPresent()) {
             final JobFieldModel fieldModel = optionalModel.get();
-            final boolean missingPermission = fieldModel.getFieldModels().stream()
-                                                  .map(model -> AuthorizationManager.generatePermissionKey(model.getContext(), model.getDescriptorName()))
-                                                  .anyMatch(permissionKey -> !authorizationManager.hasReadPermission(permissionKey));
-            if (missingPermission) {
+            final boolean hasPermissions = fieldModel.getFieldModels().stream().allMatch(model -> authorizationManager.hasReadPermission(model.getContext(), model.getDescriptorName()));
```
bamandel: These are used in multiple places. We should create a new method that is passed a function to do the specific permission check.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
psantos1113: there is the DescriptorType returned from the getType().  We should filter on that rather than name.

PR #533, opened 2019-08-08
Custom endpoint button success message

bamandel:


src/main/js/field/EndpointField.js
```diff
@@ -6,6 +6,7 @@ import LabeledField from 'field/LabeledField';
 import * as FieldModelUtilities from 'util/fieldModelUtilities';
 import { createNewConfigurationRequest } from 'util/configurationRequestBuilder';
 import { connect } from 'react-redux';
+import StatusMessage from "./StatusMessage";
```
jamesrichard91: This import path is incorrect. Please make this an absolute path like the other imports

PR #535, opened 2019-08-08
Fix(Java): Reveal better error messages for jira issue creation failures

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -204,6 +211,37 @@ public String createOrUpdateIssues(final FieldAccessor fieldAccessor, final Mess
                    .flatMap(Stream::findFirst);
     }
 
+    private IssueResponseModel createIssue(String issueCreatorEmail, String issueType, String projectName, IssueRequestModelFieldsBuilder fieldsBuilder) throws IntegrationException {
+        try {
+            return issueService.createIssue(new IssueCreationRequestModel(issueCreatorEmail, issueType, projectName, fieldsBuilder, List.of()));
+        } catch (IntegrationRestException e) {
```
bamandel: It seems strange that the method is called createIssue but the bulk functionality of the method actually deals with error handling. Could we handle this exception in the original method then call an errorHandling method instead?

psantos1113: Would it be better to have a handleException method that has the bulk of the catch block as the content.  Then the catch would look like this 
`catch(IntegrationRestException e) {
     handleException(e);
}`

Is the exception content the same for other requests not just creating an issue that the exception handling code could be used for other requests?

PR #537, opened 2019-08-09
Fix(ComponentItem): Update the notification Id check to avoid nulls.

psantos1113:
Fix the sonar cloud error for potential NPE.


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -128,7 +128,8 @@ public ComponentAttributeMap getItemsOfSameName() {
         private Set<Long> notificationIds = new LinkedHashSet<>();
 
         public ComponentItem build() throws AlertException {
-            if (null == componentName || null == componentValue || null == category || null == operation || (null == notificationIds && !notificationIds.isEmpty())) {
+            boolean idsMissing = Optional.ofNullable(notificationIds).map(Collection::isEmpty).orElse(false);
```
gkillough: `idsMissing` will be false if notificationIds is null. This seems to imply that the notificationIds being null means they aren't missing.  I'm not a huge fan of using optionals in this way, why not just do the null check with an `&&` inline.

gkillough: Was the original bug just the negation of `notificationIds.isEmpty()`?

psantos1113: sonarcloud complains about it which is the reason for this PR.

gkillough: Ok, I would just prefer we create the variables without the overhead of the Optional stuff.

psantos1113: I agree. I'm trying one more change first.

PR #538, opened 2019-08-12
Fix(Java): Make jira transition descriptions more clear

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/descriptor/JiraDistributionUIConfig.java
```diff
@@ -49,8 +49,8 @@
     public static final String DESCRIPTION_ISSUE_CREATOR = "The email of the Jira Cloud user to assign as the issue creator field of the Jira issue.";
     public static final String DESCRIPTION_JIRA_PROJECT = "The name or key of the Jira Project for which this job creates and/or updates Jira tickets.";
     public static final String DESCRIPTION_ISSUE_TYPE = "The issue type to open when creating an issue in Jira Cloud.";
-    public static final String DESCRIPTION_RESOLVE_WORKFLOW_TRANSITION = "If a transition is listed (Case sensitive), it will be used when resolving an issue.";
-    public static final String DESCRIPTION_OPEN_WORKFLOW_TRANSITION = "If a transition is listed (Case sensitive), it will be used when re-opening an issue.";
+    public static final String DESCRIPTION_RESOLVE_WORKFLOW_TRANSITION = "If a transition is listed (case sensitive), it will be used when resolving an issue (i.e. performing a DELETE operation on an open issue).";
+    public static final String DESCRIPTION_OPEN_WORKFLOW_TRANSITION = "If a transition is listed (case sensitive), it will be used when re-opening an issue (i.e. performing an ADD operation on a resolved issue).";
```
bamandel: Can we specify that this is an Alert DELETE and ADD? At first I thought we were saying we were saying a delete in Jira.

gkillough: I think the "...performing an ADD operation on a resolved issue" gets the point across, no?

bamandel: Really the DELETE was the one that made me question what we were doing. The ADD was more clear as that is specific to Alert. Just thought we may confuse users about deleting their issues.

PR #540, opened 2019-08-12
Bm notification type column

bamandel:
Added a column to jobs to display the notification types associated with the job. The same as in audit.


gkillough:
I'm confused. I thought this was supposed to be dynamic?

bamandel:
> I'm confused. I thought this was supposed to be dynamic?

In the future. The table isn't dynamically created yet, the fields are.

PR #543, opened 2019-08-13
Provider data controller paged requests

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/provider/user/ProviderUserRepository.java
```diff
@@ -24,11 +24,21 @@
...
+import org.springframework.data.repository.query.Param;
 
 public interface ProviderUserRepository extends JpaRepository<ProviderUserEntity, Long> {
     List<ProviderUserEntity> findByProvider(final String provider);
 
+    @Query(value = "SELECT DISTINCT new ProviderUserEntity(providerUser.emailAddress, providerUser.optOut, providerUser.provider) "
+                       + "FROM ProviderUserEntity providerUser "
+                       + "WHERE providerUser.provider = :provider "
+                       + "AND providerUser.emailAddress LIKE %:q%")
+    Page<ProviderUserEntity> findPageOfUsersByProvider(@Param("provider") String provider, @Param("q") String q, Pageable pageable);
```
psantos1113: Can we rename this to findPageOfUsersByProviderEmail?  The query parameter is only used on the emailAddress field.

gkillough: More accurately it would be `findPageOfUsersByProviderAndMatchingEmail` or `findPageOfUsersByProviderAndFilterEmail`. We could rename the parameter to make it more clear. Would either of these work for you?

PR #544, opened 2019-08-13
Removed all references to channel_ or provider_

bamandel:


src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -66,8 +66,8 @@ class DistributionConfiguration extends Component {
                 this.loading = false;
                 const { job } = nextProps;
                 if (job && job.fieldModels) {
-                    const channelModel = nextProps.job.fieldModels.find(model => model.descriptorName.startsWith('channel_'));
-                    const providerModel = nextProps.job.fieldModels.find(model => model.descriptorName.startsWith('provider_'));
+                    const channelModel = nextProps.job.fieldModels.find(model => FieldModelUtilities.hasKey(model, KEY_NAME));
```
gkillough: Does KEY_NAME need to be imported?

psantos1113: It does not because it is exported as a const in this javascript file.

src/main/js/distribution/Index.js
```diff
@@ -274,8 +274,8 @@ class Index extends Component {
         if (jobs) {
             jobs.forEach((job) => {
                 if (job && job.fieldModels) {
-                    const channelModel = job.fieldModels.find(fieldModel => fieldModel.descriptorName.startsWith('channel_'));
-                    const providerModel = job.fieldModels.find(fieldModel => fieldModel.descriptorName.startsWith('provider_'));
+                    const channelModel = job.fieldModels.find(fieldModel => FieldModelUtilities.hasKey(fieldModel, KEY_NAME));
```
psantos1113: KEY_NAME maps to the common name for the distribution.  That isn't channel or provider specific.  I think you want to import and use KEY_CHANNEL_NAME, and KEY_PROVIDER_NAME.

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -66,8 +66,8 @@ class DistributionConfiguration extends Component {
...
+                    const providerModel = nextProps.job.fieldModels.find(model => !FieldModelUtilities.hasKey(model, KEY_NAME));
```
psantos1113: I think you want to use KEY_PROVIDER_NAME

PR #548, opened 2019-08-15
Auto-generated Dynamic value drop down list

bamandel:
Converted ProviderDataSelectField to EndpointSelectField where we can now dynamically populate select fields with data based off of any needed fields instead of requiring only Provider.

the 'Additional Email Addresses' field temporarily uses this new field but will eventually be changed to a new table field that allows large sets of data to be displayed in the UI.


src/main/js/field/EndpointSelectField.js
```diff
@@ -0,0 +1,87 @@
...
+                    });
+                });
+            }
+        });
+    }
+
+    render() {
+        return (
+            <div>
+                <DynamicSelectInput onChange={this.props.onChange} onFocus={this.onSendClick} options={this.state.options} {...this.props} />
```
gkillough: I still think we should be using AsyncSelect here.

src/main/js/field/input/DynamicSelect.js
```diff
@@ -53,12 +54,13 @@ function DynamicSelectInput(props) {
...
             options={options}
             placeholder={placeholder}
-            value={value}
+            value={selectValue}
             isMulti={multiSelect}
             closeMenuOnSelect={!multiSelect}
             components={components}
             isDisabled={readOnly}
             noOptionsMessage={() => 'No options available'}
+            onFocus={onFocus}
```
gkillough: I'm not sold on only taking action onFocus. It seems like a hack.

bamandel: If we were to not use AsyncSelect, what would be the better option here? We don't want to be making requests every single state change and this seems like a clean way to resolve this.

src/main/js/field/EndpointSelectField.js
bamandel: Just looked more into AsyncSelect, the only real benefit would be async functionality. I don't really see any other benefits from it. If we use the AsyncSelect, we'll need to add more code for styling, state changes, content and other things which is already done in DynamicSelectInput.

psantos1113: The benefit of the AsyncSelect is that it will handle loading the data by submitting a request.  It will handle displaying that data is being loaded.  It's a component that has been built for this purpose.  DynamicSelectInput is our own home grown input field for select inputs that we are now retro-fitting to be an asynchronous select field.

src/main/js/field/input/DynamicSelect.js
psantos1113: If we do not use the AsyncSelect we need to use an event to initiate a request to fetch data.  onFocus is one event on the component that could be used.  The other is potentially onClick.  Ideally if we can add an onClick handler to the button used to display the list of options to select that would work better.

src/main/js/field/EndpointSelectField.js
bamandel: The only real change we made to DynamicSelectInput was to allow someone to pass the underlying Select an additional property. For the AsyncSelect, we still had to make the request and pass it through. The displaying is now being handled by our own Select. So it still seems to me like we're only missing out on async functionality here.

src/main/js/field/input/DynamicSelect.js
bamandel: After looking at this https://react-select.com/props, I am not sure that we have the option of an onClick for the drop down itself. Only for the options of the drop down. So I think onFocus would still be preferred.

PR #549, opened 2019-08-16
Ps swarm support

psantos1113:
Changes to update the deployment to avoid issues with volumes and handle certificate issues as well deploying in swarm.


deployment/docker-compose/docker-compose.yml
```diff
@@ -18,10 +18,12 @@ services:
...
     healthcheck:
-      test: [CMD, /usr/local/bin/docker-healthcheck.sh, 'https://localhost:8443/alert/api/about']
+      test: [CMD, /usr/local/bin/docker-healthcheck.sh, 'https://localhost:8443/alert/api/about',
+             /opt/blackduck/alert/security/root.crt, /opt/blackduck/alert/security/blackduck_system.crt,
+             /opt/blackduck/alert/security/blackduck_system.key]
       interval: 30s
-      timeout: 10s
-      retries: 5
+      timeout: 60s
+      retries: 15
```
gkillough: Why the huge jump? Shouldn't this just be a one or the other tweak?

psantos1113: This is primarily to be consistent with the black duck containers since we are potentially included in the same sandbox. We are also communicating over SSL in the healthcheck scripts therefore it may take longer to establish the connection over SSL with the new script.  When running in swarm mode containers can exit more frequently so the retry count was increased. It's mainly for consistency.

PR #550, opened 2019-08-16
Ek windows build

ekerwin:
improving the build with a little consistency and getting to work on windows

also, added a --nosplash command line arg for the runServer task to provide some default values to bypass the splash screen


.gitignore
```diff
@@ -14,7 +14,7 @@ out/
 
 # build tool
 .gradle/
-**/build/*
+build/
```
gkillough: Will this still ignore the build directories in alert-common and alert-database?

build.gradle
```diff
@@ -165,19 +163,19 @@ task createKeystore(type: Exec) {
...
+    workingDir "${project.buildDir}/certs"
     commandLine commandArray
 }
 
 task createTruststore(type: Copy) {
     from "${System.getProperty('java.home')}/lib/security/"
     include 'cacerts'
     into "${project.buildDir}/certs/"
-    rename 'cacerts', "${project.buildDir}/certs/blackduck-alert.truststore"
+    rename 'cacerts', 'blackduck-alert.truststore'
```
gkillough: Is this only used in the runServer task?

psantos1113: yes this is only used to create the trust store that alert will use when running.

PR #552, opened 2019-08-19
Bm hide related fields

bamandel:
New field type that allows us to hide other fields depending on the state of the field.

Nothing uses this at the moment but will once we add the last new field type.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/HideCheckboxConfigField.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/HideCheckboxConfigField.java
@@ -28,4 +28,12 @@ public class HideCheckboxConfigField extends CheckboxConfigField {
         relatedHiddenFields.addAll(Stream.of(key).collect(Collectors.toList()));
         return this;
     }
+
+    public List<String> getRelatedHiddenFields() {
+        return relatedHiddenFields;
+    }
+
+    public void setRelatedHiddenFields(final List<String> relatedHiddenFields) {
+        this.relatedHiddenFields = relatedHiddenFields;
+    }
 }
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/HideCheckboxConfigField.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/HideCheckboxConfigField.java
@@ -1,0 +1,31 @@
+package com.synopsys.integration.alert.common.descriptor.config.field;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.synopsys.integration.alert.common.enumeration.FieldType;
+
+public class HideCheckboxConfigField extends CheckboxConfigField {
+    private List<String> relatedHiddenFields;
+
+    public static HideCheckboxConfigField create(final String key, final String label, final String description) {
+        return new HideCheckboxConfigField(key, label, description, false);
+    }
+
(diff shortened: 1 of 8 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/HideCheckboxConfigField.java
```diff
@@ -0,0 +1,39 @@
...
+        super(key, label, description, FieldType.HIDE_CHECKBOX_INPUT, required);
+        relatedHiddenFields = new LinkedList<>();
+    }
+
+    public HideCheckboxConfigField addHiddenFieldKey(String key) {
+        relatedHiddenFields.add(key);
+        return this;
+    }
+
+    public HideCheckboxConfigField addHiddenFieldKeys(String... key) {
```
psantos1113: I would consider adding addRelatedHiddenFieldKeys to match the get method.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/HideCheckboxConfigField.java
```diff
@@ -0,0 +1,39 @@
...
+    public static HideCheckboxConfigField create(final String key, final String label, final String description) {
+        return new HideCheckboxConfigField(key, label, description, false);
+    }
+
+    public HideCheckboxConfigField(final String key, final String label, final String description, final boolean required) {
+        super(key, label, description, FieldType.HIDE_CHECKBOX_INPUT, required);
+        relatedHiddenFields = new LinkedList<>();
+    }
+
+    public HideCheckboxConfigField addHiddenFieldKey(String key) {
```
psantos1113: I would consider adding addRelatedHiddenFieldKeys to match the get method.

src/main/js/field/FieldsPanel.js
```diff
@@ -67,7 +84,16 @@ class FieldsPanel extends React.Component {
...
+        const { keyToValues } = this.props.currentConfig;
+        const hasValues = Object.keys(panelFields).some((fieldKey) => {
+            const field = panelFields[fieldKey];
+            const { key, type } = field;
+            if (!this.state.hiddenFieldKeys.includes(key)) {
+                return (type === 'CheckboxInput' || type === 'HideCheckboxInput') ? FieldModelUtilities.checkboxHasValue(keyToValues[key]) : FieldModelUtilities.hasValuesOrIsSet(keyToValues[key]);
+            }
+            return false;
+        });
+        console.log(`${panelName} has values: ${hasValues}`);
```
psantos1113: remove this console statement

src/main/js/field/FieldsPanel.js
```diff
@@ -104,8 +130,10 @@ class FieldsPanel extends React.Component {
 
         fields.forEach((field) => {
             const fieldKey = field.key;
-            const newField = FieldMapping.createField(field, currentConfig, fieldErrors[fieldKey], this.handleChange);
-            createdFields.push(newField);
+            if (!this.state.hiddenFieldKeys.includes(fieldKey)) {
```
psantos1113: You might want to look into creating an isHiddenField function since you use this.state.hiddenFieldKeys.includes a few times might make it a little more readable.

bamandel: At the moment I believe it's the only place that uses this functionality. I could add a function for it now, but it'd contain just one line and that seems strange to me if we're not reusing it yet.

psantos1113: Ok sounds reasonable to me.  I wasn't overly concerned about it.

PR #553, opened 2019-08-19
Fix(SAML): Fix the authentication to display the Alert UI with no roles

psantos1113:
If SAML authentication is used and no alert roles are defined don't display the login page just display the Alert UI.


src/main/js/store/actions/session.js
```diff
@@ -1,14 +1,5 @@
...
-    SESSION_CANCEL_LOGOUT,
-    SESSION_CONFIRM_LOGOUT,
-    SESSION_INITIALIZING,
-    SESSION_LOGGED_IN,
-    SESSION_LOGGED_OUT,
-    SESSION_LOGGING_IN,
-    SESSION_LOGIN_ERROR,
-    SESSION_LOGOUT
-} from 'store/actions/types';
+import { SAML_ENABLED, SESSION_CANCEL_LOGOUT, SESSION_CONFIRM_LOGOUT, SESSION_INITIALIZING, SESSION_LOGGED_IN, SESSION_LOGGED_OUT, SESSION_LOGGING_IN, SESSION_LOGIN_ERROR, SESSION_LOGOUT } from 'store/actions/types';
```
bamandel: Do we want this collapsed like this or expanded as it previously was? It is a lot easier to read all these if they're expanded in my opinion.

src/main/java/com/synopsys/integration/alert/web/controller/HomeController.java
```diff
@@ -60,17 +64,23 @@ public String index() {
         final HttpServletRequest httpRequest = request;
         final CsrfToken csrfToken = csrfTokenRespository.loadToken(request);
         final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-        final boolean hasAlertRole = authorizationManager.hasAlertRole();
-        final boolean authorized = authentication.isAuthenticated() && hasAlertRole && csrfToken != null;
+        final boolean isAnonymous = authentication.getAuthorities().stream()
+                                        .map(authority -> authority.getAuthority())
```
gkillough: This could be a method reference.

src/main/java/com/synopsys/integration/alert/web/controller/HomeController.java
```diff
@@ -60,17 +64,23 @@ public String index() {
...
+                                        .collect(Collectors.toSet())
```
gkillough: Instead of collecting, you can use `.anyMatch(...)`

src/main/java/com/synopsys/integration/alert/web/controller/HomeController.java
```diff
@@ -60,17 +64,23 @@ public String index() {
...
+                                        .contains("ROLE_ANONYMOUS");
```
gkillough: Where does this role come from? Should it be a constant?

src/main/java/com/synopsys/integration/alert/web/controller/HomeController.java
```diff
@@ -60,17 +64,23 @@ public String index() {
...
         } else {
-            final String body = "{\"message\":\"Authenticated\"}";
+            final ResponseBodyBuilder responseBody = new ResponseBodyBuilder("Authenticated");
+            responseBody.put("saml_enabled", samlContext.isSAMLEnabled());
+
             final HttpHeaders headers = new HttpHeaders();
             headers.add(csrfToken.getHeaderName(), csrfToken.getToken());
-            return new ResponseEntity<>(body, headers, HttpStatus.OK);
+
+            return responseFactory.createResponse(HttpStatus.OK, headers, responseBody.build());
```
gkillough: Maybe we could return a 204 or something here rather than driving our UI with the `"Authenticated"` String from above?

src/main/js/store/actions/session.js
```diff
@@ -77,7 +86,14 @@ export function verifyLogin() {
                 dispatch(loggedOut());
             } else {
                 const token = response.headers.get('X-CSRF-TOKEN');
-                dispatch(loggedIn({ csrfToken: token }));
+                response.json().then(body => {
+                    const { message, saml_enabled } = body;
+                    if (message === 'Authenticated') {
```
gkillough: We could check for a 204 here or something instead of looking for a String message body.

src/main/java/com/synopsys/integration/alert/web/controller/HomeController.java
psantos1113: It is the default assigned.  I will make it a constant in the class.

src/main/js/store/actions/session.js
psantos1113: I'll check the status code.

PR #554, opened 2019-08-21
Ps startup logging

psantos1113:
Log the configuration fields for each descriptor.


src/main/java/com/synopsys/integration/alert/workflow/startup/component/AlertStartupInitializer.java
```diff
@@ -133,6 +133,7 @@ private void initializeConfiguration(final Collection<String> descriptorNames, f
             logger.info("---------------------------------");
             logger.info("Descriptor: {}", descriptorName);
             logger.info("---------------------------------");
+            logger.info("  Start Descriptor Initialization...");
```
gkillough: To match the other verbs that we log, this should probably be "Starting".

PR #555, opened 2019-08-21
Ek msteams init

ekerwin:
Adding MS Teams


src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsChannel.java
```diff
@@ -0,0 +1,70 @@
...
+import freemarker.template.TemplateException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.util.HashMap;
+
+@Component(value = MsTeamsChannel.COMPONENT_NAME)
+public class MsTeamsChannel extends NamedDistributionChannel {
+    public static final String COMPONENT_NAME = "channel_msteams";
```
bamandel: Did you try without the channel prefix? That should be possible now.

src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsEventParser.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.StringWriter;
+import java.util.Map;
+
+@Component
+public class MsTeamsEventParser {
+    private FreemarkerTemplatingService freemarkerTemplatingService;
+    private Configuration freemarkerConfiguration;
+    private Template msTeamsTemplate;
+
+    public MsTeamsEventParser(FreemarkerTemplatingService freemarkerTemplatingService) throws IntegrationException {
```
bamandel: Should this be autowired?

src/main/resources/db/changelog/alert/5.1.0/changelog-global-msteams.xml
```diff
@@ -0,0 +1,33 @@
...
+                schemaName="ALERT"
+                tableName="REGISTERED_DESCRIPTORS">
+            <column name="NAME" value="channel_msteams"/>
+            <column name="TYPE_ID" valueComputed="GET_ID_FOR_DESCRIPTOR_TYPE('CHANNEL')"/>
+        </insert>
+    </changeSet>
+    <changeSet author="ekerwin" id="1565734102193-2">
+        <sql dbms="h2" stripComments="true">
+            CALL ASSIGN_DEFAULT_ROLE_PERMISSIONS('channel_msteams');
+        </sql>
```
bamandel: We can change this to the new stored proc that Paulo created.

src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsEventParser.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public MsTeamsMessage createMessage(DistributionEvent distributionEvent) {
+        MessageContentGroup messageContentGroup = distributionEvent.getContent();
+
+        MsTeamsMessage msTeamsMessage = new MsTeamsMessage();
+        messageContentGroup
+                .getSubContent()
+                .stream()
+                .forEach(content -> msTeamsMessage.addContent(content));
```
gkillough: This should be a method reference: `.forEach(msTeamsMessage::addContent)`.

src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsMessage.java
```diff
@@ -0,0 +1,85 @@
...
+        sections.add(msTeamsSection);
+
+        List<MsTeamsComponent> components = new ArrayList<>(providerMessageContent.getComponentItems().size());
+        for (ComponentItem componentItem : providerMessageContent.getComponentItems()) {
+            MsTeamsComponent msTeamsComponent = new MsTeamsComponent();
+            msTeamsComponent.setCategory(componentItem.getCategory());
+            msTeamsComponent.setOperation(componentItem.getOperation().toString());
+
+            String componentText = componentItem.getComponent().getValue();
+            if (componentItem.getSubComponent().isPresent()) {
```
gkillough: Rather than having an if-statement, you could potentially do something like this:
```java
1. StringBuilder componentTextBuilder = new StringBuilder(componentItem.getComponent().getValue());
2. componentItem.getSubComponent().map(item -> "/" + item.getValue()).ifPresent(componentTextBuilder::add);
3. msTeamsComponent.setText(componentTextBuilder.toString());
```
I tend to be wary of lines like this: `componentItem.getSubComponent().get().getValue()`. We will have to go back later to fix it for SonarCloud. Using ifPresent with Optionals tends to be a better way to avoid bugs. If you have readability concerns, you can put the fluent code on new lines (our convention is all or nothing with this).

src/main/resources/db/changelog/alert/5.1.0/changelog-global-msteams.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gkillough: This file should probably be called `changelog-distribution-msteams.xml`.

src/main/resources/templates/channel/msteams/message_content.ftl
```diff
@@ -0,0 +1,16 @@
+{
+  "@type": "MessageCard",
+  "@context": "https:\/\/schema.org\/extensions",
+  "summary": "new content from Alert",
```
gkillough: If this line is user-visible, we should probably capitalize the first letter.

src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsEventParser.java
jamesrichard91: Im confused, when is this constructor being called? The channel is just auto-wiring this class in. So when is the FreemarkerTemplatingService being passed in?

src/main/java/com/synopsys/integration/alert/channel/msteams/descriptor/MsTeamsDescriptor.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class MsTeamsDescriptor extends ChannelDescriptor {
+    public static final String KEY_WEBHOOK = "channel.msteams.webhook";
+
+    public static final String MSTEAMS_LABEL = "MS Teams";
+    public static final String MSTEAMS_URL = "msteams";
+    public static final String MSTEAMS_DESCRIPTION = "Configure MS Teams for Alert.";
+    public static final String MSTEAMS_ICON = "fab/msteams";
```
jamesrichard91: I know how much you love icons, and this is not a real Font Awesome icon. you could use "fab/microsoft" instead

PR #556, opened 2019-08-22
Fix(deployment): Migrate 5.1.0 deployment fixes to 5.0.0

psantos1113:
Backport of 5.1.0 deployment file changes.

See https://github.com/blackducksoftware/blackduck-alert/pull/549 for analogous changes.

There is one file different in Kubernetes that is different because the healthcheck was not implemented in master for 5.1.0.  This PR fixes that issue.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -230,9 +230,9 @@ importBlackDuckSystemCertificateIntoKeystore() {
     exitCode=$?
     if [ $exitCode -eq 0 ];
     then
-        echo "Successfully imported BlackDuck system certificate into Java truststore."
+        echo "Successfully imported BlackDuck system certificate into Java keystore."
     else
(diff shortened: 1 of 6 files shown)


docker-entrypoint.sh
```diff
@@ -141,47 +190,91 @@ trustProxyCertificate() {
...
+  $certificateManagerDir/certificate-manager.sh trust-java-cert \
+                        --store $keystoreFilePath \
+                        --password changeit \
+                        --cert $securityDir/blackduck_system.crt \
+                        --certAlias blackduck_system
+
+    exitCode=$?
+    if [ $exitCode -eq 0 ];
+    then
+        echo "Successfully imported BlackDuck system certificate into Java truststore."
```
jamesrichard91: This log is confusing, since we are importing the certificate into the Keystore. Is the log wrong? or should we import the certificate into the trust store?

psantos1113: yes copy paste error here.  I will update it.

PR #557, opened 2019-08-22
Bm auto select table

bamandel:


src/main/js/field/input/TableSelectInput.js
```diff
@@ -0,0 +1,289 @@
...
+    retrieveTableData() {
+        this.setState({
+            progress: true,
+            success: false
+        });
+        const {
+            fieldKey, csrfToken, currentConfig, endpoint
+        } = this.props;
+
+        const request = createNewConfigurationRequest(`/alert${endpoint}/${fieldKey}`, csrfToken, currentConfig);
```
gkillough: Endpoints are always preceded by '/'?

bamandel: The code in Java is expected to pass a url of the form /custom/endpoint/blah/blah which is then inserted here. I could make this friendlier by processing the endpoint url a bit, but figured people often won't be passing their own urls here.

PR #559, opened 2019-08-22
Updating how we handle error reporting around Jira Cloud missing transitions

jamesrichard91:


src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -186,12 +187,23 @@ public String createOrUpdateIssues(final FieldAccessor fieldAccessor, final Mess
...
                 }
             } catch (JiraMissingTransitionException e) {
-                missingTransitions.add(e.getTransition());
+                if (missingTransitionToIssues.containsKey(e.getTransition())) {
+                    missingTransitionToIssues.get(e.getTransition()).add(e.getIssueKey());
+                } else {
+                    Set<String> newIssueSet = new HashSet<>();
+                    newIssueSet.add(e.getIssueKey());
+                    missingTransitionToIssues.put(e.getTransition(), newIssueSet);
+                }
```
bamandel: We have this sort of functionality in multiple places around Alert (Map of string to set and adding to the set if string exists). I think we should create a wrapper class that performs this functionality with basic methods.

bamandel:
Should this functionality be in 5.0.0 as well or only in master?

gkillough:
Are these changes meant to be in 5.0.0 or master?

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -186,12 +187,23 @@ public String createOrUpdateIssues(final FieldAccessor fieldAccessor, final Mess
                     }
                 }
             } catch (JiraMissingTransitionException e) {
-                missingTransitions.add(e.getTransition());
+                if (missingTransitionToIssues.containsKey(e.getTransition())) {
```
gkillough: Our code is really getting bloated in this class. Are there any abstractions we can make around the error handling to cut down on this?

jamesrichard91:
> Are these changes meant to be in 5.0.0 or master?

Thank you for catching that. I changed the base to 5.0.0

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
psantos1113: Couldn't we use some of the new methods computeIfAbsent here?

alert-common/src/main/java/com/synopsys/integration/alert/common/SetMap.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class SetMap<K, S> {
+    private final Map<K, Set<S>> map;
+
+    public SetMap(final Map<K, Set<S>> map) {
+        this.map = map;
+    }
```
bamandel: Can we also create a default constructor that doesn't require a map to be passed and just starts empty?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
```diff
@@ -95,16 +95,14 @@ public DefaultConfigurationAccessor(final RegisteredDescriptorRepository registe
...
-            if (!jobMap.containsKey(entityJobId)) {
-                jobMap.put(entityJobId, new HashSet<>());
-            }
-            jobMap.get(entityJobId).add(entity);
+            jobMap.add(entityJobId, entity);
         }
 
-        return jobMap.entrySet()
+        return jobMap.getMap()
+                   .entrySet()
```
bamandel: We call .getMap().entrySet() multiple times. We should add a helper method to SetMap that gets this.

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -187,18 +188,12 @@ public String createOrUpdateIssues(final FieldAccessor fieldAccessor, final Mess
...
-                    newIssueSet.add(e.getIssueKey());
-                    missingTransitionToIssues.put(e.getTransition(), newIssueSet);
-                }
+                missingTransitionToIssues.add(e.getTransition(), e.getIssueKey());
             }
         }
         if (!missingTransitionToIssues.isEmpty()) {
             final StringBuilder missingTransitions = new StringBuilder();
-            for (Map.Entry<String, Set<String>> entry : missingTransitionToIssues.entrySet()) {
+            for (Map.Entry<String, Set<String>> entry : missingTransitionToIssues.getMap().entrySet()) {
```
bamandel: We can use the new entrySet helper method here

alert-common/src/main/java/com/synopsys/integration/alert/common/SetMap.java
psantos1113: Why not make this class `extend Map<K,Set<S>>`? This way it conforms to a Map.

alert-common/src/main/java/com/synopsys/integration/alert/common/SetMap.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public Set<S> getValue(K key) {
+        return map.get(key);
+    }
+
+    public Set<S> add(K key, S value) {
+        if (map.containsKey(key)) {
```
psantos1113: Use computeIfAbsent here and then you can just call set.add after the call to computeIfAbsent.  It would collapse the code to a couple of lines.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
psantos1113: If SetMap was a Map you would just call .entrySet().

bamandel:
> You could have SetMap extend Map

It's a decent idea, but there are a lot of fields to configure if you go down that path. I feel like it'd simply be easier to just have a wrapper like what's already here.

alert-common/src/main/java/com/synopsys/integration/alert/common/SetMap.java
```diff
@@ -0,0 +1,54 @@
+package com.synopsys.integration.alert.common;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class SetMap<K, S> {
```
psantos1113: Would calling this call MapOfSets be more descriptive?

gkillough:
> > You could have SetMap extend Map
> 
> It's a decent idea, but there are a lot of fields to configure if you go down that path. I feel like it'd simply be easier to just have a wrapper like what's already here.

You could have it extend HashMap. That way the methods are already implemented.

PR #563, opened 2019-08-23
Ps custom message

psantos1113:
Changes to make the Test endpoint send custom messages.


src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -141,6 +145,18 @@ class DistributionConfiguration extends Component {
...
+        this.setState({
+            showSendMessage: true
+        });
+    }
+
+    hideSendMessage() {
+        this.setState({
+            showSendMessage: false
+        });
+    }
```
bamandel: We can probably pass in the boolean as a parameter. When calling the fuction as a prop, we should be able to pass true or false.

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -182,7 +198,7 @@ class DistributionConfiguration extends Component {
         const displayTest = !currentChannel.readOnly && DescriptorUtilities.isOperationAssigned(currentChannel, OPERATIONS.EXECUTE);
         const displaySave = !currentChannel.readOnly && DescriptorUtilities.isOneOperationAssigned(currentChannel, [OPERATIONS.CREATE, OPERATIONS.WRITE]);
         const isReadOnly = currentChannel.readOnly;
-
+        const channelDescriptorName = channelConfig ? channelConfig.descriptorName : null;
```
bamandel: I believe this is equal to `const channelDescriptorName = channelConfig && channelConfig.descriptorName`

src/main/js/dynamic/DistributionConfiguration.js
psantos1113: Not a bad idea.

src/main/js/dynamic/DistributionConfiguration.js
psantos1113: I will update it that should work.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -220,47 +219,18 @@ public String testJob(final JobFieldModel jobFieldModel, final String destinatio
             final Optional<TestAction> testActionOptional = descriptorProcessor.retrieveTestAction(channelFieldModel);
             if (testActionOptional.isPresent()) {
                 final Map<String, ConfigurationFieldModel> fields = createFieldsMap(channelFieldModel, otherJobModels);
-
+                // TODO the custom message fields are not written to the database or defined fields in the database.  Need to manually add them.
```
gkillough: I think the TODO should be on the second comment, that way the first one does not get deleted when the TODO is resolved. Was the intention to resolve this TODO in the PR?

psantos1113: I will move the TODO to the second line.  The intention is not to solve it in this PR but in a future enhancement as tech debt.

PR #578, opened 2019-08-30
Change DistributionChannel's sendMessage method to return an object

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/actions/JiraDistributionTestAction.java
```diff
@@ -83,7 +84,7 @@ public String testConfig(String jobId, String destination, FieldAccessor fieldAc
             toStatus = "Resolve";
             logger.debug("Sending additional DELETE test message...");
             final DistributionEvent reResolveIssueEvent = createChannelTestEvent(jobId, fieldAccessor, ItemOperation.DELETE, messageId);
-            String reResolveResult = getDistributionChannel().sendMessage(reResolveIssueEvent);
+            MessageResult reResolveResult = getDistributionChannel().sendMessage(reResolveIssueEvent);
```
jamesrichard91: Why do we only care about the initial and the last results?

gkillough: The initial results let us know that the issue was created. The last results let us know that we made it all the way to the end without erroring out.

PR #579, opened 2019-08-30
Updating the AlertFieldException handling

jamesrichard91:


src/main/java/com/synopsys/integration/alert/channel/slack/SlackChannel.java
```diff
@@ -87,8 +86,23 @@ public String sendMessage(final DistributionEvent event) throws IntegrationExcep
...
+            Map<String, String> fieldErrors = new HashMap();
+            if (webhookOptional.isEmpty()) {
+                fieldErrors.put(SlackDescriptor.KEY_WEBHOOK, "Missing Webhook URL");
+            }
+            if (channelNameOptional.isEmpty()) {
+                fieldErrors.put(SlackDescriptor.KEY_CHANNEL_NAME, "Missing channel name");
+            }
+            throw new AlertFieldException(fieldErrors);
+        }
+        String webhook = webhookOptional.get();
```
gkillough: I really prefer using `.isPresent()` with `.get()` that way it is nested in the code-block.

jamesrichard91: So you would prefer something like this?
```
String webhook = "";
        String channelName = "";

        if (webhookOptional.isPresent() && channelNameOptional.isPresent()) {
            webhook = webhookOptional.get();
            channelName = channelNameOptional.get();
        } else {
            Map<String, String> fieldErrors = new HashMap();
            if (webhookOptional.isEmpty()) {
                fieldErrors.put(SlackDescriptor.KEY_WEBHOOK, "Missing Webhook URL");
            }
            if (channelNameOptional.isEmpty()) {
                fieldErrors.put(SlackDescriptor.KEY_CHANNEL_NAME, "Missing channel name");
            }
        }
```

gkillough: I would prefer all of the requiring the optional fields to be inside of the if block.

src/main/java/com/synopsys/integration/alert/channel/slack/SlackChannel.java
```diff
@@ -87,8 +86,23 @@ public String sendMessage(final DistributionEvent event) throws IntegrationExcep
 
     public List<Request> createRequests(final DistributionEvent event) throws IntegrationException {
         final FieldAccessor fields = event.getFieldAccessor();
-        final String webhook = fields.getString(SlackDescriptor.KEY_WEBHOOK).orElseThrow(() -> new AlertException("Missing Webhook URL"));
-        final String channelName = fields.getString(SlackDescriptor.KEY_CHANNEL_NAME).orElseThrow(() -> new AlertException("Missing channel name"));
+
+        final Optional<String> webhookOptional = fields.getString(SlackDescriptor.KEY_WEBHOOK);
+        final Optional<String> channelNameOptional = fields.getString(SlackDescriptor.KEY_CHANNEL_NAME);
```
bamandel: I think these methods should return orElse(""). We then check if the string is blank or not and throw an exception. Since we are in the Channel class, we expect certain values to always be given to us. These two fields must have a value for our code to properly function otherwise we fail.

PR #584, opened 2019-09-03
Get cvss3 severity from Black Duck if it exists

gkillough:
I also added this in a previous commit on master. Before, it was the same without the if-statement.
```java
    public static final ComponentItemPriority findPriority(String priority) {
        String upperCasePriority = priority.toUpperCase();
        if ("CRITICAL".equals(upperCasePriority) || "BLOCKER".equals(upperCasePriority)) {
            return HIGHEST;
        }

        try {
            return ComponentItemPriority.valueOf(upperCasePriority);
        } catch (IllegalArgumentException ex) {
            // couldn't find the enum value default to STANDARD
            return NONE;
        }
    }
```

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckVulnerabilityCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckVulnerabilityCollector.java
@@ -36,6 +36,8 @@ import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 import com.jayway.jsonpath.TypeRef;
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckVulnerabilityCollector.java
```diff
@@ -176,4 +182,19 @@ private LinkableItem getSeverity(String vulnerabilityUrl) {
         return severityItem;
     }
 
+    private Optional<String> getCvss3Severity(VulnerabilityView vulnerabilityView) {
+        Boolean useCvss3 = vulnerabilityView.getUseCvss3();
+        if (null != useCvss3 && useCvss3) {
+            JsonObject vulnJsonObject = vulnerabilityView.getJsonElement().getAsJsonObject();
```
jamesrichard91: This looks fine to me. Could we add a TODO to replace this when blackduck-common-api is updated to expose the cvss3 and its severity as Java objects?

PR #587, opened 2019-09-05
adding AutoActionable to automatically instantiate Action classes that require no custom behavior

ekerwin:


ekerwin:
I would like to discuss this code before anyone considers merging it.

src/main/java/com/synopsys/integration/alert/channel/msteams/SimpleChannelActions.java
```diff
@@ -0,0 +1,14 @@
+package com.synopsys.integration.alert.channel.msteams;
+
+import com.synopsys.integration.alert.common.channel.ChannelKey;
+import com.synopsys.integration.alert.common.channel.DistributionChannel;
+
+/**
+ * This lets the DescriptorProcessor create the appropriate ChannelDistributionTestAction/ConfigurationAction combination when no customization is needed.
+ */
+public interface SimpleChannelActions {
```
gkillough: I find this name misleading. The actions are not methods of this interface, this interface marks the class as something that has "simple" actions. Something along the lines of `ActionableSimpleChannel` might get the message across better, though I'm not in love with that name either.

jamesrichard91: This name is confusing to me especially since you have the SlackChannel extending this interface.

PR #589, opened 2019-09-05
Refactor descriptors to use DescriptorKey rather than constant

gkillough:


src/main/java/com/synopsys/integration/alert/ApplicationConfiguration.java
```diff
@@ -73,13 +75,24 @@ public EncryptionUtility encryptionUtility() {
     }
 
     @Bean
-    public ProxyManager proxyManager(final ConfigurationAccessor configurationAccessor) {
-        return new ProxyManager(configurationAccessor);
+    public SettingsDescriptorKey settingsDescriptorKey() {
```
jamesrichard91: Is this to move away from declaring @Component on the classes?

src/main/java/com/synopsys/integration/alert/channel/email/EmailChannelKey.java
```diff
@@ -27,7 +27,7 @@
 import com.synopsys.integration.alert.common.channel.ChannelKey;
 
 @Component
-public class EmailChannelKey extends ChannelKey {
+public final class EmailChannelKey extends ChannelKey {
     private static final String COMPONENT_NAME = "channel_email";
```
jamesrichard91: This String should be deleted based on your changes

src/main/java/com/synopsys/integration/alert/channel/jira/JiraChannelKey.java
```diff
@@ -27,7 +27,7 @@
 import com.synopsys.integration.alert.common.channel.ChannelKey;
 
 @Component
-public class JiraChannelKey extends ChannelKey {
+public final class JiraChannelKey extends ChannelKey {
     private static final String COMPONENT_NAME = "channel_jira_cloud";
```
jamesrichard91: This String should be deleted based on your changes

PR #593, opened 2019-09-06
Show missing items in table select fields

bamandel:


src/main/js/field/input/TableSelectInput.js
```diff
@@ -10,32 +10,44 @@ import { createNewConfigurationRequest } from 'util/configurationRequestBuilder'
...
 
 const container = ({ children, getValue, ...props }) => {
     const length = getValue().length;
-    return (length <= 5) ?
-        <ValueContainer {...props}>
-            {children}
-        </ValueContainer>
-        :
+    const error = (
+        <span className="missingBlackDuckData">
```
jamesrichard91: We should rename the CSS class missingBlackDuckData. This field can apply to more than just black duck data

PR #594, opened 2019-09-09
Ps jira size limit

psantos1113:
Separate the description into comments.  Take the operation comment and split it into multiple comments.


src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssueFormatHelper.java
```diff
@@ -59,47 +66,89 @@ public String createTitle(final String provider, final LinkableItem topic, final
         // FIXME make this provider-agnostic
         String prettyPrintedKey;
         if (!componentKeys.getCategory().contains("Vuln")) {
-            prettyPrintedKey = componentKeys.prettyPrint(true);
+            prettyPrintedKey = componentKeys.prettyPrint(false);
```
psantos1113: We would add all the vulnerabilities which caused the title to be greater than 255 characters.  Is there any information other than project, project version, component, component version, and category that we need in the title of the ticket?

psantos1113: Otherwise I can collapse this if else.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/MessageSplitter.java
```diff
@@ -0,0 +1,135 @@
...
+
+        final List<String> messages = new ArrayList<>();
+        messages.add(preSplit);
+        messages.addAll(splitMessage(postSplit));
+
+        return messages;
+    }
+
+    private int getSplitIndex(final String message) {
+        final char bracket = '[';
```
gkillough: Should this be an attribute of the class?

src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssueFormatHelper.java
```diff
@@ -31,13 +32,19 @@
...
+import com.synopsys.integration.alert.common.channel.MessageSplitter;
+import com.synopsys.integration.alert.common.enumeration.ItemOperation;
 import com.synopsys.integration.alert.common.exception.AlertRuntimeException;
 import com.synopsys.integration.alert.common.message.model.ComponentItem;
 import com.synopsys.integration.alert.common.message.model.ComponentKeys;
 import com.synopsys.integration.alert.common.message.model.LinkableItem;
 
 public class JiraIssueFormatHelper {
     private static final int TITLE_LIMIT = 255;
+    private static final int TEXT_LIMIT = 32767;
```
gkillough: Let's make this smaller than the actual number to account for edge cases. Maybe 25,000 or 30,000.

src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssueFormatHelper.java
gkillough: Policy Name

Also, your current code will set it to `false` every time. Maybe it is a problem with the collector? Do the vulnerability `LinkableItems` for policy tickets get flagged as `partOfKey`?

src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssueFormatHelper.java
jamesrichard91: I would agree with 30,000

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/MessageSplitter.java
```diff
@@ -0,0 +1,135 @@
...
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.channel;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+//TODO use this for SlackChannel to create message chunks.
+public final class MessageSplitter {
```
bamandel: Do we want this as a component for DependencyInjection? We could pass the limit and lineSeparator when they're needed via the method being used. This seems kind of random to not be used as a component.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/MessageSplitter.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public final String getLineSeparator() {
+        return lineSeparator;
+    }
+
+    public final List<String> splitMessages(final Collection<String> messagePieces) {
+        return splitMessages(messagePieces, false);
+    }
+
+    public final List<String> splitMessages(final Collection<String> messagePieces, boolean separateChunksByLineSeparator) {
```
bamandel: Do you think there's a library that might already do a lot of these things? If so, it might be easier to just use the library instead of creating our own chunker.

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -294,33 +302,22 @@ private void addIssueProperties(String issueKey, String provider, LinkableItem t
...
+    private IssueContentModel createContentModel(ComponentItem arbitraryItem, Collection<ComponentItem> componentItems, LinkableItem commonTopic, Optional<LinkableItem> subTopic, String provider) {
         final JiraIssueFormatHelper jiraChannelFormatHelper = new JiraIssueFormatHelper();
-        final IssueRequestModelFieldsBuilder fieldsBuilder = new IssueRequestModelFieldsBuilder();
         final String title = jiraChannelFormatHelper.createTitle(provider, commonTopic, subTopic, arbitraryItem.getComponentKeys());
-        final String description = jiraChannelFormatHelper.createDescription(commonTopic, subTopic, componentItems, provider);
-        fieldsBuilder.setSummary(title);
-        fieldsBuilder.setDescription(description);
+        final IssueContentModel descriptionModel = jiraChannelFormatHelper.createDescription(commonTopic, subTopic, componentItems, provider);
+        return IssueContentModel.of(title, descriptionModel.getDescription(), descriptionModel.getAdditionalComments());
```
bamandel: This seems slightly redundant with the line right above it. Could we have the createDescription just return the proper model with the title added?

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/MessageSplitter.java
psantos1113: There might be but this is a bug fix.  I want to use the code we have than introduce a new dependency at this point.   Maybe for the next release.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/MessageSplitter.java
psantos1113: We can evaluate that after the release.  This is implemented to initially fix the issue.

PR #597, opened 2019-09-09
Bm clean custom message

bamandel:
I was going through TODOs and saw some code that could be cleaned up. The custom message functionality has been slightly modified and I changed the message content field from a textField to a textArea.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
@@ -208,7 +208,6 @@ public class JobConfigActions {
         return "Valid";
     }
 
-    // TODO abstract duplicate functionality
     public String testJob(final JobFieldModel jobFieldModel, final String destination) throws IntegrationException {
         validateJob(jobFieldModel);
         final Collection<FieldModel> otherJobModels = new LinkedList<>();
--- a/src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
@@ -61,6 +61,7 @@ import com.synopsys.integration.exception.IntegrationException;
 
 @Component
 public class JobConfigActions {
+    public static final String CUSTOM_MESSAGE = "-CUSTOM_MESSAGE";
     private static final Logger logger = LoggerFactory.getLogger(JobConfigActions.class);
     private final ConfigurationAccessor configurationAccessor;
     private final FieldModelProcessor fieldModelProcessor;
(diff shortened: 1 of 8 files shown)


src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -61,6 +61,7 @@
 
 @Component
 public class JobConfigActions {
+    public static final String CUSTOM_MESSAGE = "-CUSTOM_MESSAGE";
```
gkillough: Why the dash? The name of this variable really doesn't give me any insight into how it's used. Maybe go with `CUSTOM_MESSAGE_SUFFIX`.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
```diff
@@ -288,7 +287,15 @@ public JobConfigController(final JobConfigActions jobConfigActions, final Respon
     private boolean hasRequiredPermissions(Collection<FieldModel> fieldModels, BiFunction<String, String, Boolean> permissionChecker) {
         return fieldModels
                    .stream()
-                   .allMatch(model -> permissionChecker.apply(model.getContext(), model.getDescriptorName()));
+                   .allMatch(model -> permissionChecker.apply(model.getContext(), trimCustomMessageName(model.getDescriptorName())));
+    }
+
+    private String trimCustomMessageName(String descriptorName) {
+        final int index = descriptorName.indexOf(JobConfigActions.CUSTOM_MESSAGE);
```
gkillough: What is with this?

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -208,63 +209,59 @@ public String validateJob(final JobFieldModel jobFieldModel) throws AlertFieldEx
...
-            final FieldModel updatedFieldModel = fieldModelProcessor.createCustomMessageFieldModel(fieldModel);
-            if (descriptor.filter(foundDescriptor -> DescriptorType.CHANNEL.equals(foundDescriptor.getType())).isPresent()) {
-                channelFieldModel = updatedFieldModel;
-            } else {
-                otherJobModels.add(updatedFieldModel);
-            }
+    private Optional<FieldModel> getCustomMessageFieldModelAndRemoveIt(JobFieldModel jobFieldModel) {
+        Optional<FieldModel> customFieldModel = jobFieldModel.getFieldModels()
+                                                    .stream()
+                                                    .filter(fieldModel -> fieldModel.getDescriptorName().endsWith(CUSTOM_MESSAGE))
```
gkillough: Why are we using the descriptor name here? Didn't we just explicitly remove all assumptions about the contents of descriptor names with the ChannelKey stuff?

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
jamesrichard91: Why are we trimming the descriptor name? When would he descriptor name ever include `-CUSTOM_MESSAGE` as a suffix?

src/main/js/dynamic/JobCustomMessageModal.js
```diff
@@ -29,10 +30,8 @@ class JobCustomMessageModal extends Component {
         this.handleHide = this.handleHide.bind(this);
     }
 
-    createCustomMessageFieldModel(jobFieldModel) {
-        const fieldModel = jobFieldModel.fieldModels.find(model => model.descriptorName === this.props.channelDescriptorName);
-        let newModel = FieldModelUtilities.createEmptyFieldModel([TOPIC_ID, MESSAGE_ID], CONTEXT_TYPE.DISTRIBUTION, this.props.channelDescriptorName);
-        newModel = FieldModelUtilities.combineFieldModels(newModel, fieldModel);
+    createCustomMessageFieldModel() {
+        let newModel = FieldModelUtilities.createEmptyFieldModel([TOPIC_ID, MESSAGE_ID], CONTEXT_TYPE.DISTRIBUTION, `${this.props.channelDescriptorName}-CUSTOM_MESSAGE`);
```
jamesrichard91: Why would we append -CUSTOM_MESSAGE here?

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
jamesrichard91: It looks like all of this is designed to send custom messages to channels. Each job can only have one channel, so why don't we just look for the channel section of the job model instead appending a string and then stripping it out again later?

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
psantos1113: The UI is adding a third descriptor in the Job Configuration as payload.

src/main/js/dynamic/JobCustomMessageModal.js
psantos1113: we're adding a new field model to the JobModel.  A JobModel is a collection of field models for each descriptor.

bamandel:
I put this together as an alternative to see what people thought. It seems the general consensus is that it's too confusing using the suffix for a custom message FieldModel.

PR #600, opened 2019-09-10
Allow PROJECT and PROJECT_VERSION notifications to resolve Jira Cloud issues

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -106,59 +105,38 @@ public JiraMessageResult createOrUpdateIssues(JiraIssueConfig jiraIssueConfig, M
...
-                final Optional<IssueComponent> existingIssueComponent = retrieveExistingIssue(jiraProjectKey, providerName, topic, subTopic, arbitraryItem, trackingKey);
+                ComponentItem arbitraryItem = combinedItems
+                                                  .stream()
+                                                  .findAny()
+                                                  .orElseThrow(() -> new AlertException("Unable to successfully combine component items for Jira Cloud issue handling."));
+                ItemOperation operation = arbitraryItem.getOperation();
+                String trackingKey = createAdditionalTrackingKey(arbitraryItem);
+
+                List<IssueComponent> existingIssues;
+                if (BlackDuckProjectVersionCollector.CATEGORY_TYPE.equals(arbitraryItem.getCategory())) {
```
bamandel: Can we add a FIXME here to remind us to remove this Provider check?

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -106,59 +105,38 @@ public JiraMessageResult createOrUpdateIssues(JiraIssueConfig jiraIssueConfig, M
...
+                    existingIssues = retrieveExistingIssues(jiraIssueConfig.getProjectComponent().getKey(), providerName, topic, subTopic, null, null);
+                } else {
+                    existingIssues = retrieveExistingIssues(jiraIssueConfig.getProjectComponent().getKey(), providerName, topic, subTopic, arbitraryItem, trackingKey);
```
bamandel: An alternative is to set arbitratyItem and trackingKey to null if you the check passes and just call the retrieveExistingIssues method a single time.

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -192,12 +169,64 @@ public JiraMessageResult createOrUpdateIssues(JiraIssueConfig jiraIssueConfig, M
...
+        String issueCreator = jiraIssueConfig.getIssueCreator();
+        IssueResponseModel issue = null;
+        try {
+            issue = issueService.createIssue(new IssueCreationRequestModel(issueCreator, jiraIssueConfig.getIssueType(), jiraIssueConfig.getProjectComponent().getName(), initialFieldsBuilder, List.of()));
+            logger.debug("Created new Jira Cloud issue: {}", issue.getKey());
+        } catch (IntegrationRestException e) {
+            handleIssueCreationRestException(e, issueCreator);
+        }
+        if (issue == null || StringUtils.isBlank(issue.getKey())) {
+            throw new AlertException("There was an problem when creating this issue.");
```
bamandel: Is there a case where an issue can be created without a key? If not, we probably just want to throw this exception in the catch statement above.

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
```diff
@@ -192,12 +169,64 @@ public JiraMessageResult createOrUpdateIssues(JiraIssueConfig jiraIssueConfig, M
...
+                   .findIssues(jiraProjectKey, provider, topic, subTopic.orElse(null), componentItem, alertIssueUniqueId)
                    .map(IssueSearchResponseModel::getIssues)
-                   .map(List::stream)
-                   .flatMap(Stream::findFirst);
+                   .orElse(List.of());
+    }
+
+    // Only the DELETE operation can update more than one issue.
+    // If an ADD or UPDATE operation has more than one issue it could operate on, then the issue properties are either not unique enough or the property indexer is not installed.
+    private List<IssueComponent> filterUpdatableIssues(List<IssueComponent> issues, ItemOperation operation) {
```
bamandel: Most of this method can be deleted and you can instead modify the list right before line 129 then pass the proper list through the methods instead of one that will just be modified later.

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
gkillough: This code was moved from a different method.

src/main/java/com/synopsys/integration/alert/channel/jira/JiraIssueHandler.java
gkillough: I think the abstraction makes it more readable. This would have been > 4 levels of nesting.

PR #603, opened 2019-09-11
Fix(BomEdit): If there's policy data create a component item

psantos1113:
Create component items for policy data.  We were only creating data for policies components that have vulnerabilities.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
@@ -34,6 +34,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/BlackDuckCollector.java
```diff
@@ -236,4 +246,11 @@ protected LinkableItem getSeverity(String vulnerabilityUrl, boolean partOfKey) {
         return vulnerabilityViewMap;
     }
 
+    protected ComponentItemPriority getPolicyPriority(String severity) {
+        if (StringUtils.isBlank(severity) || !policyPriorityMap.containsKey(severity.trim().toLowerCase())) {
```
gkillough: We could check if it is not blank and then store the result of `.trim()` and `.toLowerCase()` before getting from the map. Then we could make `return ComponentItemPriority.NONE` the last line of the method outside of the if-statement.

PR #613, opened 2019-09-13
Ps encryption read fix

psantos1113:
Try to read the new salt secret first otherwise attempt to read the old secret file name.

Added unit tests to the encryption utility class.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/security/EncryptionUtility.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/security/EncryptionUtility.java
@@ -22,6 +22,7 @@
  */
 package com.synopsys.integration.alert.common.security;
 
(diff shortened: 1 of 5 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/security/EncryptionUtility.java
```diff
@@ -161,9 +164,21 @@ private String getGlobalSaltFromFile() {
     private Optional<String> readGlobalSaltFromSecretsFile() {
         try {
             return Optional.ofNullable(filePersistenceUtil.readFromSecretsFile(SECRETS_ENCRYPTION_SALT));
+        } catch (FileNotFoundException ex) {
+            // ignore so we can attempt the old file.
```
gkillough: Can we put a TODO in here for removing this in the next major version?

PR #619, opened 2019-09-19
Ps saml fileupload

psantos1113:
FilePersistenceUtil I would like to refactor in the future because we use it to manage files in our data volume.  I would like to refactor the API to make it clearer what you can and cannot do.  A bit beyond the scope of this PR. 

The changes here work similarly to the custom endpoint manager where you register your upload target file.  The UI config defines the mime types that are accepted for the UI file picker to only allow XML files to be selected. The code also validates that the file uploaded is valid XML.


bamandel:
It looks like the only custom functionality we're passing to the upload endpoint is a validation function. If this is true, then we may be able to get away with not needing the UploadEndpointManager class (We'll still need an endpoint for uploading files). ConfigFields already have validationFunctions associated with them and are called automatically. We might be able to create a "basic" ConfigField for file upload and simply pass the validation there. We would move all the file upload functionality in UploadEndpointManager to the controller or create a new UploadActions class which is called by the controller.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
```diff
@@ -0,0 +1,169 @@
...
+            filePersistenceUtil.writeFileToUploadsDirectory(fileName, inputStream);
+        } catch (IOException ex) {
+            logger.error("Error uploading file - file: {}, context: {}, descriptor: {} ", fileName);
+            logger.error("Caused by: ", ex);
+            throw ex;
+        }
+    }
+
+    private class UploadTarget {
+
```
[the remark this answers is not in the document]
bamandel: I think this means there's a space there

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/FilePersistenceUtil.java
```diff
@@ -49,17 +54,35 @@ public FilePersistenceUtil(final AlertProperties alertProperties, final Gson gso
...
+        writeToFile(createFile(fileName), content.getBytes(StandardCharsets.UTF_8));
+    }
+
+    public void writeToFile(File destination, byte[] data) throws IOException {
+        // destination is a path to a file.  Make sure all parent directories exist before writing.
+        destination.getParentFile().mkdirs();
+        Files.write(destination.toPath(), data, StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.TRUNCATE_EXISTING);
+    }
+
+    public void writeToFile(File destination, InputStream inputStream) throws IOException {
```
bamandel: Should this method technically be called copy as we're copying a file?

src/main/java/com/synopsys/integration/alert/component/settings/descriptor/SettingsUIConfig.java
```diff
@@ -95,6 +100,9 @@
     private static final String SETTINGS_SAML_METADATA_URL_DESCRIPTION = "The Metadata URL from the external Identity Provider.";
     private static final String SETTINGS_SAML_ENTITY_ID_DESCRIPTION = "The Entity ID of the Service Provider. EX: This should be the Audience defined in Okta.";
     private static final String SETTINGS_SAML_ENTITY_BASE_URL_DESCRIPTION = "This should be the URL of the Alert system.";
+    private static final String SETTINGS_SAML_METADATA_FILE_DESCRIPTION = "The file to upload to the server containing the Metadata from the external Identity Provider.";
+
+    private static final String BUTTON_LABEL_SAML_METADATA_FILE = "Upload Metadata File";
```
bamandel: Since this seems to be the button text, should we simplify this to simply upload? The label should do the job of describing

src/main/java/com/synopsys/integration/alert/component/settings/descriptor/SettingsUIConfig.java
```diff
@@ -197,4 +211,31 @@ public SettingsUIConfig() {
...
+        }
+        return List.of();
+    }
+
+    private Collection<String> validateMetaDataFile(FieldValueModel fieldToValidate, FieldModel fieldModel) {
+        Optional<FieldValueModel> samlEnabledField = fieldModel.getFieldValueModel(SettingsDescriptor.KEY_SAML_ENABLED);
+        boolean samlEnabled = samlEnabledField.flatMap(field -> field.getValue())
+                                  .map(value -> Boolean.valueOf(value))
+                                  .orElse(false);
+        if (samlEnabled) {
```
bamandel: What will happen if someone uploads a file without enabling SAML? Will a file upload but we simply do nothing with it?

src/main/js/field/UploadFileButtonField.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+UploadFileButtonField.propTypes = {
+    endpoint: PropTypes.string.isRequired,
+    buttonLabel: PropTypes.string.isRequired,
+    currentConfig: PropTypes.object.isRequired,
+    fieldKey: PropTypes.string.isRequired,
+    csrfToken: PropTypes.string.isRequired,
+    onChange: PropTypes.func.isRequired,
+    fields: PropTypes.array,
```
bamandel: I see fields listed here but not used. You can probably remove it.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/UploadFileButtonField.java
```diff
@@ -0,0 +1,102 @@
...
+
+import com.synopsys.integration.alert.common.action.UploadEndpointManager;
+import com.synopsys.integration.alert.common.descriptor.config.field.ConfigField;
+import com.synopsys.integration.alert.common.descriptor.config.field.ConfigValidationFunction;
+import com.synopsys.integration.alert.common.enumeration.FieldType;
+
+public class UploadFileButtonField extends ConfigField {
+    private final String buttonLabel;
+    private final String endpoint;
+    private final List<ConfigField> subFields;
```
bamandel: After looking at the JS, it doesn't look like the subFields are actually being used. You can probably remove this field.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/FilePersistenceUtil.java
jamesrichard91: The inputStream that we are "copying" into the destination file does not have to be a file that exists. It could be a stream in memory. We are just putting the data from the stream into the destination file. I think the method is named appropriately.

psantos1113: James is correct here.  Files calls the method copy because you are copying the contents of the input stream, whether it exists or not.

PR #621, opened 2019-09-20
Added a couple tests of previously uncovered code

bamandel:
also removed some unnecessary code.


alert-common/src/main/java/com/synopsys/integration/alert/common/annotation/SensitiveFieldFinder.java
```diff
@@ -35,16 +32,6 @@ private SensitiveFieldFinder() {
         throw new IllegalStateException("Utility class");
     }
 
-    public static Set<Field> findSensitiveFields(final Class<?> clazz) {
```
gkillough: If this method was the primary reason this class existed, why keep the class around?

alert-common/src/main/java/com/synopsys/integration/alert/common/annotation/SensitiveFieldFinder.java
```diff
@@ -35,16 +32,6 @@ private SensitiveFieldFinder() {
...
-        final Set<Field> fields = new HashSet<>();
-        for (final Field field : clazz.getDeclaredFields()) {
-            if (field.isAnnotationPresent(SensitiveField.class) || hasParentSensitiveAnnotation(field.getAnnotations())) {
-                fields.add(field);
-            }
-        }
-        return fields;
-    }
-
     public static boolean hasParentSensitiveAnnotation(final Annotation[] annotations) {
```
jamesrichard91: If this method is still used, does it still make sense to be in this class? or do we have other classes that deal with fields that this could move into?

bamandel: IMO this class should be removed. It is used on a single field to prevent its toString from printing a secure value. We should instead modify that specific objects toString to ignore that field and then remove all classes related to this sensitive field annotation.

PR #624, opened 2019-09-23
Make messages less verbose

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/util/JiraIssueHandler.java
```diff
@@ -208,10 +208,10 @@ private IssueResponseModel createIssue(IssueRequestModelFieldsBuilder initialFie
                    .orElse(List.of());
     }
 
-    // Only the DELETE operation can update more than one issue.
+    // Only the DELETE or INFO operations can update more than one issue.
     // If an ADD or UPDATE operation has more than one issue it could operate on, then the issue properties are either not unique enough or the property indexer is not installed.
     private List<IssueComponent> filterUpdatableIssues(List<IssueComponent> issues, ItemOperation operation) {
-        if (issues.size() == 1 || ItemOperation.DELETE.equals(operation)) {
+        if (issues.size() == 1 || ItemOperation.INFO.equals(operation) || ItemOperation.DELETE.equals(operation)) {
```
psantos1113: You should be able to use == operator instead of.equals if you think that will improve readability.

PR #625, opened 2019-09-24
Bm resolve todos

bamandel:


src/main/java/com/synopsys/integration/alert/workflow/processor/MessageContentAggregator.java
```diff
@@ -146,7 +146,6 @@ public MessageContentAggregator(final ConfigurationAccessor jobConfigReader, fin
     private Map<String, MessageContentCollector> createCollectorMap(final Set<MessageContentCollector> providerMessageContentCollectors) {
         final Map<String, MessageContentCollector> collectorMap = new HashMap<>();
 
-        //TODO need better performance to map the notification type to the processor
```
jamesrichard91: If we haven't improved the performance, why are we removing the TODO?

bamandel: For all of the TODOs I've removed, they're either represented as a ticket in Jira, or no longer necessary. This one is one that is now in Jira.

PR #626, opened 2019-09-24
Fix(Java): Remove necessity for provider specific code in channels

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/ChannelMessageParser.java
```diff
@@ -133,6 +143,9 @@ protected String createLinkableItemString(LinkableItem linkableItem, boolean bol
     }
 
     protected List<String> createComponentAttributeMessagePieces(Collection<ComponentItem> componentItems) {
+        if (componentItems.isEmpty()) {
```
bamandel: It seems we're mainly looping through this collection which makes this line seem unnecessary. We could remove this line and add an if statement at line 174 to check if attributeStrings is empty then perform the add if it is not.

PR #628, opened 2019-09-26
Ps upload delete

psantos1113:
Added permissions specific to upload because I did not want to add the delete permission for the admin role for the settings component.  The reason is because I do not want someone to be able to delete the settings configuration via the rest API.  So I added new AccessOperations for file upload to separate the functionality since it changes the file system on the alert server.


src/main/resources/db/changelog/alert/5.1.0/changelog-authentication-stored-procedures.xml
```diff
@@ -62,7 +62,7 @@
 
                         statement.execute("CALL ASSIGN_ROLE_PERMISSION_ACCESS('ALERT_JOB_MANAGER', '" + descriptorName + "', 'global', 'READ');");
                         statement.execute("CALL ASSIGN_ROLE_PERMISSION_ACCESS('ALERT_JOB_MANAGER', '" + descriptorName + "', 'global', 'EXECUTE');");
-                        statement.execute("CALL ASSIGN_ROLE_PERMISSION_ALL_ACCESS('ALERT_JOB_MANAGER', '" + descriptorName + "', 'distribution');");
+                        statement.execute("CALL ASSIGN_ROLE_PERMISSION_ALL_NO_UPLOAD_ACCESS('ALERT_JOB_MANAGER', '" + descriptorName + "', 'distribution');");
 
                         statement.execute("CALL ASSIGN_ROLE_PERMISSION_ACCESS('ALERT_USER', '" + descriptorName + "', 'distribution', 'READ');");
```
jamesrichard91: Don't we want User's to have no upload access as well?

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/AccessOperation.java
```diff
@@ -27,5 +27,8 @@
     DELETE,
     READ,
     WRITE,
-    EXECUTE
+    EXECUTE,
+    UPLOAD_FILE_READ,
```
gkillough: Is the reason that these are different because of the way we currently have custom endpoints set up?

src/main/resources/db/changelog/alert/5.1.0/changelog-authentication-stored-procedures.xml
psantos1113: They only have read access on distribution jobs.

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/AccessOperation.java
psantos1113: I didn't want to allow delete because if we allow delete on the component_settings global context then through the rest API you can delete the global component_settings configuration which could potentially break the authentication into the UI and being able to read and write encrypted data.

psantos1113: Because of this I decided to make upload a special operation.

PR #629, opened 2019-09-26
Template and resources update

jamesrichard91:
* Deprecated the template directory property, since we dont support custom user provided templates.

* Moved the templates and images to directories that should make more sense.

* Fixed an issue with pattern handling. Project names could have special characters without being a pattern and was causing an issue with the filtering.

* Updated the production code to no longer load any templates or images from src/main/resources


src/main/java/com/synopsys/integration/alert/component/settings/PasswordResetService.java
```diff
@@ -90,7 +90,7 @@ public void resetPassword(String username) throws AlertException {
 
     private void handleSendAndUpdateDatabase(EmailProperties emailProperties, Map<String, Object> templateFields, String emailAddress, String username, String tempPassword) throws AlertException {
         try {
-            final String imageDirectoryPath = alertProperties.getAlertImagesDirPath();
+            final String imageDirectoryPath = alertProperties.getAlertLogo();
```
gkillough: Should we rename the variable to match the method call?

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/UpdateEmailService.java
```diff
@@ -122,7 +122,7 @@ private boolean wasEmailAlreadySentForVersion(String updateVersion) {
 
     private void handleSendAndUpdateDatabase(EmailProperties emailProperties, Map<String, Object> templateFields, String emailAddress) throws AlertException {
         try {
-            final String imageDirectoryPath = alertProperties.getAlertImagesDirPath();
+            final String imageDirectoryPath = alertProperties.getAlertLogo();
```
gkillough: Variable name?

src/main/java/com/synopsys/integration/alert/workflow/filter/JsonFieldFilterBuilder.java
```diff
@@ -53,9 +54,19 @@ public JsonFieldFilterBuilder(final JsonExtractor jsonExtractor, final JsonField
...
+                    if (contentValue.equals(value.trim())) {
+                        logger.debug("Match: {}", contentValue);
+                        return true;
+                    }
+                    try {
+                        if (contentValue.matches(value)) {
+                            logger.debug("Match: {}", contentValue);
+                            return true;
+                        }
+                    } catch (PatternSyntaxException e) {
```
gkillough: Do we validate the pattern when testing distribution jobs?

jamesrichard91: The problem is, this could either be a Project name OR the Project name pattern from the job configuration. And the issue occurs when you have a project name that has special characters that are not a valid pattern.

gkillough: Can we put a TODO here when we move away from Json Extraction?

jamesrichard91: I am adding a comment and will create a ticket to address this when we move away from Json Extraction

PR #633, opened 2019-09-30
Removing code from the NotificationFilter that was only being used in Tests

jamesrichard91:


jamesrichard91:
Gavin's changes will remove the NotificationFilter class entirely.

PR #639, opened 2019-10-02
Ps policy violation

psantos1113:
There's a lot of code here in the policy violation that will need to be encapsulated better so it can be shared for the other policy violation types and Bom Edit.  This gets the notification processed in the new processor.


src/main/java/com/synopsys/integration/alert/provider/blackduck/new_collector/PolicyViolationMessageBuilder.java
```diff
@@ -0,0 +1,456 @@
...
+        ComponentService componentService = blackDuckServicesFactory.createComponentService();
+        RuleViolationNotificationContent violationContent = notificationView.getContent();
+        ItemOperation operation = ItemOperation.ADD;
+        try {
+            ProviderMessageContent.Builder projectVersionMessageBuilder = new ProviderMessageContent.Builder()
+                                                                              .applyProvider(getProviderName(), blackDuckServicesFactory.getBlackDuckHttpClient().getBaseUrl())
+                                                                              .applyTopic("Project", violationContent.getProjectName())
+                                                                              .applySubTopic("Project Version", violationContent.getProjectVersionName(), violationContent.getProjectVersion())
+                                                                              .applyAction(operation)
+                                                                              .applyNotificationId(notificationId)
```
gkillough: We don't need to apply the action and notification id at the top level if there are component items. It won't have any affect in this case, but if there's ever a case where no component items are present, channels will treat it like a PROJECT or PROJECT_VERSION notification.

src/main/java/com/synopsys/integration/alert/provider/blackduck/new_collector/PolicyViolationMessageBuilder.java
```diff
@@ -0,0 +1,456 @@
...
+        } catch (Exception ex) {
+            logger.info("Error building policy component for notification {}, operation {}, component {}, component version {}", notificationId, operation, componentItem, componentVersionItem);
+            logger.error("Error building policy component cause ", ex);
+            return Optional.empty();
+        }
+    }
+
+    protected LinkableItem createPolicyNameItem(PolicyInfo policyInfo) {
+        String policyName = policyInfo.getPolicyName();
+        return new LinkableItem(BlackDuckContent.LABEL_POLICY_NAME, policyName, null);
```
gkillough: Is there a 2-arg constructor?

src/main/java/com/synopsys/integration/alert/provider/blackduck/new_collector/PolicyViolationMessageBuilder.java
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    protected Optional<LinkableItem> createPolicySeverityItem(PolicyInfo policyInfo) {
+        String severity = policyInfo.getSeverity();
+        if (StringUtils.isNotBlank(severity)) {
+            final LinkableItem severityItem = new LinkableItem(BlackDuckContent.LABEL_POLICY_SEVERITY_NAME, severity, null);
```
gkillough: 2-arg constructor and ~final~

src/main/java/com/synopsys/integration/alert/provider/blackduck/new_collector/util/VulnerabilityUtil.java
```diff
@@ -0,0 +1,97 @@
...
+import com.synopsys.integration.exception.IntegrationException;
+
+@Component
+public class VulnerabilityUtil {
+
+    public List<VulnerabilityView> getVulnerabilitiesForComponent(BlackDuckService blackDuckService, VulnerableComponentView vulnerableComponentView) {
+        try {
+            return blackDuckService.getAllResponses(vulnerableComponentView, VulnerableComponentView.VULNERABILITIES_LINK_RESPONSE);
+        } catch (IntegrationException ex) {
+
```
gkillough: Can we add a comment here or log why we're ignoring this?

src/main/java/com/synopsys/integration/alert/provider/blackduck/new_collector/util/BlackDuckResponseCache.java
```diff
@@ -60,4 +65,61 @@ public BlackDuckResponseCache(final BlackDuckBucketService blackDuckBucketServic
         return Optional.empty();
     }
 
+    public Optional<String> getProjectComponentQueryLink(String projectVersionUrl, String link, String componentName) {
+        Optional<String> projectLink = getProjectLink(projectVersionUrl, link);
+        return projectLink.flatMap(optionalProjectLink -> getProjectComponentQueryLink(optionalProjectLink, componentName));
+    }
+
+    public Optional<String> getProjectComponentQueryLink(String projectLink, String componentName) {
```
gkillough: There doesn't seem to be a need for this to be Optional.

PR #645, opened 2019-10-03
Notification processing moved to common

gkillough:


src/test/java/com/synopsys/integration/alert/workflow/filter/NotificationToDistributionEventConverterTestIT.java
```diff
@@ -8,14 +8,14 @@
...
 
-import com.synopsys.integration.alert.channel.util.NotificationToDistributionEventConverter;
 import com.synopsys.integration.alert.common.event.DistributionEvent;
 import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.message.model.MessageContentGroup;
 import com.synopsys.integration.alert.common.message.model.ProviderMessageContent;
 import com.synopsys.integration.alert.common.persistence.accessor.ConfigurationAccessor;
 import com.synopsys.integration.alert.common.persistence.model.ConfigurationFieldModel;
 import com.synopsys.integration.alert.common.persistence.model.ConfigurationJobModel;
+import com.synopsys.integration.alert.common.workflow.processor.NotificationToDistributionEventConverter;
```
bamandel: Would this test make more sense in the common test package now?

gkillough: Since it is an IT Test, it won't work in common.

PR #647, opened 2019-10-04
Bm authentication component

bamandel:
This separates authentication from settings making everything a little easier to read. It also migrates all of the data from the settings component to a new auth config if present.

This removes authentication from the startup page. It was expected and would have made this PR significantly larger if I fixed it immediately. It will be added in a new way soon.


src/main/java/com/synopsys/integration/alert/component/authentication/actions/AuthenticationApiAction.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private FieldModel handleNewAndUpdatedConfig(final FieldModel fieldModel) {
+        addSAMLMetadata(fieldModel);
+        return fieldModel;
+    }
+
+    private void addSAMLMetadata(final FieldModel fieldModel) {
+        final Boolean samlEnabled = fieldModel.getFieldValueModel(AuthenticationDescriptor.KEY_SAML_ENABLED)
+                                        .map(fieldValueModel -> fieldValueModel.getValue()
```
gkillough: Should this be `flatMap`?

gkillough:
```java
fieldModel.getFieldValueModel(AuthenticationDescriptor.KEY_SAML_ENABLED)
          .flatMap(FieldValueModel::getValue)
          .orElse(false);
```
We should also store `samlEnabled` as a `boolean` primitive according to SonarCloud.

src/test/java/com/synopsys/integration/alert/component/settings/SettingsGlobalApiActionTest.java
```diff
@@ -188,22 +189,23 @@ public void testValidateFieldsIsSetFalseHasValue() {
...
-        fieldModel.putField(SettingsDescriptor.KEY_LDAP_SERVER, new FieldValueModel(List.of(""), false));
-        final HashMap<String, String> fieldErrors = new HashMap<>();
-        final Map<String, ConfigField> configFieldMap = settingsUIConfig.createFields().stream()
-                                                            .collect(Collectors.toMap(ConfigField::getKey, Function.identity()));
-        final FieldValidationAction fieldValidationAction = new FieldValidationAction();
-        fieldValidationAction.validateConfig(configFieldMap, fieldModel, fieldErrors);
-        assertFalse(fieldErrors.isEmpty());
-        assertEquals(SettingsDescriptor.FIELD_ERROR_LDAP_SERVER_MISSING, fieldErrors.get(SettingsDescriptor.KEY_LDAP_SERVER));
-    }
+    // TODO move this to the AuthenticationAction test class
```
gkillough: Don't forget to do this before you merge.

PR #651, opened 2019-10-08
Configuration timestamp (Backend)

jamesrichard91:
This is the Java code and the Database changes to support the created and last updated times


src/main/resources/db/changelog/alert/5.2.0/changelog-configuration-timestamp.xml
```diff
@@ -0,0 +1,13 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="jrichard" id="2019-10-07-15-15-58-375-1">
+         <addColumn tableName="DESCRIPTOR_CONFIGS" schemaName="ALERT">
+             <column name="CREATED_AT" type="TIMESTAMP" valueComputed="SYSDATE"/>
+        </addColumn>
+        <addColumn tableName="DESCRIPTOR_CONFIGS" schemaName="ALERT">
+            <column name="LAST_UPDATED" type="TIMESTAMP" valueComputed="SYSDATE"/>
+        </addColumn>
+    </changeSet>
```
bamandel: Should we consider adding Timestamp data to all configurations in the upgrade? This way all pages will have the created/modified data present when viewing the UI instead of random pages showing this data and others not.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
```diff
@@ -356,21 +366,27 @@ private ConfigurationModel createConfigForRelevantFields(final String descriptor
...
         }
         return configs;
     }
 
-    private ConfigurationModel createConfigModel(final Long descriptorId, final Long configId, final Long contextId) throws AlertDatabaseConstraintException {
+    private ConfigurationModel createConfigModel(Long descriptorId, Long configId, Date createdAt, Date lastUpdated, Long contextId) throws AlertDatabaseConstraintException {
         final String configContext = getContextById(contextId);
-        final ConfigurationModel newModel = new ConfigurationModel(descriptorId, configId, configContext);
+
+        SimpleDateFormat formatter = new SimpleDateFormat(DATE_FORMAT);
```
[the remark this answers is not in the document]
psantos1113: We use SimpleDateFormat in multiple places.  I think that should be a separate PR to consolidate the usage of SimpleDateFormat because we have a few different format strings.

src/main/java/com/synopsys/integration/alert/web/config/ConfigActions.java
```diff
@@ -86,6 +86,7 @@ public boolean doesConfigExist(final Long id) throws AlertException {
                 }
             }
 
+            // TODO investigate why this is here instead of in an else if
```
psantos1113: The initial setup screen would not display because we had an empty list of configurations but no field models which caused issues generating the UI.  This may no longer be needed.

src/main/resources/db/changelog/alert/5.2.0/changelog-configuration-timestamp.xml
psantos1113: The defaultValueComputed="SYSDATE" attribute on the column may handle creating a default value of the current time.

src/main/java/com/synopsys/integration/alert/workflow/startup/component/AlertStartupInitializer.java
```diff
@@ -207,17 +207,17 @@ private void updateConfigurationFields(final String descriptorName, final boolea
...
+        final FieldModel fieldModel = new FieldModel(descriptorName, ConfigContextEnum.GLOBAL.name(), createdAt, lastUpdated, fieldValueModelMap);
         final FieldModel updatedFieldModel = fieldModelProcessor.performBeforeUpdateAction(fieldModel);
         return modelConverter.convertToConfigurationFieldModelMap(updatedFieldModel).values();
     }
 
-    private Collection<ConfigurationFieldModel> saveAction(final String descriptorName, final Collection<ConfigurationFieldModel> configurationFieldModels) throws AlertException {
+    private Collection<ConfigurationFieldModel> saveAction(String descriptorName, Collection<ConfigurationFieldModel> configurationFieldModels) throws AlertException {
         final Map<String, FieldValueModel> fieldValueModelMap = modelConverter.convertToFieldValuesMap(configurationFieldModels);
-        final FieldModel fieldModel = new FieldModel(descriptorName, ConfigContextEnum.GLOBAL.name(), fieldValueModelMap);
+        final FieldModel fieldModel = new FieldModel(descriptorName, ConfigContextEnum.GLOBAL.name(), null, null, fieldValueModelMap);
```
gkillough: Should these be initialized, or does the valueComputed take care of it?

jamesrichard91: No, good catch. I need to set the date here

PR #654, opened 2019-10-09
Bm settings util

bamandel:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/SettingsUtility.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Optional;
+
+import com.synopsys.integration.alert.common.exception.AlertException;
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationFieldModel;
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
+import com.synopsys.integration.alert.common.persistence.model.DefinedFieldModel;
+
+public interface SettingsUtility {
+
+    String getSettingsName();
```
bamandel: I was questioning adding this item to this interface. I feel this utility should perform the actions necessary instead of letting people figure out the name and doing it themselves. Should we keep this? There were a couple classes that would require a bit of a rework if they didn't know the settings descriptor name.

gkillough: Why not have the universal key?

src/main/java/com/synopsys/integration/alert/web/actions/SystemActions.java
```diff
@@ -91,10 +91,10 @@ public boolean isSystemInitialized() {
 
     public FieldModel getCurrentSystemSetup() {
         final Map<String, FieldValueModel> valueMap = new HashMap<>();
-        FieldModel model = new FieldModel(settingsDescriptorKey.getUniversalKey(), ConfigContextEnum.GLOBAL.name(), valueMap);
+        FieldModel model = new FieldModel(settingsUtility.getSettingsName(), ConfigContextEnum.GLOBAL.name(), valueMap);
```
gkillough: If we're not gonna keep the descriptor key here, then why aren't we using the Utility to get the FieldModel?

src/main/java/com/synopsys/integration/alert/web/security/authentication/saml/AlertWebSSOProfileOptions.java
```diff
@@ -44,7 +43,7 @@ public Boolean getForceAuthN() {
         try {
             final ConfigurationModel currentConfiguration = samlContext.getCurrentConfiguration();
             return samlContext.getFieldValueBoolean(currentConfiguration, AuthenticationDescriptor.KEY_SAML_FORCE_AUTH);
-        } catch (final AlertDatabaseConstraintException | AlertLDAPConfigurationException e) {
+        } catch (final AlertException e) {
             logger.error("Could not get the SAML force AuthN.", e);
```
gkillough: Since this line was exposed by this PR, can we make it use a real word? "Authentication"?

bamandel: Didn't even notice that. Good catch

PR #656, opened 2019-10-10
Last updated timestamp

jamesrichard91:


src/main/css/app.scss
```diff
@@ -274,6 +274,10 @@ form.form-horizontal.loginForm {
   border-bottom: $navigation-border;
 }
 
+.timeStampContainer {
+  float: right;
+}
```
bamandel: Does this time stamp text look exactly the same as all other text? I feel it should be a little more opague or something to draw less attention to it

PR #670, opened 2019-10-14
Ps initial settings removal

psantos1113:
Remove the initial settings screen.


src/main/java/com/synopsys/integration/alert/web/controller/SystemController.java
```diff
@@ -100,55 +89,22 @@ public SystemController(final SystemActions systemActions, final ContentConverte
...
-    public ResponseEntity<String> getInitialSystemSetup(final HttpServletRequest request) {
-        final Optional<ResponseEntity<String>> previousSetupResponse = checkForInitialSetup(request.getServletContext().getContextPath());
-        if (previousSetupResponse.isPresent()) {
-            return previousSetupResponse.get();
-        }
-
-        return responseFactory.createOkContentResponse(contentConverter.getJsonString(systemActions.getCurrentSystemSetup()));
+    @Deprecated
+    public ResponseEntity<String> getInitialSystemSetup() {
+        // FIXME: 6.0.0 remove this method
```
psantos1113: I have this here to avoid breaking changes but I could remove the method.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/PasswordConfigField.java
```diff
@@ -22,55 +22,68 @@
...
 
-    public static PasswordConfigField createRequired(final String key, final String label, final String description, final String panel, final ConfigValidationFunction validationFunction) {
-        return new PasswordConfigField(key, label, description, true, validationFunction);
+    private void createValidators(ConfigValidationFunction[] validationFunctions) {
+        List<ConfigValidationFunction> validators = new ArrayList<>();
+        validators.add(encryptionValidation);
+        if (null != validationFunctions) {
+            validators.addAll(Arrays.asList(validationFunctions));
+        }
+        this.setValidationFunctions(validators.stream().collect(Collectors.toUnmodifiableList()));
```
bamandel: We have almost this exact method in most ConfigFields. Perhaps we should add a method or a new ctor param to the parent class to add a default validation function?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/SelectConfigField.java
```diff
@@ -165,15 +168,14 @@ public void setClearable(final boolean clearable) {
...
-            validationFunctions = List.of(this::validateIsValidOption);
+    private void createValidators(ConfigValidationFunction[] validationFunctions) {
+        List<ConfigValidationFunction> validators = new ArrayList<>();
+        validators.add(this::validateIsValidOption);
+        if (null != validationFunctions) {
+            validators.addAll(Arrays.asList(validationFunctions));
         }
-        return validate(fieldValueModel, fieldModel, validationFunctions);
+        this.setValidationFunctions(validators.stream().collect(Collectors.toUnmodifiableList()));
+        ;
```
bamandel: extra semicolon

src/main/java/com/synopsys/integration/alert/component/settings/descriptor/SettingsUIConfig.java
```diff
@@ -127,4 +136,29 @@ public SettingsUIConfig(FilePersistenceUtil filePersistenceUtil) {
...
     }
+
+    private class EncryptionFieldsSetValidator extends EncryptionValidator {
+        @Override
+        public Collection<String> apply(final FieldValueModel fieldValueModel, final FieldModel fieldModel) {
+            Optional<FieldValueModel> pwdField = fieldModel.getFieldValueModel(SettingsDescriptor.KEY_ENCRYPTION_PWD);
+            Optional<FieldValueModel> saltField = fieldModel.getFieldValueModel(SettingsDescriptor.KEY_ENCRYPTION_GLOBAL_SALT);
+
+            boolean encryptionFieldsSet = pwdField.map(field -> field.hasValues() || field.isSet()).orElse(false)
+                                              && saltField.map(field -> field.hasValues() || field.isSet()).orElse(false);
```
bamandel: The optional fields here can be converted to booleans when initially stored then you can do a very basic check here instead of this busy one.

src/main/java/com/synopsys/integration/alert/component/settings/descriptor/SettingsUIConfig.java
```diff
@@ -127,4 +136,29 @@ public SettingsUIConfig(FilePersistenceUtil filePersistenceUtil) {
...
+                return List.of();
+            }
+            return List.of(ConfigField.REQUIRED_FIELD_MISSING);
+        }
+    }
+
+    private class EncryptionFieldValidator extends EncryptionValidator {
+        @Override
+        public Collection<String> apply(final FieldValueModel fieldValueModel, final FieldModel fieldModel) {
+            if (!fieldValueModel.hasValues() && !fieldValueModel.isSet()) {
```
bamandel: I believe you can call the fieldValueModel.containsNoData() method here instead

src/main/java/com/synopsys/integration/alert/web/controller/SystemController.java
bamandel: I'm all for removing this as I can't imagine anyone was using any of these endpoints besides us. But it would technically be breaking.

src/main/java/com/synopsys/integration/alert/web/security/HttpPathManager.java
```diff
@@ -71,7 +70,6 @@
         PATH_PASSWORD_RESET,
         PATH_PASSWORD_RESET_ROOT,
         PATH_SYSTEM_MESSAGES_LATEST,
-        PATH_SYSTEM_SETUP_INITIAL,
         PATH_SYSTEM_SETUP_DESCRIPTOR,
```
bamandel: Would this Descriptor also be no longer required and the one below on line 88?

src/main/java/com/synopsys/integration/alert/component/settings/descriptor/SettingsUIConfig.java
bamandel: You might also be able to use the !containsNoData here but I'm not sure if it's the same check or not

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/UploadFileButtonField.java
```diff
@@ -36,8 +36,18 @@
...
 
-    private UploadFileButtonField(String key, String label, String description, boolean required, String buttonLabel, List<String> accept, String capture, Boolean multiple, ConfigValidationFunction validationFunction) {
-        super(key, label, description, FieldType.UPLOAD_FILE_BUTTON, required, false, validationFunction);
+    private UploadFileButtonField(String key, String label, String description, boolean required, String buttonLabel, List<String> accept, String capture, Boolean multiple, ConfigValidationFunction... validationFunction) {
+        super(key, label, description, FieldType.UPLOAD_FILE_BUTTON, required, false);
+        this.buttonLabel = buttonLabel;
+        this.endpoint = UploadEndpointManager.UPLOAD_ENDPOINT_URL;
+        this.accept = accept;
+        this.capture = capture;
+        this.multiple = multiple;
```
gkillough: Can we name this variable something more Boolean-esque? `isMultiFileUpload` or something.

psantos1113: The variables here match exactly the attribute names of the HTML file input.  So I used the attribute naming of the HTML input field in order to make it easier to map the variables to attributes on the Javascript side.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/PasswordConfigField.java
psantos1113: Because we're using method references with the this pointer I cannot add them in the constructor.  Unless I create a small private class that implements ConfigValidationFunction.

PR #674, opened 2019-10-15
Added fields to be passed as data to custom endpoint fields

bamandel:
This isn't a change that is super required. But it does make a lot of the requests smaller.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomButtonField.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomButtonField.java
@@ -1,3 +1,25 @@
+/**
+ * alert-common
+ *
+ * Copyright (c) 2019 Synopsys, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package com.synopsys.integration.alert.common.descriptor.config.field.endpoint;
 
 import java.util.HashSet;
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomButtonField.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomButtonField.java
@@ -1,0 +1,38 @@
+package com.synopsys.integration.alert.common.descriptor.config.field.endpoint;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.synopsys.integration.alert.common.descriptor.config.field.ConfigField;
+import com.synopsys.integration.alert.common.enumeration.FieldType;
+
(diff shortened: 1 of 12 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomButtonField.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.synopsys.integration.alert.common.descriptor.config.field.ConfigField;
+import com.synopsys.integration.alert.common.enumeration.FieldType;
+
+public abstract class CustomButtonField extends ConfigField {
+    private String buttonLabel;
+    private String endpoint;
```
gkillough: Why would this class, which does not have "endpoint" in its name have the endpoint field as its member? `EndpointButtonField` seems to be a better place for this, no?

gkillough: I realize the package is called "endpoint", but then why prefix any classes in the package with "Endpoint"? There needs to be a name change somewhere.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
```diff
@@ -93,7 +93,8 @@ public ProviderDistributionUIConfig(final String label, final String urlName, fi
         final ConfigField projectNamePattern = TextInputConfigField.create(KEY_PROJECT_NAME_PATTERN, LABEL_PROJECT_NAME_PATTERN, DESCRIPTION_PROJECT_NAME_PATTERN, this::validateProjectNamePattern);
         final ConfigField configuredProject = EndpointTableSelectField.createSearchable(KEY_CONFIGURED_PROJECT, LABEL_PROJECTS, DESCRIPTION_PROJECTS, this::validateConfiguredProject)
                                                   .addColumn(new TableSelectColumn("name", "Project Name", true, true))
-                                                  .addColumn(new TableSelectColumn("description", "Project Description", false, false));
+                                                  .addColumn(new TableSelectColumn("description", "Project Description", false, false))
+                                                  .addDataFieldKey(ChannelDistributionUIConfig.KEY_PROVIDER_NAME);
```
psantos1113: The naming of addDataFieldKey is a little confusing here.   It's used to create an initial empty model and then the EndpointTableSelectField will populate the additional fields from the current config so that you build a full model.  Should this be called a requiredFieldKey so the method would change to addRequiredFieldKey?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomButtonField.java
```diff
@@ -0,0 +1,60 @@
...
+    private Set<String> returnedDataFieldKeys;
+
+    public CustomButtonField(final String key, final String label, final String description, final FieldType type, final boolean required, final boolean sensitive, final String buttonLabel, final String endpoint) {
+        super(key, label, description, type, required, sensitive);
+        this.buttonLabel = buttonLabel;
+        this.endpoint = endpoint;
+        this.returnedDataFieldKeys = new HashSet<>();
```
psantos1113: This variable is used to create an empty model so this is defining other required field keys for the component.  Perhaps requiredFieldKeys may be a better name?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/EndpointField.java
```diff
@@ -0,0 +1,60 @@
...
+package com.synopsys.integration.alert.common.descriptor.config.field.endpoint;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.synopsys.integration.alert.common.descriptor.config.field.ConfigField;
+import com.synopsys.integration.alert.common.enumeration.FieldType;
+
+public abstract class EndpointField extends ConfigField {
+    private String buttonLabel;
```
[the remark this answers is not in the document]
bamandel: Yes, the assumption is that anything that has an endpoint will have some sort of button that is clicked to hit the endpoint.

PR #678, opened 2019-10-17
Abstract Issue Handler

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/IssueHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        Set<String> issueKeys = new HashSet<>();
+        String projectName = issueConfig.getProjectName();
+
+        SetMap<String, String> missingTransitionToIssues = SetMap.createDefault();
+        for (Set<ComponentItem> componentItems : groupedComponentItems.values()) {
+            try {
+                ComponentItem arbitraryItem = componentItems
+                                                  .stream()
+                                                  .findAny()
+                                                  .orElseThrow(() -> new AlertException(String.format("No actionable component items were found. Provider: %s, Topic: %s, SubTopic: %s", providerName, topic, nullableSubTopic)));
```
jamesrichard91: Does it make sense to throw this exception? Or should we log a warning and continue to the next group?

gkillough: This exception will never actually be thrown. But throwing an exception is better than returning null because of our SonarCloud rules.

PR #687, opened 2019-10-23
Bm reformat ms

bamandel:
I essentially modified the message parser to convert the message into a JSON blob that will be sent. It might be a little too messy.

Haven't been able to test an actual message yet due to the power outage.

*DO NOT MERGE UNTIL DC1 IS BACK UP AND THIS IS TESTED*


src/test/java/com/synopsys/integration/alert/channel/msteams/MsTeamsChannelTest.java
```diff
@@ -32,22 +31,21 @@
     @Tag(TestTags.DEFAULT_INTEGRATION)
     @Tag(TestTags.CUSTOM_EXTERNAL_CONNECTION)
     public void sendMessageTestIT() throws IOException, IntegrationException {
```
gkillough: Don't forget to delete the old json file that Eric created for this.

src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsMessageParser.java
```diff
@@ -0,0 +1,100 @@
...
+        return "\r\n";
+    }
+
+    @Override
+    protected String getSectionSeparator() {
+        return "";
+    }
+
+    @Override
+    public String createHeader(MessageContentGroup messageContentGroup) {
```
gkillough: I don't know if this is correct. Every piece of message that we send will require this wrapping (e.g. if we split it into multiple requests). This should exist outside of the message parser.

bamandel: From what I saw, the only place has multiple is the components themselves. The header won't ever be duplicate.

bamandel: I think I see what you mean. But the way we do it is to split it out into cards. I will check tomorrow on our server to see if the cards can handle large data loads (Seamonkey). If it can, then I think it's fine as is

PR #688, opened 2019-10-23
Ps jira server install

psantos1113:
Implement the changes to install the marketplace app into Jira Server.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
@@ -22,6 +22,7 @@
  */
 package com.synopsys.integration.alert.channel.jira.server;
 
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 
 import com.google.gson.Gson;
@@ -53,8 +54,12 @@ public class JiraServerProperties {
 
     public JiraServerRestConfig createJiraServerConfig() throws AlertException {
         JiraServerRestConfigBuilder jiraServerConfigBuilder = new JiraServerRestConfigBuilder();
+        String jiraUrl = url;
+        if (!StringUtils.endsWith(url, "/jira")) {
+            jiraUrl = jiraUrl + "/jira";
+        }
 
-        jiraServerConfigBuilder.setUrl(url);
+        jiraServerConfigBuilder.setUrl(jiraUrl);
         jiraServerConfigBuilder.setAuthPassword(password);
         jiraServerConfigBuilder.setAuthUsername(username);
         try {
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
@@ -22,7 +22,6 @@
  */
 package com.synopsys.integration.alert.channel.jira.server;
 
-import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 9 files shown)


src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
```diff
@@ -0,0 +1,90 @@
...
+    public JiraServerProperties(String url, String password, String username) {
+        this.url = url;
+        this.password = password;
+        this.username = username;
+    }
+
+    public JiraServerRestConfig createJiraServerConfig() throws AlertException {
+        JiraServerRestConfigBuilder jiraServerConfigBuilder = new JiraServerRestConfigBuilder();
+        String jiraUrl = url;
+        if (!StringUtils.endsWith(url, "/jira")) {
```
jamesrichard91: We force the URL to end in '/jira'? For what reason?

src/main/java/com/synopsys/integration/alert/channel/jira/server/descriptor/JiraServerGlobalUIConfig.java
```diff
@@ -24,26 +24,46 @@
...
 import com.synopsys.integration.alert.common.descriptor.config.field.ConfigField;
+import com.synopsys.integration.alert.common.descriptor.config.field.PasswordConfigField;
 import com.synopsys.integration.alert.common.descriptor.config.field.TextInputConfigField;
+import com.synopsys.integration.alert.common.descriptor.config.field.endpoint.EndpointButtonField;
+import com.synopsys.integration.alert.common.descriptor.config.field.validators.EncryptionValidator;
 import com.synopsys.integration.alert.common.descriptor.config.ui.UIConfig;
 
 @Component
 public class JiraServerGlobalUIConfig extends UIConfig {
+    public static final String LABEL_SERVER_ADMIN_USER_NAME = "Admin Email Address";
```
gkillough: Should this be username?

src/main/java/com/synopsys/integration/alert/channel/jira/server/descriptor/JiraServerGlobalUIConfig.java
```diff
@@ -24,26 +24,46 @@
...
+    public static final String LABEL_SERVER_ADMIN_PASSWORD = "Admin Password";
+    public static final String LABEL_SERVER_CONFIGURE_PLUGIN = "Configure Jira Server plugin";
+    public static final String DESCRIPTION_SERVER_ADMIN_USER_NAME = "The email address of the admin used to log into the Jira Cloud server that has generated the API token.";
```
gkillough: This string should be updated to correctly reflect the auth method.

src/main/java/com/synopsys/integration/alert/channel/jira/server/descriptor/JiraServerGlobalUIConfig.java
```diff
@@ -24,26 +24,46 @@
...
+    public static final String DESCRIPTION_SERVER_ADMIN_PASSWORD = "The admin API token used to send API requests to the Jira Cloud server.";
```
gkillough: This string should also be updated.

src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
psantos1113: in order to install the indexing app in a jira server the URL must have /jira in the path.  I can make a AppServices class in int-jira-common that extends PluginManagerService that will create the url with the "/jira" in the path.  The /jira in the path is used as the context to be able to install apps into jira server.  Otherwise you get 404 errors 
Jira cloud does not require "/jira" in the path.  I will insulate Alert from that knowledge.

src/main/java/com/synopsys/integration/alert/channel/jira/server/descriptor/JiraServerGlobalUIConfig.java
psantos1113: yes need to change that thanks.

src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
psantos1113: I removed this code.  We'll just expect the URL to have /jira in it when they configure the channel.

PR #694, opened 2019-10-28
Create AuthenticationPerformer to genericize the way we handle username/password auth

gkillough:


src/main/java/com/synopsys/integration/alert/web/security/authentication/AuthenticationPerformer.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public abstract Authentication authenticateWithProvider(Authentication pendingAuthentication);
+
+    @Override
+    public int compareTo(AuthenticationPerformer other) {
+        return this.priority().compareTo(other.priority());
+    }
+
+    protected abstract AuthenticationPriority priority();
```
bamandel: instead of an abstract method this can be replaced with a getter and the value passed in through the Constructor.

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/AuthenticationPriority.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.enumeration;
+
+public enum AuthenticationPriority {
```
jamesrichard91: Does this make sense? How do we handle new authentication mechanisms? Like authenticating via GitHub? Would we keep adding time related enums? soonish? before-standard?
Instead of an enum we may just want to use an integer to denote the provider's priority. OR we can have an authentication manager that registers all of the providers and maintains the order instead of each provider setting its own priority.

PR #695, opened 2019-10-28
Fix(customEndpoint): Fix the creation of the model from requested fie…

psantos1113:
Fix the EndpointButton to create a model that has the isSet attribute set correctly.

### DIFF
--- a/src/main/js/field/EndpointButtonField.js
+++ b/src/main/js/field/EndpointButtonField.js
@@ -43,13 +43,8 @@ class EndpointButtonField extends Component {
         const {
             fieldKey, csrfToken, onChange, currentConfig, endpoint, requestedDataFieldKeys
         } = this.props;
-        let newFieldModel = FieldModelUtilities.createEmptyFieldModel(requestedDataFieldKeys, currentConfig.context, currentConfig.descriptorName);
-        requestedDataFieldKeys.forEach((field) => {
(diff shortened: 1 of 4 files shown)


src/main/js/field/EndpointButtonField.js
```diff
@@ -43,13 +43,8 @@ class EndpointButtonField extends Component {
         const {
             fieldKey, csrfToken, onChange, currentConfig, endpoint, requestedDataFieldKeys
         } = this.props;
-        let newFieldModel = FieldModelUtilities.createEmptyFieldModel(requestedDataFieldKeys, currentConfig.context, currentConfig.descriptorName);
-        requestedDataFieldKeys.forEach((field) => {
-            const values = FieldModelUtilities.getFieldModelValues(currentConfig, field);
-            newFieldModel = FieldModelUtilities.updateFieldModelValues(newFieldModel, field, values);
-        });
+        const newFieldModel = FieldModelUtilities.createFieldModelFromRequestedFields(currentConfig, requestedDataFieldKeys);
```
bamandel: This change can probably be implemented in the other endpoint classes as well (TableSelectInput.js and EndpointSelectField.js).

psantos1113: I will make those changes before merging.

PR #696, opened 2019-10-28
MsTeams message reformatted to match other message boards

bamandel:
Added encoding to prevent some of the text from looking bad in markup.
Added a message splitter to send larger messages. (This makes it a bit hard to track all of the data, may improve it some)


src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsMessageParser.java
```diff
@@ -0,0 +1,129 @@
...
+import com.synopsys.integration.alert.common.message.model.MessageContentGroup;
+import com.synopsys.integration.alert.common.message.model.ProviderMessageContent;
+
+@Component
+public class MsTeamsMessageParser extends ChannelMessageParser {
+    private Set<Character> reservedChars = Set.of('*', '~', '#', '-', '_');
+
+    @Override
+    protected String encodeString(String txt) {
+        StringBuilder newTxt = new StringBuilder(txt);
```
gkillough: I know we discussed, but it may be worthwhile adding an EscapeUtils class or something. Check out `JiraIssuePropertyHandler.escapeSearchString(...)`.

src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsEventParser.java
```diff
@@ -53,12 +57,37 @@ public MsTeamsEventParser(FreemarkerTemplatingService freemarkerTemplatingServic
         }
     }
 
-    public MsTeamsMessage createMessage(DistributionEvent distributionEvent) {
-        return distributionEvent
-                   .getContent()
-                   .getSubContent()
-                   .stream()
-                   .collect(MsTeamsMessage::new, MsTeamsMessage::addContent, MsTeamsMessage::addAllContent);
+    public List<MsTeamsMessage> splitMessages(MsTeamsMessage msTeamsMessage) {
```
gkillough: Was the MessageSplitter class not useful here?

PR #697, opened 2019-10-29
Ps issue type validation

psantos1113:
Validate the issue types belong to the project.


src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerIssueConfigValidator.java
```diff
@@ -138,14 +141,23 @@ private String validateIssueCreator(FieldAccessor fieldAccessor, Map<String, Str
...
-                return issueType;
+                if (optionalProjectName.isPresent()) {
+                    String projectName = optionalProjectName.get();
+                    boolean isValidForProject = issueMetaDataService.doesProjectContainIssueType(projectName, issueType);
+                    if (isValidForProject) {
+                        return issueType;
+                    } else {
+                        fieldErrors.put(JiraServerDescriptor.KEY_ISSUE_TYPE, String.format("The issue type '%s' not assigned to project '%s'", issueType, projectName));
+                    }
+                }
```
bamandel: Looks like we're starting to duplicate code between the Jira instances. Would be nice if we could split this out somewhere.

PR #700, opened 2019-10-31
Feat(Java): Add job enabled field

gkillough:


src/main/java/com/synopsys/integration/alert/web/audit/AuditEntryActions.java
```diff
@@ -104,7 +104,12 @@ public AuditEntryActions(AuditUtility auditUtility, NotificationManager notifica
                 logger.warn("The Distribution Job with Id {} could not be found. This notification could not be sent", commonConfigId);
                 return new AlertJobMissingException("The Distribution Job with this id could not be found.", commonConfigId);
             });
-            distributionEvents = notificationProcessor.processNotifications(commonDistributionConfig, List.of(notificationContent));
+            if (commonDistributionConfig.isEnabled()) {
```
bamandel: imo we should probably still try to resend messages even if it's disabled.

gkillough: @bamandel and I discussed this offline and the reason this code exists is because when a job is disabled, the message won't be sent. This code is just a way to expose to the user why the notification was not resent.

PR #702, opened 2019-11-01
Created a more detailed error for PKIX issues.

bamandel:
Link are parsed out of the info text and converted to links.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/config/PKIXErrorResponseFactory.java
+++ b/src/main/java/com/synopsys/integration/alert/web/config/PKIXErrorResponseFactory.java
@@ -86,7 +86,7 @@ public class PKIXErrorResponseFactory {
     }
 
     private String createTitle() {
-        return "To resolve this issue, use the appropriate link below to properly install your certificates and restart Alert.";
+        return "To resolve this issue, use the appropriate link below to properly install your certificates and then restart Alert.";
     }
 
     private List<String> createInfo() {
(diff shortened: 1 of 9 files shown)


src/main/java/com/synopsys/integration/alert/web/config/PKIXErrorResponseFactory.java
```diff
@@ -0,0 +1,103 @@
...
+
+        return false;
+    }
+
+    private String createHeader() {
+        return "There were issues with your Certificates.";
+    }
+
+    private String createTitle() {
+        return "To resolve this issue, redeploy Alert and use the appropriate link below to properly install your certificates:";
```
jamesrichard91: Maybe this should be: "To resolve this issue, use the appropriate link below to properly install your certificates and restart Alert."

Unfortunately I think the term redeploy may confuse Users. They may think we are telling them to start from scratch.

bamandel: I would love to hear everyones inputs for what they think should be said here. I just threw together what I thought would be helpful

src/main/java/com/synopsys/integration/alert/web/config/PKIXErrorResponseFactory.java
```diff
@@ -0,0 +1,103 @@
...
+
+        return false;
+    }
+
+    private String createHeader() {
+        return "There were issues with your Certificates.";
+    }
+
+    private String createTitle() {
+        return "To resolve this issue, use the appropriate link below to properly install your certificates and restart Alert.";
```
jamesrichard91: This is really knit picky but could you change "and restart Alert" to "and then restart Alert". Last change I swear!

PR #704, opened 2019-11-04
Fixing a possible NoSuchElementException on startup.

jamesrichard91:


src/main/java/com/synopsys/integration/alert/workflow/startup/component/AlertStartupInitializer.java
```diff
@@ -167,76 +167,78 @@ private void initializeConfiguration(final Collection<String> descriptorNames, f
...
+            boolean hasEnvironmentValue = hasEnvironmentValue(convertedKey);
             logger.info("    {}", convertedKey);
             logger.debug("         Environment Variable Found - {}", hasEnvironmentValue);
             String defaultValue = null;
             if (existingConfiguredFields.containsKey(key)) {
-                defaultValue = existingConfiguredFields.get(key).getFieldValue().get();
+                Optional<String> fieldValue = existingConfiguredFields.get(key).getFieldValue();
+                if (fieldValue.isPresent()) {
+                    defaultValue = fieldValue.get();
+                }
```
jamesrichard91: This was the actual change. The other changes were just removing unnecessary final's

bamandel: I'm finding a lot of our PR's are bloated because of this. We should get in the habit of pointing people where to focus until the final keywords are mostly removed.

PR #709, opened 2019-11-05
UI Updates

jamesrichard91:
Feat: Updating the distribution job to only show the common fields until both the channel type and provider have been chosen.

Fix: Removing the old icons that we no longer use to clean up some UI errors.

### DIFF
--- a/src/main/js/component/AboutInfo.js
+++ b/src/main/js/component/AboutInfo.js
@@ -7,7 +7,7 @@ import { getAboutInfo } from 'store/actions/about';
 import * as DescriptorUtilities from 'util/descriptorUtilities';
 import ConfigurationLabel from 'component/common/ConfigurationLabel';
 import { NavLink } from 'react-router-dom';
-import LabeledField from '../field/LabeledField';
+import LabeledField from 'field/LabeledField';
 
 class AboutInfo extends React.Component {
     componentDidMount() {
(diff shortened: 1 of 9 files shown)


src/main/js/component/AboutInfo.js
```diff
@@ -6,21 +6,14 @@ import ReadOnlyField from 'field/ReadOnlyField';
 import { getAboutInfo } from 'store/actions/about';
 import * as DescriptorUtilities from 'util/descriptorUtilities';
 import ConfigurationLabel from 'component/common/ConfigurationLabel';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import * as IconUtility from 'util/iconUtility';
-import { NavLink } from "react-router-dom";
-import LabeledField from "../field/LabeledField";
+import { NavLink } from 'react-router-dom';
+import LabeledField from '../field/LabeledField';
```
bamandel: This should be absolute

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -223,14 +230,23 @@ class DistributionConfiguration extends Component {
...
-                                descriptorFields={currentChannel.fields}
+                                descriptorFields={commonFields}
+                                currentConfig={channelConfig}
+                                fieldErrors={this.props.fieldErrors}
+                                self={this}
+                                stateName="channelConfig"
+                            />
+                            {currentChannel && selectedProvider &&
+                            <FieldsPanel
+                                descriptorFields={channelFields}
```
jamesrichard91: This change should always show the common fields and the channel and provider fields will only be shown when the channel type and provider have been selected.

PR #710, opened 2019-11-05
Fixed MSTeams Urls

bamandel:


src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsMessageParser.java
```diff
@@ -90,7 +88,7 @@ public String createHeader(MessageContentGroup messageContentGroup) {
     protected String createLinkableItemValueString(LinkableItem linkableItem) {
         String itemUrl = linkableItem.getUrl().orElse("");
         if (StringUtils.isNotBlank(itemUrl) && itemUrl.contains(" ")) {
-            String encodedUrl = URLEncoder.encode(itemUrl, Charsets.UTF_8);
+            String encodedUrl = itemUrl.replace(" ", "%20");
```
gkillough: Are spaces the only potential problem? What about '+' and '%'? Those are pretty common and need to be encoded.

PR #712, opened 2019-11-05
Bm descriptor spinner

bamandel:
Spinner spins in main content page and the items in the left bar are removed until the descriptors are loaded.


gkillough:
I don't know how this slipped my mind when you were talking about this yesterday, but what if we had "tool drawers" for navigation? We could display "Providers", "Channels", and "Administration" (which would be mapped to components). Then they could load in the background and the drawers could have a spinner if they were hovered over while still fetching.

PR #713, opened 2019-11-06
Fixed vulnerable component link change

jamesrichard91:
Fix: Fixing an issue where the VulnerableComponentView links have changed. The vulnerabilities link is now the href as of Black Duck 2019.10.0


gkillough:
FWIW this seems like a bug in Black Duck

PR #714, opened 2019-11-06
Added a condition to check if we should add a comma to the MSTeams Freemarker template.

bamandel:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -1,18 +1,18 @@
 /**
  * alert-database
- * <p>
```
gkillough: Does someone have the wrong version of our save actions? I've seen this a bunch recently.

bamandel: I too have seen this a lot recently. Not sure why it changes a few files each time.

psantos1113: The copyright should be coming from the common-gradle-plugin when you run a build.  Should not be from the save actions as far as I knew.

PR #715, opened 2019-11-06
Allowing the Users to skip the phone home, and updating the log level handling

jamesrichard91:


src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeTask.java
```diff
@@ -128,6 +131,20 @@ public String scheduleCronExpression() {
         return CRON_EXPRESSION;
     }
 
+    public Map<String, String> getPhoneHomeVariables() {
+        Map<String, String> environment = System.getenv();
```
gkillough: Should we have an Alert property for this as well and check that first? We're using spring which should be managing our environment.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeTask.java
```diff
@@ -128,6 +131,20 @@ public String scheduleCronExpression() {
...
+        String phoneHomeURLOverride = environment.get(PhoneHomeClient.PHONE_HOME_URL_OVERRIDE_VARIABLE);
+        Map<String, String> phoneHomeVariables = new HashMap<>();
+        if (StringUtils.isNotBlank(skipPhoneHome)) {
+            phoneHomeVariables.put(PhoneHomeClient.SKIP_PHONE_HOME_VARIABLE, skipPhoneHome);
+        }
+        if (StringUtils.isNotBlank(phoneHomeURLOverride)) {
+            phoneHomeVariables.put(PhoneHomeClient.PHONE_HOME_URL_OVERRIDE_VARIABLE, phoneHomeURLOverride);
+        }
+        return phoneHomeVariables;
+    }
```
bamandel: If we're passing a map regardless, we might as well just pass the entire System.getenv() map and let Phonehome deal with it. This seems like unnecessary work.
If your goal is to allow this method for testing purposes, then just return System.getenv() here.

psantos1113: I think we want to use BlackDuckPhoneHomeHelper here.  We also get the black duck data for phone home so that might be what we want to use here.

PR #718, opened 2019-11-07
Skip phone home if variable is set

jamesrichard91:


src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeTask.java
```diff
@@ -100,6 +101,19 @@ public PhoneHomeTask(TaskScheduler taskScheduler, AboutReader aboutReader, Defau
         this.blackDuckProperties = blackDuckProperties;
     }
 
+    @Override
+    public boolean shouldScheduleTask() {
+        Map<String, String> environmentVariables = System.getenv();
+        Boolean skipPhoneHome = BooleanUtils.toBoolean(environmentVariables.get(PhoneHomeClient.SKIP_PHONE_HOME_VARIABLE));
```
jamesrichard91: Is this alright? Or should I add an @Value to this class to get this variable?

gkillough: I'm in favor of the `@Value` even if it feels like overkill.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/StartupScheduledTask.java
```diff
@@ -31,18 +31,26 @@
...
 
-    public StartupScheduledTask(final TaskScheduler taskScheduler, final String taskName, final TaskManager taskManager) {
+    public StartupScheduledTask(TaskScheduler taskScheduler, String taskName, TaskManager taskManager) {
         super(taskScheduler, taskName);
         this.taskManager = taskManager;
     }
 
     public abstract String scheduleCronExpression();
 
+    public boolean shouldScheduleTask() {
```
bamandel: I think adding enable/disable functionality would be better here. If we plan on adding Tasks in theUI, we'll need enable/disable anyway.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/StartupScheduledTask.java
```diff
@@ -31,18 +31,26 @@
...
+        return true;
+    }
+
     public void startTask() {
+        boolean shouldScheduleTask = shouldScheduleTask();
+        if (!shouldScheduleTask) {
```
bamandel: Can we put a log in this if statement to specify why the task didn't run? Perhaps Debug?

PR #724, opened 2019-11-08
Added required symbols to required fields in UI

bamandel:


src/main/js/field/LabeledField.js
```diff
@@ -16,10 +16,11 @@ class LabeledField extends Component {
     render() {
         const { showDescription } = this.state;
         const {
-            field, labelClass, description, showDescriptionPlaceHolder, label, errorName, errorValue
+            field, labelClass, description, showDescriptionPlaceHolder, label, errorName, errorValue, required
         } = this.props;
 
         const labelClasses = `${labelClass} text-right`;
+        const requiredClasses = (required) ? `${labelClasses} required` : labelClasses;
```
jamesrichard91: I would prefer if `labelClasses` became `baseClasses` and `requiredClasses` became `labelClasses`

PR #726, opened 2019-11-08
Added a column to the Jobs Distribution table

bamandel:


src/main/js/distribution/Index.js
```diff
@@ -184,6 +184,18 @@ class Index extends Component {
         return <IconTableCellFormatter handleButtonClicked={this.copyButtonClicked} currentRowSelected={row} buttonIconName="copy" buttonText="Copy" />;
     }
 
+    enabledState(cell, row) {
+        const icon = (cell == 'true') ? 'check' : 'times';
+        const color = (cell == 'true') ? 'green' : 'red';
```
gkillough: Do we have constants for these colors? I think we have slight variations with each to comply with Synopsys branding requirements.

PR #727, opened 2019-11-11
Ps changelock fix

psantos1113:
A patch where upon startup of Alert liquibase will release the changelocks on the database.

A new changelocks directory is created with an empty changeset.
The jar files for the liquibase command line are copied into the lib/ directory of the deployment.
The entrypoint script executes the liquibase command line to release the locks on the database.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,9 +121,13 @@ bootJar {
 distributions {
     boot {
         contents {
-            from("$project.buildDir/libs") {
+            from("$project.buildDir/libs/liquibase") {
+                include 'h2*.jar'
                 include 'liquibase-core*.jar'
-                into 'lib'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -121,12 +121,39 @@ bootJar {
...
+    from findJar('slf4j-api')
+    from findJar('snakeyaml')
+    into "${project.buildDir}/libs/liquibase"
+}
+
+def findJar(prefix) {
+    configurations.runtimeClasspath.filter {
+        it.name.startsWith(prefix)
+    }
+}
```
jamesrichard91: This should take in an array of prefix's to prevent iterating through the same list multiple times.

psantos1113: Will make the improvement in the master branch.

PR #735, opened 2019-11-13
Feat(Java): Add user update method to UserAccessor

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
```diff
@@ -70,49 +71,56 @@ public DefaultUserAccessor(UserRepository userRepository, UserRoleRepository use
...
     @Override
-    public UserModel addUser(String userName, String password, String emailAddress) {
-        return addOrUpdateUser(UserModel.of(userName, password, emailAddress, Collections.emptySet()));
+    public UserModel updateUser(UserModel user, boolean passwordEncoded) throws AlertDatabaseConstraintException {
+        Long userId = user.getId();
+        if (null == userId) {
+            throw new AlertDatabaseConstraintException("A user id must be specified");
+        }
+
+        userRepository.findById(userId)
```
bamandel: Is this check only to verify that a user exists? We could use the query to check for existence instead

src/main/java/com/synopsys/integration/alert/web/security/authentication/event/AuthenticationEventListener.java
```diff
@@ -51,9 +50,12 @@ public String getDestinationName() {
...
     public void handleEvent(AlertAuthenticationEvent event) {
         UserModel user = event.getUser();
-        Optional<UserModel> dbUser = userAccessor.getUser(user.getName());
-        if (dbUser.isEmpty()) {
-            userAccessor.addOrUpdateUser(user, true);
+        if (null != user) {
+            try {
+                userAccessor.addUser(user, true);
+            } catch (AlertDatabaseConstraintException ignored) {
+                // User already exists. Nothing to do.
```
bamandel: It's strange that we go through the effort of generating this exception for addUser then completely ignore it when it occurs. Since there's already a constraint on the DB level in regards to this, can we get away with not forcing this exception and perhaps adding a does user exist check when necessary?

psantos1113: Isn't this for synching up users from LDAP or SAML.  So if the user has already logged in and the user exists we don't care.  But when we add a user through the API we will want the exception thrown there.  I think that's why this is ignored here.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
gkillough: It's to check that the id is valid before attempting to continue.

src/main/java/com/synopsys/integration/alert/web/security/authentication/event/AuthenticationEventListener.java
gkillough: If you look at the original code, we attempt to get the user and, if absent, add the user to the DB. In the case where we added a new user, there would be two overlapping database api calls, now there is only one.

PR #744, opened 2019-11-19
Bm permission bitmask

bamandel:
This replaces the AccessOperations table and instead uses a bitmask for access permissions. A lot of the java remains the same, and our data has been migrated. This helps cut down on rows and queries.


src/main/resources/db/changelog/alert/5.3.0/permissions-matrix.xml
```diff
@@ -0,0 +1,283 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="bmandel" id="2019-11-18-14-48-50-829">
+        <renameColumn tableName="PERMISSION_MATRIX" oldColumnName="ACCESS_OPERATION_ID" newColumnName="OPERATIONS" schemaName="ALERT"/>
+        <modifyDataType tableName="PERMISSION_MATRIX" columnName="OPERATIONS" schemaName="ALERT" newDataType="int"/>
+        <dropTable tableName="ACCESS_OPERATIONS" schemaName="ALERT"/>
+        <delete tableName="PERMISSION_MATRIX" schemaName="ALERT"/>
```
gkillough: Do this before modifying the table.

src/main/resources/db/changelog/alert/5.3.0/permissions-matrix.xml
```diff
@@ -0,0 +1,283 @@
...
+
+        <!-- Admin permissions -->
+
+        <insert tableName="PERMISSION_MATRIX" schemaName="ALERT">
+            <column name="ROLE_ID" valueComputed="SELECT ID FROM ALERT.ROLES WHERE ROLENAME = 'ALERT_ADMIN'"/>
```
gkillough: Do we have a stored procedure to get the role id?

gkillough: GET_ID_FOR_USER_ROLE

psantos1113: I would also use `GET_ID_FOR_REGISTERED_DESCRIPTOR_NAME('provider_blackduck)` for the DESCRIPTOR_ID column also.  Also use `GET_ID_FOR_CONFIG_CONTEXT` for the CONFIG context

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/AccessOperation.java
```diff
@@ -22,13 +22,45 @@
...
+    CREATE(0),
+    DELETE(1),
+    READ(2),
+    WRITE(3),
+    EXECUTE(4),
+    UPLOAD_FILE_READ(5),
+    UPLOAD_FILE_WRITE(6),
+    UPLOAD_FILE_DELETE(7);
+
+    private int bit;
```
gkillough: We should make this final so that we don't accidentally modify state.

PR #747, opened 2019-11-20
Sync all Black Duck users (and make sure they're active!)

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -196,69 +158,112 @@ public void mapUsersToProjectByEmail(String projectHref, Collection<String> emai
...
-        List<ProviderUserModel> savedUsers = users
-                                                 .stream()
-                                                 .map(user -> convertToUserEntity(descriptorKey, user))
-                                                 .map(providerUserRepository::save)
-                                                 .map(entity -> convertToUserModel(entity))
-                                                 .collect(Collectors.toList());
-        return savedUsers;
+    public void updateUserData(ProviderKey providerKey, Set<String> emailAddresses) {
+        updateUserDB(providerKey, emailAddresses);
     }
```
jamesrichard91: Do we need the updateProjectDB method? If this is just a pass through, shouldnt we just move the content of the updateProjectDB method into updateUserData?

gkillough: This PR wasn't quite ready. I was gonna add you as a reviewer once it was done.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -196,69 +158,113 @@ public void mapUsersToProjectByEmail(String projectHref, Collection<String> emai
...
+        updateProjectDB(providerKey, projectToUserData.keySet());
         Set<String> userData = projectToUserData.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
-        updateUserDB(descriptorKey, userData);
+        userData.addAll(additionalRelevantUsers);
+        updateUserDB(providerKey, userData);
         updateUserProjectRelations(projectToUserData);
     }
 
+    // ========================
+    // Non-interface public API
```
gkillough: These are used for the database subproject and tests.

bamandel: Looking at most of these methods, they are used within this class or strictly in tests. Are we sure we want to make them public? Generally we want to avoid methods specifically for testing.

PR #750, opened 2019-11-22
Refactor(jira): Extract a separate sub project for Jira.

psantos1113:
Move the Jira Issue Tracker code into it's own project so it can be extracted as a library. No logic changes just updates to build and package structure.


build.gradle
```diff
@@ -79,6 +79,7 @@ dependencies {
     implementation project(':alert-common')
     implementation project(':alert-database')
     implementation project(':alert-issuetracker')
+    implementation project(':alert-issuetracker-jira')
```
bamandel: is the alert-issuetracker import still necessary if we're pulling in the jira version?

alert-issuetracker-jira/src/main/java/com/synopsys/integration/issuetracker/jira/cloud/JiraCloudContext.java
```diff
@@ -20,7 +20,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package com.synopsys.integration.alert.issuetracker.jira.cloud;
```
gkillough: If these are still in alert, should we keep that in the package name?

psantos1113: I'm looking to move the sub projects into their own repositories this week.  That's the reason for the package change there.

build.gradle
psantos1113: yes I should be able just include alert-issuetracker-jira and get the issuetracker project as a transitive.

PR #751, opened 2019-11-22
Fix role stored procedures

gkillough:


src/main/resources/db/changelog/alert/5.3.0/permissions-matrix.xml
```diff
@@ -241,6 +242,20 @@
             <column name="OPERATIONS" value="255"/>
         </insert>
 
+        <insert tableName="PERMISSION_MATRIX" schemaName="ALERT">
+            <column name="ROLE_ID" valueComputed="GET_ID_FOR_USER_ROLE('ALERT_ADMIN')"/>
```
bamandel: Is this supposed to be ALERT_JOB_MANAGER?

psantos1113: yes it should be we should fix it and merge this PR.

PR #752, opened 2019-11-25
Bm user role ui

bamandel:
Temporary UI for the new UserManagement tab. Need to create endpoints so that the buttons actually perform actions. Will also format the tables better in the future.


src/main/js/store/actions/users.js
```diff
@@ -0,0 +1,171 @@
...
+    USER_MANAGEMENT_USER_DELETED,
+    USER_MANAGEMENT_USER_DELETING,
+    USER_MANAGEMENT_USER_FETCH_ERROR_ALL,
+    USER_MANAGEMENT_USER_FETCHED_ALL,
+    USER_MANAGEMENT_USER_FETCHING_ALL,
+    USER_MANAGEMENT_USER_SAVE_ERROR,
+    USER_MANAGEMENT_USER_SAVED,
+    USER_MANAGEMENT_USER_SAVING
+} from 'store/actions/types'
+import * as ConfigRequestBuilder from "util/configurationRequestBuilder";
```
jamesrichard91: Please make sure your imports are single quotes

src/main/js/dynamic/loaded/users/UserManagement.js
```diff
@@ -0,0 +1,60 @@
...
+        super(props);
+
+        const descriptor = DescriptorUtilities.findDescriptorByNameAndContext(this.props.descriptors, DescriptorUtilities.DESCRIPTOR_NAME.COMPONENT_USERS, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)[0];
+
+        this.state = {
+            descriptor: descriptor
+        };
+    }
+
+    checkPermissions(operation) {
```
psantos1113: We do this in multiple places we should probably add a utility to check permissions and put it in the util directory like the FieldModel utilities.

src/main/js/dynamic/loaded/users/UserTable.js
```diff
@@ -0,0 +1,121 @@
...
+                header: 'username',
+                headerLabel: 'Username',
+                isKey: true
+            },
+            {
+                header: 'email_address',
+                headerLabel: 'Email',
+                isKey: false
+            },
+            {
```
psantos1113: Have to double check here but we may only want username and email address shown.  The expired, locked, and password_expired fields are there because that is the type of information the user models can support.  However in Alert right now we don't have any of those features to expire a user or their password to my knowledge.  We may need to implement that.

src/main/js/field/TableDisplay.js
```diff
@@ -0,0 +1,223 @@
...
+                onModalClose()
+            }}>
+                <Modal.Header closeButton>
+                    <Modal.Title>{this.props.modalTitle}</Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    <form className="form-horizontal" onSubmit={this.handleSubmit} noValidate>
+                        {this.props.children}
+                        <ConfigButtons
+                            cancelId="job-cancel"
```
psantos1113: I think the cancelId and showId need to change to be unique.  Looks like a copy paste error.

src/main/js/field/TableDisplay.js
```diff
@@ -0,0 +1,223 @@
...
+                if (isSelect) {
+                    return '#e8e8e8';
+                }
+                return null;
+            }
+        };
+
+        const content = (
+            <
+                div>
```
psantos1113: clean up the whitespace difference.

src/main/js/dynamic/loaded/users/UserManagement.js
```diff
@@ -0,0 +1,60 @@
...
+        const { descriptor } = this.state;
+        if (descriptor) {
+            return DescriptorUtilities.isOperationAssigned(descriptor, operation)
```
bamandel: This needs a semicolon

src/main/js/dynamic/loaded/users/UserTable.js
bamandel: What about the enabled field? Will this be hidden as well?

PR #757, opened 2019-12-03
Collapse TopLevelAction

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/combiner/AbstractMessageCombiner.java
```diff
@@ -0,0 +1,156 @@
...
+public class AbstractMessageCombiner implements MessageCombiner {
+    public final List<ProviderMessageContent> combine(List<ProviderMessageContent> messages) {
+        SetMap<ContentKey, ProviderMessageContent> messagesGroupedByKey = SetMap.createLinked();
+        messages.forEach(message -> messagesGroupedByKey.add(message.getContentKey(), message));
+
+        List<ProviderMessageContent> combinedMessages = new ArrayList<>();
+        for (Set<ProviderMessageContent> groupedMessages : messagesGroupedByKey.values()) {
+            combineGroupedMessages(groupedMessages)
+                .ifPresent(combinedMessages::add);
+        }
```
bamandel: Are we going through a bunch of sets we just went through and added to a collection? It looks like we can instead do this originally, right?

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/combiner/AbstractMessageCombiner.java
```diff
@@ -0,0 +1,156 @@
...
+            Set<LinkableItem> componentAttributes = componentItem.getComponentAttributes();
+            String key = componentItem.createKey(true, true);
+            ComponentItem oldItem = keyToItems.get(key);
+
+            Set<LinkableItem> combinedAttributes = new LinkedHashSet<>();
+            if (null != oldItem) {
+                combinedAttributes.addAll(oldItem.getComponentAttributes());
+                combinedAttributes.addAll(componentAttributes);
+            } else {
+                combinedAttributes.addAll(componentAttributes);
```
bamandel: this can be simplified. Make the null check and add the oldItems if not null. Then always add the rest of componentAttributes.

PR #759, opened 2019-12-04
Bm user endpoints

psantos1113:
Branch with the initial endpoints to create users and roles.


src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
```diff
@@ -0,0 +1,88 @@
...
+    private AuthorizationUtility authorizationUtility;
+    private UserAccessor userAccessor;
+
+    public UserActions(AuthorizationUtility authorizationUtility, UserAccessor userAccessor) {
+        this.authorizationUtility = authorizationUtility;
+        this.userAccessor = userAccessor;
+    }
+
+    public Collection<UserRoleModel> getUsers() {
+        return userAccessor.getUsers().stream()
```
gkillough: Does `userAccessor.getUsers()` include the roles?

psantos1113: Yes the UserModel returned by userAccessor.getUsers() contains the role names associated with the user.  We have two classes named UserRoleModel in different packages. That may cause confusion.  One is the definition of the Role and lives under the persistence package.  So it should be named RoleModel.

PR #767, opened 2019-12-10
Ps user role assignment

psantos1113:
Fix the assignment of roles for a newly created user.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
@@ -26,6 +26,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import javax.transaction.Transactional;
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
```diff
@@ -67,15 +71,23 @@ public UserActions(UserAccessor userAccessor) {
...
+        String emailAddress = userConfig.getEmailAddress();
+
         Map<String, String> fieldErrors = new HashMap<>();
         validateCreationUserName(fieldErrors, userName);
         validatePasswordLength(fieldErrors, password);
         validateRequiredField(FIELD_KEY_USER_MGMT_EMAILADDRESS, fieldErrors, emailAddress);
         if (!fieldErrors.isEmpty()) {
             throw new AlertFieldException(fieldErrors);
         }
+        Collection<UserRoleModel> roleNames = userConfig.getRoleNames().stream().map(UserRoleModel::of).collect(Collectors.toList());
```
bamandel: UserRoleModel::of will mark the roles as custom. Is that what we want?

PR #768, opened 2019-12-10
Bm edit copy

bamandel:


src/main/java/com/synopsys/integration/alert/web/user/RoleController.java
```diff
@@ -73,6 +74,20 @@ public RoleController(ContentConverter contentConverter, ResponseFactory respons
         return responseFactory.createOkContentResponse(contentConverter.getJsonString(roleActions.getRoles()));
     }
 
+    @PutMapping
+    public ResponseEntity<String> updateRole(@RequestBody RolePermissionModel rolePermissionModel) {
+        if (!hasPermission(authorizationManager::hasCreatePermission)) {
```
gkillough: Should this be checking `create`? Do we have another permission type?

src/main/js/util/configurationRequestBuilder.js
```diff
@@ -71,6 +71,19 @@ export function createUpdateRequest(apiUrl, csrfToken, configurationId, fieldMod
     });
 }
 
+export function createUpdateWithoutIdRequest(apiUrl, csrfToken, model) {
```
gkillough: Do we validate that the role name is unique on the back-end? If not, we should.

bamandel: We do :)

src/main/java/com/synopsys/integration/alert/web/user/RoleController.java
bamandel: Changed to write

PR #770, opened 2019-12-11
Bm fix test message

bamandel:
Test message response now displays descriptor name instead of key.

WARNING: There are a lot of finals removed in this PR since I had to touch a lot of files. The majority of changes are in the first few files.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannelKey.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannelKey.java
@@ -24,6 +24,7 @@ package com.synopsys.integration.alert.channel.jira.server;
 
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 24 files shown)


src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannelKey.java
```diff
@@ -32,4 +33,9 @@
     public String getUniversalKey() {
         return "channel_jira_server";
     }
+
+    @Override
+    public String getDisplayName() {
+        return JiraServerDescriptor.JIRA_LABEL;
```
psantos1113: There's no circular dependency here? The descriptors use the keys.  Why not move the label constants into the key here and then use that anywhere JiraServerDescriptor.JIRA_LABEL.

gkillough: If it's static, it doesn't belong to an instance of the class, so I think this should be ok.

PR #772, opened 2019-12-12
Ps user

psantos1113:
Use the user id instead of the user name to update the user.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
@@ -132,10 +132,16 @@ public class DefaultUserAccessor implements UserAccessor {
     public boolean changeUserPassword(String username, String newPassword) {
         Optional<UserEntity> entity = userRepository.findByUserName(username);
         if (entity.isPresent()) {
-            UserEntity oldEntity = entity.get();
-            UserEntity updatedEntity = new UserEntity(oldEntity.getUserName(), defaultPasswordEncoder.encode(newPassword), oldEntity.getEmailAddress());
-            updatedEntity.setId(oldEntity.getId());
-            return userRepository.save(updatedEntity) != null;
+            return changeUserPassword(entity.get(), newPassword);
+        }
+        return false;
+    }
+
+    @Override
+    public boolean changeUserPassword(Long userId, String newPassword) {
+        Optional<UserEntity> entity = userRepository.findById(userId);
+        if (entity.isPresent()) {
+            return changeUserPassword(entity.get(), newPassword);
         }
         return false;
     }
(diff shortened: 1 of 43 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
```diff
@@ -156,13 +150,36 @@ public boolean changeUserEmailAddress(String username, String emailAddress) {
...
+    private boolean changeUserEmailAddress(UserEntity oldEntity, String emailAddress) {
+        UserEntity updatedEntity = new UserEntity(oldEntity.getUserName(), oldEntity.getPassword(), emailAddress);
+        updatedEntity.setId(oldEntity.getId());
+        return userRepository.save(updatedEntity) != null;
+    }
+
+    private void deleteUser(UserEntity userEntity) throws AlertDatabaseConstraintException {
+        if (!RESERVED_USER_IDS.contains(userEntity.getId())) {
+            assignRoles(userEntity.getUserName(), Set.of());
+            userRepository.delete(userEntity);
```
bamandel: All of these methods can be used with an ID as opposed to Username. In the public delete method above, we use the ID to retrieve the entity. We could just use the ID given and never query the DB for the entire entity. If they have a large amount of users, this could be slow.

src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
```diff
@@ -98,61 +81,53 @@ public UserConfig createUser(UserConfig userConfig) throws AlertDatabaseConstrai
...
-        Set<String> configuredRoleNames = userConfig.getRoleNames();
-        if (null != configuredRoleNames && !configuredRoleNames.isEmpty() && userModel.isPresent()) {
-            Collection<UserRoleModel> roleNames = authorizationUtility.getRoles().stream()
-                                                      .filter(role -> configuredRoleNames.contains(role.getName()))
-                                                      .collect(Collectors.toList());
-            authorizationUtility.updateUserRoles(userModel.get().getId(), roleNames);
-        }
+            if (passwordHasValue) {
+                UserModel newUserModel = UserModel.existingUser(existingUser.getId(), userName, password, emailAddress, existingUser.getRoles());
+                userAccessor.updateUser(newUserModel, false);
```
bamandel: This object creation and user upating can instead be done outside of the if statement. We can use the if statement to update the password value. The passwordEncoded field will just be equal to negation of passwordHasValue

src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
```diff
@@ -98,61 +81,53 @@ public UserConfig createUser(UserConfig userConfig) throws AlertDatabaseConstrai
...
-        }
-        return userAccessor.changeUserEmailAddress(userName, emailAddress);
+    public void deleteUser(Long userId) throws AlertDatabaseConstraintException {
+        userAccessor.deleteUser(userId);
     }
 
     private UserConfig convertToCustomUserRoleModel(UserModel userModel) {
         // converting to an object to return to the client; remove the password field.
-        return new UserConfig(
+        return new UserConfig(userModel.getId().toString(),
```
bamandel: Can we move this param to the next line?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
psantos1113: In the exception we include the username so the entity is needed.  I will change the operations to use the user id.

PR #778, opened 2019-12-16
Fix the user table issues to display errors

psantos1113:
Display the validation issues for creating and updating users.  Also ensure the table modal displays data for the correct user when editing.


src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
```diff
@@ -155,12 +156,19 @@ private void validateRequiredField(String fieldKey, Map<String, String> fieldErr
...
     }
 
-    private void validateUserName(Map<String, String> fieldErrors, String userName) {
+    private void validateUserExistsByName(Map<String, String> fieldErrors, String userName) {
         validateRequiredField(FIELD_KEY_USER_MGMT_USERNAME, fieldErrors, userName);
         Optional<UserModel> userModel = userAccessor.getUser(userName);
         userModel.ifPresent(user -> fieldErrors.put(FIELD_KEY_USER_MGMT_USERNAME, "A user with that username already exists."));
     }
 
+    private void validateUserExistsById(Map<String, String> fieldErrors, Long userId, String userName) {
```
bamandel: This method is strange to me, JPA offers a method that checks if an ID exists in the table. You should be able to shorten this to that single line.

PR #782, opened 2019-12-16
No duplicate permissions allowed.

bamandel:


bamandel:
> Are we doing the validation on the front-end because of the multiple modals?

We're doing validation here as this modal doesn't actually update anything in the DB. This only modifies state. The other modals will update values in the DB and perform server validation.

PR #783, opened 2019-12-18
Ps validate provider

psantos1113:
Display an error icon in the table if the job has an error.  When saving job configurations validate the provider and channel global configuration if present.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -70,6 +79,9 @@ public ChannelDistributionUIConfig(ChannelKey channelKey, String label, String u
         ConfigField channelName = new EndpointSelectField(KEY_CHANNEL_NAME, LABEL_CHANNEL_NAME, DESCRIPTION_CHANNEL_NAME)
                                       .applyClearable(false)
                                       .applyRequired(true);
+        if (null != channelConfigValidator) {
+            channelName.applyValidationFunctions(channelConfigValidator);
```
gkillough: We might want to change this variable to `channelNameField` so it doesn't appear to be a string.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigController.java
```diff
@@ -224,6 +231,30 @@ public JobConfigController(JobConfigActions jobConfigActions, ResponseFactory re
         }
     }
 
+    @GetMapping("/validate")
+    public ResponseEntity<String> validateConfig() {
+        try {
+            List<PermissionKey> keys = new LinkedList<>();
+            for (Map.Entry<DescriptorKey, Descriptor> entry : descriptorMap.getDescriptorMap().entrySet()) {
```
bamandel: Since we're not really benefitting from the use of DescriptorMap here, you can change the passed variable to be a list of descriptors and then just loop through that. You can get the descriptorKey from the descriptor itself.

PR #788, opened 2019-12-19
Added a more detailed description to the Format field

bamandel:

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
@@ -63,7 +63,7 @@ public abstract class ProviderDistributionUIConfig extends UIConfig {
     private static final String LABEL_NOTIFICATION_TYPES = "Notification Types";
     private static final String LABEL_FORMAT = "Format";
     private static final String DESCRIPTION_NOTIFICATION_TYPES = "Select one or more of the notification types. Only these notification types will be included for this distribution job.";
-    private static final String DESCRIPTION_FORMAT = "Select the format of the message that will be created.";
+    private static final String DESCRIPTION_FORMAT = "Select the format of the message that will be created: ";
 
     private final ProviderContent providerContent;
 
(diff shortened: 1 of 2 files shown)


bamandel:
I would like feedback on the descriptions.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
```diff
@@ -117,9 +116,23 @@ private LabelValueSelectOption convertToLabelValueOption(ProviderNotificationTyp
...
+    }
+
+    private String createFormatDescription() {
+        StringBuilder formatDescription = new StringBuilder();
+        for (FormatType format : providerContent.getSupportedContentFormats()) {
+            String label = format.getLabel();
+            String description = format.getDescription();
+
+            formatDescription.append(label);
+            formatDescription.append(": ");
```
jamesrichard91: Isn't the ':' already there? The DESCRIPTION_FORMAT now ends in ':'

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/FormatType.java
```diff
@@ -23,8 +23,23 @@
 package com.synopsys.integration.alert.common.enumeration;
 
 public enum FormatType {
-    DEFAULT,
-    DIGEST,
-    SUMMARY
+    DEFAULT("Default", "The message will contain all the relevant data found in your selected provider."),
+    DIGEST("Digest", "The message will contain a delta of the content found in your selected provider since it was last queried."),
```
jamesrichard91: Can we change the Digest format to "The message will contain a delta of the content found in your selected provider since it was last queried. Add and Delete operations will cancel each other out depending on the order they occurred."

alert-common/src/main/java/com/synopsys/integration/alert/common/enumeration/FormatType.java
```diff
@@ -23,8 +23,23 @@
...
+    SUMMARY("Summary", "The message contains only a small and summarized overview of what was found in your selected provider.");
```
jamesrichard91: Can we change the Summary format to "The message contains only a summarized form of the Digest data."

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
bamandel: This colon is located after the format name whereas the other colon is after the generic format description before we describe each one.

PR #790, opened 2019-12-20
Ps external user roles

psantos1113:
Allow the assignment of internal roles to external users.
Change the UI to make the fields in user management readonly for external users.
Allow an external user to be deleted since only roles can be modified.  when the user logs in they will be re-added to Alert.

I will have a different PR to remove the external boolean to be a more descriptive if the user has logged in via SAML or LDAP.  This is the initial change to add internal roles to external users.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
@@ -118,14 +118,18 @@ public class DefaultUserAccessor implements UserAccessor {
         UserEntity existingUser = userRepository.findById(userId)
                                       .orElseThrow(() -> new AlertDatabaseConstraintException(String.format("No user found with id '%s'", userId)));
         Long existingUserId = existingUser.getId();
-
-        String password = passwordEncoded ? user.getPassword() : defaultPasswordEncoder.encode(user.getPassword());
-        UserEntity newEntity = new UserEntity(user.getName(), password, user.getEmailAddress(), user.isExpired(), user.isLocked(), user.isPasswordExpired(), user.isEnabled(), existingUser.isExternal());
-        newEntity.setId(existingUserId);
+        UserEntity savedEntity = existingUser;
+        // if it isn't an external user then update username, password, and email.
+        if (!existingUser.isExternal()) {
+            String password = passwordEncoded ? user.getPassword() : defaultPasswordEncoder.encode(user.getPassword());
+            UserEntity newEntity = new UserEntity(user.getName(), password, user.getEmailAddress(), user.isExpired(), user.isLocked(), user.isPasswordExpired(), user.isEnabled(), existingUser.isExternal());
+            newEntity.setId(existingUserId);
+            savedEntity = userRepository.save(newEntity);
+        }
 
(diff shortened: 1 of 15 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
```diff
@@ -118,14 +118,18 @@ public UserModel updateUser(UserModel user, boolean passwordEncoded) throws Aler
         UserEntity existingUser = userRepository.findById(userId)
                                       .orElseThrow(() -> new AlertDatabaseConstraintException(String.format("No user found with id '%s'", userId)));
         Long existingUserId = existingUser.getId();
-
-        String password = passwordEncoded ? user.getPassword() : defaultPasswordEncoder.encode(user.getPassword());
-        UserEntity newEntity = new UserEntity(user.getName(), password, user.getEmailAddress(), user.isExpired(), user.isLocked(), user.isPasswordExpired(), user.isEnabled(), existingUser.isExternal());
-        newEntity.setId(existingUserId);
+        UserEntity savedEntity = existingUser;
+        // if it isn't an external user then update username, password, and email.
```
bamandel: Should we throw an exception if it is an external user and they try to update any of those?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
```diff
@@ -118,14 +118,18 @@ public UserModel updateUser(UserModel user, boolean passwordEncoded) throws Aler
...
+        if (!existingUser.isExternal()) {
```
gkillough: Should we expose something to the user to let them know that the username, password, and email were not updated?

src/main/java/com/synopsys/integration/alert/web/security/authentication/UserManagementAuthoritiesPopulator.java
```diff
@@ -107,17 +116,31 @@ public String getSAMLRoleAttributeName(String defaultName) {
...
+        Optional<UserModel> userModel = userAccessor.getUser(userName);
+        Set<String> roleNames = userModel.map(UserModel::getRoleNames).orElse(Set.of());
+        Set<String> newRoleNames = new LinkedHashSet<>(roleNames.size());
+        for (String roleName : roleNames) {
+            newRoleNames.add(function.apply(roleName));
+        }
+        return newRoleNames;
+    }
+
+    @Deprecated
```
bamandel: Do we need to deprecate private methods? I imagine that we just remove these methods when they're no longer used.

src/main/js/dynamic/loaded/users/UserTable.js
```diff
@@ -7,6 +7,7 @@ import { connect } from 'react-redux';
 import { clearUserFieldErrors, createNewUser, deleteUser, fetchUsers, updateUser } from 'store/actions/users';
 import DynamicSelectInput from 'field/input/DynamicSelect';
 import { fetchRoles } from 'store/actions/roles';
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
```
bamandel: quotes should be single quotes

PR #792, opened 2020-01-07
Ps auth type

psantos1113:
Add the authentication type as a string to the user table.  DATABASE, LDAP, or SAML are the current values.  Display the authentication type in the UI.


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultUserAccessor.java
```diff
@@ -121,7 +113,7 @@ public UserModel updateUser(UserModel user, boolean passwordEncoded) throws Aler
         Long existingUserId = existingUser.getId();
         UserEntity savedEntity = existingUser;
         // if it isn't an external user then update username, password, and email.
-        if (existingUser.isExternal()) {
+        if (!existingUser.getAuthenticationType().equals("DATABASE")) {
```
jamesrichard91: We may want to change this to `AuthenticationType.DATABASE.name().equals(existingUser.getAuthenticationType())`. This way if we change DATABASE we only have to change it in one place.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/AuthenticationType.java
```diff
@@ -0,0 +1,46 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.persistence.model;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class AuthenticationType extends AlertSerializableModel {
```
gkillough: Why not make this an enum and then let the accessor worry about converting to/from Long?

psantos1113: Use an enum now to handle the mapping of the database id to the enumeration.  The accessor uses methods on the enum to be able to get the correct type.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuthenticationTypeAccessor.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Optional<AuthenticationTypeDetails> getAuthenticationTypeDetails(AuthenticationType authenticationType) {
+        Optional<AuthenticationTypeEntity> authenticationTypeEntity = authenticationTypeRepository.findById(authenticationType.getId());
+        return authenticationTypeEntity.map(entity -> new AuthenticationTypeDetails(entity.getId(), entity.getName()));
+    }
+
+    @Override
+    public AuthenticationType getAuthenticationType(Long id) {
+        return AuthenticationType.getById(id);
```
gkillough: Should we allow this to throw unchecked exceptions if the id is invalid?

gkillough: Actually, it looks like this reads from a map, so no exception, but it can return null. We might want to wrap it in an optional.

PR #793, opened 2020-01-07
Adding the providerUrl to the issue property

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/IssueHandler.java
```diff
@@ -172,11 +172,20 @@ protected abstract T createIssue(IssueConfig issueConfig, String providerName, L
         LinkableItem topic = messageContent.getTopic();
         LinkableItem nullableSubTopic = messageContent.getSubTopic().orElse(null);
 
+        String providerUrl = messageContent.getProvider().getUrl()
+                                 .map(url -> {
```
gkillough: Can we make this its own function?

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/IssueHandler.java
```diff
@@ -172,11 +172,20 @@ protected abstract T createIssue(IssueConfig issueConfig, String providerName, L
         LinkableItem topic = messageContent.getTopic();
         LinkableItem nullableSubTopic = messageContent.getSubTopic().orElse(null);
 
+        String providerUrl = messageContent.getProvider().getUrl()
```
gkillough: We should make sure that the URL is actually being set during processing.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/IssueHandler.java
```diff
@@ -172,15 +172,30 @@ protected abstract T createIssue(IssueConfig issueConfig, String providerName, L
...
+                                 .map(url -> formatUrl(url))
```
gkillough: This can now be a method reference. `this:: formatUrl`

src/main/java/com/synopsys/integration/alert/channel/jira/common/model/AlertJiraIssueProperties.java
```diff
@@ -26,6 +26,7 @@
 
 public class AlertJiraIssueProperties extends AlertSerializableModel {
     private String provider;
+    private String providerUrl;
```
psantos1113: When merging this you will need to update the isuetracker-jira library to include the provider URL as well since this is now an issue property.  The class will be the JiraIssueSearchProperties class. Also will this need to be added to the Jira Plugin that includes the feature to migrate issues.  Add the Black Duck Server URL as the provider URL in the Jira Plugin migration issue.

PR #795, opened 2020-01-07
Bm saving spinner

bamandel:


src/main/js/dynamic/loaded/users/RoleTable.js
```diff
@@ -121,8 +121,14 @@ class RoleTable extends Component {
     }
 
     savePermissions(permission) {
-        const { role } = this.state;
+        const { role, incrementalId } = this.state;
         const { permissions } = role;
+        if (!permission.id) {
+            permission.id = incrementalId;
+            this.setState({
+                incrementalId: incrementalId + 1
```
jamesrichard91: What is the purpose of the incrementalId?

bamandel: The permissions are stored in a table in state. Because of this, there is no id associated with any of the rows. In order to edit a row and know which one needs to be modified, we need an ID. This incrementalId is used to track rows in state to know what rows are being modified.

PR #799, opened 2020-01-09
Ps settings fix

psantos1113:
Migrate the old ALERT_COMPONENT_SETTINGS_ environment variables for authentication to the new authentication component configuration in the database.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/workflow/startup/component/AuthenticationSettingsMigration.java
+++ b/src/main/java/com/synopsys/integration/alert/workflow/startup/component/AuthenticationSettingsMigration.java
@@ -1,0 +1,39 @@
+/**
+ * blackduck-alert
+ *
+ * Copyright (c) 2020 Synopsys, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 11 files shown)


src/main/java/com/synopsys/integration/alert/workflow/startup/component/AuthenticationSettingsMigration.java
```diff
@@ -0,0 +1,161 @@
...
+            List<ConfigurationModel> foundConfigurationModels = fieldConfigurationAccessor.getConfigurationByDescriptorKeyAndContext(authenticationDescriptorKey, ConfigContextEnum.GLOBAL);
+
+            Map<String, ConfigurationFieldModel> existingConfiguredFields = new HashMap<>();
+            foundConfigurationModels.forEach(config -> existingConfiguredFields.putAll(config.getCopyOfKeyToFieldMap()));
+
+            Set<ConfigurationFieldModel> configurationModels = createFieldModelsFromDefinedFields(fieldsForDescriptor, existingConfiguredFields);
+            boolean overwriteConfig = settingsUtility.getConfiguration()
+                                          .flatMap(configurationModel -> configurationModel.getField(SettingsDescriptor.KEY_STARTUP_ENVIRONMENT_VARIABLE_OVERRIDE))
+                                          .flatMap(ConfigurationFieldModel::getFieldValue)
+                                          .map(value -> Boolean.valueOf(value)).orElse(Boolean.FALSE);
```
bamandel: This can be replaced with a method reference. Also, I generally prefer the orElse on it's own line.

src/main/java/com/synopsys/integration/alert/workflow/startup/component/ConfigurationLogger.java
```diff
@@ -35,27 +35,27 @@
 import com.synopsys.integration.alert.common.rest.ProxyManager;
 
 @Component
-@Order(4)
+@Order(5)
```
bamandel: Perhaps we should make our standard steps increments of 10? That way we can fit temporary/new steps in between existing without re-ordering everything every time

psantos1113: I will implement that change in 5.3.0 as an enhancement when this change is merged.

PR #801, opened 2020-01-10
Ps startup range

psantos1113:
In case we need to add items to the startup order the main existing components are set in increments of 10.


src/main/java/com/synopsys/integration/alert/workflow/startup/component/AlertStartupInitializer.java
```diff
@@ -56,7 +56,7 @@
 import com.synopsys.integration.alert.web.config.FieldModelProcessor;
 
 @Component
-@Order(1)
+@Order(10)
```
gkillough: Will things with order always get initialized before things without order?

psantos1113: Components without the Order annotation will be assigned the lowest priority and get initialized after the components with the Order annotation.
From the Spring docs:
https://docs.spring.io/spring/docs/5.0.1.RELEASE/javadoc-api/org/springframework/core/OrderComparator.html

PR #807, opened 2020-01-14
Fix(jira): Fix the description of the user name.

psantos1113:
The description of the user name field mentions an email which is not valid.  The Jira server has it's own users so you need to use the username of the user not the email.  Clarify in the description.


gkillough:
Why was this email previously? Is if because the admin user needs to provide their email?

src/main/java/com/synopsys/integration/alert/channel/jira/server/descriptor/JiraServerDistributionUIConfig.java
```diff
@@ -43,7 +43,7 @@
     private static final String LABEL_RESOLVE_WORKFLOW_TRANSITION = "Resolve Transition";
     private static final String LABEL_OPEN_WORKFLOW_TRANSITION = "Re-open Transition";
     private static final String DESCRIPTION_ADD_COMMENTS = "If true, this will add comments to the Jira ticket with data describing the latest change.";
-    private static final String DESCRIPTION_ISSUE_CREATOR = "The email of the Jira Server user to assign as the issue creator field of the Jira issue.";
+    private static final String DESCRIPTION_ISSUE_CREATOR = "The username of the Jira Server user to assign as the issue creator field of the Jira issue.";
```
psantos1113: This was email before due to a copy paste error from the Jira Cloud implementation. Changed to a single word.

PR #813, opened 2020-01-15
Ps sonar fixes

psantos1113:
Fixes for Sonar Cloud issues.  Creating pull request for a build confirmation that all the issues are fixed.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
@@ -26,8 +26,8 @@ import java.util.Optional;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.CompareToBuilder;
-import org.apache.commons.lang3.builder.RecursiveToStringStyle;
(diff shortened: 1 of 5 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/LinkableItem.java
```diff
@@ -98,7 +98,7 @@ public int compareTo(final LinkableItem otherItem) {
 
     @Override
     public String toString() {
-        ReflectionToStringBuilder builder = new ReflectionToStringBuilder(this, RecursiveToStringStyle.JSON_STYLE);
+        ReflectionToStringBuilder builder = new ReflectionToStringBuilder(this, ToStringStyle.JSON_STYLE.JSON_STYLE);
```
gkillough: I'm pretty sure we use the RecursiveToStringStyle in our libraries. What was this issue with it?

psantos1113: The sonar cloud issue was that we were using a child class to reference a static in a parent class.  The fix is to use the parent class in the static reference.

PR #818, opened 2020-01-20
Certificate Storage

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessor.java
```diff
@@ -0,0 +1,97 @@
...
+        if (StringUtils.isBlank(alias)) {
+            throw new AlertDatabaseConstraintException("The field 'alias' cannot be blank");
+        }
+
+        String certificateContent = certificateModel.getCertificateContent();
+        if (StringUtils.isBlank(certificateContent)) {
+            throw new AlertDatabaseConstraintException("The field 'certificateContent' cannot be blank");
+        }
+
+        CustomCertificateEntity entityToSave = new CustomCertificateEntity(alias, certificateContent);
```
jamesrichard91: I'm not sure that we want to store the certificateContent in our DB. What would the use be? It should just be stored in the Java truststore

psantos1113: Currently the certificates get regenerated each time the container is started.  So we would need to have a startup component to import the certificates stored in the database into the truststore.  We could look into saving the certificate into the volume but if the cfssl container gets updated and generates new certificates with different attributes then the Alert container may not be able to communicate with the other containers in a deployment with Black Duck.

PR #822, opened 2020-01-22
Ps cert rest api

psantos1113:
Initial changes to provide a Rest API for certificate management.


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/CustomCertificateAccessor.java
```diff
@@ -23,13 +23,16 @@
...
 import java.util.List;
+import java.util.Optional;
 
 import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
 import com.synopsys.integration.alert.common.persistence.model.CustomCertificateModel;
 
 public interface CustomCertificateAccessor {
     List<CustomCertificateModel> getAll();
 
+    Optional<CustomCertificateModel> getCertificate(Long id);
```
gkillough: I don't know why I didn't notice this earlier, but this class has a wide variety of naming styles used for methods. Would you mind making all the methods use the full word 'certificate', including the 'getAll'?

psantos1113: Will do no worries.

PR #825, opened 2020-01-23
Fix(messages): Fix Sonar cloud issues with statics.

psantos1113:
Use the classes that define the static constants rather than a child class to fix sonar cloud issues.


psantos1113:
The issue will be addressed in the API generator.  The API generator should encapsulate the static links into the unversioned view object.  Because it's been decided to address the issue in the API generator we are not going to fix this in Alert for the time being.

PR #826, opened 2020-01-23
Ps certificate tests

psantos1113:
Added some tests for the CertificateActions class.  Also includes fixes from manually testing.


src/main/java/com/synopsys/integration/alert/web/certificates/CertificateActions.java
```diff
@@ -76,9 +76,15 @@ public CertificateModel createCertificate(CertificateModel certificateModel) thr
...
-        CustomCertificateModel storedCertificate = certificateAccessor.storeCertificate(certificateToStore);
-        certificateUtility.importCertificate(storedCertificate);
-        return convertFromDatabaseModel(storedCertificate);
+        try {
+            CustomCertificateModel storedCertificate = certificateAccessor.storeCertificate(certificateToStore);
+            certificateUtility.importCertificate(storedCertificate);
+            return convertFromDatabaseModel(storedCertificate);
+        } catch (AlertException ex) {
+            certificateAccessor.deleteCertificate(certificateToStore.getAlias());
+            certificateUtility.removeCertificate(certificateToStore.getAlias());
```
gkillough: This line may also throw an exception and I'm not sure how we want to handle it.

PR #828, opened 2020-01-24
Ps cert controller tests

psantos1113:
Added Permission checks for the a user to have global access to use the controller.  Added an integration test for the controller.

To define permissions I needed to add a new descriptor which will be used in 6.0.0 when we have a UI component to manage certificates.


src/main/java/com/synopsys/integration/alert/web/certificates/CertificatesController.java
```diff
@@ -49,24 +54,34 @@
...
+        this.descriptorKey = descriptorKey;
+        this.authorizationManager = authorizationManager;
         this.responseFactory = responseFactory;
         this.contentConverter = contentConverter;
         this.actions = actions;
     }
 
     @GetMapping
     public ResponseEntity<String> readCertificates() {
+        if (!hasPermission(authorizationManager::hasDeletePermission)) {
```
gkillough: Why is this checking `delete` permissions?

psantos1113: fixed this

PR #829, opened 2020-01-27
Update black duck response cache with api object

crowleySynopsys:
# Pull Request template

* Noticed TODO in BlackDuckResponseCache to replace Json object with API object

* Updating cvss3Severity to its new (and very wordy) equivalent


alert-common/build.gradle
```diff
@@ -2,7 +2,7 @@ ext.moduleName = 'com.synopsys.integration.alert.common'
 
 dependencies {
     api 'com.synopsys.integration:issuetracker-jira:0.0.3'
-    api 'com.blackducksoftware.integration:blackduck-common:45.0.6'
+    api 'com.synopsys.integration:blackduck-common:45.1.1-SNAPSHOT'
```
bamandel: Do we want a snapshot for now?

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
```diff
@@ -145,19 +146,9 @@ public String getSeverity(String vulnerabilityUrl) {
...
-                JsonElement cvss3Severity = cvss3.getAsJsonObject().get("severity");
-                if (null != cvss3Severity) {
-                    return Optional.of(cvss3Severity.getAsString());
-                }
-            }
-        }
-        return Optional.empty();
+    private Optional<ProjectVersionVulnerableBomComponentsItemsVulnerabilityWithRemediationSeverityType> getCvss3Severity(VulnerabilityView vulnerabilityView) {
+        ProjectVersionVulnerableBomComponentsItemsVulnerabilityWithRemediationSeverityType severity = vulnerabilityView.getCvss3().getSeverity();
+        return Optional.of(severity);
```
jamesrichard91: Can you change this to `Optional.ofNullable(severity)` ? Just in case the severity is null

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
```diff
@@ -105,14 +106,14 @@ public String getProjectComponentQueryLink(String projectLink, String componentN
         return Optional.empty();
     }
 
-    public String getSeverity(String vulnerabilityUrl) {
-        String severity = "UNKNOWN";
+    public ProjectVersionVulnerableBomComponentsItemsVulnerabilityWithRemediationSeverityType getSeverity(String vulnerabilityUrl) {
```
jamesrichard91: Why did we change the return type from String to ProjectVersionVulnerableBomComponentsItemsVulnerabilityWithRemediationSeverityType?  Instead of the users of the method calling the name() method on the response why dont we do that at the end of this method?

alert-common/build.gradle
gkillough: I'm thinking no.

crowleySynopsys: oops that's my bad I missed that one.  Will update

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
crowleySynopsys: certainly

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/VulnerabilityNotificationMessageBuilder.java
```diff
@@ -151,9 +152,9 @@ public String getNotificationType() {
 
             LinkableItem vulnIdItem = new LinkableItem(MessageBuilderConstants.LABEL_VULNERABILITIES, vulnerabilityId, vulnerabilityUrl);
             // TODO in Black Duck 2019.10.0, the VulnerabilitySourceQualifiedId includes the severity so we don't need to make a request to get it
-            String notificationSeverity = blackDuckResponseCache.getSeverity(vulnerabilityUrl);
+            ProjectVersionVulnerableBomComponentsItemsVulnerabilityWithRemediationSeverityType notificationSeverity = blackDuckResponseCache.getSeverity(vulnerabilityUrl);
             if (vulnerabilityFilters.isEmpty() || vulnerabilityFilters.contains(notificationSeverity)) {
-                LinkableItem severity = new LinkableItem(MessageBuilderConstants.LABEL_VULNERABILITY_SEVERITY, notificationSeverity);
+                LinkableItem severity = new LinkableItem(MessageBuilderConstants.LABEL_VULNERABILITY_SEVERITY, notificationSeverity.name());
```
gkillough: Based on the `TODO`, it looks like the item the loop is on (`VulnerabilitySourceQualifiedId vulnerabilityItem`) should now have the severity. Can we replace the line getting it from the `blackDuckResponseCache`? If so, let's remove the `TODO`.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
gkillough: We should probably also check that `getCvss3()` is not null. Alternatively we can keep looking at the `useCvss3` flag in the object.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
crowleySynopsys: good question, I'll change that

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/VulnerabilityNotificationMessageBuilder.java
crowleySynopsys: the object is manually maintained in the API, and does not have a severity field.  do we know who made the TODO/how they heard that this change would happen in 2019.10.0?

gkillough: We would need to search the issue tracker on that version. I'll discuss more offline.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
```diff
@@ -145,17 +148,10 @@ public String getSeverity(String vulnerabilityUrl) {
     }
 
     // TODO update this code with an Object from blackduck-common-api when available
```
gkillough: We can remove this TODO now.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
```diff
@@ -121,7 +124,7 @@ public String getSeverity(String vulnerabilityUrl) {
             logger.debug("Error fetching vulnerability view", e);
         }
 
-        return severity;
+        return severity.name();
```
gkillough: Could this throw an NPE?

crowleySynopsys: good call, if vulnerabilityView is an empty optional then it would throw an NPE.  how do you guys feel about returning inline conditionals?  or would you rather getSeverity return an Optional?

PR #831, opened 2020-01-27
Fix(tests): Fix controller tests and add more.

psantos1113:
All the tests were issuing get requests.  Updated the tests to issue the correct controller REST calls.

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/web/certificates/CertificateActionsTestIT.java
+++ b/src/test/java/com/synopsys/integration/alert/web/certificates/CertificateActionsTestIT.java
@@ -1,5 +1,6 @@
 package com.synopsys.integration.alert.web.certificates;
 
+import static com.synopsys.integration.alert.web.certificates.CertificateTestUtil.TEST_ALIAS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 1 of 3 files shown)


src/test/java/com/synopsys/integration/alert/web/certificates/CertificateActionsTestIT.java
```diff
@@ -29,6 +31,19 @@ public void createCertificateTest() throws Exception {
...
+    @Test
+    public void createCertificateIdTest() throws Exception {
+        String certificateContent = readCertificateContents();
+        try {
+            CertificateModel certificate = new CertificateModel("alias", certificateContent);
+            certificate.setId("badId");
+            certificateActions.createCertificate(certificate);
+            fail();
+        } catch (AlertDatabaseConstraintException ex) {
+            assertTrue(ex.getMessage().contains("id cannot be present"));
```
gkillough: I used to be in favor of checking the substrings in messages, but now I worry that it causes us to need to update the tests too often, which, in turn, causes the tests to be less valuable. I would be more comfortable having the fail at line 41 and nothing in the catch block (maybe an assertNotNull).

psantos1113: I have changed to an assertNotNull for the message.

PR #833, opened 2020-01-30
Update the environment variables for LDAP and SAML to have the correct prefix

gkillough:


jamesrichard91:
Do we want this merged into master or 5.3.0?

PR #835, opened 2020-01-30
Use the correct stored procedure to initialize the certificates components

gkillough:


src/main/resources/db/changelog/alert/5.3.0/custom-certificates.xml
```diff
@@ -14,9 +14,13 @@
         <sql dbms="h2" stripComments="true">
             CALL REGISTER_COMPONENT('component_certificates');
```
gkillough: This gives default permissions to roles (ALERT_ADMIN: 255), so we don't have to do anything for ADMIN in the block below:

PR #839, opened 2020-02-03
Fix(SQL): Add 'if not exists' to custom_certificates table creation command

gkillough:


gkillough:
I created this PR more for the sake of recording the change than for an actual review itself.

PR #844, opened 2020-02-07
Fix vulnerability severity by only using cvss3 when applicable

gkillough:


src/main/java/com/synopsys/integration/alert/web/audit/AuditEntryActions.java
```diff
@@ -117,14 +117,15 @@ public AuditEntryActions(AuditUtility auditUtility, NotificationManager notifica
         if (distributionEvents.isEmpty()) {
             logger.warn("This notification could not be sent. Make sure you have a Distribution Job configured to handle this notification.");
         }
-        distributionEvents.forEach(event -> {
```
gkillough: We should never use `forEach` unless it's for 1) a method reference 2) at the end of a series of stream operations.

PR #845, opened 2020-02-07
Fix Jira Cloud formatting

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -158,19 +158,17 @@ protected String createLinkableItemString(LinkableItem linkableItem, boolean bol
...
-            formattedString = String.format("%s: %s", name, value);
+        if (bold) {
+            name = emphasize(name);
+            value = emphasize(value);
         }
 
-        if (bold) {
-            formattedString = emphasize(formattedString);
+        if (optionalUrl.isPresent()) {
+            // The nuance around stylizing links adds too much complexity for too little value to worry about emphasizing the it.
```
jamesrichard91: Is the end of the comment right? shouldn't it be "emphasizing it" not "emphasizing the it"?

gkillough: Haha, changed it from `value` to `it` because I didn't want to be redundant, but of course a forgot the `the`.

PR #846, opened 2020-02-07
fix: Update SAML settings after database save and log exceptions.

psantos1113:
This allows the SAML to be disabled because the database settings are saved first before the runtime changes to the SAML.  If there's an exception log it which will catch PKIX exceptions that prevent SAML from being disabled because it isn't a caught exception and it bubbles up.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/component/settings/actions/SettingsGlobalApiAction.java
+++ b/src/main/java/com/synopsys/integration/alert/component/settings/actions/SettingsGlobalApiAction.java
@@ -36,6 +36,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.synopsys.integration.alert.common.action.ApiAction;
+import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.persistence.model.UserModel;
 import com.synopsys.integration.alert.common.rest.model.FieldModel;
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/alert/component/settings/actions/SettingsGlobalApiAction.java
```diff
@@ -75,96 +76,111 @@ public FieldModel afterGetAction(final FieldModel fieldModel) {
...
-        final FieldModel modelToSave = new FieldModel(fieldModel.getDescriptorName(), fieldModel.getContext(), fields);
+        FieldModel modelToSave = new FieldModel(fieldModel.getDescriptorName(), fieldModel.getContext(), fields);
         modelToSave.setId(fieldModel.getId());
         return modelToSave;
     }
 
-    private FieldModel handleNewAndUpdatedConfig(final FieldModel fieldModel) {
+    @Override
+    public FieldModel afterSaveAction(FieldModel fieldModel) throws AlertException {
+        addSAMLMetadata(fieldModel);
```
jamesrichard91: Do we need to call `addSAMLMetadata(fieldModel);` in both the afterSaveAction and in the afterUpdateAction?

psantos1113: yes the save is for a POST and update is for the PUT rest call.

PR #848, opened 2020-02-10
Ensure descriptors with no global config do not fail global config validation

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/GlobalConfigExistsValidator.java
```diff
@@ -42,23 +44,33 @@
...
-        String descriptorDisplayName = descriptorKeys.stream()
-                                           .filter(descriptorKey -> descriptorKey.getUniversalKey().equals(descriptorName))
-                                           .map(descriptorKey -> descriptorKey.getDisplayName())
-                                           .findFirst()
-                                           .orElse(descriptorName);
+
+        Optional<DescriptorKey> optionalDescriptorKey = descriptors
+                                                            .stream()
+                                                            .filter(desc -> desc.getDescriptorKey().getUniversalKey().equals(descriptorName))
+                                                            .filter(this::hasGlobalConfig)
```
gkillough: Validating global configs is more expensive than comparing the descriptorName which is why the map happens later.

PR #849, opened 2020-02-11
Add provider config id to notifications

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/ProviderDescriptor.java
```diff
@@ -27,6 +27,10 @@
 import com.synopsys.integration.alert.common.provider.ProviderKey;
 
 public abstract class ProviderDescriptor extends Descriptor {
+    // FIXME these must be registered with all present and future providers
```
jamesrichard91: We have stored procedures for registering channels that do this type of thing by default. We should create similar procedures for registering a Provider.

gkillough: That will likely happen after we solve our test/runServer problem with postgres.

PR #855, opened 2020-02-12
refactor(phoneHome): Create a handler interface for provider specific…

psantos1113:
Create a class to update a phone home request builder with provider specific data.  
Create a class to lookup handlers by provider.
Update Phone home to iterate over the providers and their configurations to phone home data.


src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeHandlerMap.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
+import com.synopsys.integration.alert.common.provider.Provider;
+import com.synopsys.integration.alert.common.provider.ProviderKey;
+import com.synopsys.integration.alert.common.provider.ProviderPhoneHomeHandler;
+import com.synopsys.integration.phonehome.PhoneHomeRequestBody;
+
+@Component
+public class PhoneHomeHandlerMap {
```
gkillough: This class seems unnecessary. Why not just autowire a `List< ProviderPhoneHomeHandler>` in the PhoneHomeTask and then loop over that. You can still look up the provider configs because `ProviderPhoneHomeHandler` has a `ProviderKey`.

psantos1113: This is just for a more efficient way to look up the provider handler.  We would have to have another for loop or a filter stream to find the handler for each provider.  With this it would be a constant time lookup of the provider key.  I can simply put in a stream to filter the handlers based on key.

PR #857, opened 2020-02-13
Startup System Validation

psantos1113:
Update the System Validation of the providers on Startup of Alert.  

Fix the SystemValidator tests and make them all unit tests. 
Fix the accumulator tests due to a null pointer exception.


src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckValidator.java
```diff
@@ -37,27 +37,32 @@
...
 import com.synopsys.integration.alert.common.exception.AlertRuntimeException;
 import com.synopsys.integration.alert.common.persistence.accessor.SystemMessageUtility;
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
 import com.synopsys.integration.alert.common.provider.ProviderValidator;
 import com.synopsys.integration.blackduck.configuration.BlackDuckServerConfig;
 import com.synopsys.integration.exception.IntegrationException;
 import com.synopsys.integration.log.IntLogger;
 import com.synopsys.integration.log.Slf4jIntLogger;
 
 @Component
```
gkillough: Rather than this being a component, we could demand it as part of a Provider's contract by having a `validate()` method on the Provider itself. Then it could choose where/ how to implement it.

The Provider could pass its properties and a configurationModel to a different class if necessary. That way, it is guaranteed that a new Provider would have a validator.

PR #858, opened 2020-02-14
Add provider config id to data tables

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -58,56 +54,52 @@
...
+        try {
+            Optional<Long> optionalConfigId = configurationAccessor.getProviderConfigurationByName(providerConfigName)
+                                                  .map(ConfigurationModel::getConfigurationId);
+            if (optionalConfigId.isPresent()) {
+                return providerProjectRepository.findByProviderConfigId(optionalConfigId.get())
+                           .stream()
+                           .map(this::convertToProjectModel)
+                           .collect(Collectors.toList());
+            }
+        } catch (AlertDatabaseConstraintException ignored) {
```
jamesrichard91: Why are we ignoring this exception?

gkillough: If the configId is present, then the database constraint is fulfilled, meaning that specific exception will not be thrown. Because this method is more of a "search for provider configs by name", maybe it would be better to return an `Optional` than a `List`.

gkillough: Not to mention that because this is a search for projects, we don't necessarily want to throw an exception, but instead return an empty `List`.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -130,78 +122,51 @@ public void deleteProjects(ProviderKey providerKey, Collection<ProviderProject>
...
+    public List<ProviderUserModel> getAllUsers(Long providerConfigId) {
+        return providerUserRepository.findByProviderConfigId(providerConfigId)
                    .stream()
                    .map(this::convertToUserModel)
                    .collect(Collectors.toList());
     }
 
     @Override
-    @Transactional(readOnly = true, isolation = Isolation.READ_COMMITTED)
-    public AlertPagedModel<ProviderUserModel> getPageOfUsers(String providerName, Integer pageNumber, Integer pageSize, String q) throws AlertDatabaseConstraintException {
```
gkillough: This was never used in production.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -58,56 +54,52 @@
...
+    public List<ProviderProject> findByProviderConfigId(Long providerConfigId) {
+        return providerProjectRepository.findByProviderConfigId(providerConfigId)
                    .stream()
                    .map(this::convertToProjectModel)
                    .collect(Collectors.toList());
     }
 
     @Override
-    public void deleteProjects(ProviderKey providerKey, Collection<ProviderProject> providerProjects) {
+    public void deleteProjects(Collection<ProviderProject> providerProjects) {
```
psantos1113: Are we missing `@Transactional` on this method?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
jamesrichard91: You may want to add a comment to avoid future confusion. But given your explanation this seems fine

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
gkillough: The class is annotated with `@Transactional`, so this method will be. The read committed methods just override the class level annotation.

PR #862, opened 2020-02-19
Fix after save and task names

gkillough:


src/main/java/com/synopsys/integration/alert/web/config/FieldModelProcessor.java
```diff
@@ -50,122 +50,122 @@
...
 
-    public void performAfterDeleteAction(final String descriptorName, final String context) throws AlertException {
-        final Optional<ApiAction> optionalApiAction = descriptorProcessor.retrieveApiAction(descriptorName, context);
+    public void performAfterDeleteAction(FieldModel fieldModel) throws AlertException {
+        Optional<ApiAction> optionalApiAction = descriptorProcessor.retrieveApiAction(fieldModel.getDescriptorName(), fieldModel.getContext());
         if (optionalApiAction.isPresent()) {
-            final ApiAction apiAction = optionalApiAction.get();
-            apiAction.afterDeleteAction(descriptorName, context);
+            ApiAction apiAction = optionalApiAction.get();
+            apiAction.afterDeleteAction(fieldModel);
```
gkillough: Actual change.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
```diff
@@ -46,36 +46,41 @@
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final TaskScheduler taskScheduler;
-    private final String taskName;
     private ScheduledFuture<?> future;
 
-    public ScheduledTask(final TaskScheduler taskScheduler, final String taskName) {
+    public static String computeTaskName(Class<? extends ScheduledTask> clazz) {
```
psantos1113: Is this used by any other class should this be private

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
```diff
@@ -46,36 +46,41 @@
...
+
+    public ScheduledTask(TaskScheduler taskScheduler) {
         this.taskScheduler = taskScheduler;
-        this.taskName = taskName;
     }
 
-    public String getTaskName() {
-        return taskName;
+    public String computeTaskName() {
+        return ScheduledTask.computeTaskName(getClass());
```
psantos1113: Couldn't we keep getTaskName() and in the constructor of Scheduled task do the following:
`taskName = computeTaskName();`

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
gkillough: It is used in the static method computeProviderTaskName() which is currently only called by the SchedulingGlobalApiAction.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
gkillough: We can do this for ScheduledTask, but not for ProviderTask which extends it. This is because ProviderTask would need ProviderProperties passed to its constructor which is currently part of a different ticket.

PR #863, opened 2020-02-20
Change task name to use ID instead of config name

psantos1113:


alert-common/src/main/java/com/synopsys/integration/alert/common/provider/lifecycle/ProviderTask.java
```diff
@@ -34,11 +34,11 @@
     private ProviderProperties providerProperties;
 
     // TODO reevaluate if this is needed once scheduling supports multiple providers
-    public static String computeProviderTaskName(ProviderKey providerKey, String providerConfigName, Class<? extends ProviderTask> providerTaskClass) {
+    public static String computeProviderTaskName(ProviderKey providerKey, Long providerConfigId, Class<? extends ProviderTask> providerTaskClass) {
         String superTaskName = ScheduledTask.computeTaskName(providerTaskClass);
         String providerUniversalKey = providerKey.getUniversalKey();
 
-        return String.format("%s::Provider[%s]::Configuration[%s]", superTaskName, providerUniversalKey, providerConfigName);
+        return String.format("%s::Provider[%s]::Configuration[%d]", superTaskName, providerUniversalKey, providerConfigId);
```
gkillough: Do we want to have something like "...Configuration[id:%d]"

PR #864, opened 2020-02-20
 Create a task factory to create the tasks for black duck provider

psantos1113:


src/main/java/com/synopsys/integration/alert/provider/blackduck/factories/BlackDuckTaskFactory.java
```diff
@@ -0,0 +1,69 @@
...
+        this.blackDuckDataAccessor = blackDuckDataAccessor;
+        this.configurationAccessor = configurationAccessor;
+        this.notificationManager = notificationManager;
+        this.filePersistenceUtil = filePersistenceUtil;
+    }
+
+    @Override
+    public List<ProviderTask> createTasks(ProviderProperties providerProperties) {
+        BlackDuckAccumulator accumulator = new BlackDuckAccumulator(blackDuckProviderKey, taskScheduler, notificationManager, filePersistenceUtil);
+        accumulator.setProviderPropertiesForRun(providerProperties);
```
gkillough: We want to get rid of the getter/setter for properties and pass them in through the constructor.

gkillough: That is part of the reason why it can't be a component.

src/main/java/com/synopsys/integration/alert/provider/blackduck/factories/BlackDuckTaskFactory.java
```diff
@@ -0,0 +1,69 @@
...
+        BlackDuckDataSyncTask syncTask = new BlackDuckDataSyncTask(blackDuckProviderKey, taskScheduler, blackDuckDataAccessor, configurationAccessor);
```
gkillough: This doesn't seem to have the properties set at all.

alert-common/src/main/java/com/synopsys/integration/alert/common/provider/lifecycle/ProviderTask.java
```diff
@@ -62,10 +62,6 @@ public String getTaskName() {
         return ProviderTask.computeProviderTaskName(providerKey, getProviderProperties().getConfigId(), getClass());
     }
 
-    public final void setProviderPropertiesForRun(ProviderProperties providerProperties) {
-        this.providerProperties = providerProperties;
-    }
-
     protected ProviderProperties getProviderProperties() {
```
gkillough: I think `validateProviderProperties` and `invalidateProviderProperties` should also be deleted now.

PR #865, opened 2020-02-20
Add provider task properties table and fix BlackDuckAccumulator

gkillough:


alert-database/src/main/resources/liquibase/6.0.0/provider-task-properties.xml
```diff
@@ -0,0 +1,35 @@
...
+                baseTableName="provider_task_properties"
+                baseColumnNames="provider_config_id"
+                constraintName="provider_task_properties_provider_config_id_fk"
+                referencedTableSchemaName="alert"
+                referencedTableName="descriptor_configs"
+                referencedColumnNames="id"
+                onDelete="CASCADE"
+        />
+    </changeSet>
+    <!-- FIXME create upgrade script for existing tasks -->
```
gkillough: There is a separate ticket for this.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ProviderTaskPropertiesAccessor.java
```diff
@@ -0,0 +1,12 @@
+package com.synopsys.integration.alert.common.persistence.accessor;
+
+import java.util.Optional;
+
+import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
+
+public interface ProviderTaskPropertiesAccessor {
+    Optional<String> getTaskProperty(String taskName, String propertyKey);
```
psantos1113: Should the getTaskProperty also take the configId?  Even though the task name is unique because it currently has the config id in the task name.

PR #867, opened 2020-02-21
Migrate the accumulator task last search date from a file to the DB

gkillough:


alert-database/src/main/resources/liquibase/6.0.0/provider-fields.xml
```diff
@@ -70,7 +70,7 @@
         <insert schemaName="alert" tableName="field_values">
             <column name="config_id" valueComputed="GET_OLD_BLACK_DUCK_PROVIDER_CONFIG_ID()"/>
             <column name="field_id" valueComputed="GET_FIELD_ID('provider.common.config.enabled')"/>
-            <column name="field_value" value="true"/>
+            <column name="field_value" valueBoolean="true"/>
```
gkillough: This is wrong because field_value should be a VARCHAR.

PR #873, opened 2020-02-24
Helm Chart: Allow the user to set the volumeName in the PVC resource

mphammer:
# PR Overview:

* N/A

* Users should be able to use Volumes that they have already created (aka not have them dynamically created by the cluster). They need to specify the volumeName in order to do this.

* Add pvcVolumeName to Values.yaml and set it in the Persistent Volume Claim
*
*

### DIFF
--- a/deployment/helm/values.yaml
+++ b/deployment/helm/values.yaml
@@ -27,6 +27,7 @@ cfssl:
 enablePersistentStorage: false
 pvcSize: "5G"
 storageClassName: ""
+volumeName: ""
 
 # Expose Alert's User Interface
 exposeui: true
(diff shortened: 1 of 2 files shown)


deployment/helm/values.yaml
```diff
@@ -27,6 +27,7 @@ cfssl:
 enablePersistentStorage: false
 pvcSize: "5G"
 storageClassName: ""
+pvcVolumeName: ""
```
yashbhutwala: small nit, but can we keep this the same as standard pvc key [`volumeName`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#persistentvolumeclaimspec-v1-core)?

mphammer: Done

psantos1113:
This is approved I will merge into master.

PR #874, opened 2020-02-24
Helm Chart: Have the resources in Values.yaml match the k8s format

mphammer:
# PR Overview:

* N/A

* The User can specify any k8s resource configurations in the Values.yaml

* Have the format for taking in resources in the Values.yaml file match the format of k8s objects. Then pass the values directly into the deployment manifest for Alert and Cfssl

### DIFF
--- a/deployment/helm/values.yaml
+++ b/deployment/helm/values.yaml
@@ -5,8 +5,11 @@
 # alert - configurations for the Alert Pod
 alert:
   image: "docker.io/blackducksoftware/blackduck-alert:VERSION_TOKEN"
-  limitMemory: "2560M"
-  requestMemory: "2560M"
+  resources:
+    limits:
+      memory: "2560M"
+    requests:
+      memory: "2560M"
   nodeSelector: {}
   tolerations: []
   affinity: {}
@@ -17,8 +20,11 @@ deployAlertWithBlackDuck: false
 enableStandalone: true
 cfssl:
   image: "docker.io/blackducksoftware/blackduck-cfssl:1.0.0"
-  limitMemory: "640M"
-  requestMemory: "640M"
+  resources:
+    limits:
+      memory: "640M"
+    requests:
+      memory: "640M"
   nodeSelector: {}
   tolerations: []
   affinity: {}
--- a/deployment/helm/values.yaml
+++ b/deployment/helm/values.yaml
@@ -7,9 +7,9 @@ alert:
   image: "docker.io/blackducksoftware/blackduck-alert:VERSION_TOKEN"
   resources:
     limits:
-      memory: "2560M"
+      memory: "2560Mi"
     requests:
-      memory: "2560M"
+      memory: "2560Mi"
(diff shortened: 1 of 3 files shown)


deployment/helm/values.yaml
```diff
@@ -5,8 +5,11 @@
 # alert - configurations for the Alert Pod
 alert:
   image: "docker.io/blackducksoftware/blackduck-alert:VERSION_TOKEN"
-  limitMemory: "2560M"
-  requestMemory: "2560M"
+  resources:
+    limits:
+      memory: "2560M"
+    requests:
+      memory: "2560M"
```
yashbhutwala: @mphammer this is a bug, similar to the one for Black Duck. https://github.com/blackducksoftware/blackduck-alert/blob/5.2.1/deployment/docker-compose/standalone/docker-compose.yml#L28

M in docker means Mi in k8s

See: https://github.com/blackducksoftware/synopsys-operator/pull/1100/files and https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/

deployment/helm/values.yaml
```diff
@@ -17,8 +20,11 @@ deployAlertWithBlackDuck: false
 enableStandalone: true
 cfssl:
   image: "docker.io/blackducksoftware/blackduck-cfssl:1.0.0"
-  limitMemory: "640M"
-  requestMemory: "640M"
+  resources:
+    limits:
+      memory: "640M"
+    requests:
+      memory: "640M"
```
yashbhutwala: Same bug as above 😄

yashbhutwala: Also, I'm not sure why we use 640, whereas the docker-compose file specifies 512: https://github.com/blackducksoftware/blackduck-alert/blob/5.2.1/deployment/docker-compose/standalone/docker-compose.yml#L13

yashbhutwala: Interesting, so this may also be a bug, since Black Duck also specifies 512 for the cfssl container: https://github.com/blackducksoftware/hub/blob/v2019.12.1/docker-compose/docker-compose.yml#L107

deployment/helm/values.yaml
mphammer: Nice find! I switched it to Mi

deployment/helm/values.yaml
mphammer: Nice find! I switched it to Mi

mphammer: Interesting that it's 640 in the docker-swarm file: https://github.com/blackducksoftware/blackduck-alert/blob/master/deployment/docker-swarm/standalone/docker-compose.yml#L16

yashbhutwala: Same for Black Duck: https://github.com/blackducksoftware/hub/blob/v2019.12.1/docker-swarm/docker-compose.yml#L125

:exploding_head:

psantos1113:
This PR is approved.  Merging the PR.

PR #883, opened 2020-02-26
fix: Correct issues with cancel of modals and saving permissions.

psantos1113:


src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -42,7 +42,7 @@ class PermissionTable extends Component {
         const { name, value, type, checked } = e.target;
         const { permissionsData } = this.state;
         const updatedValue = type === 'checkbox' ? checked.toString()
-            .toLowerCase() === 'true' : value;
+        .toLowerCase() === 'true' : value;
```
bamandel: Do we prefer this sort of indentation?

src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -113,7 +113,7 @@ class PermissionTable extends Component {
         const splitPermissions = permissionsColumn.split('-');
 
         const prettyNameObject = this.createDescriptorOptions()
-            .find(option => descriptorName === option.label);
+        .find(option => descriptorName === option.label);
```
bamandel: Indentation here as well

src/main/js/dynamic/loaded/users/RoleTable.js
```diff
@@ -33,7 +33,7 @@ class RoleTable extends Component {
         const { name, value, type, checked } = e.target;
         const { role } = this.state;
         const updatedValue = type === 'checkbox' ? checked.toString()
-            .toLowerCase() === 'true' : value;
+        .toLowerCase() === 'true' : value;
```
bamandel: indenting

src/main/js/field/TableDisplay.js
```diff
@@ -134,9 +136,25 @@ class TableDisplay extends Component {
         this.updateData();
     }
 
+    handleInsertModalSubmit(event, onModalClose) {
+        if (event) {
+            event.preventDefault()
+            event.stopPropagation();
+        }
```
bamandel: Do we need this if statement if we're calling it in the handleSubmit as well?

psantos1113: Have to stop the event propagation before the onModalClose otherwise all the modals are closed on this one event. Also handleSubmit is called by other functions.  So we have to stop the even propagation in both places.

PR #889, opened 2020-02-28
Making the SAMLContext Serializable. Removing the images field from DockerTagModel because we do not use it.

jamesrichard91:


src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/model/DockerTagModel.java
```diff
@@ -22,16 +22,14 @@
...
-
 import com.google.gson.annotations.SerializedName;
 import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 
 public class DockerTagModel extends AlertSerializableModel {
     private String name;
     @SerializedName("full_size")
     private Long fullSize;
-    private List<Object> images;
+    //private List<Object> images;
```
bamandel: Should this be commented?

src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/model/DockerTagModel.java
```diff
@@ -65,9 +62,9 @@ public Long getFullSize() {
     }
 
     // If necessary, this can return a list of more specific Objects.
-    public List<Object> getImages() {
-        return images;
-    }
+    //    public List<Object> getImages() {
+    //        return images;
+    //    }
```
bamandel: Should this be commented?

jamesrichard91:
I left the images field commented out in case we decide we want to use it in the future. I can just remove it if that reduces confusion.

PR #895, opened 2020-03-04
Fix issues found from a development smoke test

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -77,14 +79,24 @@ public String createHeader(MessageContentGroup messageContentGroup) {
         return headerSeparator;
     }
 
-    public String getCommonTopic(MessageContentGroup messageContentGroup) {
-        return createLinkableItemString(messageContentGroup.getCommonTopic(), true) + getLineSeparator();
+    public String getCommonTopic(MessageContentGroup messageContent, ItemOperation nullableTopLevelAction) {
```
jamesrichard91: Why not add @Nullable, javax.annotation.Nullable, to the nullableTopLevelAction?

``` java
public String getCommonTopic(MessageContentGroup messageContent, @Nullable ItemOperation nullableTopLevelAction) {
```

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -77,14 +79,24 @@ public String createHeader(MessageContentGroup messageContentGroup) {
...
+            commonTopic = new LinkableItem(commonTopic.getName(), commonTopic.getValue());
+        }
+        return createLinkableItemString(commonTopic, true) + getLineSeparator();
     }
 
-    public String getComponentSubTopic(ProviderMessageContent messageContent) {
-        return messageContent.getSubTopic()
-                   .map(item -> createLinkableItemString(item, true))
-                   .orElse("");
+    public String getComponentSubTopic(ProviderMessageContent messageContent, ItemOperation nullableTopLevelAction) {
```
jamesrichard91: Why not add `@Nullable`, javax.annotation.Nullable, to the nullableTopLevelAction?

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -329,4 +341,14 @@ protected String createLinkableItemValueString(LinkableItem linkableItem) {
                    .findAny();
     }
 
+    protected ItemOperation getNullableTopLevelAction(MessageContentGroup messageContentGroup) {
```
jamesrichard91: Why not add `@Nullable`, javax.annotation.Nullable, to the returned ItemOperation?
``` java
protected @Nullable ItemOperation getNullableTopLevelAction(MessageContentGroup messageContentGroup) {
```

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -77,14 +79,24 @@ public String createHeader(MessageContentGroup messageContentGroup) {
         return headerSeparator;
     }
 
-    public String getCommonTopic(MessageContentGroup messageContentGroup) {
-        return createLinkableItemString(messageContentGroup.getCommonTopic(), true) + getLineSeparator();
+    public String getCommonTopic(MessageContentGroup messageContent, ItemOperation nullableTopLevelAction) {
+        LinkableItem commonTopic = messageContent.getCommonTopic();
+        if (ItemOperation.DELETE.equals(nullableTopLevelAction)) {
+            commonTopic = new LinkableItem(commonTopic.getName(), commonTopic.getValue());
+        }
```
bamandel: When should we use the ternary operator vs something like this?

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -77,14 +79,24 @@ public String createHeader(MessageContentGroup messageContentGroup) {
...
-        return messageContent.getSubTopic()
-                   .map(item -> createLinkableItemString(item, true))
-                   .orElse("");
+    public String getComponentSubTopic(ProviderMessageContent messageContent, ItemOperation nullableTopLevelAction) {
+        Optional<LinkableItem> optionalSubTopic = messageContent.getSubTopic();
+        if (optionalSubTopic.isPresent()) {
+            LinkableItem subTopic = optionalSubTopic.get();
+            if (ItemOperation.DELETE.equals(nullableTopLevelAction)) {
+                subTopic = new LinkableItem(subTopic.getName(), subTopic.getValue());
+            }
```
bamandel: Another possible location for ternary operator

src/main/java/com/synopsys/integration/alert/provider/blackduck/actions/BlackDuckGlobalApiAction.java
```diff
@@ -77,8 +81,8 @@ public void afterDeleteAction(FieldModel fieldModel) {
         FieldValueModel fieldValueModel = keyToValues.get(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME);
         String blackDuckGlobalConfigName = fieldValueModel.getValue().orElse("");
 
-        Long configId = Long.parseLong(Objects.requireNonNullElse(fieldModel.getId(), "-1"));
-        providerLifecycleManager.unscheduleTasksForProviderConfig(blackDuckProvider, configId);
+        Long configId = Long.parseLong(fieldModel.getId());
```
jamesrichard91: By removing the default "-1" aren't you opening this up to possible NPE's?

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
gkillough: We don't have a standardized use of this and even if it is null, it is used in a null-safe way.

src/main/java/com/synopsys/integration/alert/provider/blackduck/actions/BlackDuckGlobalApiAction.java
gkillough: There is no way for this config to be deleted without the id, so it will always be present or error out before it gets here.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
jamesrichard91: Or the return type should be `Optional<ItemOperation>`

gkillough: Perhaps.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
gkillough: This is gonna sound weird, but I think a "three dot" maximum should be the rule of thumb. If dereferencing is happening more than three times, it should not be part of a ternary operation.

gkillough: That goes for the calling class as well, think of method calls as `this.doSomething()`.

PR #896, opened 2020-03-05
role dialog fixes

psantos1113:
When clicking on the add method to create a new role clear the permissionData object.
Need to transition from inProgress to not in progress to close the permissions dialog.

### DIFF
--- a/src/main/js/dynamic/loaded/users/PermissionTable.js
+++ b/src/main/js/dynamic/loaded/users/PermissionTable.js
@@ -241,7 +241,13 @@ class PermissionTable extends Component {
             });
             return false;
         }
-        return this.props.saveRole(permissionsData);
+        const saved = this.props.saveRole(permissionsData);
+        if (saved) {
+            this.setState({
+                permissionsData: {}
+            });
+        }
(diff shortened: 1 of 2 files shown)


src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -233,15 +234,27 @@ class PermissionTable extends Component {
         );
     }
 
-    onSavePermissions() {
+    async onSavePermissions() {
+        await this.setState({
```
psantos1113: In 6.0.0 I may consider using promises here and refactor this code to use the .then function to control order.

src/main/js/dynamic/loaded/users/RoleTable.js
```diff
@@ -67,9 +67,9 @@ class RoleTable extends Component {
         });
     }
 
-    onSave() {
+    async onSave() {
         const { role } = this.state;
-        this.props.saveRole(role);
+        await this.props.saveRole(role);
```
psantos1113: May look into using Promises here rather than async and await.

PR #897, opened 2020-03-05
user password fix

psantos1113:
check if the user confirm password error is populated.  If it has then edit dialog should be shown with the error.


src/main/js/dynamic/loaded/users/UserTable.js
```diff
@@ -74,7 +74,7 @@ class UserTable extends Component {
         const { user } = this.state;
 
         const updatedValue = type === 'checkbox' ? checked.toString()
-            .toLowerCase() === 'true' : value;
+        .toLowerCase() === 'true' : value;
```
bamandel: Not sure if this was intentional or not. Did we decide to keep this format?

PR #903, opened 2020-03-06
Validate that a job name is not a duplicate while updating

gkillough:


gkillough:
Closing to fix in 5.3.0

PR #905, opened 2020-03-09
Fix: compare existing job name with all other jobs to prevent duplicate names

jamesrichard91:


src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -172,20 +174,28 @@ public JobFieldModel updateJob(UUID id, JobFieldModel jobFieldModel) throws Aler
...
                 List<ConfigurationJobModel> jobs = configurationAccessor.getAllJobs();
+
                 boolean foundDuplicateName = jobs.stream()
+                                                 .filter(job -> {
+                                                     if (null != currentJobId) {
+                                                         return !job.getJobId().equals(currentJobId);
+                                                     } else {
+                                                         return true;
+                                                     }
+                                                 })
```
gkillough: Either abstract this into its own method, or do the check before this stream. This is a little to cluttered IMO.

PR #911, opened 2020-03-12
certificates ui

psantos1113:
create the UI to import certificates into Alert.


docker-entrypoint.sh
```diff
@@ -297,36 +297,6 @@ createDataBackUp(){
     fi
 }
 
-checkVolumeDirectories() {
```
jamesrichard91: Was this not needed anymore?

src/main/js/store/actions/certificates.js
```diff
@@ -0,0 +1,195 @@
...
+    CERTIFICATES_FETCHED_ALL,
+    CERTIFICATES_FETCHING_ALL,
+    CERTIFICATES_SAVE_ERROR,
+    CERTIFICATES_SAVED,
+    CERTIFICATES_SAVING
+} from "store/actions/types";
+
+import * as RequestUtilities from 'util/RequestUtilities'
+import { verifyLoginByStatus } from "./session";
+import { fetchUsers } from "./users";
```
jamesrichard91: Can you make these absolute paths?

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
```diff
@@ -0,0 +1,197 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import PropTypes from "prop-types";
+import { clearCertificateFieldErrors, deleteCertificate, fetchCertificates, saveCertificate } from "store/actions/certificates";
+import ConfigurationLabel from "component/common/ConfigurationLabel";
+import TableDisplay from "field/TableDisplay";
+import TextInput from "field/input/TextInput";
+import TextArea from "field/input/TextArea";
+import ReadOnlyField from "field/ReadOnlyField";
```
bamandel: These should be single quotes

docker-entrypoint.sh
psantos1113: no with the database container defining the volume PGDATA and docker-compose isn't going to be supported we don't have to verify that we can write to the volume. The encryption data is stored as secrets.  The accumulator has the last runtime stored in the database.  So we should be good here to remove this.

src/main/js/store/actions/certificates.js
psantos1113: will do

PR #919, opened 2020-03-17
Helm Chart: Values.environs are passed through a map

mphammer:
# Pull Request template

* N/A
* JIRA Ticket: https://jira-sig.internal.synopsys.com/browse/CLOUD-664

* This is done to match the conventions in the Black Duck Helm Chart (which uses a map)

* Change the Values.environs field to be a map
* Modify how they are transcribed into the ConfigMap and Secret

### DIFF
--- a/deployment/helm/values.yaml
+++ b/deployment/helm/values.yaml
@@ -39,23 +39,22 @@ volumeName: ""
 exposeui: true
 exposedServiceType: NodePort # ClusterIP | NodePort | LoadBalancer
 
-# environs is a list of *additional* environs to add to the Alert's ConfigMap
+# environs is a map of *additional* environs to add to the Alert's ConfigMap
 # The follwing environs are already set in the ConfigMap
 # at /templates/alert-environ-resources.yaml (go there to set them):
-#   - HUB_WEBAPP_HOST
-#   - HUB_CFSSL_HOST
-#   - ALERT_SERVER_PORT
-# Format: array of strings delimited by commas; each string has the format "KEY:VALUE" e.g. ["HUB_WEBAPP_HOST:localhost", "ALERT_SERVER_PORT:443"]
-environs: []
+#   HUB_WEBAPP_HOST
+#   HUB_CFSSL_HOST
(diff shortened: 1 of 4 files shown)


deployment/helm/values.yaml
```diff
@@ -45,15 +45,13 @@ exposedServiceType: NodePort # ClusterIP | NodePort | LoadBalancer
 #   - HUB_WEBAPP_HOST
 #   - HUB_CFSSL_HOST
 #   - ALERT_SERVER_PORT
-# Format: array of strings delimited by commas; each string has the format "KEY:VALUE" e.g. ["HUB_WEBAPP_HOST:localhost", "ALERT_SERVER_PORT:443"]
-environs: []
+environs: {}
```
yashbhutwala: can we just add examples here in the comments similar to blackduck: 
# if you are setting the value using set flag in helm command, do --set environs.* = ""; i.e.: --set environs.HUB_POSTGRES_CONNECTION_ADMIN="blackduck@pg-server-name"

deployment/helm/values.yaml
```diff
@@ -45,15 +45,13 @@ exposedServiceType: NodePort # ClusterIP | NodePort | LoadBalancer
...
 
 # secretEnvirons is a list of environs to add to the Alert's Secret
 # The follwing environs are handled by setting .Values.setEncryptionSecretData:
 #   - ALERT_ENCRYPTION_PASSWORD
 #   - ALERT_ENCRYPTION_GLOBAL_SALT
-# Format: array of strings delimited by commas; each string has the format "KEY:VALUE" e.g. ["KEY:VALUE", "KEY:VALUE"]
-secretEnvirons: []
+secretEnvirons: {}
```
yashbhutwala: same note as above about the example.  also s/follwing/following 🙃

deployment/helm/values.yaml
mphammer: done!

deployment/helm/values.yaml
mphammer: done!

PR #920, opened 2020-03-17
Prevent duplicate permissions

jamesrichard91:


src/main/java/com/synopsys/integration/alert/web/user/RoleActions.java
```diff
@@ -184,4 +188,16 @@ private void validateCreationRoleName(Map<String, String> fieldErrors, String ro
         }
     }
 
+    private void validatePermissions(Set<PermissionModel> permissionModels) throws AlertConfigurationException {
+        Set<PermissionKey> descriptorContexts = new HashSet<>();
+        for (PermissionModel permissionModel : permissionModels) {
```
gkillough: Would it be worth it to add a method to PermissionModel that checks if the descriptorContext is equal to that of another PermissionModel.

PR #924, opened 2020-03-19
purge fix

psantos1113:
The audit entry table had a row deleted from it.  The audit_notification_relation table has an entry for the notification id with the missing audit entry.  This causes an issue when purging.  It doesn't appear to be an issue in 5.3.0 at all. The changes here are to resolve a particular customer issue where they are in the state where an audit entry is missing from the database.  The purge data tries to run and cannot complete due to a database constraint failure.

The solution is to log the error message and try to delete the audit relation by notification id.


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
```diff
@@ -175,25 +179,35 @@ public PageRequest getPageRequestForNotifications(final Integer pageNumber, fina
...
-    private void deleteAuditEntries(final Long notificationId) {
-        final List<AuditNotificationRelation> foundRelations = auditNotificationRepository.findByNotificationId(notificationId);
-        final Function<AuditNotificationRelation, Long> transform = AuditNotificationRelation::getAuditEntryId;
-        final List<Long> auditIdList = foundRelations.stream().map(transform).collect(Collectors.toList());
-        auditNotificationRepository.deleteAll(foundRelations);
-        final List<AuditEntryEntity> auditEntryList = auditEntryRepository.findAllById(auditIdList);
-        auditEntryRepository.deleteAll(auditEntryList);
+    private void deleteAuditEntries(Long notificationId) {
+        List<AuditNotificationRelation> foundRelations;
+        try {
```
psantos1113: Here's where the change really is.  try catch block for the old code that was failing.  Attempt to delete from the audit relation by notification id.  Otherwise if it fails log it and move on to deleting the notifications.

gkillough: Does this have to do with foreign keys?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
```diff
@@ -175,25 +179,35 @@ public PageRequest getPageRequestForNotifications(final Integer pageNumber, fina
...
+            foundRelations = auditNotificationRepository.findByNotificationId(notificationId);
+            Function<AuditNotificationRelation, Long> transform = AuditNotificationRelation::getAuditEntryId;
+            List<Long> auditIdList = foundRelations.stream().map(transform).collect(Collectors.toList());
```
gkillough: Each method that returns a `Stream` should be on its own line.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
psantos1113: Yes I believe it has to do with the foreign key constraints between the audit_notification_relation, audit_entry, and raw_notification_content tables.  Currently the customer is in a state

I can't produce the problem in 5.1.0 and 5.3.0. Because the database is in this odd state in 5.0.0 this change will be able to allow notifications to be purged to clean up the database. Alert may have been missing some delete on cascade constraints.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
```diff
@@ -175,25 +179,35 @@ public PageRequest getPageRequestForNotifications(final Integer pageNumber, fina
...
-        final Function<AuditNotificationRelation, Long> transform = AuditNotificationRelation::getAuditEntryId;
-        final List<Long> auditIdList = foundRelations.stream().map(transform).collect(Collectors.toList());
-        auditNotificationRepository.deleteAll(foundRelations);
-        final List<AuditEntryEntity> auditEntryList = auditEntryRepository.findAllById(auditIdList);
-        auditEntryRepository.deleteAll(auditEntryList);
+    private void deleteAuditEntries(Long notificationId) {
+        List<AuditNotificationRelation> foundRelations;
+        try {
+            foundRelations = auditNotificationRepository.findByNotificationId(notificationId);
+            Function<AuditNotificationRelation, Long> transform = AuditNotificationRelation::getAuditEntryId;
```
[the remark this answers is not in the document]
psantos1113: I think I had a different solution in mind when I first wrote the change.  I will look to clean up.

PR #928, opened 2020-03-23
Add a delete button to the global configurations.

psantos1113:
Added a delete button with a confirmation modal to the config buttons.  The global configurations, provider and channel, will display a delete button to delete the configurations.

### DIFF
--- a/src/main/js/component/common/ConfirmModal.js
+++ b/src/main/js/component/common/ConfirmModal.js
@@ -1,0 +1,86 @@
+import React, { Component } from 'react';
+import { Modal } from 'react-bootstrap';
+import PropTypes from 'prop-types';
+
+class ConfirmModal extends Component {
+    constructor(props) {
+        super(props);
+        this.handleAffirmativeClick = this.handleAffirmativeClick.bind(this);
+        this.handleNegativeClick = this.handleNegativeClick.bind(this);
+        this.resetState = this.resetState.bind(this);
+        this.state = {
+            show: props.showModal,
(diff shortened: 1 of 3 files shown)


src/main/js/component/common/ConfirmModal.js
```diff
@@ -0,0 +1,86 @@
...
+                    <Modal.Title>{title}</Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    {message}
+                </Modal.Body>
+                <Modal.Footer>
+                    <button id="confirmNegative" type="button" className="btn btn-link" onClick={this.handleNegativeClick}>{negativeButtonText}</button>
+                    <button id="confirmAffirmative" type="button" className="btn btn-danger" onClick={this.handleAffirmativeClick}>{affirmativeButtonText}</button>
+                </Modal.Footer>
+            </Modal>
```
bamandel: Do we have a modal component that can be used effectively here?

gkillough: I think this component is exactly that.

psantos1113: You're talking about Popup.js which should move into the component folder.  But that uses our config buttons.  This is to confirm an action and it has the same look and feel as LogoutConfirmation. Which is also a confirmation dialog.

bamandel: We could try building on top of the pop up button a couple different types of pop ups. But that would be for the next version

PR #933, opened 2020-03-25
Addressing the missing default admin system message update. Removing outdated code.

jamesrichard91:


src/main/java/com/synopsys/integration/alert/web/user/UserActions.java
```diff
@@ -88,6 +98,7 @@ public UserConfig createUser(UserConfig userConfig) throws AlertDatabaseConstrai
             authorizationUtility.updateUserRoles(userId, roleNames);
         }
         userModel = userAccessor.getUser(userId).orElse(userModel);
+        validateSysadminUser(userId);
```
gkillough: Do we definitely want to do this for GETs?

psantos1113: I think this is just defensive coding in case someone tries to sent a payload of data with the sysadmin user id.We want to validate the sysadmin user to not potentially change the sysadmin user.

jamesrichard91: Github displays this in a weird way. This is in createUser not in getUsers. You need to expand the code to see that.

PR #934, opened 2020-03-25
provider table

psantos1113:
Changes to add the provider table for configuring multiple providers.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderGlobalUIConfig.java
```diff
@@ -58,4 +74,29 @@ public ProviderKey getProviderKey() {
         return providerKey;
     }
 
+    private Collection<String> validateDuplicateNames(FieldValueModel fieldToValidate, FieldModel fieldModel) {
+        try {
+            List<ConfigurationModel> configurations = configurationAccessor.getConfigurationsByDescriptorType(DescriptorType.PROVIDER);
+            if (!configurations.isEmpty()) {
```
bamandel: I don't think we really need this empty check. If the content is empty, we'll stream through an empty list creating another empty list which we're checking later on.

src/main/js/field/TableDisplay.js
```diff
@@ -150,25 +151,57 @@ class TableDisplay extends Component {
...
+        };
+        this.props.onConfigTest(callback);
     }
 
     createEditModal() {
         const { currentRowSelected } = this.state;
-        const { modalTitle, newConfigFields, inProgress } = this.props;
-        const showModal = Boolean(currentRowSelected) || this.isShowModal();
+        const { modalTitle, newConfigFields, inProgress, testButton, testButtonLabel, errorDialogMessage, actionMessage } = this.props;
+        const showModal = this.isShowModal();
```
bamandel: Probably don't need this variable anymore

src/main/js/providers/ProviderTable.js
```diff
@@ -0,0 +1,363 @@
...
+        const { name, value, type, checked } = e.target;
+        const { providerConfig } = this.state;
+
+        const updatedValue = type === 'checkbox' ? checked.toString().toLowerCase() === 'true' : value;
+        const newProviderConfig = Object.assign(providerConfig, FieldModelUtilities.updateFieldModelSingleValue(providerConfig, name, updatedValue));
+
+        this.setState({
+            providerConfig: newProviderConfig
+        });
+    }
```
bamandel: HandleChange seems to be very similar everywhere. Could this be made common somewhere? If it's similar to the TableDisplay, perhaps we add most of this code there?

bamandel: FieldsPanel handles its on handleChange similar to this. Perhaps we could do it that way.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderGlobalUIConfig.java
```diff
@@ -58,4 +74,29 @@ public ProviderKey getProviderKey() {
...
+                List<ConfigurationModel> modelsWithName = configurations.stream()
+                                                              .filter(configurationModel ->
+                                                                          configurationModel.getField(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME)
+                                                                              .flatMap(ConfigurationFieldModel::getFieldValue)
+                                                                              .filter(configName -> configName.equals(fieldToValidate.getValue().orElse("")))
+                                                                              .isPresent())
+                                                              .collect(Collectors.toList());
+                if (modelsWithName.size() > 1) {
+                    return List.of(ERROR_DUPLICATE_PROVIDER_NAME);
```
gkillough: Should this be a database constraint? I don't think this is the appropriate place to handle this error.

src/main/js/Navigation.js
```diff
@@ -40,12 +40,18 @@ class Navigation extends Component {
...
         const nav = (
             <Fragment>
-                {providers}
+                <li className="navHeader">
+                    Providers
+                </li>
+                <li>
+                    <NavLink to="/alert/providers/blackduck" activeClassName="activeNav">
+                        Black Duck
+                    </NavLink>
```
gkillough: Anything we can do to make this more dynamic? We should have a display name in the provider metadata no?

src/main/js/store/actions/globalConfiguration.js
```diff
@@ -152,6 +167,27 @@ export function refreshConfig(id) {
     };
 }
 
+export function getAllConfigs(descriptorName) {
+    return (dispatch, getState) => {
+        dispatch(fetchingConfig());
+        const { csrfToken } = getState().session;
```
gkillough: Are we storing the CSRF token as a cookie?

src/main/js/providers/ProviderTable.js
psantos1113: I will look into this with another feature I will add.  I will consider it for a different PR.

src/main/js/Navigation.js
psantos1113: I should be able to iterate through the providers to make it more dynamic.

src/main/js/store/actions/globalConfiguration.js
psantos1113: The CSRF token is stored in the redux state for the application.  When you login we update the redux state with the CSRF token for the application.  It isn't stored as a cookie.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderGlobalUIConfig.java
jamesrichard91: I think this is defensive coding and saves us from running a few lines of code that we may not want.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderGlobalUIConfig.java
jamesrichard91: We could move this duplicate validation into the database accessors.

src/main/js/providers/ProviderTable.js
```diff
@@ -0,0 +1,363 @@
...
+            });
+        }
+
+        selectedConfig.id = null;
+        this.setState({
+            providerConfig: selectedConfig
+        }, callback);
+    }
+
+    checkJobPermissions(operation) {
```
jamesrichard91: I dont think this method name makes sense. Maybe `checkGlobalPermissions` ?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderGlobalUIConfig.java
psantos1113: A database exception won't get associated as a field exception and then we can't display the error of a duplicate name with the name text field in the form.  We perform the same validation with job names and construct the the field exception in the JobConfigActions.  If I throw a database exception it won't be tied to a field.  We should add a constraint to the database that the configuration name be unique

gkillough: My bad, I realize this is the validator, not the endpoint.

PR #935, opened 2020-03-25
Validation updates

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/security/EncryptionUtility.java
```diff
@@ -113,7 +113,11 @@ private String getEncodedSalt() {
     }
 
     public boolean isPasswordSet() {
-        return null != getPassword();
+        return StringUtils.isNotBlank(getPassword());
```
gkillough: This is a good change, but we should double check that we are clear about blank values not being allowed for these settings in the doc.

alert-common/src/main/java/com/synopsys/integration/alert/common/system/BaseSystemValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import com.synopsys.integration.alert.common.persistence.accessor.SystemMessageUtility;
+
+public class BaseSystemValidator {
+    private SystemMessageUtility systemMessageUtility;
+
+    public BaseSystemValidator(SystemMessageUtility systemMessageUtility) {
+        this.systemMessageUtility = systemMessageUtility;
+    }
+
+    public boolean addSystemMessageIfHasError(String errorMessage, SystemMessageSeverity systemMessageSeverity, SystemMessageType messageType, boolean hasError) {
```
gkillough: I know this method name was my suggestion, but `ForError` might be less clunky than `IfHasError`.

src/main/java/com/synopsys/integration/alert/component/settings/SettingsValidator.java
```diff
@@ -35,45 +31,43 @@
...
+                encryptionUtility.isPasswordMissing());
+            if (encryptionError) {
+                valid = false;
+                logger.error(SettingsDescriptor.FIELD_ERROR_ENCRYPTION_PWD);
+            }
+            boolean saltError = addSystemMessageIfHasError(SettingsDescriptor.FIELD_ERROR_ENCRYPTION_GLOBAL_SALT, SystemMessageSeverity.ERROR, SystemMessageType.ENCRYPTION_CONFIGURATION_ERROR,
+                encryptionUtility.isGlobalSaltMissing());
+            if (saltError) {
+                valid = false;
+                logger.error(SettingsDescriptor.FIELD_ERROR_ENCRYPTION_GLOBAL_SALT);
```
gkillough: We do this a few times, should we add an overloaded version of the method with a `logError` flag?

jamesrichard91: The problem is we are not always logging at error level, some of them are warnings which makes it more complicated.

PR #936, opened 2020-03-27
Add validation of the certificate fields.

psantos1113:


alert-common/src/main/java/com/synopsys/integration/alert/common/security/CertificateUtility.java
```diff
@@ -145,6 +145,10 @@ public File getAndValidateTrustStoreFile() throws AlertConfigurationException {
         }
     }
 
+    public void validateCertificateContent(CustomCertificateModel customCertificateModel) throws AlertException {
```
gkillough: This method and the method below this both take a customCertificateModel. How does a user of the API know which validate to call?

psantos1113: The method below is private so they wouldn't know it exists. I can change the public method to validateCertificateContents.  I am just validating that the content is a in fact a certificate we can parse and potentially load.

gkillough: Ah, I see. I would still prefer if there was more distinction between what the methods do.

psantos1113: I have a change I will push soon to better distinguish between the two.  It's a valid point.  Naming things is always the fun part.

PR #938, opened 2020-03-31
task view

psantos1113:
Implement a read only table to display the tasks running on the system.  You can double click on the item and get more detail.


alert-common/src/main/java/com/synopsys/integration/alert/common/provider/lifecycle/ProviderTask.java
```diff
@@ -50,6 +53,16 @@ public String getTaskName() {
         return taskName;
     }
 
+    @Override
+    public TaskMetaData createTaskMetaData() {
```
gkillough: Can we return the specific type here?

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/TaskMetaData.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package com.synopsys.integration.alert.common.workflow.task;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class TaskMetaData extends AlertSerializableModel {
+    private static final long serialVersionUID = -3249768131233749231L;
+    private String taskName;
+    private String name;
+    private String fullyQualifiedName;
```
gkillough: What is the difference between `taskName` and `name`? Which does `fullyQualifiedName` refer to? How can we rename these variables to make that self-revealing?

src/main/js/dynamic/loaded/tasks/TaskManagement.js
```diff
@@ -0,0 +1,141 @@
...
+        return [
+            {
+                header: 'taskName',
+                headerLabel: 'Task Name',
+                isKey: true,
+                hidden: true
+            },
+            {
+                header: 'name',
+                headerLabel: 'Name',
```
gkillough: If you end up renaming these in Java, don't forget to rename them here.

alert-common/src/main/java/com/synopsys/integration/alert/common/provider/lifecycle/ProviderTask.java
psantos1113: Will do

PR #940, opened 2020-04-02
 environment changes

psantos1113:
Changes to how environment variables are handled.  

1. The environment variables are only used to initialize configurations if they are not present. 
2. The environment override flag has been removed.  
3. New environment variables for resetting authentication has been added.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/DescriptorGlobalConfigUtility.java
```diff
@@ -0,0 +1,104 @@
...
+        this.configurationFieldModelConverter = configurationFieldModelConverter;
+    }
+
+    public DescriptorKey getKey() {
+        return key;
+    }
+
+    public boolean doesConfigurationExist() {
+        try {
+            return !configurationAccessor.getConfigurationByDescriptorKeyAndContext(key, context).isEmpty();
```
gkillough, line 64: I was initially confused because I thought this returned an Optional. Let's change the method name from `getConfigurationByDescriptorKeyAndContext` to `getConfigurationsByDescriptorKeyAndContext` (making configuration plural).

jamesrichard91, line 64: You access key and context by referencing the fields directly but in other places in this class you use the getKey() method. Can we make this more consistent? Either access the fields or use the get methods.

deployment/blackduck-alert.env
```diff
@@ -21,15 +26,14 @@ ALERT_ENCRYPTION_GLOBAL_SALT=
 
 # -- BlackDuck Provider Settings
 
+#ALERT_PROVIDER_BLACKDUCK_PROVIDER_COMMON_CONFIG_NAME=Default Config
```
jamesrichard91, line 29: I know this is probably not part of this PR but why is this `COMMON_CONFIG_NAME` and not `DEFAULT_CONFIG_NAME`?

deployment/blackduck-alert.env
```diff
@@ -21,15 +26,14 @@ ALERT_ENCRYPTION_GLOBAL_SALT=
...
+#ALERT_PROVIDER_BLACKDUCK_PROVIDER_COMMON_CONFIG_ENABLED=true
```
jamesrichard91, line 30: Same as above, why isnt this `DEFAULT_CONFIG_ENABLED` ?

psantos1113: see the answer above.

deployment/blackduck-alert.env
psantos1113: COMMON_CONFIG_NAME for the environment variable gets turned into common.config.name which is the registered field name in the database for the configuration.  Each provider configuration you add has a common.config.name field.  Using this name requires no additional code to be added to AlertStartupInitializer to handle creating an initial Black Duck Provider configuration.  If we use a different environment variable name then we will need to add special logic to the AlertStartupInitializer class to process the blackduck provider and look for a specific environment variable name rather than use our existing environment variable naming convention.

PR #941, opened 2020-04-09
fix for helm lint to match the Chart name with folder name

msenmurugan:
resolves #IALERT-1676

* To make the helm lint command to pass
* fix for Synopsys Operator migration

* As part of the Jenkins pipeline, helm validations need to be added. One of the validation is to add helm lint command. In order to pass the Helm lint command, the chart name need to match with folder name. 
* Current chart name is alert-helm, which won't be nice, instead rename it to synopsys-alert to match the application name
* remove "-alert" to all resources
* added the label component type to all labels


mphammer:
Resolves JIRA Ticket: https://jira-sig.internal.synopsys.com/browse/IALERT-1677
Duplicate PR: https://github.com/blackducksoftware/blackduck-alert/pull/950

PR #942, opened 2020-04-09
fix: Fix the failing test and SonarCloud link.

psantos1113:


src/main/java/com/synopsys/integration/alert/workflow/scheduled/update/UpdateChecker.java
```diff
@@ -195,10 +195,11 @@ private int compareVersions(String firstVersion, String secondVersion) {
...
         String resultToken = versionToken;
-        if (versionToken.contains(SNAPSHOT)) {
-            resultToken = versionToken.substring(0, versionToken.indexOf(SNAPSHOT));
-        } else if (versionToken.contains(QA_BUILD)) {
+        // currently we get -SIGQA1-SNAPSHOT so we need to look for SIGQA first.
+        if (versionToken.contains(QA_BUILD)) {
             resultToken = versionToken.substring(0, versionToken.indexOf(QA_BUILD));
+        } else if (versionToken.contains(SNAPSHOT)) {
+            resultToken = versionToken.substring(0, versionToken.indexOf(SNAPSHOT));
         }
```
jamesrichard91, line 203: Should we change this to something like this, since contains uses indexOf anyway?
```java
int indexOfQABuild = versionToken.indexOf(QA_BUILD);
int indexOfSnapshot = versionToken.indexOf(SNAPSHOT);
if (indexOfQABuild > 0) {
            resultToken = versionToken.substring(0, indexOfQABuild);
 } else if (indexOfSnapshot > 0) {
            resultToken = versionToken.substring(0, indexOfSnapshot);
}
```

PR #943, opened 2020-04-09
Sonarcloud code smells

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/AlertProperties.java
```diff
@@ -136,8 +137,8 @@ public Boolean getH2ConsoleEnabled() {
         return h2ConsoleEnabled;
     }
 
-    public Boolean getSslEnabled() {
-        return sslEnabled;
+    public boolean getSslEnabled() {
+        return BooleanUtils.isTrue(sslEnabled);
```
jamesrichard91, line 141: SonarCloud pushes to use boolean when possible to avoid issues with `null` objects

alert-common/src/main/java/com/synopsys/integration/alert/common/action/ApiAction.java
```diff
@@ -27,29 +27,50 @@
 
 public abstract class ApiAction {
 
+    /**
+     * @throws AlertException Overriding classes may throw this exception
+     */
     public FieldModel beforeSaveAction(FieldModel fieldModel) throws AlertException {
```
jamesrichard91, line 33: SonarCloud does not want a method to throw an exception if the body of code does not throw that exception. In our case we do this so that overriding methods can throw the exception. The Javadoc allows SonarCloud to ignore this.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/DistributionChannel.java
```diff
@@ -64,7 +64,7 @@ public void sendAuditedMessage(DistributionEvent event) throws IntegrationExcept
             auditUtility.setAuditEntrySuccess(event.getAuditIds());
         } catch (IntegrationRestException irex) {
             auditUtility.setAuditEntryFailure(event.getAuditIds(), irex.getMessage(), irex);
-            logger.error("{} : {}", irex.getHttpStatusCode(), irex.getHttpStatusMessage());
```
gkillough, line 67: What is the specific error SonarCloud is complaining about here? This may be a suggestion we want to ignore.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -140,7 +140,7 @@ public boolean collapseOnCategory() {
         return getCollapseOnCategory();
     }
 
-    public LinkedHashSet<LinkableItem> getComponentAttributes() {
+    public Set<LinkableItem> getComponentAttributes() {
```
gkillough, line 143: Is there a way we can keep the explicit declaration that this set is an ordered set (not necessarily sorted, but ordered)?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/CustomEndpoint.java
```diff
@@ -35,6 +35,9 @@ public CustomEndpoint(String fieldKey, CustomEndpointManager customEndpointManag
         customEndpointManager.registerFunction(fieldKey, this::createResponse);
     }
 
+    /**
+     * @param fieldModel This FieldModel may be use for further computation in overriding classes
+     */
     protected Optional<ResponseEntity<String>> preprocessRequest(FieldModel fieldModel) {
```
jamesrichard91, line 41: SonarCloud does not want a method to have unused parameters. In our case we do this so that overriding methods can use the parameter. The Javadoc allows SonarCloud to ignore this.

alert-common/src/main/java/com/synopsys/integration/alert/common/email/EmailMessagingService.java
```diff
@@ -110,7 +110,6 @@ public void sendEmailMessage(EmailTarget emailTarget) throws AlertException {
             sendMessages(emailProperties, session, messages);
         } catch (MessagingException | IOException | IntegrationException ex) {
             String errorMessage = "Could not send the email. " + ex.getMessage();
-            logger.error(errorMessage, ex);
```
jamesrichard91, line 113: This gets logged later, we don't need to log this multiple times.

alert-common/src/main/java/com/synopsys/integration/alert/common/security/CertificateUtility.java
```diff
@@ -72,7 +72,6 @@ public void importCertificate(CustomCertificateModel customCertificate) throws A
                 trustStore.store(stream, getTrustStorePassword());
             }
         } catch (IOException | CertificateException | NoSuchAlgorithmException | KeyStoreException e) {
-            logger.debug("Error importing certificate.", e);
```
jamesrichard91, line 75: These errors get logged again later, the exception message is sufficient.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -85,6 +85,7 @@ public DefaultProviderDataAccessor(ProviderProjectRepository providerProjectRepo
                            .collect(Collectors.toList());
             }
         } catch (AlertDatabaseConstraintException ignored) {
+            // ignoring this exception
```
jamesrichard91, line 88: SonarCloud does not like an empty catch block unless there is a comment as to why

src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraTestIssueRequestCreator.java
```diff
@@ -77,14 +77,12 @@ public IssueTrackerRequest createRequest(IssueOperation operation, String messag
 
             switch (operation) {
                 case RESOLVE: {
-                    IssueContentModel contentModel = jiraMessageParser.createIssueContentModel(providerName, IssueResolutionRequest.OPERATION, topicItem, subTopicItem, componentItems, arbitraryItem);
-                    return IssueResolutionRequest.of(issueSearchProperties, contentModel);
+                    return createResolveIssueRequest(providerName, topicItem, subTopicItem, componentItems, arbitraryItem, issueSearchProperties);
```
jamesrichard91, line 80: SonarCloud does not like code in the switch cases, to make the cases easier to understand moved the code to a method

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/StartupScheduledTask.java
```diff
@@ -47,7 +48,7 @@ public void checkTaskEnabled() {
     public void startTask() {
         checkTaskEnabled();
         String taskName = getTaskName();
-        if (!getEnabled()) {
+        if (BooleanUtils.isFalse(getEnabled())) {
```
gkillough, line 51: Would returning a `boolean` instead of a `Boolean` resolve this? I'd rather not use `BooleanUtils` everywhere.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
gkillough: I think we as a team agreed to the convention of naming the exception `ignored`. We should remove this SonarCloud rule.

src/main/java/com/synopsys/integration/alert/web/certificates/CertificateActions.java
```diff
@@ -120,7 +121,7 @@ private void deleteByAlias(CustomCertificateModel certificateModel) {
             certificateAccessor.deleteCertificate(certificateModel.getAlias());
             certificateUtility.removeCertificate(certificateModel.getAlias());
         } catch (AlertException deleteEx) {
-            logger.error("Error deleting certificate with alias {}", certificateModel.getAlias());
+            logger.error("Error deleting certificate with alias {}. Error: ", certificateModel.getAlias(), deleteEx.getMessage());
```
gkillough: I think this is still wrong. Missing `{}` after "Error: "

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/DistributionChannel.java
jamesrichard91: I undid this particular change, but the issue is SonarCloud wants you to either log and handle the exception OR re-throw the exception but it does not want you to do both. In this case we want to log the extra fields from the IntegrationRestException so we will ignore this smelly smell.

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
```diff
@@ -80,7 +80,7 @@ public JiraServerCustomEndpoint(JiraServerChannelKey jiraChannelKey, CustomEndpo
             String username = jiraProperties.getUsername();
             String password = jiraProperties.getPassword();
             Response response = jiraAppService.installMarketplaceServerApp(JiraConstants.JIRA_APP_KEY, username, password);
-            if (response.isStatusCodeError()) {
+            if (BooleanUtils.isTrue(response.isStatusCodeError())) {
```
gkillough, line 83: Does this have to do with using `Boolean` rather than `boolean`? If so, we should fix that instead.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
jamesrichard91: Where we create the Set we are creating a LinkedHashSet. SonarCloud reasonably points out that User's should not worry about the exact implementation of the Set so just return the generic Set type.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
```diff
@@ -134,10 +133,6 @@ public String createFooter(MessageContentGroup messageContentGroup) {
 
     protected abstract String getLineSeparator();
 
-    protected String getListItemPrefix() {
```
gkillough, line 137: The reason this exists is to be overridable in different channel implementations. SonarCloud doesn't know that.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/StartupScheduledTask.java
jamesrichard91: Good catch

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
jamesrichard91: We would need to fix Response in integration-rest to resolve this. Which we should but until that gets fixed, this resolves any null issues that would come up. The current implementation of `response.isStatusCodeError()` does allow returning a null.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/message/ChannelMessageParser.java
jamesrichard91: I think we may want to add this method back if we ever need to override it. Currently there is nowhere in Alert that overrides the method.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
gkillough: My worry is that it is imperative to the consumer of this object that order be maintained. If someone consumes this object by streaming over it and collecting to a generic set, then the order is lost. The API must somehow explicitly declare that this is ordered.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/StartupScheduledTask.java
psantos1113: I believe SonarCloud won't complain if you do the following:
`boolean enabled = getEnabled();
if(!enabled) {`

Although you still have the same underlying problem with Null.  We should favor the primitive type, but the Boolean Object might be needed for the databased entity objects.  I'd have to refresh my memory on that.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
jamesrichard91: Alright, thats not a bad criticism, I can change this back and ignore the rule violation.

alert-database/src/main/java/com/synopsys/integration/alert/database/system/DefaultSystemMessageUtility.java
```diff
@@ -102,8 +103,8 @@ public void removeSystemMessagesByType(SystemMessageType messageType) {
     @Transactional
     public void deleteSystemMessages(List<SystemMessageModel> messagesToDelete) {
         List<SystemMessage> convertedMessages = messagesToDelete.stream()
-                                                          .map(this::convertToSystemMessage)
-                                                          .filter(message -> message != null).collect(Collectors.toList());
+                                                    .map(this::convertToSystemMessage)
+                                                    .filter(Objects::nonNull).collect(Collectors.toList());
```
psantos1113, line 107: collect should be on it's own line like map.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
bamandel: Is there a way for us to handle this type of Exception differently? It's all over the code base and almost always has the same content. Would handling these exceptions at the source be a better option?

src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraTestIssueRequestCreator.java
```diff
@@ -95,6 +93,18 @@ public IssueTrackerRequest createRequest(IssueOperation operation, String messag
...
+        IssueSearchProperties issueSearchProperties) {
+        IssueContentModel contentModel = jiraMessageParser.createIssueContentModel(providerName, IssueResolutionRequest.OPERATION, topicItem, subTopicItem, componentItems, arbitraryItem);
+        return IssueResolutionRequest.of(issueSearchProperties, contentModel);
+    }
+
+    private IssueCreationRequest createCreateOrUpdateIssueRequest(String providerName, LinkableItem topicItem, LinkableItem subTopicItem, Set<ComponentItem> componentItems, ComponentItem arbitraryItem,
+        IssueSearchProperties issueSearchProperties) {
+        IssueContentModel contentModel = jiraMessageParser.createIssueContentModel(providerName, IssueCreationRequest.OPERATION, topicItem, subTopicItem, componentItems, arbitraryItem);
+        return IssueCreationRequest.of(issueSearchProperties, contentModel);
+    }
```
bamandel, line 106: I think we could take this opportunity to use a bifunction here. The only difference is the Class calling of.

PR #947, opened 2020-04-09
 Remove the next run time for the provider accumulator.

psantos1113:


src/main/java/com/synopsys/integration/alert/component/scheduling/descriptor/SchedulingDescriptor.java
```diff
@@ -31,9 +31,8 @@
 public class SchedulingDescriptor extends ComponentDescriptor {
     public static final String SCHEDULING_LABEL = "Scheduling";
     public static final String SCHEDULING_URL = "scheduling";
-    public static final String SCHEDULING_DESCRIPTION = "This page shows when the scheduled tasks will run next as well as allow you to configure the frequency of the tasks.";
+    public static final String SCHEDULING_DESCRIPTION = "This page shows when system scheduled tasks will run next, as well as, allow you to configure the frequency of the system tasks.";
```
gkillough, line 34: I think because "as well as" is being used as for a conjunction and not an appositive, the first comma is correct, but the second one is not.

PR #948, opened 2020-04-10
 Remove template directory and load from jar.

psantos1113:
load templates from the classpath.


build.gradle
```diff
@@ -130,8 +130,6 @@ distributions {
             from('alert-database/src/main/resources/scripts/export_h2_tables.sql') { into 'upgradeResources' }
             from('alert-database/src/main/resources/scripts/import_postgres_tables.sql') { into 'upgradeResources' }
             from('src/main/resources/images/') { into 'images/' }
-            from('src/main/resources/templates/email') { into 'templates/email' }
-            from('src/main/resources/templates/msteams') { into 'templates/msteams' }
```
gkillough, line 134: Don't we still have our own templates to copy?

jamesrichard91: We don't need to copy them because they are inside the jar

PR #952, opened 2020-04-14
deployment fixes

psantos1113:
Create a new volume for the database container.
Remove the alert specific user.  
Create a separate service for the alert database in Kubernetes.
Fixes 5.3.0 to 6.0.0 upgrades in docker swarm.

### DIFF
--- a/alert-database/Dockerfile
+++ b/alert-database/Dockerfile
@@ -5,7 +5,7 @@ ENV POSTGRES_USER sa
 ENV POSTGRES_PASSWORD blackduck
 
 ENV DOCKER_ENTRYPPOINT_INITD_DIR /docker-entrypoint-initdb.d
-ENV ALERT_DB_DIR /opt/blackduck/alert/alert-config/data/postgres/alertdb
+ENV ALERT_DB_DIR /var/lib/postgresql/data/postgres/alertdb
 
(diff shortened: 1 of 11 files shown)


alert-database/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM postgres:11.6-alpine
+FROM postgres:11.7-alpine
```
gkillough, line 1: Let's make sure we upgrade our gradle dependency as well.

deployment/kubernetes/hub/3-alertdb.yml
```diff
@@ -0,0 +1,68 @@
...
+      template:
+        metadata:
+          name: alertdb
+          labels:
+            app: alertdb
+            tier: alertdb
+        spec:
+          volumes:
+            # NOTE: replace the emptyDir volume with a persistent volume or all the data will be lost when the pod is destroyed.
+            - emptyDir: {}
```
gkillough, line 23: Should we document and/or parameterize these since they are shared?

psantos1113: I think once the helm charts are finalized we should document that approach.  I will add to the comment here.

PR #954, opened 2020-04-16
database password

psantos1113:
Allow database credentials to be set via environment variables or by docker secrets.


src/main/resources/application.properties
```diff
@@ -16,8 +16,8 @@ spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.datasource.url=jdbc:postgresql://alertdb:5432/alertdb
 spring.datasource.hikari.jdbc-url=jdbc:postgresql://alertdb:5432/alertdb
-spring.datasource.username=sa
-spring.datasource.password=blackduck
+spring.datasource.username=${ALERT_DB_USERNAME:sa}
+spring.datasource.password=${ALERT_DB_PASSWORD:blackduck}
```
gkillough, line 20: Are these pulled from the docker-entrypoint?

psantos1113: You can set them as environment variables on the environment section of the overrides file.  Or you can use secrets files.  The docker entrypoint checks if the secrets file exists and then uses that value and exports the environment variables so the environment variable are accessible to the application.

gkillough: Right, I was just trying to figure out how it worked with the secrets in this file. That makes sense though.

PR #956, opened 2020-04-17
Jr fix job validation

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
```diff
@@ -86,6 +86,11 @@ public ChannelDistributionUIConfig(ChannelKey channelKey, String label, String u
         return Stream.concat(configFields.stream(), channelDistributionFields.stream()).collect(Collectors.toList());
     }
 
+    @Override
+    public boolean hasFields() {
```
gkillough, line 90: Can we make this a little bit clearer? Some UIConfigs inherit fields, but don't defined any.

gkillough: Could we make it a final method that checks if `createFields()` is not empty?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/GlobalConfigExistsValidator.java
```diff
@@ -75,6 +75,11 @@ public GlobalConfigExistsValidator(ConfigurationAccessor configurationAccessor,
             List<ConfigurationModel> configurations = configurationAccessor.getConfigurationsByDescriptorNameAndContext(descriptorName, ConfigContextEnum.GLOBAL);
             if (configurations.isEmpty()) {
                 return Optional.of(String.format(GLOBAL_CONFIG_MISSING, descriptorDisplayName));
+            } else {
+                boolean configurationsAreEmpty = configurations.stream().filter(configuration -> configuration.getCopyOfFieldList().size() > 0).findFirst().isEmpty();
```
gkillough, line 79: Can we put all streams on their own line here?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
jamesrichard91: I wanted to avoid creating the fields just to throw them away and not use them.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/GlobalConfigExistsValidator.java
```diff
@@ -70,11 +70,22 @@ public GlobalConfigExistsValidator(ConfigurationAccessor configurationAccessor,
...
-        String descriptorDisplayName = optionalDescriptorKey.map(DescriptorKey::getDisplayName).orElse(descriptorName);
+        String descriptorDisplayName = optionalDescriptorKey
+                                           .map(DescriptorKey::getDisplayName)
+                                           .orElse(descriptorName);
         try {
             List<ConfigurationModel> configurations = configurationAccessor.getConfigurationsByDescriptorNameAndContext(descriptorName, ConfigContextEnum.GLOBAL);
             if (configurations.isEmpty()) {
                 return Optional.of(String.format(GLOBAL_CONFIG_MISSING, descriptorDisplayName));
+            } else {
+                boolean configurationsAreEmpty = configurations
```
bamandel, line 81: Moving this stream to line 77.5 will allow you to have only the original if statement as streaming over an empty list will hardly have an impact to begin with.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
```diff
@@ -48,6 +48,8 @@ public UIConfig(String label, String description, String urlName) {
 
     public abstract List<ConfigField> createFields();
 
+    public abstract boolean hasFields();
```
bamandel, line 51: Since this is just a boolean, should we have it passed through the constructor and remove abstract from this method?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ChannelDistributionUIConfig.java
bamandel: I feel our code is more bug prone if hard code true and false. If a user changes one method they have to remember the other one exists and maybe change it accordigly.

PR #957, opened 2020-04-17
upgrade fix

psantos1113:
Fix the procedure to get the provider configuration id.

Remove the not null constraint on the provider_config_id column.  Because valueComputed is used the first statement that liquibase issues is 
ALTER TABLE raw_notification_content ADD COLUMN 'provider_config_id' NOT NULL;
But there is no default value defined so NULL is used but the column can't be null.
May be related to this liquibase bug which has similar behavior:
https://liquibase.jira.com/browse/CORE-3028


psantos1113:
Before merging I will look into adding a changeset to add the constraint of not null to the column after the modified changeset.

PR #958, opened 2020-04-20
Increasing coverage on database apis

ChomickiM:
Increasing the test coverage of our alert.database.api package by adding in new unit tests. In the future I will split a new MR into fewer new test classes in order to make future PRs smaller.


alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/MockSettingsKeyRepository.java
```diff
@@ -0,0 +1,152 @@
...
+
+import org.springframework.data.domain.Example;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+
+import com.synopsys.integration.alert.database.settings.SettingsKeyEntity;
+import com.synopsys.integration.alert.database.settings.SettingsKeyRepository;
+
+public class MockSettingsKeyRepository implements SettingsKeyRepository {
```
ChomickiM, line 17: We created these "mock" repositories for test purposes so that we can control the functionality within repository and do not have to mock out the returns from each of these methods since these repositories were easier to implement than others.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/MockSystemStatusRepository.java
```diff
@@ -0,0 +1,141 @@
...
+
+import org.springframework.data.domain.Example;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+
+import com.synopsys.integration.alert.database.system.SystemStatus;
+import com.synopsys.integration.alert.database.system.SystemStatusRepository;
+
+public class MockSystemStatusRepository implements SystemStatusRepository {
```
ChomickiM, line 15: Same as above.

alert-database/src/main/java/com/synopsys/integration/alert/database/user/AuthenticationTypeEntity.java
```diff
@@ -31,13 +31,18 @@
 @Entity
 @Table(schema = "alert", name = "authentication_type")
 public class AuthenticationTypeEntity extends DatabaseEntity {
+    private static final long serialVersionUID = -2308934073112918860L;
```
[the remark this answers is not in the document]
ChomickiM: I believe this was added as a save action, when I try to delete the serialVersionUID and save, it is added back in automatically.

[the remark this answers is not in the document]
jamesrichard91: Yes it was added by the save action.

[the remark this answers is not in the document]
jamesrichard91: We may want to remove this rule from the save actions. It's not clear if we are getting any value from the serialVersionUID's as Gavin pointed out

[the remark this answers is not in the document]
psantos1113: You will get a compiler warning if the serialVersionUID is missing on classes that implement Serializable.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultDescriptorAccessorTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+        Mockito.when(descriptorTypeRepository.findFirstByType(descriptorType.name())).thenReturn(Optional.empty());
+        Mockito.when(descriptorTypeRepository.save(Mockito.any())).thenReturn(descriptorTypeEntity);
+        Mockito.when(registeredDescriptorRepository.findByTypeId(Mockito.any())).thenReturn(List.of(registeredDescriptorEntity));
+        Mockito.when(descriptorTypeRepository.findById(registeredDescriptorEntity.getTypeId())).thenReturn(Optional.of(descriptorTypeEntity));
+
+        DefaultDescriptorAccessor descriptorAccessor = new DefaultDescriptorAccessor(registeredDescriptorRepository, null, null, descriptorTypeRepository);
+        List<RegisteredDescriptorModel> registeredDescriptorModelList = descriptorAccessor.getRegisteredDescriptorsByType(descriptorType);
+
+        assertEquals(1, registeredDescriptorModelList.size());
```
jamesrichard91, line 149: How is this any different than the previous test? It looks like the assertions do not match the test name

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultDescriptorAccessorTest.java
```diff
@@ -0,0 +1,289 @@
...
+        DescriptorTypeRepository descriptorTypeRepository = Mockito.mock(DescriptorTypeRepository.class);
+
+        Mockito.when(registeredDescriptorRepository.findById(descriptorId)).thenReturn(Optional.of(registeredDescriptorEntity));
+        Mockito.when(descriptorTypeRepository.findById(Mockito.any())).thenReturn(Optional.of(descriptorTypeEntity));
+
+        DefaultDescriptorAccessor descriptorAccessor = new DefaultDescriptorAccessor(registeredDescriptorRepository, null, null, descriptorTypeRepository);
+        Optional<RegisteredDescriptorModel> registeredDescriptorModel = descriptorAccessor.getRegisteredDescriptorById(descriptorId);
+
+        assertTrue(registeredDescriptorModel.isPresent());
+        assertEquals(typeId, registeredDescriptorModel.get().getId());
```
jamesrichard91, line 189: Instead of calling `registeredDescriptorModel.get()` over and over, you could create a local variable for the return of this method and use that variable instead. 
``` java
Optional<RegisteredDescriptorModel> registeredDescriptorModelOptional = descriptorAccessor.getRegisteredDescriptorById(descriptorId);
RegisteredDescriptorModel registeredDescriptorModel = registeredDescriptorModelOptional.get();
assertTrue(registeredDescriptorModel.isPresent());
assertEquals(typeId, registeredDescriptorModel.getId());
assertEquals(name, registeredDescriptorModel.getName());
assertEquals(descriptorType, registeredDescriptorModel.getType());
```

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultDescriptorAccessorTest.java
```diff
@@ -0,0 +1,289 @@
...
+        Mockito.when(configContextRepository.findFirstByContext(invalidConfigContextEnum.name())).thenReturn(Optional.empty());
+        Mockito.when(configContextRepository.save(Mockito.any())).thenReturn(configContextEntity);
+        Mockito.when(definedFieldRepository.findByDescriptorIdAndContext(Mockito.any(), Mockito.any())).thenReturn(List.of(definedFieldEntity));
+
+        DefaultDescriptorAccessor descriptorAccessor = new DefaultDescriptorAccessor(registeredDescriptorRepository, definedFieldRepository, configContextRepository, descriptorTypeRepository);
+        List<DefinedFieldModel> definedFieldModelList = descriptorAccessor.getFieldsForDescriptor(descriptorKey, configContextEnum);
+        List<DefinedFieldModel> emptyConfigContextDefinedFieldModelList = descriptorAccessor.getFieldsForDescriptor(descriptorKey, invalidConfigContextEnum);
+
+        assertEquals(1, definedFieldModelList.size());
+        assertEquals(definedFieldsKey, definedFieldModelList.get(0).getKey());
```
jamesrichard91, line 227: Instead of calling `definedFieldModelList.get(0)` repeatedly, try calling it once and create a local variable and reuse the variable.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/MockSettingsKeyRepository.java
```diff
@@ -0,0 +1,152 @@
...
+
+    Map<String, SettingsKeyEntity> settingsKeyEntities = new HashMap<>();
```
jamesrichard91, line 19: I understand this is a Mock object for tests but I think it would be best to follow good conventions and make these fields private and create setters or some methods to add to these Map's.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/MockSystemStatusRepository.java
```diff
@@ -0,0 +1,141 @@
...
+
+    //Only methods that are used by a test are currently implemented, all others are left default.
+    SystemStatus systemStatus = new SystemStatus(Boolean.FALSE, new Date());
```
jamesrichard91, line 18: Same as my comment above. We should make this field private and expose a setter for this method. We should also avoid hardcoding the fields. This reduces the usability of the mock object in the tests.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultDescriptorAccessorTest.java
ChomickiM: This test is testing for an Optional.empty case that is used by one of the private methods to increase its coverage, it is being passed on line 141 and as a result needed some other mocks. Because it will create a new descriptorTypeEntity I used the same setup as in the previous test after that.

PR #960, opened 2020-04-21
change PVC claim size representation from G to Gi

msenmurugan:
# Pull Request template
Resolves: 
* change PVC claim size representation from G to Gi
* If you specify limits alone on kubernetes container resources, then internally Kubernetes will set requests as same as limits and hence removed requests from resources in order to remove the duplicates.

### DIFF
--- a/deployment/helm/synopsys-alert/values.yaml
+++ b/deployment/helm/synopsys-alert/values.yaml
@@ -16,10 +16,8 @@ alert:
   resources:
     limits:
       memory: "2560Mi"
-    requests:
-      memory: "2560Mi"
   persistentVolumeClaimName: ""
-  claimSize: "5G"
+  claimSize: "5Gi"
   storageClass: ""
   volumeName: ""
   nodeSelector: {}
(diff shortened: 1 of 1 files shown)


psantos1113:
@msenmurugan Does this change need to also merge into 5.3.1?

msenmurugan:
@psantos1113 I'm not sure whether it will have prob during upgrades.. It's better if we add it now..

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -37,8 +35,6 @@ cfssl:
   resources:
     limits:
       memory: "640Mi"
-    requests:
-      memory: "640Mi"
```
jamesrichard91, line 41: Why are we removing the memory here and above?

psantos1113: @msenmurugan you will be able to answer this better.

msenmurugan: @psantos1113 I have already specified in the description of the PR

If you specify limits alone on kubernetes container resources, then internally Kubernetes will set requests as same as limits and hence removed requests from resources in order to remove the duplicates

If customer want to keep different requests and limits, they can provide those values as part of alert.resources.requests.memory flag

PR #966, opened 2020-04-29
helm alertdb

psantos1113:
Helm Chart support for Alert 6.0.0
The alert container connects to the database via environment variables passed into the container.
This allows Alert to support an on premise postgres database deployed with Alert or an external Alert database to be deployed.

I will release a new version of the alertdb container after this PR has been merged.  You will see a 1.0.2-SNAPSHOT for the alertdb container.  I could not get the same container to work in both helm and swarm.  The healthcheck of the postgresql-12-centos7 image would not return properly in docker swarm and the container would never be healthy. 

@mphammer and  @msenmurugan please review the helm charts in this PR.  This chart works for fresh installs and upgrades from 5.3.1.


build.gradle
```diff
@@ -216,8 +217,8 @@ task runServer(type: com.synopsys.integration.alert.build.RunServerTask, depends
                                 '--spring.datasource.username=sa',
                                 '--spring.datasource.password=blackduck',
                                 '--spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver',
-                                '--spring.datasource.url=jdbc:tc:postgresql:11.7:///alertdb?TC_INITSCRIPT=file:src/test/resources/testDatabase/init_test_db.sql&TC_TMPFS=/testtmpfs:rw',
-                                '--spring.datasource.hikari.jdbc-url=jdbc:tc:postgresql:11.7:///alertdb?TC_INITSCRIPT=file:src/test/resources/testDatabase/init_test_db.sql&TC_TMPFS=/testtmpfs:rw',
+                                '--spring.datasource.url=jdbc:tc:postgresql:12.2:///alertdb?TC_INITSCRIPT=file:src/test/resources/testDatabase/init_test_db.sql&TC_TMPFS=/testtmpfs:rw',
+                                '--spring.datasource.hikari.jdbc-url=jdbc:tc:postgresql:12.2:///alertdb?TC_INITSCRIPT=file:src/test/resources/testDatabase/init_test_db.sql&TC_TMPFS=/testtmpfs:rw',
```
bamandel, line 221: Perhaps a variable to store this value would be better since it's used in multiple locations and needs to be the same.

docker-entrypoint.sh
```diff
@@ -329,12 +332,47 @@ liquibaseChangelockReset() {
   echo "End releasing liquibase changeloglock."
 }
 
+validatePostgresConnection() {
+  # https://stackoverflow.com/a/58784528/6921621
```
bamandel, line 336: A brief comment about why this is relevant would help me here personally.

docker-entrypoint.sh
```diff
@@ -329,12 +332,47 @@ liquibaseChangelockReset() {
...
+      echo "Alert postgres database connection valid."
+    else
+      echo "Alert postgres connection cannot be made."
+      sleep 10
+      exit 1
+    fi
+}
+
+createPostgresDatabase() {
+  # https://stackoverflow.com/a/58784528/6921621
```
bamandel, line 349: Same as above since it's the same link

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -41,6 +41,29 @@ cfssl:
   securityContext: {}
   podSecurityContext: {}
 
+# alertdb - configurations for the alertdb Pod
+alertdb:
```
msenmurugan, line 45: Instead of alertdb, can we have it as postgres?

deployment/helm/synopsys-alert/templates/alertdb.yaml
```diff
@@ -0,0 +1,178 @@
+{{- if not .Values.alertdb.isExternal }}
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: alert
+    component: alertdb
+  name: {{ .Release.Name }}-alert-postgres
```
msenmurugan, line 8: To all the names, can you please remove _-alert_? This is because already in synopsysctl we are taking care of adding -alert to the release name.

And also the customer who uses Black Duck along with Alert, then need to mandatory add -alert to the release name in order to work. This is because you can't have 2 releases having same names for different apps in the same namespace.

deployment/helm/synopsys-alert/templates/alertdb.yaml
```diff
@@ -0,0 +1,178 @@
+{{- if not .Values.alertdb.isExternal }}
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: alert
```
msenmurugan, line 6: To all the labels, can you please add name to the label?

deployment/helm/synopsys-alert/templates/alertdb.yaml
```diff
@@ -0,0 +1,178 @@
...
+        image: {{ .Values.registry }}/postgresql-12-centos7:1
+        {{- end}}
+        imagePullPolicy: IfNotPresent
+        lifecycle:
+          preStop:
+            exec:
+              command:
+                - sh
+                - -c
+                - LD_LIBRARY_PATH=/opt/rh/rh-postgresql12/root/usr/lib64 /opt/rh/rh-postgresql12/root/usr/bin/pg_ctl -D /var/lib/pgsql/data/userdata -l logfile stop
```
msenmurugan, line 120: can you please change the -D to point to /var/lib/postgresql/data

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -41,6 +41,29 @@ cfssl:
...
+  registry: "docker.io/centos" # override the docker registry at container level
+  isExternal: true # false for running Postgres as a container and true for using External Postgres database
+  host: "localhost" # required only for external postgres, for postgres as a container, it will point to <name>-alert-postgres
+  port: 5432
+  # NOTE: do not change usernames when using postgres as a container; only configure if using external database
+  userUserName: sa
+  userPassword: blackduck
+  databaseName: alertdb
+  # pvc related parameters for postgres container
+  persistentVolumeClaimName: {} # set if you want to create your own PVC (used for migration)
```
msenmurugan, line 55: pls change it to persistentVolumeClaimName: . Remove {}.

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -41,6 +41,29 @@ cfssl:
...
   podSecurityContext: {}
 
+# alertdb - configurations for the alertdb Pod
+alertdb:
+  registry: "docker.io/centos" # override the docker registry at container level
+  isExternal: true # false for running Postgres as a container and true for using External Postgres database
+  host: "localhost" # required only for external postgres, for postgres as a container, it will point to <name>-alert-postgres
+  port: 5432
+  # NOTE: do not change usernames when using postgres as a container; only configure if using external database
+  userUserName: sa
```
msenmurugan, line 51: you won't be having 2 users in an app? one is an admin who can have access at DDL level and another user is having DML level access. The code need to use the DML level access. This is to avoid any SQL injection and also the permissions restricting to the user.

docker-entrypoint.sh
psantos1113: The alert container can start up before the database container.  If we can't connect to the database don't start alert because the alert container needs to read and write data to/from the database.

docker-entrypoint.sh
psantos1113: If we have an external database we will check to see if the alert database tables have been created in that database.  If they haven't then create the initial table schema into postgres.

docker-entrypoint.sh
gkillough: @psantos1113 I think he means a comment within the file.

src/main/java/com/synopsys/integration/alert/workflow/startup/component/LastSearchDataMigration.java
```diff
@@ -0,0 +1,91 @@
...
+@Component
+@Order(55)
+public class LastSearchDataMigration extends StartupComponent {
+    private static final Logger logger = LoggerFactory.getLogger(LastSearchDataMigration.class);
+    private static final String LAST_SEARCH_FILE = "blackduck-accumulator-task-last-search.txt";
+
+    // Because pg_stat_file requires admin privileges to run, we have to migrate the last search file data in code.
+    // We don't have admin privileges to postgres when alert starts.
+
+    private BlackDuckProviderKey providerKey;
```
gkillough, line 52: We can call this variable `blackDuckProviderKey` to make it more clear in the `initialize()` method that this is the only expected config in an upgrade.

src/main/java/com/synopsys/integration/alert/workflow/startup/component/LastSearchDataMigration.java
```diff
@@ -0,0 +1,91 @@
...
+import com.synopsys.integration.alert.common.persistence.accessor.ConfigurationAccessor;
+import com.synopsys.integration.alert.common.persistence.accessor.ProviderTaskPropertiesAccessor;
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
+import com.synopsys.integration.alert.common.persistence.util.FilePersistenceUtil;
+import com.synopsys.integration.alert.provider.blackduck.BlackDuckProviderKey;
+import com.synopsys.integration.alert.provider.blackduck.tasks.BlackDuckAccumulator;
+
+@Component
+@Order(55)
+public class LastSearchDataMigration extends StartupComponent {
```
gkillough, line 45: Let's also deprecate this file for removal in 8.0.0.

deployment/helm/synopsys-alert/values.yaml
jamesrichard91: Is there a problem with having this as alertdb? Is the name change just a preference?

deployment/helm/synopsys-alert/values.yaml
psantos1113: At the moment no.  We would need to test using the liquibase user and password properties for DDL access and the spring datasource properties for the DML access.  I will file a ticket for this.

deployment/helm/synopsys-alert/templates/alert.yaml
```diff
@@ -32,94 +32,101 @@ spec:
       name: {{ .Release.Name }}
```
msenmurugan, line 32: Can you pls add a checksum for Alert DB creds secrets?

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -41,6 +41,29 @@ cfssl:
   securityContext: {}
   podSecurityContext: {}
 
+# postgres - configurations for the postgres Pod
+postgres:
+  registry: "docker.io/centos" # override the docker registry at container level
+  isExternal: true # false for running Postgres as a container and true for using External Postgres database
+  host: "localhost" # required only for external postgres, for postgres as a container, it will point to <name>-postgres
```
msenmurugan, line 48: Whether it needs to be empty?

deployment/helm/synopsys-alert/values.yaml
msenmurugan: @jamesrichard91 just to match with the container names, so that customers won't get confused.

deployment/helm/synopsys-alert/values.yaml
mphammer: If you make it empty then you can use this to require users to set it if isExternal is true
```
{{ required "must provide --set postgres.host=\"\"" .Values.postgres.host }}
```

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -41,6 +41,29 @@ cfssl:
...
+  # NOTE: do not change usernames when using postgres as a container; only configure if using external database
+  userUserName: sa
+  userPassword: blackduck
+  databaseName: alertdb
+  # pvc related parameters for postgres container
+  persistentVolumeClaimName: "" # set if you want to create your own PVC (used for migration)
+  claimSize: "5Gi" # PVC claim size
+  storageClass: "" # PVC storage class name
+  volumeName: "" # existing persistent volume name backing the PVC
+  # path to postgresql initialization script
```
mphammer, line 59: Is there a missing value here for the postgresql initialization script?

psantos1113: we don't have one.  I will remove the comment.  The alert container handles the initialization when it starts up.

deployment/helm/synopsys-alert/values.yaml
psantos1113: I will add this

psantos1113: With the required there the postgres.isExternal must be false by default.  Which is to create a postgres container in the deployment.  That will be the default for alert.

PR #969, opened 2020-05-01
Jira app check

jamesrichard91:


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/util/BlackDuckResponseCache.java
```diff
@@ -64,7 +64,10 @@ public BlackDuckResponseCache(BlackDuckBucketService blackDuckBucketService, Bla
             Future<Optional<T>> optionalProjectVersionFuture = blackDuckBucketService.addToTheBucket(bucket, url, responseClass);
             if (bucket.hasAnyErrors()) {
                 Optional<Exception> error = bucket.getError(url);
-                error.ifPresent(exception -> logger.error(String.format("There was a problem retrieving the link '%s'.", url), exception));
+                error.ifPresent(exception -> {
+                    logger.debug(String.format("There was a problem retrieving the link '%s'. Error: %s", url, exception.getMessage()));
+                    logger.trace(exception.getMessage(), exception);
+                });
             }
```
jamesrichard91, line 71: Changed this logging because this was logging error messages and stack traces for components that were deleted. So I changed the logging to DEBUG and changed the stack trace to be logged at TRACE so that the User won't be confused and so we can still get the information if needed.

PR #972, opened 2020-05-05
Mc alertdb tests continued

ChomickiM:
Tests for DefaultCustomCertificateAccessor and ProviderTaskProperties. Included in this PR is a new method for Mocking the JPA repository. This was done to simplify the other repositories by only having to implement the required methods and overloading whichever ones are required for the tests to run.

These Mock JPA repository changes were also pushed to the previously implemented repositories.


alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+public class DefaultCustomCertificateAccessorTest {
+    private final String alias = "alias-test";
+    private final String content = "content-test";
+
+    private CustomCertificateModel expectedCustomCertificateModel = new CustomCertificateModel(alias, content, new Date().toString());
+
+    @Test
+    public void getCertificatesTest() {
+        CustomCertificateRepository customCertificateRepository = new MockCustomCertificateRepository(alias, content, new Date());
```
gkillough, line 27: `new Date()` will have a different underlying _time_ every time it is called. We should be careful when comparing these.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -0,0 +1,174 @@
...
+import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
+import com.synopsys.integration.alert.common.persistence.model.CustomCertificateModel;
+import com.synopsys.integration.alert.database.api.mock.MockCustomCertificateRepository;
+import com.synopsys.integration.alert.database.certificates.CustomCertificateRepository;
+
+public class DefaultCustomCertificateAccessorTest {
+    private final String alias = "alias-test";
+    private final String content = "content-test";
+
+    private CustomCertificateModel expectedCustomCertificateModel = new CustomCertificateModel(alias, content, new Date().toString());
```
jamesrichard91, line 23: If we want to compare the updated times, we should move the Date to its own field. And in the tests we can create a new Date if we want to use a newer time, and use the Date field when they should be the same.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
ChomickiM: when I was doing my comparisons I wasn't quite sure how to compare the time stamps so I decided to just compare on the alias and content. Instead, I could create a Date object like I did with alias and content and pass it throughout if you think thats a better alternative

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -0,0 +1,174 @@
...
+        List<CustomCertificateModel> customCertificateModelList = customCertificateAccessor.getCertificates();
+
+        assertEquals(1, customCertificateModelList.size());
+        CustomCertificateModel customCertificateModel = customCertificateModelList.get(0);
+        testCustomCertificateModel(expectedCustomCertificateModel, customCertificateModel);
+
+        customCertificateAccessor.deleteCertificate(alias);
+        customCertificateModelList = customCertificateAccessor.getCertificates();
+
+        assertEquals(0, customCertificateModelList.size());
```
bamandel, line 120: This could be an isEmpty check

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -0,0 +1,174 @@
...
+        List<CustomCertificateModel> customCertificateModelList = customCertificateAccessor.getCertificates();
+
+        assertEquals(1, customCertificateModelList.size());
+        CustomCertificateModel customCertificateModel = customCertificateModelList.get(0);
+        testCustomCertificateModel(expectedCustomCertificateModel, customCertificateModel);
+
+        customCertificateAccessor.deleteCertificate(0L);
+        customCertificateModelList = customCertificateAccessor.getCertificates();
+
+        assertEquals(0, customCertificateModelList.size());
```
bamandel, line 148: Could be isEmpty check.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/DefaultMockJPARepository.java
```diff
@@ -0,0 +1,140 @@
...
+import org.springframework.data.domain.Example;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+public class DefaultMockJPARepository<T, ID> implements JpaRepository<T,ID> {
+    // Empty implementation of a JPA repository to be extended by other mock repositories.
+    // The child classes that extend this can then override the methods they need to
+    // provide the correct testing functionality needed.
```
bamandel, line 15: These lines may not be necessary if you made the class abstract. Having the keyword there would others realize that the class should be extended.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -0,0 +1,174 @@
...
+    public void storeCertificateIdDoesNotExistTest() throws Exception {
+        CustomCertificateRepository customCertificateRepository = new MockCustomCertificateRepository();
+        DefaultCustomCertificateAccessor customCertificateAccessor = new DefaultCustomCertificateAccessor(customCertificateRepository);
+
+        try {
+            CustomCertificateModel certificateModel = new CustomCertificateModel(9L, alias, content, new Date().toString());
+            customCertificateAccessor.storeCertificate(certificateModel);
+            fail();
+        } catch (AlertDatabaseConstraintException e) {
+        }
```
jamesrichard91, line 104: Can you add a comment in the exception block or an assertion that the exception is not null. That way other people can see that we expect that exception and it wasn't a mistake.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -0,0 +1,174 @@
...
+    @Test
+    public void deleteCertificateByAliasBlankTest() throws Exception {
+        CustomCertificateRepository customCertificateRepository = new MockCustomCertificateRepository(alias, content, new Date());
+        DefaultCustomCertificateAccessor customCertificateAccessor = new DefaultCustomCertificateAccessor(customCertificateRepository);
+
+        try {
+            customCertificateAccessor.deleteCertificate("");
+            fail();
+        } catch (AlertDatabaseConstraintException e) {
+        }
```
jamesrichard91, line 132: Same as above.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/DefaultMockJPARepository.java
```diff
@@ -0,0 +1,140 @@
...
+    // To use the class correctly have a child class extend this class and also implement the JPA Repository
+    // interface to be tested.
+    //
+    // For example creating a mock for SystemStatusRepository
+    // class MockSystemStatusRepository extends DefaultMockJPARepository<SystemStatus, Long> implements SystemStatusRepository
```
bamandel, line 20: Generally, this sort of explanation would go above the class declaration.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/MockCustomCertificateRepository.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public MockCustomCertificateRepository() {
+
+    }
+
+    public MockCustomCertificateRepository(String alias, String content, Date lastUpdated) {
+        CustomCertificateEntity customCertificateEntity = new CustomCertificateEntity(alias, content, lastUpdated);
+        customCertificateEntity.setId(currentId);
+        customCertificateEntityMapById.put(currentId, customCertificateEntity);
+        currentId++;
```
jamesrichard91, line 27: You could replace these 3 lines by calling the save method.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultProviderTaskPropertiesAccessorTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test
+    public void setTaskPropertyExceptionTest() throws Exception {
+        DefaultProviderTaskPropertiesAccessor providerTaskPropertiesAccessor = new DefaultProviderTaskPropertiesAccessor(providerTaskPropertiesRepository);
+
+        try {
+            providerTaskPropertiesAccessor.setTaskProperty(null, "", "", "");
+            fail();
+        } catch (AlertDatabaseConstraintException e) {
+        }
```
jamesrichard91, line 72: Should add something in the exception block.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/mock/MockProviderTaskPropertiesRepository.java
```diff
@@ -0,0 +1,32 @@
...
+import com.synopsys.integration.alert.database.provider.task.ProviderTaskPropertiesEntity;
+import com.synopsys.integration.alert.database.provider.task.ProviderTaskPropertiesRepository;
+
+public class MockProviderTaskPropertiesRepository extends DefaultMockJPARepository<ProviderTaskPropertiesEntity, Long> implements ProviderTaskPropertiesRepository {
+
+    private Map<Long, ProviderTaskPropertiesEntity> providerTaskPropertiesEntityMap = new HashMap<>();
+
+    public MockProviderTaskPropertiesRepository(Long providerConfigId, String taskName, String propertyName, String value) {
+        ProviderTaskPropertiesEntity providerTaskPropertiesEntity = new ProviderTaskPropertiesEntity(providerConfigId, taskName, propertyName, value);
+        providerTaskPropertiesEntityMap.put(providerConfigId, providerTaskPropertiesEntity);
```
jamesrichard91, line 16: Should call the save method here, in case we need to add additional behavior for saving entities in this Mock class in the future.

alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultCustomCertificateAccessorTest.java
```diff
@@ -50,7 +53,7 @@ public void getCertificateTest() {
 
     @Test
     public void storeCertificateTest() throws Exception {
-        CustomCertificateModel certificateModel = new CustomCertificateModel(alias, content, new Date().toString());
+        CustomCertificateModel certificateModel = new CustomCertificateModel(alias, content, testDate.toString());
```
jamesrichard91, line 56: Why are we using the DateUtils above and the .toString() method here? We should be more consistent and choose one of these approaches.

PR #976, opened 2020-05-12
Fixing UIConfig, and TableDisplay usage

jamesrichard91:
Fixing the config field retrieval so state changes don't have lasting impacts. Updating the UI to make sure the correct descriptor is used.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
@@ -59,6 +59,10 @@ public abstract class UIConfig extends Stringable {
     protected abstract List<ConfigField> createFields();
 
     public List<ConfigField> getFields() {
+        return configFields;
+    }
+
+    public List<ConfigField> getMetadataFields() {
         return configFields.stream().map(SerializationUtils::clone).collect(Collectors.toList());
     }
 
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
@@ -23,9 +23,11 @@
 package com.synopsys.integration.alert.common.descriptor.config.ui;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.annotation.PostConstruct;
 
+import org.apache.commons.lang3.SerializationUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import com.synopsys.integration.alert.common.descriptor.config.field.ConfigField;
@@ -57,7 +59,7 @@ public abstract class UIConfig extends Stringable {
     protected abstract List<ConfigField> createFields();
 
     public List<ConfigField> getFields() {
-        return configFields;
+        return configFields.stream().map(SerializationUtils::clone).collect(Collectors.toList());
     }
 
     public final boolean hasFields() {
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
@@ -58,10 +58,16 @@ public abstract class UIConfig extends Stringable {
 
     protected abstract List<ConfigField> createFields();
 
+    /**
+     * This should be used when validating the fields.
+     */
     public List<ConfigField> getFields() {
         return configFields;
     }
 
+    /**
+     * This list of ConfigFields is used to define what is shown in the UI. This does not copy the validation functions.
+     */
     public List<ConfigField> getMetadataFields() {
         return configFields.stream().map(SerializationUtils::clone).collect(Collectors.toList());
     }
--- a/src/main/js/dynamic/loaded/certificates/CertificatesPage.js
+++ b/src/main/js/dynamic/loaded/certificates/CertificatesPage.js
@@ -7,6 +7,7 @@ import TableDisplay from 'field/TableDisplay';
 import TextInput from 'field/input/TextInput';
 import TextArea from 'field/input/TextArea';
 import ReadOnlyField from 'field/ReadOnlyField';
+import * as DescriptorUtilities from 'util/descriptorUtilities';
 
 class CertificatesPage extends Component {
     constructor(props) {
@@ -146,8 +147,17 @@ class CertificatesPage extends Component {
     }
 
     render() {
-        const { fetching, inProgress, certificates, certificateDeleteError, label, description, fieldErrors } = this.props;
+        const { fetching, inProgress, certificates, certificateDeleteError, label, description, fieldErrors, descriptors } = this.props;
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, DescriptorUtilities.DESCRIPTOR_NAME.COMPONENT_CERTIFICATES, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        let foundDescriptor = null;
+        if (descriptorList && descriptorList.length > 0) {
+            foundDescriptor = descriptorList[0];
+        }
+
         const hasFieldErrors = fieldErrors && Object.keys(fieldErrors).length > 0;
+        const canCreate = DescriptorUtilities.isOperationAssigned(foundDescriptor, DescriptorUtilities.OPERATIONS.CREATE);
+        const canDelete = DescriptorUtilities.isOperationAssigned(foundDescriptor, DescriptorUtilities.OPERATIONS.DELETE);
+        const canSave = DescriptorUtilities.isOperationAssigned(foundDescriptor, DescriptorUtilities.OPERATIONS.WRITE);
         return (
             <div>
                 <div>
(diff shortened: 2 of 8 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/UIConfig.java
```diff
@@ -57,7 +59,7 @@ public void setConfigFields() {
     protected abstract List<ConfigField> createFields();
 
     public List<ConfigField> getFields() {
-        return configFields;
+        return configFields.stream().map(SerializationUtils::clone).collect(Collectors.toList());
```
jamesrichard91, line 62: For User's that have READONLY access we were changing the state of the fields. This change was being persisted even when you switched user's. So now we create a copy of the list to prevent the changes from effecting the in memory fields.

src/main/js/dynamic/loaded/users/UserManagement.js
```diff
@@ -12,15 +12,18 @@ class UserManagement extends Component {
...
-        this.state = { descriptor };
     }
 
     render() {
-        const { descriptor } = this.state;
-        const { label, description } = this.props;
-        const canCreate = DescriptorUtilities.isOperationAssigned(descriptor, DescriptorUtilities.OPERATIONS.CREATE);
-        const canDelete = DescriptorUtilities.isOperationAssigned(descriptor, DescriptorUtilities.OPERATIONS.DELETE);
+        const { label, description, descriptors } = this.props;
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, DescriptorUtilities.DESCRIPTOR_NAME.COMPONENT_USERS, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
```
psantos1113, line 19: Should we have a findSingleDescriptorByNameAndContext or modify findDescriptorByNameAndContext to perform encapsulate this code into the method so that we don't have to copy and paste the if check?

Add this code to the findDescriptorByNameAndContext or create a new method:
`
let foundDescriptor = null;
if(descriptorList && descriptorList.length > 0) {
  foundDescriptor = descriptorLIst[0];
}
`

src/main/js/providers/ProviderTable.js
```diff
@@ -59,16 +58,18 @@ class ProviderTable extends Component {
...
+            const foundDescriptor = descriptorList[0];
+            const emptyConfig = FieldModelUtilities.createFieldModelWithDefaults(foundDescriptor.fields, DescriptorUtilities.CONTEXT_TYPE.GLOBAL, foundDescriptor.name);
+            const updatedFieldModel = FieldModelUtilities.combineFieldModels(emptyConfig, providerConfig);
+            if (providerConfig.id) {
+                updatedFieldModel.id = providerConfig.id;
+            }
+            return updatedFieldModel;
         }
-        return updatedFieldModel;
+        return {};
```
bamandel, line 72: Should this return an empty object? from the function name, I imagine this should return at least the original model without defaults.

src/main/js/providers/ProviderTable.js
```diff
@@ -133,9 +134,11 @@ class ProviderTable extends Component {
     }
 
     retrieveData() {
-        const { descriptor } = this.state;
-        if (descriptor) {
-            this.props.getAllConfigs(descriptor.name);
+        const { descriptors, descriptorName } = this.props
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, descriptorName, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        if (descriptorList && descriptorList.length > 0) {
+            const foundDescriptor = descriptorList[0];
```
bamandel, line 140: The lines Paulo mentioned above is used through out the PR. The new function can be used here as well

src/main/js/providers/ProviderTable.js
```diff
@@ -180,14 +183,16 @@ class ProviderTable extends Component {
...
     createModalFields() {
-        const { providerConfig, descriptor } = this.state;
-        const { fieldErrors } = this.props;
+        const { providerConfig } = this.state;
+        const { fieldErrors, descriptors, descriptorName } = this.props;
         const newConfig = this.combineModelWithDefaults(providerConfig);
-        if (descriptor) {
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, descriptorName, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        if (descriptorList && descriptorList.length > 0) {
+            const foundDescriptor = descriptorList[0];
```
bamandel, line 191: I'll just mark them all for you

src/main/js/providers/ProviderTable.js
```diff
@@ -212,9 +217,11 @@ class ProviderTable extends Component {
         const { id } = selectedRow;
         const { providerConfigs } = this.props;
         let selectedConfig = providerConfigs.find(config => config.id === id);
-        const { descriptor } = this.state;
-        if (descriptor) {
-            descriptor.fields.forEach(field => {
+        const { descriptors, descriptorName } = this.props
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, descriptorName, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        if (descriptorList && descriptorList.length > 0) {
+            const foundDescriptor = descriptorList[0];
```
bamandel, line 223: here

src/main/js/providers/ProviderTable.js
```diff
@@ -59,16 +58,18 @@ class ProviderTable extends Component {
...
-            return {};
-        }
-        const emptyConfig = FieldModelUtilities.createFieldModelWithDefaults(descriptor.fields, DescriptorUtilities.CONTEXT_TYPE.GLOBAL, descriptor.name);
-        const updatedFieldModel = FieldModelUtilities.combineFieldModels(emptyConfig, providerConfig);
-        if (providerConfig.id) {
-            updatedFieldModel.id = providerConfig.id;
+        const { descriptors, descriptorName } = this.props
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, descriptorName, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        if (descriptorList && descriptorList.length > 0) {
+            const foundDescriptor = descriptorList[0];
```
bamandel, line 64: here

src/main/js/providers/ProviderTable.js
```diff
@@ -33,22 +33,21 @@ class ProviderTable extends Component {
...
 
     componentDidMount() {
-        const descriptor = this.props.descriptors.find(descriptor => descriptor.name === this.props.descriptorName
-            && descriptor.context === DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
-        if (descriptor) {
-            const emptyConfig = FieldModelUtilities.createFieldModelWithDefaults(descriptor.fields, DescriptorUtilities.CONTEXT_TYPE.GLOBAL, descriptor.name);
+        const { descriptors, descriptorName } = this.props
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, descriptorName, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        if (descriptorList && descriptorList.length > 0) {
+            const foundDescriptor = descriptorList[0];
```
bamandel, line 45: here

src/main/js/providers/ProviderTable.js
```diff
@@ -264,20 +260,25 @@ class ProviderTable extends Component {
...
 
     render() {
-        const { descriptor } = this.state;
-        const { providerConfigs, descriptorFetching, configFetching, testInProgress, updateStatus, fieldErrors, errorMessage, actionMessage } = this.props;
-        const descriptorHeader = descriptor && (
+        const { providerConfigs, descriptorFetching, configFetching, testInProgress, updateStatus, fieldErrors, errorMessage, actionMessage, descriptors, descriptorName } = this.props;
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, descriptorName, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        let foundDescriptor = null;
+        if (descriptorList && descriptorList.length > 0) {
+            foundDescriptor = descriptorList[0];
```
bamandel, line 267: here

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
```diff
@@ -146,8 +147,17 @@ class CertificatesPage extends Component {
     }
 
     render() {
-        const { fetching, inProgress, certificates, certificateDeleteError, label, description, fieldErrors } = this.props;
+        const { fetching, inProgress, certificates, certificateDeleteError, label, description, fieldErrors, descriptors } = this.props;
+        const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, DescriptorUtilities.DESCRIPTOR_NAME.COMPONENT_CERTIFICATES, DescriptorUtilities.CONTEXT_TYPE.GLOBAL)
+        let foundDescriptor = null;
+        if (descriptorList && descriptorList.length > 0) {
+            foundDescriptor = descriptorList[0];
```
bamandel, line 154: here

src/main/js/providers/ProviderTable.js
jamesrichard91: Previously we had 
``` java
 if (!descriptor) {
            return {};
        }
```
So this maintains that behavior.

PR #982, opened 2020-05-19
upgrade validation fix

psantos1113:
Added debug logging for validating fields.
Made all static loggers a private final member to avoid memory leaks not being able to garbage collect a class due to the static logger reference.  The Logger Factory handles caching the loggers so it makes the static keyword obsolete for performance gains.
Update the liquibase script to insert the default configuration name.


alert-database/src/main/resources/liquibase/6.0.0/provider-fields.xml
```diff
@@ -103,7 +103,7 @@
                     WHERE type_id IN (
                         SELECT id
                         FROM alert.descriptor_types
-                        WHERE type = 'CHANNEL'
+                        WHERE type = 'PROVIDER'
```
psantos1113, line 106: This is the actual fix for the 3 upgrade issues that were found.  The config name was being added the channel distribution configurations and not the provider distribution configuration.

src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
```diff
@@ -36,31 +38,38 @@
 
 @Component
 public class FieldValidationAction {
-
+    private final Logger logger = LoggerFactory.getLogger(FieldValidationAction.class);
```
psantos1113, line 41: Added debug logging to diagnose the issues with the jobs.  If we have a problem with configurations being invalid we can turn on debug logging and get more information.

alert-database/src/main/resources/liquibase/6.0.0/provider-fields.xml
gkillough: Ugh. How did I mess that up?

PR #985, opened 2020-05-21
 helm docs

psantos1113:
Updated helm installation README file.  You can see how the changes look by navigating here.

https://github.com/blackducksoftware/blackduck-alert/blob/ps_helm_docs/deployment/helm/synopsys-alert/README.md

### DIFF
--- a/deployment/helm/synopsys-alert/README.md
+++ b/deployment/helm/synopsys-alert/README.md
@@ -210,7 +210,7 @@ You must have a claim created for the 'alert' service regardless of an on-premis
 For the on-premise database deployment a second Persistent Volume Claim must be created to store the database data.
 
 ##### Alert Volume Claim Configuration
-- A persistent volume must be created first
+- A persistent volume must be created first if a dynamic provisioner not being used
 - A persistent volume claim must be created in the same namespace as the Alert deployment and bound to the persistent volume for Alert
 - Configure Alert to use the volume claim
 
@@ -238,7 +238,7 @@ For the on-premise database deployment a second Persistent Volume Claim must be 
   ```
     
 ##### Using On-Premise Database
-- A persistent volume must be created first for the Postgres database
+- A persistent volume must be created first for the Postgres database if a dynamic provisioner not being used
 - A persistent volume claim must be created in the same namespace as the Alert deployment for Postgres and bound to the persistent volume for Postgres
 - Configure Postgres to use the volume claim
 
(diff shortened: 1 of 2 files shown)


deployment/helm/synopsys-alert/README.md
```diff
@@ -130,3 +190,237 @@ Alternatively, a YAML file that specifies the values for the above parameters ca
 ```console
 $ helm install . --name <name> --namespace <namespace> --set enableStandalone=true
 ```
+
+### Persistent Storage
+The section will describe the changes needed to configure persistent storage.  
+**Please note:** By default the 'values.yaml' file has persistent storage disabled in order for Alert to work out of the box.
```
mphammer, line 196: It's actually enabled right now `enablePersistentStorage: true` by default

deployment/helm/synopsys-alert/README.md
```diff
@@ -29,36 +43,63 @@ $ helm install . --name <name> --namespace <namespace>
...
+  `<INTERNAL_PORT>:<EXTERNAL_PORT>/TCP`
+  
+  For example:
+  
+  `8443:31594/TCP`
+  
+  The internal port is 8443 and the external port is 31594. When accessing the Alert UI the external port will be used in the URL.
+  Once the external port is identified the URL to access the UI will be in the following format:
+  
+  `https://my.kubernetes.host:<EXTERNAL_PORT>/alert`
```
mphammer, line 97: I believe this needs to be `https://<EXTERNAL_NODE_IP>:<EXTERNAL_PORT>/alert`. Users can get that information by doing a `kubectl get nodes -o wide` or by setting up an ingress.

deployment/helm/synopsys-alert/README.md
psantos1113: Yes I would like it to be false so that Alert can start up right out of the box.  The second file changed in this PR sets the value to false.

deployment/helm/synopsys-alert/README.md
psantos1113: I will update this.

deployment/helm/synopsys-alert/README.md
```diff
@@ -130,3 +190,237 @@ Alternatively, a YAML file that specifies the values for the above parameters ca
...
+  ```yaml
+  enablePersistentStorage: true 
+  ```
+#### With Storage Claims
+This section defines configuration using Persistent Volume Claims.  Claims can be optionally used rather than just a Persistent volume.
+You must have a claim created for the 'alert' service regardless of an on-premise or external database.
+For the on-premise database deployment a second Persistent Volume Claim must be created to store the database data.
+
+##### Alert Volume Claim Configuration
+- A persistent volume must be created first
```
mphammer, line 209: It only needs to be created first if they don't have a dynamic provisioner - which will see the pvc and create a persistent volume for it

PR #987, opened 2020-05-22
Build changes

jamesrichard91:
Moving the runServer lines into the RunServerTask. Moving the createKeystore lines into CreateKeystoreTask. Adding helm validation tasks and cleanup. Updating the deployment files to exclude the helm tmp files.

I don't have helm setup locally so I have not yet tested the helm tasks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,6 @@ mainClassName = 'com.synopsys.integration.alert.Application'
 
 version = '6.0.0-SIGQA8-SNAPSHOT'
 def testContainerPostgresVersion = '12.2'
-def reuseContainer = 'false'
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -266,6 +211,32 @@ task cleanBundles(type: Delete) {
 
 tasks.clean.finalizedBy(cleanBundles)
 
+task helmLint(type: Exec) {
```
psantos1113, line 214: If you can, can you have these run after the distributions task?
The distributions should create the deployment directory under the build directory.  You could then point the validations to the directory:
` project.buildDir/deployment/helm/synopsys-alert`
Could be something to look into after this release.

PR #992, opened 2020-05-26
Separating gradle tasks into their own files

jamesrichard91:


buildSrc/docker.gradle
```diff
@@ -0,0 +1,55 @@
...
+
+task dockerLogin(type: Exec) {
+    commandLine "docker", "login", "--username", "$System.env.DOCKER_INT_BLACKDUCK_USER", "--password", "$System.env.DOCKER_INT_BLACKDUCK_PASSWORD"
+}
+
+task pushImage(type: Exec, dependsOn: [buildDockerImage]) {
+    commandLine "docker", "push", "blackducksoftware/${project.name}:${project.version}"
+}
+
+task all(dependsOn: [build, helmValidation, createDeploymentZip, dockerLogin, buildDockerImage, pushImage]) {
```
psantos1113, line 47: Should this task be in the main gradle file since it is calling tasks from other gradle files?

jamesrichard91: Good suggestion! That also removed the inter-dependencies of the task files.

build.gradle
```diff
@@ -22,21 +22,28 @@ plugins {
 }
 
 // Copied from src/test/java - com.synopsys.integration.alert.TestTags
```
gkillough, line 24: Let's move this comment down to the line declaring the tags.

PR #993, opened 2020-05-26
duplicate email fix

psantos1113:
Need to remove old user and project relations.  We were never removing old relations only adding new relations.  Using similar logic to updating the projects where we determine what to add and what to remove.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
@@ -177,6 +177,8 @@ public class DefaultProviderDataAccessor implements ProviderDataAccessor {
         }
         logger.debug("Adding {} user relations to project {} ", userRelationsToAdd.size(), project.getName());
         logger.debug("Removing {} user relations from project {} ", userRelationsToRemove.size(), project.getName());
+        providerUserProjectRelationRepository.saveAll(userRelationsToAdd);
+        providerUserProjectRelationRepository.deleteAll(userRelationsToRemove);
     }
 
     private List<ProviderUserModel> saveUsers(Long providerConfigId, Collection<ProviderUserModel> users) {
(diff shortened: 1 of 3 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -163,12 +164,21 @@ private void mapUsersToProjectByEmail(Long providerConfigId, String projectHref,
...
                                             .orElseThrow(() -> new AlertDatabaseConstraintException("A project with the following href did not exist: " + projectHref));
         Long projectId = project.getId();
+        List<ProviderUserProjectRelation> userRelationsToRemove = providerUserProjectRelationRepository.findByProviderProjectId(projectId);
+        List<ProviderUserProjectRelation> userRelationsToAdd = new LinkedList<>();
         for (String emailAddress : emailAddresses) {
-            providerUserRepository.findByEmailAddressAndProviderConfigId(emailAddress, providerConfigId)
-                .stream()
-                .map(ProviderUserEntity::getId)
-                .forEach(userId -> providerUserProjectRelationRepository.save(new ProviderUserProjectRelation(userId, projectId)));
+            List<ProviderUserEntity> providerUserEntities = providerUserRepository.findByEmailAddressAndProviderConfigId(emailAddress, providerConfigId);
```
gkillough, line 170: Why does `findByEmailAddressAndProviderConfigId(...)` return a `List`? Shouldn't it be an optional? 
I think what we want is a `findByEmailAddressInAndProviderConfigId(List<String> emailAddresses, Long providerConfigId);

psantos1113: The same email address could be a user in multiple configs so it returns a list per the API.  I will add the in keyword and test it out.

gkillough: Right, but we are filtering the config, so that should be a unique combination, no?

psantos1113: You can also have multiple blackduck users with the same email address.  We do that for testing. The username is unique in Black Duck but not the email address when it is a local user.  That is why we need a list.

gkillough: Gotcha. Regardless, we should make this a single query.

PR #996, opened 2020-05-28
ldap password fix

psantos1113:
For sensitive fields read the values from the database and use the value if the value from the field model isn't set.

We were writing the empty string as the LDAP password each time Alert starts up. Read if the value is saved in the database and use that if the field isn't set in the new configuration to be saved.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/DefaultDescriptorGlobalConfigUtility.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/DefaultDescriptorGlobalConfigUtility.java
@@ -23,7 +23,9 @@
 package com.synopsys.integration.alert.common.descriptor.accessor;
 
 import java.util.Collection;
+import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/DefaultDescriptorGlobalConfigUtility.java
```diff
@@ -96,9 +98,33 @@ public FieldModel save(FieldModel fieldModel) throws AlertException {
...
+            Map<String, ConfigurationFieldModel> updatedValues = updateSensitiveFields(valueMap, existingConfig.get());
+            configurationModel = configurationAccessor.updateConfiguration(id, updatedValues.values());
+        } else {
+            configurationModel = configurationAccessor.createConfiguration(key, context, valueMap.values());
+        }
         FieldModel convertedFieldModel = configurationFieldModelConverter.convertToFieldModel(configurationModel);
         return apiAction.afterUpdateAction(convertedFieldModel);
     }
+
+    private Map<String, ConfigurationFieldModel> updateSensitiveFields(Map<String, ConfigurationFieldModel> values, ConfigurationModel existingConfig) {
```
gkillough, line 114: I'm surprised we don't have a utility for this. We should probably add a TODO here and an issue for 6.1.0

psantos1113: FieldModelProcessor will handle these changes.  The problem is that FieldModelProcessor is in the alert-main project and this utility is in the alert-common sub project.  Using FieldModelProcessor would create a dependency we don't want.  If we can move FieldModelProcessor into alert-common then that would resolve this issue.  I will add a TODO comment.

PR #1002, opened 2020-06-03
Add repository flush call to save method.

psantos1113:
Add a flush call before handleEvent in order to ensure the notifications are saved in the database and can be read via notification ID's in the real time event handler.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
@@ -90,7 +90,7 @@ public class DefaultNotificationManager implements NotificationManager {
                                                        .map(this::toModel)
                                                        .collect(Collectors.toList());
         notificationContentRepository.flush();
-
+        // TODO move this sendEvent call out of this class.  We can then remove the flush call since the save will be in a transaction.
         if (!savedModels.isEmpty()) {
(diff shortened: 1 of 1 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationManager.java
```diff
@@ -89,6 +89,7 @@ public DefaultNotificationManager(NotificationContentRepository notificationCont
                                                        .stream()
                                                        .map(this::toModel)
                                                        .collect(Collectors.toList());
+        notificationContentRepository.flush();
```
gkillough, line 92: Can we also add a TODO to move the event sending out of the database api?

jamesrichard91: I agree with Gavin, EventManager should be pulled out of the the NotificationManager. That would allow the transaction to save the notifications to complete before sending the events and would remove the need for the flush. But that can be done as a future improvement.

psantos1113: I added the TODO.  This morning I tried refactoring the NotificationManager to refactor the transaction boundaries.  But that wasn't working so I opted for the simple approach for now.  This should be refactored in the future.

PR #1009, opened 2020-06-08
Create and separate a mutable and immutable version of the ConfigurationModel class

ChomickiM:
After discussing with Gavin, we decided rather to use a builder to instead separate the model into having a mutable and immutable version, where the mutable is a subclass of main ConfigurationModel by removing the put method. 

Let me know if there are any questions related to the refactor or tests.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ConfigurationModel.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ConfigurationModel.java
@@ -28,13 +28,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import com.synopsys.integration.alert.common.enumeration.ConfigContextEnum;
+import com.synopsys.integration.alert.common.persistence.model.mutable.ConfigurationModelMutable;
 import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 
-public final class ConfigurationModel extends AlertSerializableModel {
+public class ConfigurationModel extends AlertSerializableModel {
     private final Long descriptorId;
     private final Long configurationId;
     private final String createdAt;
@@ -53,6 +52,10 @@ public final class ConfigurationModel extends AlertSerializableModel {
         this.lastUpdated = lastUpdated;
         this.context = context;
         configuredFields = new HashMap<>();
+    }
+
+    protected Map<String, ConfigurationFieldModel> getConfiguredFields() {
+        return configuredFields;
     }
 
     public Long getDescriptorId() {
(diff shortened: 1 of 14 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ConfigurationModel.java
```diff
@@ -88,20 +91,7 @@ public ConfigContextEnum getDescriptorContext() {
...
-            oldConfigField.setFieldValues(values);
-        } else {
-            configuredFields.put(fieldKey, configFieldModel);
-        }
-    }
-
-    private List<String> combine(ConfigurationFieldModel first, ConfigurationFieldModel second) {
-        return Stream.concat(first.getFieldValues().stream(), second.getFieldValues().stream()).collect(Collectors.toList());
+    public ConfigurationModelMutable createMutableCopy() {
+        return new ConfigurationModelMutable(descriptorId, configurationId, createdAt, lastUpdated, context);
```
gkillough, line 95: We also need the `configuredFields` from `this` set on the new instance. You could do this through the constructor or you could do something like this: 
```java
public ConfigurationModelMutable createMutableCopy() {
    ConfigurationModelMutable mutableCopy = new ConfigurationModelMutable(descriptorId, configurationId, createdAt, lastUpdated, context);
    mutableCopy.getConfiguredFields().putAll(configuredFields);
    return mutableCopy;
}
``` 
or (not 100% sure this will work)
```java
public ConfigurationModelMutable createMutableCopy() {
    ConfigurationModelMutable mutableCopy = new ConfigurationModelMutable(descriptorId, configurationId, createdAt, lastUpdated, context);
    mutableCopy.configuredFields = configuredFields;
    return mutableCopy;
}
```

alert-common/src/test/java/com/synopsys/integration/alert/common/persistence/model/ConfigurationModelTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test
+    public void getCopyOfKeyToFieldMapTest() {
+        ConfigurationModel configurationModel = createConfigurationModel();
+        assertTrue(configurationModel.getCopyOfKeyToFieldMap().isEmpty());
+    }
+
+    @Test
+    public void createMutableCopyTest() {
+        ConfigurationModel configurationModel = createConfigurationModel();
```
gkillough, line 70: Let's add a field or two to `configurationModel` to make sure that the fields are copied.

gkillough: If all of its constructors create the `Map` maybe we should consider passing it as a parameter in one of them.

ChomickiM: configurationModel does not have access to the put method so it cannot add a configuredField. I would be able to add a field on the configurationModelMutable but that test would be covered in ConfigurationModelMutableTest.

gkillough: Ok, that makes me think we will need to update the constructor(s) in `ConfigurationModel` then.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ConfigurationModel.java
psantos1113: I like the putAll approach better.  The equals operator would assign the mutable to the same object reference as the immutable instance.

PR #1011, opened 2020-06-09
Flavored Markdown Fixes

gkillough:


gkillough:
Using GitHub's "Display rich diff" feature might be helpful here.

PR #1014, opened 2020-06-11
Refactor validation functions into a ValidationResult object

ChomickiM:
Previously we used a Collection of strings to handle error messages produced by the validators. In this refactor I instead created a ValidationResult object to instead be used with several other convenience methods.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
@@ -150,13 +150,13 @@ public class UploadEndpointManager {
             if (validationFunction.isPresent()) {
                 writeFile(tempFilename, fileResource);
                 File fileToValidate = filePersistenceUtil.createUploadsFile(tempFilename);
-                ValidationResult errors = validationFunction.get().apply(fileToValidate);
+                ValidationResult validationResult = validationFunction.get().apply(fileToValidate);
                 filePersistenceUtil.delete(fileToValidate);
-                if (!errors.hasErrors()) {
+                if (!validationResult.hasErrors()) {
                     writeFile(targetFilename, fileResource);
                     return responseFactory.createCreatedResponse("", "File uploaded.");
                 }
-                return responseFactory.createBadRequestResponse("", errors.combineErrorMessages());
+                return responseFactory.createBadRequestResponse("", validationResult.combineErrorMessages());
             } else {
                 writeFile(targetFilename, fileResource);
                 return responseFactory.createCreatedResponse("", "File uploaded.");
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
@@ -25,12 +25,10 @@ package com.synopsys.integration.alert.common.action;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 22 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/ConfigField.java
```diff
@@ -194,22 +195,21 @@ public ConfigField applyValidationFunctions(ConfigValidationFunction... validati
...
-        validateLength(fieldToValidate, errors);
-        if (errors.isEmpty()) {
+    private ValidationResult validate(FieldValueModel fieldToValidate, FieldModel fieldModel, List<ConfigValidationFunction> validationFunctions) {
+        ValidationResult errors = ValidationResult.of(validateRequiredField(fieldToValidate), validateLength(fieldToValidate));
+
+        if (!errors.hasErrors()) {
             for (ConfigValidationFunction validation : validationFunctions) {
                 if (null != validation) {
-                    errors.addAll(validation.apply(fieldToValidate, fieldModel));
+                    //errors.addAll(validation.apply(fieldToValidate, fieldModel));
```
bamandel, line 208: The comment should be removed

alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
```diff
@@ -151,13 +150,13 @@ public void unRegisterTarget(String targetKey) throws AlertException {
             if (validationFunction.isPresent()) {
                 writeFile(tempFilename, fileResource);
                 File fileToValidate = filePersistenceUtil.createUploadsFile(tempFilename);
-                Collection<String> errors = validationFunction.get().apply(fileToValidate);
+                ValidationResult errors = validationFunction.get().apply(fileToValidate);
```
gkillough, line 153: It might be nice to update the variable name to `validationResult`.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public static ValidationResult of(String... errors) {
+        return new ValidationResult(Arrays.asList(errors));
+    }
+
+    public static ValidationResult of(ValidationResult... validationResults) {
+        Collection<String> newValidationResult = new ArrayList(validationResults.length);
+        for (ValidationResult result : validationResults) {
+            newValidationResult.addAll(result.getErrors());
+        }
```
bamandel, line 72: A stream would work nicely here. You can use Stream.of (or something similar) on the array validationResults and then use the map method to convert the list to errors.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -0,0 +1,75 @@
...
+        return new ValidationResult(newValidationResult);
+    }
```
gkillough, line 74: As a team convention, we tend to put `public static` methods above the constructor(s).

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
gkillough: `Arrays.stream(...)`

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public boolean hasErrors() {
+        return !errors.isEmpty();
+    }
+
+    public String combineErrorMessages() {
+        return StringUtils.join(errors, ", ");
+    }
+
+    public static ValidationResult of() {
```
jamesrichard91, line 56: Can we change the name of the is method to `success()`?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -0,0 +1,75 @@
...
+        return new ValidationResult();
+    }
+
+    public static ValidationResult of(Collection<String> errors) {
```
jamesrichard91, line 60: Can we change the name of the is method to `error(Collection<String> errors)`?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -0,0 +1,75 @@
...
+        return new ValidationResult(errors);
+    }
+
+    public static ValidationResult of(String... errors) {
```
jamesrichard91, line 64: Can we change the name of the is method to `error(String... errors)`?

PR #1018, opened 2020-06-12
unknown config fix

psantos1113:
- Create a task to check if there are no provider configurations in the system.

- On startup all system messages are removed and then system validation occurs.

- BlackDuckValidator is used to validate a Black Duck configuration and not determine if there are no provider configurations defined.

- Use the existing database structure to create provider specific system messages.  Since system message types are stored as strings the provider specific message types use the following format:
`<CONFIG_ID>_<SYSTEM_MESSAGE_TYPE>` This ensures that the Black Duck validator can remove only the system messages for that provider because the config Id is present in the message type.

As a fix for a feature release using the same database structure was used here to fix the issue.

In the future we may want to revisit the System Message database structure.  We may want to separate into multiple tables.  A table for provider messages, a table for user based messages when user profiles are introduced, and system wide messages.


alert-common/src/main/java/com/synopsys/integration/alert/common/provider/ProviderConfigMissingValidator.java
```diff
@@ -0,0 +1,82 @@
...
+import com.synopsys.integration.alert.common.enumeration.SystemMessageType;
+import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
+import com.synopsys.integration.alert.common.persistence.accessor.ConfigurationAccessor;
+import com.synopsys.integration.alert.common.persistence.accessor.SystemMessageUtility;
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
+import com.synopsys.integration.alert.common.provider.lifecycle.ProvidersMissingTask;
+import com.synopsys.integration.alert.common.system.BaseSystemValidator;
+
+@Component
+public class ProviderConfigMissingValidator extends BaseSystemValidator {
```
psantos1113, line 43: This class is the only class for determining if there are no providers configured at all.

src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckValidator.java
```diff
@@ -61,17 +60,13 @@ public boolean validate(BlackDuckProperties blackDuckProperties) {
         String configName = blackDuckProperties.getConfigName();
         logger.info("Validating Black Duck configuration '{}'...", configName);
 
-        removeNoProviderConfigMessages();
+        removeSystemMessagesByType(SystemMessageType.BLACKDUCK_PROVIDER_CONFIGURATION_MISSING);
```
psantos1113, line 63: I will remove this line I shouldn't need it anymore.

src/main/java/com/synopsys/integration/alert/workflow/startup/component/SystemMessageInitializer.java
```diff
@@ -64,6 +68,7 @@ public boolean validate() {
         logger.info("----------------------------------------");
         logger.info("Validating system configuration....");
 
+        clearOldMessages();
```
psantos1113, line 71: This is a startup component that determines if the system is in a valid state or not.  On startup clear old messages and start clean.

PR #1020, opened 2020-06-16
about page fix

psantos1113:
Create links if the user has access to descriptors with Global context otherwise list them.
If the user has no access to providers don't show the table.

### DIFF
--- a/src/main/js/component/AboutInfo.js
+++ b/src/main/js/component/AboutInfo.js
@@ -8,7 +8,7 @@ import * as DescriptorUtilities from 'util/descriptorUtilities';
 import ConfigurationLabel from 'component/common/ConfigurationLabel';
 import { NavLink } from 'react-router-dom';
 import LabeledField from 'field/LabeledField';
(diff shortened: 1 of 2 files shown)


src/main/js/component/AboutInfo.js
```diff
@@ -39,36 +43,89 @@ class AboutInfo extends React.Component {
...
+        const data = [];
+        const nameSet = new Set();
+        globalDescriptors.map(descriptor => descriptor.name)
+        .forEach(name => nameSet.add(name));
+        distributionDescriptors.map(descriptor => descriptor.name)
+        .forEach(name => nameSet.add(name));
+
+        nameSet.forEach(descriptorName => {
+            const globalDescriptor = globalDescriptors.find(descriptor => descriptor.name === descriptorName);
+            const distributionDescriptor = distributionDescriptors.find(descriptor => descriptor.name === descriptorName);
```
bamandel, line 85: Should this be done in the else statement? It doesn't look like it's used anywhere else.

psantos1113: It would save a little processing time.

PR #1021, opened 2020-06-16
Improve 404 error message on install for Jira Server

ChomickiM:


src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
```diff
@@ -79,9 +81,17 @@ public JiraServerCustomEndpoint(JiraServerChannelKey jiraChannelKey, CustomEndpo
...
-                return Optional.of(responseFactory.createBadRequestResponse("", "The Jira server responded with error code: " + response.getStatusCode()));
+            try {
+                Response response = jiraAppService.installMarketplaceServerApp(JiraConstants.JIRA_APP_KEY, username, password);
+                if (BooleanUtils.isTrue(response.isStatusCodeError())) {
+                    return Optional.of(responseFactory.createBadRequestResponse("", "The Jira server responded with error code: " + response.getStatusCode()));
+                }
+            } catch (IntegrationRestException e) {
+                if (RestConstants.NOT_FOUND_404 == e.getHttpStatusCode()) {
+                    return Optional.of(responseFactory.createNotFoundResponse(
+                        "The marketplace listing of the Alert Issue Property Indexer app may not support your version of Jira. Please install the app yourself or request a compatibility update. " + e.getMessage()));
```
jamesrichard91, line 92: Can we actually change this part of the message "Please install the app yourself or request a compatibility update." to "Please install the app manually or request a compatibility update."

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
```diff
@@ -79,9 +81,17 @@ public JiraServerCustomEndpoint(JiraServerChannelKey jiraChannelKey, CustomEndpo
             PluginManagerService jiraAppService = jiraServicesFactory.createPluginManagerService();
             String username = jiraProperties.getUsername();
             String password = jiraProperties.getPassword();
-            Response response = jiraAppService.installMarketplaceServerApp(JiraConstants.JIRA_APP_KEY, username, password);
-            if (BooleanUtils.isTrue(response.isStatusCodeError())) {
-                return Optional.of(responseFactory.createBadRequestResponse("", "The Jira server responded with error code: " + response.getStatusCode()));
+            try {
+                Response response = jiraAppService.installMarketplaceServerApp(JiraConstants.JIRA_APP_KEY, username, password);
+                if (BooleanUtils.isTrue(response.isStatusCodeError())) {
```
gkillough, line 86: What determines if the status-code is an error or if an exception is thrown?

ChomickiM: This was from the existing code, it requires access to Response object which is now created on line 82 so we included it into the try-catch block.

gkillough: Right, but isn't 404 an error status-code? Why does an exception get thrown for a 404 and not for any other error code (like 400, 401, 500, etc). 

To clarify my questions: 
1. Will `jiraAppService.installMarketplaceServerApp(...)` only throw an exception for a 404 or will it also throw an exception for other errors?
2. Which errors does `response.isStatusCodeError()` work for if an exception is being thrown for some errors?

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
```diff
@@ -79,9 +81,17 @@ public JiraServerCustomEndpoint(JiraServerChannelKey jiraChannelKey, CustomEndpo
...
+                    return Optional.of(responseFactory.createBadRequestResponse("", "The Jira server responded with error code: " + response.getStatusCode()));
+                }
+            } catch (IntegrationRestException e) {
+                if (RestConstants.NOT_FOUND_404 == e.getHttpStatusCode()) {
+                    return Optional.of(responseFactory.createNotFoundResponse(
+                        "The marketplace listing of the Alert Issue Property Indexer app may not support your version of Jira. Please install the app yourself or request a compatibility update. " + e.getMessage()));
+                }
+                throw e;
```
gkillough, line 94: We should consider wrapping this exception in an Optional and returning it like we do for other errors in this method.

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
psantos1113: From looking at the code the jira service will throw an exception for any response where the response code >= 400 a bad request.  The response.isStatusCodeError isn't needed here since if there is an error an exception will always be thrown.

gkillough: If that's the case, does that mean that before these changes, this never worked?

jamesrichard91: It was likely there just in case the library changed.

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
```diff
@@ -79,17 +80,22 @@ public JiraServerCustomEndpoint(JiraServerChannelKey jiraChannelKey, CustomEndpo
...
             String password = jiraProperties.getPassword();
-            Response response = jiraAppService.installMarketplaceServerApp(JiraConstants.JIRA_APP_KEY, username, password);
-            if (BooleanUtils.isTrue(response.isStatusCodeError())) {
-                return Optional.of(responseFactory.createBadRequestResponse("", "The Jira server responded with error code: " + response.getStatusCode()));
+            try {
+                Response response = jiraAppService.installMarketplaceServerApp(JiraConstants.JIRA_APP_KEY, username, password);
+            } catch (IntegrationRestException e) {
+                if (RestConstants.NOT_FOUND_404 == e.getHttpStatusCode()) {
+                    return Optional.of(responseFactory.createNotFoundResponse(
+                        "The marketplace listing of the Alert Issue Property Indexer app may not support your version of Jira. Please install the app manually or request a compatibility update. " + e.getMessage()));
```
gkillough, line 88: We may want to insert "Error: " in here: 
> update. " + e.getMessage()  

so that it becomes:
> update. Error: " + e.getMessage()

PR #1023, opened 2020-06-17
table id fix

psantos1113:
- Fix table display to only have one popup used for both insert and edit which prevents the duplicate ids. 
- Use a table reference to be able to change the save and test functions that are called since insert and edit call different functions.
- Add ids to the fields table display renders to help with automation.
- Fix lint issues with the classes that use TableDisplay

### DIFF
--- a/src/main/js/dynamic/loaded/certificates/CertificatesPage.js
+++ b/src/main/js/dynamic/loaded/certificates/CertificatesPage.js
@@ -1,7 +1,12 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
-import { clearCertificateFieldErrors, deleteCertificate, fetchCertificates, saveCertificate } from 'store/actions/certificates';
+import {
+    clearCertificateFieldErrors,
+    deleteCertificate,
+    fetchCertificates,
+    saveCertificate
+} from 'store/actions/certificates';
 import ConfigurationLabel from 'component/common/ConfigurationLabel';
 import TableDisplay from 'field/TableDisplay';
 import TextInput from 'field/input/TextInput';
@@ -14,7 +19,6 @@ class CertificatesPage extends Component {
         super(props);
         this.retrieveData = this.retrieveData.bind(this);
         this.clearModalFieldState = this.clearModalFieldState.bind(this);
-        this.createColumns = this.createColumns.bind(this);
         this.handleChange = this.handleChange.bind(this);
         this.onSave = this.onSave.bind(this);
         this.onConfigClose = this.onConfigClose.bind(this);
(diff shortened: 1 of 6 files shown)


src/main/js/field/TableDisplay.js
```diff
@@ -50,24 +53,35 @@ class TableDisplay extends Component {
...
         };
 
-        return this.props.columns.map(column => {
-
-            const assignedDataFormate = column['dataFormat'] ? column['dataFormat'] : defaultDataFormat;
-            const searchable = column['searchable'] ? column['searchable'] : true;
+        const { columns } = this.props;
+        return columns.map((column) => {
+            const assignedDataFormate = column.dataFormat ? column.dataFormat : defaultDataFormat;
+            const searchable = column.searchable ? column.searchable : true;
```
gkillough, line 84: I know this was already here, but won't `searchable` always be `true` because of this? Similar question with `dataFormat` above.

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
```diff
@@ -31,125 +35,157 @@ class CertificatesPage extends Component {
...
                     errorName={certificateContentKey}
-                    errorValue={fieldErrors[certificateContentKey]} />
+                    errorValue={fieldErrors[certificateContentKey]}
+                />
             </div>
         );
     }
 
-    onEdit(selectedRow, callback) {
+    handleChange(e) {
```
bamandel, line 131: We might have this exact function in another utility class. May be useful importing and using it

src/main/js/dynamic/loaded/users/RoleTable.js
```diff
@@ -80,136 +49,189 @@ class RoleTable extends Component {
...
         this.setState({
             role: {
                 permissions: []
             }
         }, callback);
-        this.props.clearFieldErrors();
+        clearFieldErrorsAction();
+    }
 
+    handleChange(e) {
```
bamandel, line 104: This can also be changed to the util handleChange

src/main/js/dynamic/loaded/users/RoleTable.js
```diff
@@ -80,136 +49,189 @@ class RoleTable extends Component {
...
-            if (!permission.id) {
-                permission.id = incrementedId;
-                incrementedId++;
+        let incrementedId = incrementalId;
+        const updatedPermissions = [];
+        permissions.forEach((permission) => {
+            const permissionCopy = JSON.parse(JSON.stringify(permission));
+            if (!permissionCopy.id) {
+                permissionCopy.id = incrementedId;
+                incrementedId += 1;
```
bamandel, line 183: Any specific reason we changed to this type of increment?

src/main/js/field/TableDisplay.js
psantos1113: The dataFormat attribute is meant to be a function to be used for rendering.  So the if check determines if the property is set otherwise use the default format.  Since searchable is meant to be a boolean that is a problem.  I need to change the logic to see if the property is defined otherwise default to true.

src/main/js/dynamic/loaded/users/UserTable.js
```diff
@@ -78,30 +80,34 @@ class UserTable extends Component {
     }
 
     retrieveData() {
-        this.props.getUsers();
+        const { getUsers } = this.props;
+        getUsers();
     }
 
     handleChange(e) {
```
bamandel, line 87: We may be able to replace this handle change as well

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
psantos1113: I think we may have talked about creating a utility for this, but I'm not finding a utility for the change event.

src/main/js/field/TableDisplay.js
```diff
@@ -195,28 +221,47 @@ class TableDisplay extends Component {
...
+            cancelFunction = tablePopupRef && tablePopupRef.onCancel;
+        }
+        let submitFunction = this.handleSubmit;
+        if (isInsertModal) {
+            submitFunction = tablePopupRef && tablePopupRef.handleSubmit;
+        }
+        let testFunction = this.handleTest;
+        if (isInsertModal) {
+            testFunction = tablePopupRef && tablePopupRef.handleTest;
+        }
```
bamandel, line 252: Can all of these if checks be condensed into 1?

src/main/js/dynamic/loaded/users/RoleTable.js
psantos1113: eslint rule does not like ++ it recommends in this case +=.  
Example:
increment x by y
`x = x + y` equivalent to `x += y` here in our case y = 1

src/main/js/field/TableDisplay.js
```diff
@@ -295,68 +323,87 @@ class TableDisplay extends Component {
...
+                    width="48"
+                    columnClassName="tableCell"
+                    dataFormat={this.copyButtonClick}
+                    thStyle={{ textAlign: 'center' }}
+                >
+                    Copy
+                </TableHeaderColumn>
+            );
+            tableColumns.push(copyColumn);
+        }
```
bamandel, line 406: Instead of creating 2 complete differnt TableHeaderColumn objects, we should change this to create a single one, but pass it different values to dataFormat and children. This would cut down on needing to change multiple places if we need to change the look of this column.

src/main/js/field/TableDisplay.js
psantos1113: done will update soon.

src/main/js/dynamic/loaded/users/RoleTable.js
bamandel: Gotcha. That's a bit surprising.

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
bamandel: I think you're right. Just took a quick look and didn't see any. Perhaps it's a nice future change.

src/main/js/field/TableDisplay.js
psantos1113: we push it into an array of table columns.  They need to be separate object instances.  I will create a function that will return a new instance.

PR #1025, opened 2020-06-17
Build: Adding task to push the helm chart

jamesrichard91:


buildSrc/deploymentTasks.gradle
```diff
@@ -56,3 +56,13 @@ task helmValidation(dependsOn: [copyDeploymentFiles, helmLint, helmDryRun, helmP
     helmPackage.mustRunAfter helmDryRun
 }
 
+task helmPushChart(type: Exec) {
+    workingDir = "${project.buildDir}/deployment/helm/synopsys-alert/"
+    if (!project.ext.isSnapshot) {
+        commandLine "helm", "repo", "add", "sig-cloudnative", "https://${System.env.ARTIFACTORY_HOST}/artifactory/sig-cloudnative", "&&",
```
jamesrichard91, line 62: We could see if we could run the `repo add` command on the Jenkins node and add a comment here with the command to run in case we need to setup Helm on the node again.

psantos1113: if we've added the repo already I am guessing the first part of the command will fail.  Can we use `https://helm.sh/docs/helm/helm_repo_list/` command to list the repos and grep if it is present?  If it isn't present run the add command and always run the push command at the end.  Otherwise consider using the `||` operator instead of `&&` So if the add repo fails because it exists it will try the push command.

jamesrichard91: I moved the `repo add...` into a comment in case we need to setup helm again. I also put this in a doLast block so the GradleException does not get thrown when the IDE is processing the tasks.

PR #1026, opened 2020-06-18
Refactor BlackDuckMessageBuilder to be an abstract class

ChomickiM:


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/BlackDuckMessageBuilder.java
```diff
@@ -33,16 +33,26 @@
 import com.synopsys.integration.blackduck.service.bucket.BlackDuckBucket;
 import com.synopsys.integration.exception.IntegrationException;
 
-public interface BlackDuckMessageBuilder<T> {
-    default String getProviderName() {
-        return "Black Duck";
+public abstract class BlackDuckMessageBuilder<T> {
+
+    private final String providerName = "Black Duck";
```
gkillough, line 38: Our integration team convention is to not leave whitespace between the class-name and the first variable/method.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/BlackDuckMessageBuilder.java
```diff
@@ -33,16 +33,26 @@
...
+        return this.providerName;
     }
 
-    String getNotificationType();
+    public String getNotificationType() {
+        return this.notificationType;
+    }
 
-    List<ProviderMessageContent> buildMessageContents(CommonMessageData commonMessageData, T notificationView, BlackDuckBucket blackDuckBucket, BlackDuckServicesFactory blackDuckServicesFactory);
+    abstract public List<ProviderMessageContent> buildMessageContents(CommonMessageData commonMessageData, T notificationView, BlackDuckBucket blackDuckBucket, BlackDuckServicesFactory blackDuckServicesFactory);
```
gkillough, line 53: Can we change this to `public abstract`? I think the latest formatting file + save actions should cover this. You might want to double check that they are configured correctly.

ChomickiM: I updated my save actions to the ones Jake recommended the other day but this did not seem to automatically fix this. Regardless, I have made the change.

PR #1028, opened 2020-06-18
about descriptor fix

psantos1113:
- Have the about model list all possible providers and channels
- Have the UI check if the user has Global context permissions and create a NavLink to that part of the UI if the user has permissions.
- Take the contents of the DescriptorController and refactor it into an Actions class
- Add a method to generate DescriptorMetadata classes without any permission information.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/AboutReader.java
+++ b/src/main/java/com/synopsys/integration/alert/AboutReader.java
@@ -23,8 +23,7 @@
 package com.synopsys.integration.alert;
 
 import java.nio.charset.StandardCharsets;
-import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -32,10 +31,11 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.google.gson.Gson;
-import com.synopsys.integration.alert.common.descriptor.DescriptorKey;
-import com.synopsys.integration.alert.common.descriptor.DescriptorMap;
+import com.synopsys.integration.alert.common.descriptor.config.ui.DescriptorMetadata;
+import com.synopsys.integration.alert.common.enumeration.DescriptorType;
 import com.synopsys.integration.alert.common.persistence.accessor.SystemStatusUtility;
 import com.synopsys.integration.alert.common.util.DateUtils;
+import com.synopsys.integration.alert.web.actions.DescriptorMetadataActions;
 import com.synopsys.integration.alert.web.model.AboutModel;
 import com.synopsys.integration.rest.RestConstants;
 import com.synopsys.integration.util.ResourceUtil;
@@ -46,13 +46,13 @@ public class AboutReader {
     private final Logger logger = LoggerFactory.getLogger(AboutReader.class);
     private final Gson gson;
     private final SystemStatusUtility systemStatusUtility;
-    private final DescriptorMap descriptorMap;
+    private final DescriptorMetadataActions descriptorActions;
 
     @Autowired
-    public AboutReader(Gson gson, SystemStatusUtility systemStatusUtility, DescriptorMap descriptorMap) {
+    public AboutReader(Gson gson, SystemStatusUtility systemStatusUtility, DescriptorMetadataActions descriptorActions) {
         this.gson = gson;
         this.systemStatusUtility = systemStatusUtility;
-        this.descriptorMap = descriptorMap;
+        this.descriptorActions = descriptorActions;
     }
 
     public AboutModel getAboutModel() {
@@ -60,12 +60,8 @@ public class AboutReader {
             String aboutJson = ResourceUtil.getResourceAsString(getClass(), "/about.txt", StandardCharsets.UTF_8.toString());
             AboutModel aboutModel = gson.fromJson(aboutJson, AboutModel.class);
             String startupDate = systemStatusUtility.getStartupTime() != null ? DateUtils.formatDate(systemStatusUtility.getStartupTime(), RestConstants.JSON_DATE_FORMAT) : "";
-            List<String> providers = descriptorMap.getProviderDescriptorMap().keySet().stream()
-                                         .map(DescriptorKey::getUniversalKey)
-                                         .collect(Collectors.toList());
-            List<String> channels = descriptorMap.getChannelDescriptorMap().keySet().stream()
-                                        .map(DescriptorKey::getUniversalKey)
-                                        .collect(Collectors.toList());
+            Set<DescriptorMetadata> providers = descriptorActions.getDescriptors(null, DescriptorType.PROVIDER.name(), null);
+            Set<DescriptorMetadata> channels = descriptorActions.getDescriptors(null, DescriptorType.CHANNEL.name(), null);
             return new AboutModel(aboutModel.getVersion(), aboutModel.getCreated(), aboutModel.getDescription(), aboutModel.getProjectUrl(), systemStatusUtility.isSystemInitialized(), startupDate, providers, channels);
         } catch (Exception e) {
             logger.error(e.getMessage(), e);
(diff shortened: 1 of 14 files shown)


src/main/java/com/synopsys/integration/alert/AboutReader.java
```diff
@@ -42,19 +46,23 @@
...
+        this.descriptorActions = descriptorActions;
     }
 
     public AboutModel getAboutModel() {
         try {
             String aboutJson = ResourceUtil.getResourceAsString(getClass(), "/about.txt", StandardCharsets.UTF_8.toString());
             AboutModel aboutModel = gson.fromJson(aboutJson, AboutModel.class);
             String startupDate = systemStatusUtility.getStartupTime() != null ? DateUtils.formatDate(systemStatusUtility.getStartupTime(), RestConstants.JSON_DATE_FORMAT) : "";
-            return new AboutModel(aboutModel.getVersion(), aboutModel.getCreated(), aboutModel.getDescription(), aboutModel.getProjectUrl(), systemStatusUtility.isSystemInitialized(), startupDate);
+            Set<DescriptorMetadata> providers = descriptorActions.getDescriptors(null, DescriptorType.PROVIDER.name(), null);
```
psantos1113, line 63: Add the provider and channel descriptor information back into the about model like we used to in older versions of Alert.

psantos1113: Note this DescriptorMetadata has no operations data contained in it.  So the UI will not display anything based on permissions if these DescriptorMetadata objects are used.

src/main/java/com/synopsys/integration/alert/web/actions/DescriptorMetadataActions.java
```diff
@@ -0,0 +1,147 @@
+/**
```
psantos1113, line 1: This class contains the contents of the DescriptorController.  This provides methods to create the descriptor metadata to be used in both the about page and the user based descriptor data.

src/main/java/com/synopsys/integration/alert/web/actions/DescriptorMetadataActions.java
```diff
@@ -0,0 +1,147 @@
...
+    private final Collection<Descriptor> descriptors;
+    private final AuthorizationManager authorizationManager;
+
+    @Autowired
+    public DescriptorMetadataActions(Collection<Descriptor> descriptors, AuthorizationManager authorizationManager) {
+        this.descriptors = descriptors;
+        this.authorizationManager = authorizationManager;
+    }
+
+    public Set<DescriptorMetadata> getDescriptors(@Nullable String name, @Nullable String type, @Nullable String context) {
```
psantos1113, line 57: Create all DescriptorMetadata objects regardless of user permission.  This is used by the AboutReader.

src/main/java/com/synopsys/integration/alert/web/actions/DescriptorMetadataActions.java
```diff
@@ -0,0 +1,147 @@
...
+        return getDescriptors(name, type, context, this::generateUIComponents);
+    }
+
+    public Set<DescriptorMetadata> getDescriptorsByPermissions(@Nullable String name, @Nullable String type, @Nullable String context) {
```
psantos1113, line 61: This is the method the DescriptorController uses.  Gets the DescriptorMetadata based on the user's permissions.  This is the same type of code that the controller executed in the past.

src/main/java/com/synopsys/integration/alert/web/actions/DescriptorMetadataActions.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private Set<Descriptor> filter(Collection<Descriptor> descriptors, Predicate<Descriptor> predicate) {
+        return descriptors
+                   .stream()
+                   .filter(predicate)
+                   .collect(Collectors.toSet());
+    }
+
+    private Set<DescriptorMetadata> generateUIComponents(Set<Descriptor> filteredDescriptors, ConfigContextEnum context) {
```
psantos1113, line 97: Create the metadata for all descriptors with UI Components.

src/main/java/com/synopsys/integration/alert/web/actions/DescriptorMetadataActions.java
```diff
@@ -0,0 +1,147 @@
...
+        for (ConfigContextEnum applicableContext : applicableContexts) {
+            for (Descriptor descriptor : filteredDescriptors) {
+                descriptor.createMetaData(applicableContext)
+                    .ifPresent(descriptorMetadata::add);
+            }
+        }
+        return descriptorMetadata;
+    }
+
+    private Set<DescriptorMetadata> generateUIComponentsByPermissions(Set<Descriptor> filteredDescriptors, ConfigContextEnum context) {
```
psantos1113, line 109: Create the DescriptorMetadata based on the user permissions.  Refactored the code from the controller because we were making an extra method call that we didn't need.  Check the permissions here.  If the user has permissions then create the DescriptorMetadata objects.

src/main/java/com/synopsys/integration/alert/web/controller/metadata/DescriptorController.java
```diff
@@ -22,117 +22,29 @@
  */
 package com.synopsys.integration.alert.web.controller.metadata;
 
-import java.util.Collection;
```
psantos1113, line 25: Removed code from this class and moved it into DescriptorMetadataActions

src/main/java/com/synopsys/integration/alert/web/user/RoleActions.java
```diff
@@ -84,7 +84,9 @@ public UserRoleModel createRole(RolePermissionModel rolePermissionModel) throws
         Set<PermissionModel> permissions = rolePermissionModel.getPermissions();
         validatePermissions(permissions);
         PermissionMatrixModel permissionMatrixModel = convertToPermissionMatrixModel(permissions);
-        return authorizationUtility.createRoleWithPermissions(roleName, permissionMatrixModel);
+        UserRoleModel userRoleModel = authorizationUtility.createRoleWithPermissions(roleName, permissionMatrixModel);
+        authorizationManager.loadPermissionsIntoCache();
```
psantos1113, line 88: This is a bug that might need to be patched in older versions.  The cache doesn't get updated when a new role is created.  Therefore this descriptor list for the user in the UI would be empty causing the original issue.

src/main/js/component/AboutInfo.js
```diff
@@ -57,59 +57,32 @@ class AboutInfo extends React.Component {
...
     }
 
-    createProviderTableData(globalDescriptors) {
+    createTableData(userBasedDescriptors, descriptors) {
         const data = [];
-        for (let key in globalDescriptors) {
-            const descriptor = globalDescriptors[key];
-            if (!data.find(existing => existing.urlName === descriptor.urlName)) {
+        for (let key in descriptors) {
+            const descriptor = descriptors[key];
```
psantos1113, line 63: For each descriptor (providers or channels) check the users list of descriptors and see if they have permission for a Global UI context.  If they do then navigate will be true and the item in the table will be a link not just the name of the descriptor.

src/main/js/util/descriptorUtilities.js
```diff
@@ -35,7 +35,7 @@ export function findDescriptorByNameAndContext(descriptorList, descriptorName, c
     if (!descriptorList) {
         return null;
     }
-    const resultList = descriptorList.filter(descriptor => descriptor.name === descriptorName && descriptor.context === context);
+    const resultList = descriptorList.filter((descriptor) => descriptor.name === descriptorName && descriptor.context === context);
```
psantos1113, line 38: Fix lint issues

src/main/js/component/AboutInfo.js
gkillough: I like this. 

A comment for the sake of other reviewers who may not be comfortable with this: 
> I think metadata should be available to any authenticated user, but the UI should be responsible for determining what metadata it displays. Only the `field_values` table contains sensitive content that we care about securing. As an open source project, all of this metadata is available here on GitHub anyway, so why complicate our API by trying to restrict it on a per-user basis.

src/main/java/com/synopsys/integration/alert/AboutReader.java
jamesrichard91: Possibly a future improvement but should we create a `getDescriptors()` method in `DescriptorMetadataActions` that only takes in a DescriptorType name?

psantos1113: Implemented your suggestion because the getDescriptorsByType is what we currently do for the about page.

PR #1032, opened 2020-06-22
Provider Callback Models

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/ComponentItem.java
```diff
@@ -302,15 +320,31 @@ public Builder applySubComponent(String subComponentName, String subComponentVal
...
             return this;
         }
 
-        public Builder addSubComponentUrl(String subComponentUrl) {
+        public Builder applySubComponentUrl(String subComponentUrl) {
             this.subComponentUrl = subComponentUrl;
             return this;
         }
 
+        public Builder applyComponentItemCallbackInfo(ComponentItemCallbackInfo callbackInfo) {
```
bamandel, line 332: Can this builder just call the other builder and cut down on duplication?

PR #1033, opened 2020-06-23
Message Builder Callback

gkillough:


src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/BomEditMessageBuilder.java
```diff
@@ -141,12 +142,14 @@ public BomEditMessageBuilder(PolicyCommonBuilder policyCommonBuilder) {
                 ComponentItem.Builder builder = new ComponentItem.Builder()
                                                     .applyCategory(MessageBuilderConstants.CATEGORY_TYPE_VULNERABILITY)
                                                     .applyOperation(ItemOperation.UPDATE)
-                                                    // FIXME get the vulnerability id(s) and create a ComponentItem from each of them
+                                                    // TODO get the vulnerability id(s) and create a ComponentItem from each of them
```
gkillough, line 145: Switched to TODO  because FIXME implies we shouldn't release until this is resolved IMO.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/BomEditMessageBuilder.java
```diff
@@ -191,6 +194,8 @@ public BomEditMessageBuilder(PolicyCommonBuilder policyCommonBuilder) {
                                                         .applyAllComponentAttributes(commonAttributes)
                                                         .applyNotificationId(notificationId);
                     ComponentBuilderUtil.applyComponentInformation(builder, blackDuckResponseCache, componentData);
+                    BlackDuckIssueTrackerCallbackUtil.createCallbackInfo(getNotificationType(), versionBomComponent)
+                        .ifPresent(builder::applyComponentItemCallbackInfo);
```
bamandel, line 198: Looks like there's some duplication going on here that we may want to cut back on. (Same as line 151)

gkillough: Yeah, this is related to the tech-debt I was talking about. We definitely need to clean up these classes.

src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/util/BlackDuckIssueTrackerCallbackUtil.java
```diff
@@ -0,0 +1,21 @@
+package com.synopsys.integration.alert.provider.blackduck.collector.builder.util;
+
+import java.util.Optional;
+
+import com.synopsys.integration.alert.common.message.model.ComponentItemCallbackInfo;
+import com.synopsys.integration.alert.provider.blackduck.BlackDuckProviderKey;
+import com.synopsys.integration.blackduck.api.generated.view.ProjectVersionComponentView;
+import com.synopsys.integration.blackduck.api.manual.enumeration.NotificationType;
+
+public final class BlackDuckIssueTrackerCallbackUtil {
```
jamesrichard91, line 10: Is there a reason we want this class to use static methods rather than using the `@Component` annotation and autowire this in where needed and autowire in the BlackDuckProviderKey?

PR #1034, opened 2020-06-24
Black Duck Callback Handler

gkillough:


src/main/java/com/synopsys/integration/alert/provider/blackduck/issues/BlackDuckProviderIssueHandler.java
```diff
@@ -0,0 +1,89 @@
...
+     *                                     or<br/>
+     *                                     {blackDuckUrl}/api/projects/{projectId}/versions/{projectVersionId}/components/{componentId}/component-versions/{componentVersionId}/issues
+     * @return a string in the format of {blackDuckUrl}/api/projects/{projectId}/versions/{projectVersionId}/issues
+     */
+    private String createIssueLookupUrl(String bomComponentVersionIssuesUrl) {
+        String projectVersionUrl = StringUtils.substringBefore(bomComponentVersionIssuesUrl, "/components");
+        return projectVersionUrl + "/issues";
+    }
+
+}
```
bamandel, line 89: This is slightly concerning to me as we're making assumptions here and assume these URLs won't change. Is there no concrete way for us to get this issues url from Blackduck?

gkillough: Not without asking a lot more of Events. This will go away when Black Duck supports GETs against the `bomComponentVersionIssuesUrl`.

alert-common/src/main/java/com/synopsys/integration/alert/common/event/ProviderCallbackEvent.java
```diff
@@ -8,9 +8,11 @@
 public class ProviderCallbackEvent extends AlertEvent {
     private final String callbackUrl;
     private final String notificationType;
+
+    // TODO consider splitting channelDestination out into something resembling "key" and "link"
     private final LinkableItem channelDestination;
     private final IssueOperation operation;
-    private final String channelEventDescription;
+    private final String channelActionSummary;
```
gkillough, line 15: I would love some variable name suggestions here.

- `channelDestination` represents the location that the channel posted to (e.g. name="IssueKey", Value="JIRA-123", url= "https://jira/browse/JIRA-123").
- `operation` is currently tightly coupled to `IssueOperation`, but we might want to change that.
- `channelActionSummary` represents a summary of the channel message (e.g. the Jira issue summary).

PR #1035, opened 2020-06-26
Add validation checks for Blackduck API token

ChomickiM:
This adds a check to see if the user configured with Alert can hit the notifications endpoint in blackduck. Currently, this requires that the user has superuser permissions in blackduck. An error message will be displayed when testing the configuration in the black duck provider setup, the BlackDuck validator which creates a system error every minute, and when testing configuration on the distribution.


src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
```diff
@@ -0,0 +1,71 @@
...
+import com.synopsys.integration.alert.provider.blackduck.BlackDuckProperties;
+import com.synopsys.integration.blackduck.exception.BlackDuckApiException;
+import com.synopsys.integration.blackduck.rest.BlackDuckHttpClient;
+import com.synopsys.integration.blackduck.service.BlackDuckServicesFactory;
+import com.synopsys.integration.blackduck.service.NotificationService;
+import com.synopsys.integration.exception.IntegrationException;
+import com.synopsys.integration.log.Slf4jIntLogger;
+import com.synopsys.integration.rest.RestConstants;
+
+public class BlackDuckApiTokenValidator {
```
bamandel, line 39: Should this class be a component? It looks like everytime we reference it, we just autowire BlackDuckProperties and use the new keyword.

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
```diff
@@ -0,0 +1,71 @@
...
+    private final BlackDuckProperties blackDuckProperties;
+    private final Logger logger = LoggerFactory.getLogger(BlackDuckApiTokenValidator.class);
+
+    public BlackDuckApiTokenValidator(BlackDuckProperties blackDuckProperties) {
+        this.blackDuckProperties = blackDuckProperties;
+    }
+
+    public boolean isApiTokenValid() {
+        Optional<BlackDuckHttpClient> optionalBlackDuckHttpClient = blackDuckProperties.createBlackDuckHttpClientAndLogErrors(logger);
+        return optionalBlackDuckHttpClient
```
bamandel, line 49: Can this be chained into a single call?

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
```diff
@@ -0,0 +1,71 @@
...
+                   .map(this::isNotificationApiAllowed)
+                   .orElse(false);
+    }
+
+    private boolean isNotificationApiAllowed(BlackDuckHttpClient httpClient) {
+        boolean valid = true;
+        try {
+            BlackDuckServicesFactory blackDuckServicesFactory = blackDuckProperties.createBlackDuckServicesFactory(httpClient, new Slf4jIntLogger(logger));
+            NotificationService notificationService = blackDuckServicesFactory.createNotificationService();
+            notificationService.getLatestNotificationDate();
```
bamandel, line 59: Wouldn't the real test be validating the contents here? We should return an expression here and return false in the catch clauses.

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
gkillough: `BlackDuckProperties` is no longer a Spring Component. With multiple configurations for Black Duck Providers, we must construct this on a per-config basis.

bamandel: I'm too old fashioned

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
```diff
@@ -0,0 +1,71 @@
...
+    public BlackDuckApiTokenValidator(BlackDuckProperties blackDuckProperties) {
+        this.blackDuckProperties = blackDuckProperties;
+    }
+
+    public boolean isApiTokenValid() {
+        Optional<BlackDuckHttpClient> optionalBlackDuckHttpClient = blackDuckProperties.createBlackDuckHttpClientAndLogErrors(logger);
+        return optionalBlackDuckHttpClient
+                   .map(this::isNotificationApiAllowed)
+                   .orElse(false);
+    }
```
gkillough, line 52:
```suggestion
    public boolean isApiTokenValid() {
    return blackDuckProperties.createBlackDuckHttpClientAndLogErrors(logger)
                   .map(this::isNotificationApiAllowed)
                   .orElse(false);
    }
```

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
gkillough, line 59: Couldn't no notifications exist yet, but the config still be valid? A clean Black Duck instance could have this state.

ChomickiM: We aren't validating the contents because we dont really care what notifications come back, just that your able to hit the /api/notifications endpoint. Any user that doesn't have the correct permissions will receive a 403 error.

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckValidator.java
```diff
@@ -100,6 +101,13 @@ public boolean validate(BlackDuckProperties blackDuckProperties) {
                     valid = false;
                 }
             }
+
+            BlackDuckApiTokenValidator blackDuckAPITokenValidator = new BlackDuckApiTokenValidator(blackDuckProperties);
+            if (!blackDuckAPITokenValidator.isApiTokenValid()) {
+                String message = "User permission failed, cannot read notifications from Black Duck.";
```
gkillough, line 107: We should probably either just pass this String directly into the method, or create a constant for it. Otherwise I believe SonarCloud will flag it as a code smell.

src/main/java/com/synopsys/integration/alert/provider/blackduck/validators/BlackDuckApiTokenValidator.java
ChomickiM: Same as the comment above, we don't really care about if there or aren't notifications. Even if it returned empty it would not give us a 403 forbidden which only occurs when the black duck user is lacking super user permissions

gkillough: Gotcha, so the status will still be 2xx?

ChomickiM: I could probably leave a comment in the code if that clears it up a bit more if you think that would be a good idae

ChomickiM: yes

gkillough:
Let's make sure this change is merged into 6.1.0 and not master.

ChomickiM:
Closing this PR against Master, will re-open against 6.1.0.

PR #1036, opened 2020-06-26
forbidden permission fix

psantos1113:
- Used the Global configuration as a model for the UI changes.  
- Use the StatusMessage component to show errors on the various Alert screens.  This includes any component using TableDisplay, the global configurations, distribution jobs, and audit.  
- Built a mechanism to clearly define an error object and what its contents can be. 
- Made the fetch code more consistent in setting up error handling.  
- The ui now uses the error object to display errors in the UI.  If they are field errors for field panels or errors creating and deleting items in the tables. The error object is used.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/user/RoleActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/user/RoleActions.java
@@ -30,8 +30,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import javax.transaction.Transactional;
-
 import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -53,7 +51,6 @@ import com.synopsys.integration.alert.web.model.PermissionModel;
 import com.synopsys.integration.alert.web.model.RolePermissionModel;
 
 @Component
-@Transactional
 public class RoleActions {
     private static final String FIELD_KEY_ROLE_NAME = "roleName";
     private final AuthorizationUtility authorizationUtility;
--- a/src/main/js/distribution/Index.js
+++ b/src/main/js/distribution/Index.js
@@ -18,6 +18,7 @@ import DistributionConfiguration, {
     KEY_NAME,
     KEY_PROVIDER_NAME
 } from 'dynamic/DistributionConfiguration';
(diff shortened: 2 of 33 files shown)


src/main/java/com/synopsys/integration/alert/web/user/RoleActions.java
```diff
@@ -53,7 +51,6 @@
 import com.synopsys.integration.alert.web.model.RolePermissionModel;
 
 @Component
-@Transactional
```
psantos1113, line 56: Deletes wouldn't work because the delete uses this transaction.  The accessor deletes the entity but then with the loadPermissionIntoCache calls a get from the accessor and PSQL complains because we are not in a read only transaction.  Use the transactions in the accessor and remove transactional here.

src/main/js/dynamic/loaded/audit/Details.js
```diff
@@ -157,6 +157,7 @@ class Details extends Component {
                                 </div>
                             </Tab>
                         </Tabs>
+                        <p name="message">{this.props.errorMessage}</p>
```
jamesrichard91, line 160: Do we not want to use the StatusMessage here and leave the action message empty?

src/main/js/distribution/Index.js
```diff
@@ -345,6 +346,8 @@ class Index extends Component {
         const content = (
             <div>
                 {this.getCurrentJobConfig()}
+                <StatusMessage actionMessage={null}
```
psantos1113, line 349: This is used to show read failures.

src/main/js/store/actions/about.js
```diff
@@ -1,6 +1,7 @@
 import { ABOUT_INFO_FETCH_ERROR, ABOUT_INFO_FETCHED, ABOUT_INFO_FETCHING } from 'store/actions/types';
 
-import { verifyLoginByStatus } from 'store/actions/session';
+import { unauthorized } from 'store/actions/session';
+import * as HTTPErrorUtils from '../../util/httpErrorUtilities';
```
jamesrichard91, line 4: The import path should be fixed

src/main/js/dynamic/loaded/audit/Details.js
psantos1113: That's the dialog for audit.  Status message is shown underneath the modal.  All modals right now show errors this way. 
So the modals are at least consistent.

PR #1038, opened 2020-06-29
Fix provider config name cascade

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/DefaultDescriptorGlobalConfigUtility.java
```diff
@@ -108,7 +108,7 @@ public FieldModel update(Long id, FieldModel fieldModel) throws AlertException {
             configurationModel = configurationAccessor.createConfiguration(key, context, valueMap.values());
         }
         FieldModel convertedFieldModel = configurationFieldModelConverter.convertToFieldModel(configurationModel);
-        return apiAction.afterUpdateAction(convertedFieldModel);
+        return apiAction.afterUpdateAction(beforeUpdateAction, convertedFieldModel);
```
bamandel, line 111: The name of beforeUpdateAction confused me here, can we change it to something like, beforeUpdateActionResult or something similar?

PR #1040, opened 2020-06-30
Ps id fixes

psantos1113:
Add id to the html elements in the UI.  Some of the ids have changed based on bugs where the ids weren't being set correctly.

### DIFF
--- a/src/main/js/component/common/ConfirmModal.js
+++ b/src/main/js/component/common/ConfirmModal.js
@@ -71,16 +71,18 @@ class ConfirmModal extends Component {
 }
 
 ConfirmModal.propTypes = {
+    id: PropTypes.string,
     affirmativeAction: PropTypes.func.isRequired,
     affirmativeButtonText: PropTypes.string,
     message: PropTypes.string.isRequired,
     negativeAction: PropTypes.func,
     negativeButtonText: PropTypes.string,
     showModal: PropTypes.bool.isRequired,
-    title: PropTypes.string.isRequired,
+    title: PropTypes.string.isRequired
 };
 
 ConfirmModal.defaultProps = {
+    id: 'confirmModal',
     affirmativeButtonText: 'Yes',
     negativeButtonText: 'No',
     negativeAction: () => true
--- a/src/main/js/component/common/ConfirmModal.js
+++ b/src/main/js/component/common/ConfirmModal.js
@@ -48,9 +48,9 @@ class ConfirmModal extends Component {
     }
 
     render() {
-        const { affirmativeButtonText, message, negativeButtonText, title } = this.props;
+        const { id, affirmativeButtonText, message, negativeButtonText, title } = this.props;
         return (
-            <Modal show={this.state.show} onHide={this.handleNegativeClick}>
+            <Modal id={id} show={this.state.show} onHide={this.handleNegativeClick}>
                 <Modal.Header closeButton>
                     <Modal.Title>{title}</Modal.Title>
                 </Modal.Header>
--- a/src/main/js/component/common/RefreshTableCellFormatter.js
+++ b/src/main/js/component/common/RefreshTableCellFormatter.js
@@ -18,9 +18,13 @@ const RefreshTableCellFormatter = ({ id, handleButtonClicked, currentRowSelected
 );
 
 RefreshTableCellFormatter.propTypes = {
-    id: PropTypes.string.isRequired,
+    id: PropTypes.string,
     currentRowSelected: PropTypes.object.isRequired,
     handleButtonClicked: PropTypes.func.isRequired
+};
+
+RefreshTableCellFormatter.defaultProps = {
+    id: 'refreshTableCellFormatterId'
 };
 
 export default RefreshTableCellFormatter;
--- a/src/main/js/component/common/RefreshTableCellFormatter.js
+++ b/src/main/js/component/common/RefreshTableCellFormatter.js
@@ -2,8 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-const RefreshTableCellFormatter = ({ handleButtonClicked, currentRowSelected }) => (
+const RefreshTableCellFormatter = ({ id, handleButtonClicked, currentRowSelected }) => (
     <button
+        id={id}
         className="btn btn-link editJobButton"
         type="button"
         title="Refresh"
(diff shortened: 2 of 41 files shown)


src/main/js/component/common/ConfirmModal.js
```diff
@@ -71,16 +71,18 @@ class ConfirmModal extends Component {
 }
 
 ConfirmModal.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 74: Is this property used in this component?

src/main/js/component/common/SystemMessage.js
```diff
@@ -35,12 +35,14 @@ class SystemMessage extends Component {
 }
 
 SystemMessage.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 38: Is this property used in this component?

src/main/js/field/EndpointButtonField.js
```diff
@@ -144,6 +144,7 @@ class EndpointButtonField extends Component {
 }
 
 EndpointButtonField.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 147: Is this property used in this component?

src/main/js/field/EndpointSelectField.js
```diff
@@ -74,6 +74,7 @@ class EndpointSelectField extends Component {
 }
 
 EndpointSelectField.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 77: Is this property used?

src/main/js/field/LabeledField.js
```diff
@@ -67,6 +67,7 @@ class LabeledField extends Component {
 }
 
 LabeledField.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 70: Is this property used?

src/main/js/field/ReadOnlyField.js
```diff
@@ -29,12 +29,14 @@ class ReadOnlyField extends Component {
 }
 
 ReadOnlyField.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 32: Is this property used?

src/main/js/field/UploadFileButtonField.js
```diff
@@ -4,7 +4,7 @@ import LabeledField from 'field/LabeledField';
 import { createDeleteRequest, createFileUploadRequest, createReadRequest } from 'util/configurationRequestBuilder';
 import { connect } from 'react-redux';
 import StatusMessage from 'field/StatusMessage';
-import GeneralButton from "./input/GeneralButton";
+import GeneralButton from './input/GeneralButton';
```
jamesrichard91, line 7: Need to fix the import path to be absolute not relative

src/main/js/field/UploadFileButtonField.js
```diff
@@ -185,6 +185,7 @@ class UploadFileButtonField extends Component {
 }
 
 UploadFileButtonField.propTypes = {
+    id: PropTypes.string,
```
jamesrichard91, line 188: Is this property used?

src/main/js/field/EndpointSelectField.js
psantos1113: Yes the DynamicSelectInput has `{..this.props}` Which will use the id of the field and set it on the DynamicSelectField which does use the id property when rendering.

src/main/js/field/EndpointButtonField.js
psantos1113: Yes line 128 the `{...this.props}` the id of this component is used in the LabeledField.  But I have to double check LabeledField is setting it.

src/main/js/field/ReadOnlyField.js
psantos1113: because of this code it will `{...this.props}` set the id property on the LabeledField.  Have to confirm LabeledField is doing the right thing.

src/main/js/field/UploadFileButtonField.js
psantos1113: The usage of `{...this.props}` will set the id on the LabeledField.  Have to confirm the LabeledField is setting it properly.

src/main/js/component/common/RefreshTableCellFormatter.js
```diff
@@ -17,6 +18,7 @@ const RefreshTableCellFormatter = ({ handleButtonClicked, currentRowSelected })
 );
 
 RefreshTableCellFormatter.propTypes = {
+    id: PropTypes.string.isRequired,
```
gkillough, line 21: Why choose to require this one?

src/main/js/distribution/Index.js
```diff
@@ -198,7 +199,8 @@ class Index extends Component {
     }
 
     copyButtonClick(cell, row) {
-        return <IconTableCellFormatter handleButtonClicked={this.copyButtonClicked} currentRowSelected={row} buttonIconName="copy" buttonText="Copy" />;
+        return <IconTableCellFormatter id="distribution" handleButtonClicked={this.copyButtonClicked}
```
gkillough, line 202: I'm not familiar with `IconTableCellFormatter`, but if it's a repeated field, it should probably have a dynamic suffix so that there are no errors related to uniqueness.

src/main/js/component/common/RefreshTableCellFormatter.js
psantos1113: Because the original has no default values whereas IconTableCellFormatter does have default values.  I will add default values for this class.

src/main/js/distribution/Index.js
gkillough: Same thing for `editButtonClick` above this.

psantos1113: Confirmed with QA that that unique ids for the cells won't benefit them as of now.  The edit button and copy button formatters will have different ids from each other.  I will update that.

src/main/js/dynamic/loaded/audit/Details.js
```diff
@@ -43,20 +43,26 @@ class Details extends Component {
...
         }
 
         return (<div className="auditJobDetails">{errorMessage}{errorStackTrace}</div>);
     }
 
     resendButton(cell, row) {
-        return (<RefreshTableCellFormatter handleButtonClicked={this.onResendClick} currentRowSelected={row}
-                                           buttonText="Re-send" />);
+        return (<RefreshTableCellFormatter
+            id={`audit-detail-refresh-cell`}
```
gkillough, line 63: Are the curly braces and back-ticks necessary here?

psantos1113: good catch no they are not

PR #1041, opened 2020-06-30
Create Initial Azure Boards Library

gkillough:


azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/HttpServiceException.java
```diff
@@ -0,0 +1,51 @@
+package com.synopsys.integration.azure.boards.common.http;
+
+import org.apache.http.HttpStatus;
+
+public class HttpServiceException extends Exception {
```
jamesrichard91, line 5: Shouldn't we extend IntegrationException to make error handling easier?

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureBoardsOAuthService.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.api.client.auth.oauth2.StoredCredential;
+import com.google.api.client.auth.oauth2.TokenResponse;
+import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.util.Base64;
+import com.google.api.client.util.store.DataStoreFactory;
+import com.synopsys.integration.exception.IntegrationException;
+
+public class AzureBoardsOAuthService {
```
gkillough, line 22: The API for this class will definitely change, but for now, I'm going to leave this as a POC.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/model/AzureSpecTemplate.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public String populateSpec() {
+        String populatedSpec = specTemplate;
+        for (Map.Entry<String, String> replacementEntry : replacementMappings.entrySet()) {
+            populatedSpec = StringUtils.replaceOnce(populatedSpec, replacementEntry.getKey(), replacementEntry.getValue());
+        }
+
+        if (!replacementMappings.isEmpty()) {
+            replacementMappings.clear();
+        }
```
gkillough, line 36: This is probably unnecessary.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureBoardsOAuthService.java
```diff
@@ -0,0 +1,86 @@
...
+        try {
+            AuthorizationCodeTokenRequest authorizationCodeTokenRequest = authorizationCodeFlow.newTokenRequest(authorizationCode);
+            TokenResponse tokenResponse = authorizationCodeTokenRequest.execute();
+            return authorizationCodeFlow.createAndStoreCredential(tokenResponse, userId);
+        } catch (IOException e) {
+            throw new IntegrationException(e);
+        }
+    }
+
+    private AuthorizationCodeFlow initializeFlow(String clientId) throws IOException {
```
psantos1113, line 59: Wouldn't this need to be public in the future?  I don't see this method used but the public methods requires an AuthorizationCodeFlow which this private method produces.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/model/AzureSpecTemplate.java
```diff
@@ -0,0 +1,44 @@
+package com.synopsys.integration.azure.boards.common.model;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class AzureSpecTemplate {
```
psantos1113, line 8: This seems like it would be well suited in our int-rest library. A utility to handle this kind of replacement for urls.  Might be something to look into in the future.  I like the concept here.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureBoardsOAuthService.java
gkillough: Yeah, I'm leaving the code in this class alone for now because we will need to determine what gets configured vs gets defaults. I just wanted to leave some code examples in the library, but this may actually become something totally different when we start to implement the channel global config.

PR #1043, opened 2020-07-01
Azure Library: Boards Service

gkillough:


azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/service/board/BoardFieldsModel.java
```diff
@@ -0,0 +1,32 @@
+package com.synopsys.integration.azure.boards.common.service.board;
+
+import com.synopsys.integration.azure.boards.common.model.FieldReferenceModel;
+
+public class BoardFieldsModel {
+    private FieldReferenceModel columnField;
+    private FieldReferenceModel doneField;
+    private FieldReferenceModel rowField;
```
gkillough, line 8: I'm not 100% sure this is the actual content of this object, but the API description is weird: 
https://docs.microsoft.com/en-us/rest/api/azure/devops/work/boards/get?view=azure-devops-rest-5.1#boardfields

psantos1113: I think you're interpretation is correct at the moment.

PR #1044, opened 2020-07-01
Azure Library Workitem Service

gkillough:


azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/util/AzureFieldsExtractor.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Optional;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class AzureFieldsExtractor {
+    private final Gson gson;
+
+    public AzureFieldsExtractor(Gson gson) {
```
jamesrichard91, line 34: What is the intended use for this class? It looks like it's only used in the tests.

jamesrichard91: Are we not able to deserialize the response into java objects correctly?

gkillough: It's a utility to extract the actual objects from `fields` fields in certain Azure responses where the `fields` object of the response is dynamic.

gkillough: We cannot deserialize them losslessly.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/service/workitem/WorkItemResponseFields.java
```diff
@@ -0,0 +1,75 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.synopsys.integration.azure.boards.common.util.AzureFieldDefinition;
+
+public final class WorkItemResponseFields {
```
gkillough, line 34: This is a non-exhaustive list of known fields that are possible on a WorkItem.

PR #1047, opened 2020-07-06
Creaing descriptor and UI configs for Azure Boards

ChomickiM:
This is the setup work for the new Azure Boards channel. Some of the descriptions and labels have placeholder values as we have not yet come up with what we will and won't need. 

If there are any suggestions you guys have to the strings that I've chosen for labels or descriptors, please let me know and I will change them.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
@@ -31,15 +31,17 @@ import com.synopsys.integration.alert.common.descriptor.ChannelDescriptor;
 @Component
 public class AzureBoardsDescriptor extends ChannelDescriptor {
     public static final String AZURE_BOARDS_PREFIX = "azure.boards";
+    public static final String AZURE_BOARDS_CHANNEL_PREFIX = "channel." + AZURE_BOARDS_PREFIX;
 
-    public static final String KEY_AZURE_PROJECT = "channel." + AZURE_BOARDS_PREFIX + ".project";
-    public static final String KEY_AZURE_BOARD = "channel." + AZURE_BOARDS_PREFIX + ".board";
-    public static final String KEY_WORK_ITEM_TYPE = "channel." + AZURE_BOARDS_PREFIX + ".work.item.type";
-    public static final String KEY_WORK_ITEM_COMPLETED_STATE = "channel." + AZURE_BOARDS_PREFIX + ".work.item.completed.state";
-    public static final String KEY_WORK_ITEM_REOPEN_STATE = "channel." + AZURE_BOARDS_PREFIX + ".work.item.reopen.state";
+    public static final String KEY_AZURE_PROJECT = AZURE_BOARDS_CHANNEL_PREFIX + ".project";
+    public static final String KEY_AZURE_BOARD = AZURE_BOARDS_CHANNEL_PREFIX + ".board";
+    public static final String KEY_WORK_ITEM_TYPE = AZURE_BOARDS_CHANNEL_PREFIX + ".work.item.type";
+    public static final String KEY_WORK_ITEM_COMPLETED_STATE = AZURE_BOARDS_CHANNEL_PREFIX + ".work.item.completed.state";
+    public static final String KEY_WORK_ITEM_REOPEN_STATE = AZURE_BOARDS_CHANNEL_PREFIX + ".work.item.reopen.state";
 
     public static final String KEY_AZURE_BOARDS_URL = AZURE_BOARDS_PREFIX + ".url";
     public static final String KEY_ORGANIZATION_NAME = AZURE_BOARDS_PREFIX + ".organization.name";
+    //TODO revisit these fields as we may only need access token for oauth
     public static final String KEY_CONSUMER_KEY = AZURE_BOARDS_PREFIX + ".consumer.key";
     public static final String KEY_PRIVATE_KEY = AZURE_BOARDS_PREFIX + ".private.key";
     public static final String KEY_OAUTH = AZURE_BOARDS_PREFIX + ".oauth";
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
@@ -1,0 +1,56 @@
+/**
+ * blackduck-alert
+ *
+ * Copyright (c) 2020 Synopsys, Inc.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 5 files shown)


src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsGlobalUIConfig.java
```diff
@@ -0,0 +1,75 @@
...
+        this.encryptionValidator = encryptionValidator;
+    }
+
+    @Override
+    public List<ConfigField> createFields() {
+        ConfigField azureBoardsUrlField = new URLInputConfigField(AzureBoardsDescriptor.KEY_AZURE_BOARDS_URL, LABEL_AZURE_BOARDS_URL, DESCRIPTION_AZURE_BOARDS_URL).applyRequired(true);
+        ConfigField organizationName = new TextInputConfigField(AzureBoardsDescriptor.KEY_ORGANIZATION_NAME, LABEL_ORGANIZATION_NAME, DESCRIPTION_ORGANIZATION_NAME).applyRequired(true);
+        ConfigField consumerKey = new PasswordConfigField(AzureBoardsDescriptor.KEY_CONSUMER_KEY, LABEL_CONSUMER_KEY, DESCRIPTION_CONSUMER_KEY, encryptionValidator).applyRequired(true);
+        ConfigField privateKey = new PasswordConfigField(AzureBoardsDescriptor.KEY_PRIVATE_KEY, LABEL_PRIVATE_KEY, DESCRIPTION_PRIVATE_KEY, encryptionValidator).applyRequired(true);
+        ConfigField configureOAuth = new EndpointButtonField(AzureBoardsDescriptor.KEY_OAUTH, LABEL_OAUTH, DESCRIPTION_OAUTH, BUTTON_LABEL_OAUTH);
```
psantos1113, line 70: There isn't a custom endpoint registered in this PR to handle the oauth connection?  Is that intended for a different PR.  Right now this field will not do anything on the server side.  You could at the least register the custom endpoint for this field and log a message to ensure that the server recieves the request when this button is pushed.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
```diff
@@ -0,0 +1,56 @@
...
+    public static final String KEY_AZURE_BOARDS_URL = "azure.boards.url";
+    public static final String KEY_ORGANIZATION_NAME = "azure.boards.organization.name";
+    public static final String KEY_CONSUMER_KEY = "azure.boards.consumer.key";
+    public static final String KEY_PRIVATE_KEY = "azure.boards.private.key";
+    public static final String KEY_OAUTH = "azure.boards.oauth";
+    public static final String KEY_ACCESS_TOKEN = "azure.boards.access.token";
+
+    public static final String AZURE_BOARDS_LABEL = "Azure Boards";
+    public static final String AZURE_BOARDS_URL = "azure_boards";
+    public static final String AZURE_BOARDS_DESCRIPTION = "Configure the Azure Boards instance that Alert will send issue updates to.";
```
bamandel, line 48: We might want to use some variables for prefixes. This method of writing the entire KEY name is prone to error.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsGlobalUIConfig.java
```diff
@@ -0,0 +1,75 @@
...
+@Component
+public class AzureBoardsGlobalUIConfig extends UIConfig {
+    public static final String LABEL_AZURE_BOARDS_URL = "Url";
+    public static final String LABEL_ORGANIZATION_NAME = "Organization Name";
+    public static final String LABEL_CONSUMER_KEY = "Consumer Key";
+    public static final String LABEL_PRIVATE_KEY = "Private Key";
+    public static final String LABEL_OAUTH = "Initialize OAuth";
+    public static final String LABEL_ACCESS_TOKEN = "Access Token";
+
+    public static final String DESCRIPTION_AZURE_BOARDS_URL = "FILL OUT THIS DESCRIPTION";
```
bamandel, line 47: At least a FIXME should be added here just so we are aware this isn't finished yet.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
jamesrichard91: Please create static strings to use as the prefixes for the keys. That way each key only needs to add its unique part

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsGlobalUIConfig.java
jamesrichard91: I agree with Paulo, can you setup a custom endpoint so when the button is pushed it logs a message for now, that way we can easily add the implementation later.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
```diff
@@ -0,0 +1,58 @@
...
+
+@Component
+public class AzureBoardsDescriptor extends ChannelDescriptor {
+    public static final String AZURE_BOARDS_PREFIX = "azure.boards";
+
+    public static final String KEY_AZURE_PROJECT = "channel." + AZURE_BOARDS_PREFIX + ".project";
+    public static final String KEY_AZURE_BOARD = "channel." + AZURE_BOARDS_PREFIX + ".board";
+    public static final String KEY_WORK_ITEM_TYPE = "channel." + AZURE_BOARDS_PREFIX + ".work.item.type";
+    public static final String KEY_WORK_ITEM_COMPLETED_STATE = "channel." + AZURE_BOARDS_PREFIX + ".work.item.completed.state";
+    public static final String KEY_WORK_ITEM_REOPEN_STATE = "channel." + AZURE_BOARDS_PREFIX + ".work.item.reopen.state";
```
gkillough, line 39: Do we have a constant we can use for the ".channel" prefix?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsDescriptor.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public static final String KEY_AZURE_BOARDS_URL = AZURE_BOARDS_PREFIX + ".url";
+    public static final String KEY_ORGANIZATION_NAME = AZURE_BOARDS_PREFIX + ".organization.name";
+    public static final String KEY_CONSUMER_KEY = AZURE_BOARDS_PREFIX + ".consumer.key";
+    public static final String KEY_PRIVATE_KEY = AZURE_BOARDS_PREFIX + ".private.key";
+    public static final String KEY_OAUTH = AZURE_BOARDS_PREFIX + ".oauth";
+    public static final String KEY_ACCESS_TOKEN = AZURE_BOARDS_PREFIX + ".access.token";
```
gkillough, line 46: We may only end up needing the access token because it is OAuth 2.0. Do you mind adding a TODO above line 43 to revisit these fields?

PR #1060, opened 2020-07-09
Azure Library: Work Item Types

gkillough:
To be reviewed after #1058


azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/service/state/WorkItemTypeStateService.java
```diff
@@ -0,0 +1,77 @@
...
+        return azureHttpService.get(requestSpec, responseType);
+    }
+
+    public AzureArrayResponseModel<WorkItemTypeProcessStateResponseModel> getStatesForProcess(String organizationName, String processId, String workItemTypeRefName) throws HttpServiceException {
+        String requestSpec = API_SPEC_ORGANIZATION_PROCESS_WORKITEMTYPE_STATES
+                                 .defineReplacement("{organization}", organizationName)
+                                 .defineReplacement("{processId}", processId)
+                                 .defineReplacement("{witRefName}", workItemTypeRefName)
+                                 .populateSpec();
+        requestSpec = String.format("%s?%s=%s", requestSpec, AzureHttpService.AZURE_API_VERSION_QUERY_PARAM_NAME, "5.1-preview.1");
```
jamesrichard91, line 61: Are you sure we should be using a version type of *-preview?

gkillough: There are several pieces of functionality that require using preview versions. Luckily these are not subject to change, but they may not be fully tested.

PR #1061, opened 2020-07-09
Method called in a for loop so flush the repository at the end.

psantos1113:
The mapUsersToProjectByEmail method is called inside a forloop in a transaction.  It uses the batch methods deleteAll and saveAll.  Need to flush after those methods are called so the next iteration of the loop can find the repository data.


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -177,6 +177,7 @@ private void mapUsersToProjectByEmail(Long providerConfigId, String projectHref,
         logger.debug("Removing {} user relations from project {} ", userRelationsToRemove.size(), project.getName());
         providerUserProjectRelationRepository.saveAll(userRelationsToAdd);
         providerUserProjectRelationRepository.deleteAll(userRelationsToRemove);
+        providerUserProjectRelationRepository.flush();
```
gkillough, line 180: What happens to the transaction when flush is called?

psantos1113: The transaction still protects us. 
https://stackoverflow.com/questions/9908082/does-a-hibernate-transaction-rollback-delete-session-flushed-entities

PR #1066, opened 2020-07-14
Fix Jira Server Plugin Installation Validation

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/cloud/web/JiraCustomEndpoint.java
```diff
@@ -95,8 +96,6 @@ public JiraCustomEndpoint(JiraChannelKey jiraChannelKey, CustomEndpointManager c
             Thread.currentThread().interrupt();
             return Optional.of(responseFactory.createInternalServerErrorResponse("", "Thread was interrupted while validating Jira plugin installation: " + e.getMessage()));
         }
-
-        return Optional.empty();
```
gkillough, line 99: Having the `return Optional.empty()` after the try-catch makes it easy to forget to return/throw errors.

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
```diff
@@ -92,15 +92,14 @@ public JiraServerCustomEndpoint(JiraServerChannelKey jiraChannelKey, CustomEndpo
             if (!jiraPluginInstalled) {
                 return Optional.of(responseFactory.createNotFoundResponse("Was not able to confirm Jira server successfully installed the Jira Server plugin. Please verify the installation on you Jira server."));
             }
+            return Optional.empty();
         } catch (IntegrationException e) {
-            createBadRequestIntegrationException(e);
+            return createBadRequestIntegrationException(e);
```
gkillough, line 97: The actual fix.

src/main/java/com/synopsys/integration/alert/channel/jira/cloud/web/JiraCustomEndpoint.java
```diff
@@ -87,6 +87,7 @@ public JiraCustomEndpoint(JiraChannelKey jiraChannelKey, CustomEndpointManager c
             if (!jiraPluginInstalled) {
                 return Optional.of(responseFactory.createNotFoundResponse("Was not able to confirm Jira Cloud successfully installed the Jira Cloud plugin. Please verify the installation on you Jira Cloud server."));
             }
+            return Optional.empty();
```
jamesrichard91, line 90: Is this change so we are consistent with the jira server change?

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomEndpoint.java
jamesrichard91, line 97: Was this the cause of the issue?

src/main/java/com/synopsys/integration/alert/channel/jira/cloud/web/JiraCustomEndpoint.java
gkillough: Yes. Just to ensure that we don't forget to actually return the error message.

PR #1069, opened 2020-07-16
Fix BD Callback PUT and issue link

gkillough:


src/main/java/com/synopsys/integration/alert/provider/blackduck/issues/BlackDuckProviderIssueHandler.java
```diff
@@ -59,6 +59,7 @@ public void createOrUpdateBlackDuckIssue(String bomComponentVersionIssuesUrl, Bl
             issueRequestModel.setIssueDescription(existingIssue.getIssueDescription());
             issueRequestModel.setIssueCreatedAt(existingIssue.getIssueCreatedAt());
             issueRequestModel.setIssueUpdatedAt(currentDate);
+            bomComponentVersionIssuesUrl = existingIssue.getMeta().getHref();
```
jamesrichard91, line 62: If bomComponentVersionIssuesUrl is passed in, why are we overriding it here? Can you add a comment to explain? Or create a new variable to pass into performRequest(...) that would reveal the intent?

gkillough: A PUT requires the id be appended to the request, but the only way the API reveals that is through the href. I'll update the variable names to make it more clear.

PR #1072, opened 2020-07-20
Auto-refresh Improvements

ChomickiM:
The following changes are updates and improvements to the auto-refresh, bringing the refresh rate up to 30 seconds as well as pausing it on modals that it previously did not effect such as some delete modals. 

As a note, currently the insert modal for the distribution jobs will continue to refresh the table in the background as we cannot get the state of the table easily. After discussing it with Paulo, the solution for this table will be to change update the distributions page to use TableDisplay to simplify the modals in a future ticket.

Laslty, several style changes were made to use object destructuring.

### DIFF
--- a/src/main/js/component/common/AutoRefresh.js
+++ b/src/main/js/component/common/AutoRefresh.js
@@ -20,7 +20,8 @@ class AutoRefresh extends Component {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        const { autoRefresh } = this.props;
+        const { autoRefresh, isEnabled } = this.props;
+
         if ((prevProps.autoRefresh !== autoRefresh)) {
             this.setState({
                 refresh: this.props.autoRefresh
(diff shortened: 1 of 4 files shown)


src/main/js/component/common/AutoRefresh.js
```diff
@@ -20,17 +20,23 @@ class AutoRefresh extends Component {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        const { autoRefresh } = this.props;
+        const { refresh } = this.state;
+        const { autoRefresh, isEnabled } = this.props;
```
ChomickiM, line 24: These are the style improvements

src/main/js/component/common/AutoRefresh.js
```diff
@@ -39,20 +45,24 @@ class AutoRefresh extends Component {
     }
 
     toggleTimer() {
-        if (this.state.refresh) {
+        const { refresh } = this.state;
+        const { refreshRate, isEnabled } = this.props;
```
ChomickiM, line 49: these are also style improvements

PR #1075, opened 2020-07-21
Enhance transition error messages

gkillough:


alert-jira/src/main/java/com/synopsys/integration/alert/jira/server/util/JiraServerTransitionHandler.java
```diff
@@ -23,29 +23,32 @@
...
 import com.synopsys.integration.exception.IntegrationException;
 import com.synopsys.integration.jira.common.model.components.IdComponent;
 import com.synopsys.integration.jira.common.model.components.StatusDetailsComponent;
 import com.synopsys.integration.jira.common.model.components.TransitionComponent;
 import com.synopsys.integration.jira.common.model.request.IssueRequestModel;
-import com.synopsys.integration.jira.common.model.response.TransitionsResponseModel;
 import com.synopsys.integration.jira.common.server.builder.IssueRequestModelFieldsBuilder;
 import com.synopsys.integration.jira.common.server.service.IssueService;
 
 public class JiraServerTransitionHandler extends JiraTransitionHandler {
```
bamandel, line 37: Unless I'm missing something, this class looks identical to the JiraCloudTransitionHandler. Instead of forcing a hierarchy here, could we just put this funcitonality into JiraTransitionHandler?

gkillough: Cloud: `import com.synopsys.integration.jira.common.cloud.service.IssueService`
Server: `import com.synopsys.integration.jira.common.server.service.IssueService`

bamandel: Checked them out, I think we should make an interface for IssueService that would be used in place of those 2 imports. But that would be out of scope.

PR #1079, opened 2020-07-22
Validate before accumulating

ChomickiM:
Removed the BlackDuckValidatorTask and moved its functionality into the BlackDuckAccumulator to be called prior to the accumulate method.


src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckAccumulator.java
```diff
@@ -77,6 +80,7 @@ public String formatDate(OffsetDateTime date) {
 
     @Override
     public void runProviderTask() {
+        blackDuckValidator.validate(getProviderProperties());
```
psantos1113, line 83: I would check the boolean value returned by validate.  If validate returns true then I would call the accumulate method.  Otherwise don't bother trying to accumulate.

psantos1113: This is just an enhancement not a blocking issue for this PR.

PR #1080, opened 2020-07-23
 helm hostname

psantos1113:
add the ability to configure the hostname for the  alert service.  Also made the port configurable.


deployment/helm/synopsys-alert/templates/alert-environ-configmap.yaml
```diff
@@ -1,13 +1,20 @@
 apiVersion: v1
 kind: ConfigMap
 data:
+  {{- if .Values.alert.hostname }}
+  ALERT_HOSTNAME: {{ .Values.alert.hostname | quote }}
+  {{- else }}
+  ALERT_HOSTNAME: {{ .Release.Name | quote }}-alert
```
mphammer, line 7: I don't think we need to put "-alert"

This PR removed the "-alert" from other areas of the Chart: https://github.com/blackducksoftware/blackduck-alert/pull/941

deployment/helm/synopsys-alert/templates/alert-environ-configmap.yaml
```diff
@@ -1,13 +1,20 @@
...
+  {{- end}}
+  {{- if .Values.alert.port }}
+  ALERT_SERVER_PORT: {{ .Values.alert.port | quote }}
```
mphammer, line 10: I don't think "ALERT_SERVER_PORT" *needs* to be configurable. But this is okay if you want it :) 

What do you think @msenmurugan ?

psantos1113: I have removed '-alert' from the hostname.  If it should be added back then I can add it back.

jamesrichard91: Will that impact existing customers using the helm charts? Would changing the host name impact the volumes or any of the data?

psantos1113: the original host name was always localhost and that is the default so this should not affect existing customers.  The <release name>-alert was a way to provide a different default host name unique to alert if the hostname was missing from values.yml

deployment/helm/synopsys-alert/templates/alert-environ-configmap.yaml
```diff
@@ -1,13 +1,20 @@
...
 kind: ConfigMap
 data:
+  {{- if .Values.alert.hostname }}
+  ALERT_HOSTNAME: {{ .Values.alert.hostname | quote }}
+  {{- else }}
+  ALERT_HOSTNAME: {{ .Release.Name | quote }}
+  {{- end}}
+  {{- if .Values.alert.port }}
+  ALERT_SERVER_PORT: {{ .Values.alert.port | quote }}
+  {{- else }}
```
msenmurugan, line 11: @psantos1113 can you pls remove if and else for ALERT_SERVER_PORT and set them to values.alert.port always? Since the default is 8443, we don't need if-else

deployment/helm/synopsys-alert/templates/alert-environ-configmap.yaml
```diff
@@ -1,13 +1,20 @@
 apiVersion: v1
 kind: ConfigMap
 data:
+  {{- if .Values.alert.hostname }}
+  ALERT_HOSTNAME: {{ .Values.alert.hostname | quote }}
+  {{- else }}
+  ALERT_HOSTNAME: {{ .Release.Name | quote }}
```
msenmurugan, line 7: Since the default value is provided as localhost, it will always go to if condition. And also, if someone want to set a different hostname, then they can set them in the flag. So ideally, we can remove the if-else condition and always read it from .Values.alert.hostname

PR #1081, opened 2020-07-23
Azure Boards Context

gkillough:


src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureBoardsChannel.java
```diff
@@ -74,8 +79,22 @@ public IssueTrackerResponse sendRequests(IssueTrackerContext context, List<Issue
...
-        return null;
+        String azureProjectName = fieldAccessor.getStringOrNull(AzureBoardsDescriptor.KEY_AZURE_PROJECT);
+        String workItemCreatorEmail = fieldAccessor.getStringOrNull(AzureBoardsDescriptor.KEY_WORK_ITEM_CREATOR_EMAIL);
+        String workItemTypeName = fieldAccessor.getStringOrNull(AzureBoardsDescriptor.KEY_WORK_ITEM_TYPE);
+        boolean commentOnWorkItems = fieldAccessor.getBooleanOrFalse(AzureBoardsDescriptor.KEY_WORK_ITEM_COMMENT);
+        String completedStateName = fieldAccessor.getStringOrNull(AzureBoardsDescriptor.KEY_WORK_ITEM_COMPLETED_STATE);
+        String reopenStateName = fieldAccessor.getStringOrNull(AzureBoardsDescriptor.KEY_WORK_ITEM_REOPEN_STATE);
+        return new IssueConfig(
+            azureProjectName,
+            azureProjectName,
```
ChomickiM, line 90: should azureProjectName be here twice? Are projectName and projectKey the same?

gkillough: The intention of this class was not clear. The three fields are: 
- projectKey 
- projectName 
- projectId 

There is no concept of projectKey vs projectName in Azure. There is a projectId (UUID type), but it would have to be requested through the API.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureHttpServiceFactory.java
```diff
@@ -30,6 +30,8 @@
 
 public class AzureHttpServiceFactory {
     public static final String DEFAULT_BASE_URL = "https://dev.azure.com";
+    public static final String DEFAULT_AUTHORIZATION_URL = "https://app.vssps.visualstudio.com/oauth2/authorize&response_type=Assertion";
+    public static final String DEFAULT_TOKEN_URL = "https://app.vssps.visualstudio.com/oauth2/token?client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer&grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer";
```
bamandel, line 34: We could add another variable with the Authorization and Token URL hostnames, cut down on duplication.

psantos1113, line 34: Are these URLs always going to be the same?

psantos1113: These are the two URLs will never change for OAuth?

gkillough: As far as I can tell, yes. The base url might be different if they aren't hosted, but it's unclear how OAuth works in that case.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/descriptor/AzureBoardsGlobalUIConfig.java
```diff
@@ -39,20 +37,15 @@
 public class AzureBoardsGlobalUIConfig extends UIConfig {
     public static final String LABEL_AZURE_BOARDS_URL = "Url";
     public static final String LABEL_ORGANIZATION_NAME = "Organization Name";
-    public static final String LABEL_CONSUMER_KEY = "Consumer Key";
-    public static final String LABEL_PRIVATE_KEY = "Private Key";
```
bamandel, line 43: If we're doing OAuth, shouldn't Consumer and private keys always be needed in the handshake?

gkillough: Not for OAuth 2.0.

PR #1084, opened 2020-07-27
Make the status message errors take precedence over action messages

psantos1113:


src/main/js/dynamic/loaded/users/RoleTable.js
```diff
@@ -292,7 +292,7 @@ RoleTable.propTypes = {
 const mapStateToProps = (state) => ({
     roles: state.roles.data,
     descriptors: state.descriptors.items,
-    roleError: state.roles.roleError,
+    roleError: state.roles.error.message,
```
psantos1113, line 295: Roles table was not showing forbidden permission message.

src/main/js/field/StatusMessage.js
```diff
@@ -17,8 +17,7 @@ class StatusMessage extends Component {
     componentDidUpdate(prevProps) {
         if (this.props.errorMessage !== prevProps.errorMessage && !this.state.showError) {
             this.setState({ showError: true });
-        }
-        if (this.props.actionMessage !== prevProps.actionMessage && !this.state.showMessage) {
+        } else if (this.props.actionMessage !== prevProps.actionMessage && !this.state.showMessage) {
```
psantos1113, line 20: Show either the error message or the action message.  Error message takes precedence.

src/main/js/store/actions/roles.js
```diff
@@ -30,7 +30,7 @@ function fetchedAllRoles(roles) {
 function fetchingAllRolesError(message) {
     return {
         type: USER_MANAGEMENT_ROLE_FETCH_ERROR_ALL,
-        roleFetchError: message
+        message
```
psantos1113, line 33: Roles table was not showing forbidden permission message.

src/main/js/store/actions/users.js
```diff
@@ -30,7 +30,7 @@ function fetchedAllUsers(users) {
 function fetchingAllUsersError(message) {
     return {
         type: USER_MANAGEMENT_USER_FETCH_ERROR_ALL,
-        userFetchError: message
+        message
```
psantos1113, line 33: user table was not showing forbidden permission message.

src/main/js/store/actions/roles.js
gkillough: How does this work with no LHS property?

gkillough: Also, this is the roles table.

psantos1113: Javascript expands the message variable to be` 'message': { message } ` it's shorthand if the LHS matches the variable name.

psantos1113: The roles and users table was not displaying the permission denied banner correctly.  It was not setting the correct attributes to be shown in the UI.  That's why the change was made.

PR #1090, opened 2020-07-29
Merge feat warnings into feature branch

ChomickiM:
# Pull Request template

*

*

*
*
*

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
+++ b/src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
@@ -63,6 +63,9 @@ public class FieldValidationAction {
                 if (validationResult.hasErrors()) {
                     fieldErrors.put(key, validationResult.combineErrorMessages());
                 }
+                if (validationResult.hasWarnings()) {
+                    fieldErrors.put(key, validationResult.combineWarningMessages());
+                }
             }
         }
         logger.debug("Finished validating fields in configuration field model.");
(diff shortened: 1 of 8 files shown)


src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
```diff
@@ -63,6 +63,9 @@ public void validateConfig(Map<String, ConfigField> descriptorFields, FieldModel
                 if (validationResult.hasErrors()) {
                     fieldErrors.put(key, validationResult.combineErrorMessages());
                 }
+                if (validationResult.hasWarnings()) {
+                    fieldErrors.put(key, validationResult.combineWarningMessages());
```
jamesrichard91, line 67: Should we be adding warnings to the fieldErrors?

psantos1113: This is a work in progress.  We are keeping the same functionality for now and making sure that the code is in place to be able to add warnings even though they are treated as errors. The next branch and refactoring will to convert to the objects that contain information about warnings or errors. Taking a phased approach to this refactor.

psantos1113: That's our next step.   We know this code works with the existing code and provides the pieces to add warnings.  Next step is to incorporate all of it together.

jamesrichard91: Can we add a TODO in the places that we know need to change?

psantos1113: We will add TODOs as we move forward.  We are working on a feature branch so the feature branch will ultimately be reviewed by everyone.

PR #1101, opened 2020-08-03
Seperate field errors into Warnings and errors

ChomickiM:
This PR is for setting up the groundwork for us to differentiate between warnings and errors in Alert. Currently, with this PR all the functionality should remain the same as all the fields that are created are errors. After this we will review the fields we want to be converted to warnings and make the appropriate changes.


alert-common/src/main/java/com/synopsys/integration/alert/common/exception/AlertFieldException.java
```diff
@@ -27,43 +27,57 @@
...
         this.fieldErrors = fieldErrors;
     }
 
-    public Map<String, String> getFieldErrors() {
+    public Map<String, AlertFieldStatus> getFieldErrors() {
         return fieldErrors;
     }
 
+    //TODO Remove convertToAlertFieldStatus in the future once AlertFieldException has been refactored
+    /*
```
jamesrichard91, line 70: If this is commented out and not needed, it should just be deleted.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/service/IssueCreatorTestAction.java
```diff
@@ -90,17 +91,17 @@ public IssueTrackerResponse testConfig(IssueTrackerContext issueTrackerContext)
         String toStatus = "Resolve";
         Optional<String> possibleSecondIssueKey = Optional.empty();
         try {
-            Map<String, String> transitionErrors = new HashMap<>();
+            Map<String, AlertFieldStatus> transitionErrors = new HashMap<>();
```
bamandel, line 94: I think it would be a good time to stop using a map here and start using an object that handles all this. It would clear up some code and make it easier to understand.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -31,45 +31,76 @@
...
         Collection<String> validationErrors = Arrays.stream(validationResults)
                                                   .map(ValidationResult::getErrors)
                                                   .flatMap(Collection::stream)
                                                   .collect(Collectors.toList());
-        return new ValidationResult(validationErrors);
+        Collection<String> validationWarnings = Arrays.stream(validationResults)
+                                                    .map(ValidationResult::getWarnings)
+                                                    .flatMap(Collection::stream)
+                                                    .collect(Collectors.toList());
+        return new ValidationResult(validationErrors, validationWarnings);
```
bamandel, line 66: You can abstract this further by creating a private method that is passed a function. The function will either do getErrors or getWarnings. You'd then call the method twice here with different functions passed.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/ResponseBodyBuilder.java
```diff
@@ -61,12 +68,16 @@ public ResponseBodyBuilder put(String key, String value) {
...
 
-    public ResponseBodyBuilder putErrors(Map<String, String> errors) {
+    public ResponseBodyBuilder putErrors(Map<String, AlertFieldStatus> errors) {
         JsonObject element = new JsonObject();
-        for (Entry<String, String> entry : errors.entrySet()) {
-            element.addProperty(entry.getKey(), entry.getValue());
+        for (Entry<String, AlertFieldStatus> entry : errors.entrySet()) {
+            AlertFieldStatus alertFieldStatus = entry.getValue();
+            JsonObject statusObject = new JsonObject();
+            statusObject.addProperty(PROPERTY_KEY_SEVERITY, alertFieldStatus.getSeverity().name());
```
bamandel, line 76: You can add a method to AlertFieldsStatus called getSeverityName that does this
alertFieldStatus.getSeverity().name()

src/main/js/field/LabeledField.js
```diff
@@ -21,7 +21,14 @@ class LabeledField extends Component {
 
         const baseClasses = `${labelClass} text-right`;
         const labelClasses = (required) ? `${baseClasses} required` : baseClasses;
-
+        let severity = 'ERROR';
+        let fieldMessage = '';
```
bamandel, line 25: These can be converted into const if you using a ternary operator like you did on line 30 and 31

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/service/IssueCreatorTestAction.java
ChomickiM: Speaking with Paulo, we think this might be a good change to make as well. Since this is a feature branch we will make these changes in another branch and merge into this one rather than splitting it into two separate PRs.

gkillough: Should `AlertFieldStatus` just have a field for `fieldName`? Then we deal with a list everywhere?

src/main/js/field/LabeledField.js
```diff
@@ -21,7 +21,14 @@ class LabeledField extends Component {
...
+        if (errorValue) {
+            severity = errorValue.severity;
+            fieldMessage = errorValue.fieldMessage;
```
gkillough, line 28: { severity, fieldMessage } = errorValue;

psantos1113: That was the original code I had.  However errorValue can be null because there are no errors.  When errorValue is null object destructuring fails.  So that's why we have the code that is there.  If we change errorValue to {} then the check for errorValue will always be true.  We would have to test that

PR #1104, opened 2020-08-04
Fix audit-resend persistence warning

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultAuditUtility.java
```diff
@@ -137,9 +137,10 @@ public DefaultAuditUtility(AuditEntryRepository auditEntryRepository, AuditNotif
         for (Long notificationId : allMessageNotificationIds) {
             AuditEntryEntity auditEntryEntity = new AuditEntryEntity(jobId, DateUtils.createCurrentDateTimestamp(), null, null, null, null);
 
+            boolean didAuditEntryExist = false;
```
jamesrichard91, line 140: When is this ever true?

jamesrichard91: It appears to get set to false and then never changed

gkillough: Line 143

jamesrichard91: Ah! I see. Setting it in the if condition is interesting...

gkillough: Yeah, I'm not sure if it's a convention we are comfortable with.

psantos1113: I myself don't care for the assignment in the if statement.  In my opinion the if statement should just be checking a boolean condition not also assigning the value to the variable that it ultimately checks.  I would assign it beforehand and then check if the variable is true then find the existing item.

ChomickiM: I agree with Paulo, like James I missed it the first time I was reviewing and missed where it was getting set because I'm not used to seeing those get set as part of a conditional

PR #1105, opened 2020-08-06
Azure AD SAML Fixes

psantos1113:
Fix the SAML functionality to work with Azure AD SAML.  Needed to disable signature verification to avoid the certificate errors.  Then needed to add an additional metadata provider to create the SAML hand shake.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/security/authentication/saml/SAMLManager.java
+++ b/src/main/java/com/synopsys/integration/alert/web/security/authentication/saml/SAMLManager.java
@@ -147,6 +147,8 @@ public class SAMLManager {
         return Optional.of(createDelegate(provider));
     }
 
+    // This needs to be created in order for Azure AD SAML configuration to work. The entity id in the metadata is different
+    // than the entity id configured in Azure.  This allows the the entity id to get mapped and found correctly for the application.
     private Optional<ExtendedMetadataDelegate> createMemoryProvider() throws MetadataProviderException {
         EntityDescriptor descriptor = metadataGenerator.generateMetadata();
         MetadataMemoryProvider provider = new MetadataMemoryProvider(descriptor);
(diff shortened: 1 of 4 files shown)


src/main/java/com/synopsys/integration/alert/web/security/authentication/saml/SAMLManager.java
```diff
@@ -142,10 +147,23 @@ public void setupMetadataManager(String metadataURL, String entityId, String ent
...
+        EntityDescriptor descriptor = metadataGenerator.generateMetadata();
+        MetadataMemoryProvider provider = new MetadataMemoryProvider(descriptor);
+        provider.initialize();
+        return Optional.of(createDelegate(provider));
+    }
+
     private ExtendedMetadataDelegate createDelegate(MetadataProvider provider) {
         ExtendedMetadataDelegate delegate = new ExtendedMetadataDelegate(provider, extendedMetadata);
-        delegate.setMetadataTrustCheck(true);
+        delegate.setMetadataTrustCheck(false);
```
gkillough, line 161: Is there something Alert can do to act as a proxy for this check (i.e. if Alert has the certificate imported) or is this not an SSL trust check?

psantos1113: Not exactly an SSL trust check.  The entities can have an X509 tag which is a certificate used to create the entity.  We need to use that signature for validation.  However that requires the configuration of a KeyManager class that is specific to SAML.  It can point to a keystore file to see if the signature is trusted, but there are other constraints with setting that up.  For instance we need to have a default entity id that we look up which we don't know on startup.  How to load the map of entities at runtime wasn't initially clear to get the key validation working.  We should have an enhancement in the future to solve the problem though

PR #1106, opened 2020-08-06
Expose filter-chain servlet request and response to custom endpoints

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomEndpointManager.java
```diff
@@ -24,37 +24,45 @@
...
+    private final Map<String, BiFunction<FieldModel, HttpServletContentWrapper, ResponseEntity<String>>> endpointFunctions = new HashMap<>();
 
     public boolean containsFunction(String functionKey) {
         return endpointFunctions.containsKey(functionKey);
     }
 
     public void registerFunction(String functionKey, Function<FieldModel, ResponseEntity<String>> endpointFunction) throws AlertException {
+        if (containsFunction(functionKey)) {
+            throw new AlertException("A custom endpoint is already registered for " + functionKey);
+        }
```
bamandel, line 50: This part is duplicated between methods. Can we call the method below here?

gkillough: Ugh, meant to do this on line 51.

PR #1107, opened 2020-08-07
Field Message refactor for feat_field_warnings

ChomickiM:
This PR is to merge the changes that were suggested back into feat_field_warnings before merging them both into master. Let me know if there are any questions or opinions with the changes I made from the original PR.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
```diff
@@ -103,4 +99,11 @@ public String combineWarningMessages() {
     private String combineMessages(Collection<String> messages) {
         return StringUtils.join(messages, ", ");
     }
+
+    private static Collection<String> getStatus(Function<ValidationResult, Collection<String>> mapResults, ValidationResult... validationResults) {
```
jamesrichard91, line 103: Is the idea for this method to turn the Array of ValidationResult's to a Collection?

alert-common/src/main/java/com/synopsys/integration/alert/common/exception/AlertFieldException.java
```diff
@@ -22,52 +22,52 @@
...
 public class AlertFieldException extends AlertException {
     private static final long serialVersionUID = 7993564907680483145L;
 
-    private final Map<String, AlertFieldStatus> fieldErrors;
+    private final List<AlertFieldStatus> fieldErrors;
 
-    public static AlertFieldException singleFieldError(String fieldKey, AlertFieldStatus fieldError) {
-        return new AlertFieldException(Map.of(fieldKey, fieldError));
+    public static AlertFieldException singleFieldError(String fieldKey, String fieldError) {
+        //return new AlertFieldException(Map.of(fieldKey, fieldError));
```
jamesrichard91, line 35: We should delete the commented out lines of code.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
gkillough, line 103: Can we change the variable name `mapResults` to `extractValidationResults`.

src/main/java/com/synopsys/integration/alert/component/authentication/actions/AuthenticationTestAction.java
```diff
@@ -67,9 +67,9 @@ public MessageResult testConfig(String configId, FieldModel fieldModel, FieldAcc
         boolean samlEnabled = registeredFieldValues.getBooleanOrFalse(AuthenticationDescriptor.KEY_SAML_ENABLED);
         if (!ldapEnabled && !samlEnabled) {
             String errorMessage = "Enable LDAP or SAML authentication.";
-            Map<String, AlertFieldStatus> errorsMap = Map.of(
-                AuthenticationDescriptor.KEY_LDAP_ENABLED, AlertFieldStatus.error(errorMessage),
-                AuthenticationDescriptor.KEY_SAML_ENABLED, AlertFieldStatus.error(errorMessage));
+            List<AlertFieldStatus> errorsMap = List.of(
```
jamesrichard91, line 70: We should update the variable name since it is no longer a Map.

src/main/java/com/synopsys/integration/alert/component/authentication/actions/AuthenticationTestAction.java
```diff
@@ -112,7 +112,7 @@ private void performSAMLTest(FieldAccessor registeredFieldValues) throws Integra
         Optional<ConfigurationFieldModel> metaDataFileField = registeredFieldValues.getField(AuthenticationDescriptor.KEY_SAML_METADATA_FILE);
         boolean testMetaDataURL = metaDataURLField.map(ConfigurationFieldModel::isSet).orElse(false);
         boolean testMetaDataFile = metaDataFileField.map(ConfigurationFieldModel::isSet).orElse(false);
-        Map<String, AlertFieldStatus> errorsMap = new HashMap<>();
+        List<AlertFieldStatus> errorsMap = new ArrayList<>();
```
jamesrichard91, line 115: Same

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validators/ValidationResult.java
ChomickiM: this was created to simplify calling getErrors and getWarnings in the static of method above. We are using the `ValidationResult(Collection<String> errors, Collection<String> warnings)` constructor to set the errors and warnings collections.

src/main/java/com/synopsys/integration/alert/component/authentication/actions/AuthenticationTestAction.java
```diff
@@ -89,15 +89,15 @@ private void performLdapTest(FieldModel fieldModel, FieldAccessor registeredFiel
         String userName = fieldModel.getFieldValue(AuthenticationUIConfig.TEST_FIELD_KEY_USERNAME).orElse("");
         Optional<LdapAuthenticationProvider> ldapProvider = ldapManager.createAuthProvider(registeredFieldValues);
         String errorMessage = String.format("Ldap Authentication test failed for the test user %s.  Please check the LDAP configuration.", userName);
-        Map<String, AlertFieldStatus> errorsMap = new HashMap<>();
+        List<AlertFieldStatus> errorsMap = new ArrayList<>();
```
jamesrichard91, line 92: Update variable name

src/main/java/com/synopsys/integration/alert/component/authentication/actions/AuthenticationTestAction.java
gkillough, line 70: Let's rename this to just `errors` because it is no longer a Map. It also helps if we change the type in the future so that it's not coupled to a specific data structure.

src/main/java/com/synopsys/integration/alert/component/authentication/actions/AuthenticationTestAction.java
gkillough, line 92: `errors`

gkillough:
I'm very tempted to suggest we ditch the List/Map approach entirely and manage the collection ourselves in an `AlertFieldStatuses` object or something. It seems like 90% of the calls are creating an error or warning nested in a `List.add`. Might be nice to be able to call `AlertFieldStatuses.addError` or `AlertFieldStatuses.addWarning`. Of course at that point we could wrap a `Map` or any `Collection` to suit the needs of the class.

src/test/java/com/synopsys/integration/alert/util/AlertFieldStatusConverter.java
```diff
@@ -0,0 +1,22 @@
+package com.synopsys.integration.alert.util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import com.synopsys.integration.alert.common.descriptor.config.field.errors.AlertFieldStatus;
+
+public class AlertFieldStatusConverter {
```
bamandel, line 10: We have DataStructureUtils which essentially does these conversions I believe. I would verify whether this class is absolutely necessary.

PR #1108, opened 2020-08-11
Fixing Sonar Cloud issues

ChomickiM:
Previously we would be checking a List<AlertFieldStatus> to see if it contains a String. It's now updated to see if the key exists by the getFieldName

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
+++ b/src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
@@ -53,7 +53,9 @@ public class FieldValidationAction {
                 fieldErrors.add(AlertFieldStatus.error(key, ConfigField.REQUIRED_FIELD_MISSING));
             }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/synopsys/integration/alert/web/config/FieldValidationAction.java
```diff
@@ -53,7 +53,9 @@ public void validateConfig(Map<String, ConfigField> descriptorFields, FieldModel
                 fieldErrors.add(AlertFieldStatus.error(key, ConfigField.REQUIRED_FIELD_MISSING));
             }
 
-            if (!fieldErrors.contains(key) && optionalFieldValue.isPresent()) {
+            boolean containsKey = fieldErrors.stream()
+                                      .anyMatch(fieldErrorName -> key.equals(fieldErrorName.getFieldName()));
```
gkillough, line 57: Could this be changed to the following?
```suggestion
                                      .map(AlertFieldStatus::getFieldName)
                                      .anyMatch(key::equals);
```

PR #1111, opened 2020-08-12
Initial oauth support

psantos1113:
Initial set of changes to provide OAuth support for Azure Boards.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/oauth/OAuthCustomEndpoint.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.gson.Gson;
+import com.synopsys.integration.alert.common.action.CustomEndpointManager;
+import com.synopsys.integration.alert.common.descriptor.config.field.endpoint.ButtonCustomEndpoint;
+import com.synopsys.integration.alert.common.exception.AlertException;
+import com.synopsys.integration.alert.common.rest.HttpServletContentWrapper;
+import com.synopsys.integration.alert.common.rest.ResponseFactory;
+import com.synopsys.integration.alert.common.rest.model.FieldModel;
+
+public abstract class OAuthCustomEndpoint extends ButtonCustomEndpoint {
```
gkillough, line 35: I think it might be best if we make this extend `CustomEndpoint` so that we aren't bound to return a `String` later. We should start moving away from `ResponseEntity<String>` IMO. If we do that, we can just return a `OAuthEndpointResponse` rather than wrapping and manually serializing with `Gson`.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/ProxyManager.java
```diff
@@ -113,4 +115,16 @@ public ProxyInfo createProxyInfo() throws IllegalArgumentException {
         return settingsConfiguration.flatMap(configurationModel -> configurationModel.getField(key)).flatMap(ConfigurationFieldModel::getFieldValue);
     }
 
+    public Proxy createProxy() {
+        if (!getProxyHost().isPresent()) {
```
gkillough, line 119: Why not `isEmpty()`? SonarCloud?

psantos1113: Yeah not sure if the SonarCloud issues has been updated yet.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/oauth/OAuthCustomEndpoint.java
psantos1113: We will need to refactor the CustomEndpointManager and CustomEndpoint.  Although CustomEndpoint uses a generic R the method always returns a Response<String>

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
```diff
@@ -22,37 +22,118 @@
...
+            if (!clientId.isPresent()) {
+                return new OAuthEndpointResponse(HttpStatus.BAD_REQUEST.value(), false, "", "client id not found.");
+            }
+            Optional<String> alertServerUrl = alertProperties.getServerUrl();
+
+            if (!alertServerUrl.isPresent()) {
+                return new OAuthEndpointResponse(HttpStatus.BAD_REQUEST.value(), false, "", "Could not determine the alert server url for the callback.");
+            }
+            String authUrl = createAuthURL(clientId.get(), alertServerUrl.get());
+            logger.info("Authenticating Azure OAuth URL: " + authUrl);
```
gkillough, line 87: This should probably be logged at DEBUG.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
```diff
@@ -22,37 +22,118 @@
...
+    }
+
+    private FieldAccessor createFieldAccessor(FieldModel fieldModel) {
+        Map<String, ConfigurationFieldModel> fields = new HashMap<>();
+        try {
+            fields.putAll(modelConverter.convertToConfigurationFieldModelMap(fieldModel));
+            // check if a configuration exists because the client id is a sensitive field and won't have a value in the field model if updating.
+            if (StringUtils.isNotBlank(fieldModel.getId())) {
+                Optional<ConfigurationModel> configurationFieldModel = configurationAccessor.getConfigurationById(Long.valueOf(fieldModel.getId()));
+                configurationFieldModel.ifPresent(model -> fields.putAll(model.getCopyOfKeyToFieldMap()));
```
gkillough, line 105: Could we do something like this?
```suggestion
                configurationAccessor.getConfigurationById(Long.valueOf(fieldModel.getId()))
                .map(ConfigurationModel::getCopyOfKeyToFieldMap)
                .ifPresent(fields::putAll);
```

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
```diff
@@ -22,37 +22,118 @@
...
+            }
+        } catch (AlertDatabaseConstraintException ex) {
+            logger.error("Error creating field acessor for Azure authentication", ex);
+        }
+        return new FieldAccessor(fields);
     }
 
+    private String createAuthURL(String clientId, String alertServerUrl) {
+        StringBuilder authUrlBuilder = new StringBuilder(300);
```
gkillough, line 114: Is the 300 here the max length? We should probably throw an exception if we violate that constraint if it's required by Microsoft.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
```diff
@@ -22,37 +22,118 @@
...
+        authUrlBuilder.append(AzureHttpServiceFactory.DEFAULT_AUTHORIZATION_URL);
+        authUrlBuilder.append(createQueryString(clientId, alertServerUrl));
+        return authUrlBuilder.toString();
+    }
+
+    private String createQueryString(String clientId, String alertServerUrl) {
+        String authorizationUrl = String.format("%s%s", alertServerUrl, AzureOauthCallbackController.AZURE_OAUTH_CALLBACK_PATH);
+        StringBuilder queryBuilder = new StringBuilder(250);
```
gkillough, line 122: Same thought here. If this is the length requirement, we should be more explicit about it.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
```diff
@@ -22,37 +22,118 @@
...
+        queryBuilder.append("&client_id=");
```
gkillough, line 123: Should this first string start with '?', or is that from a previous constant?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
```diff
@@ -22,37 +22,118 @@
...
+        queryBuilder.append(clientId);
+        //TODO have an object that stores the request keys and purges them after some amount of time.
+        //TODO also store a redirect URL if possible
+        queryBuilder.append("&state=");
+        queryBuilder.append(createRequestKey());
+        queryBuilder.append("&scope=vso.work%20vso.code_write");
```
gkillough, line 129: I believe we need to request more scopes than this. They are listed on the confluence design page.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
psantos1113: The 250 isn't a max length.  It's a best guess for initial capacity.  My test URL was 237 characters long. This is so the StringBuilder allocates a string with enough capacity initially instead of resizing more than once with the append calls.  it will resize dynamically like an ArrayList.  So this just makes it perform a little better.  The string underneath is backed by a character array.  So the character array will be created with a good enough size to avoid resizing during the append calls.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
psantos1113: The URL for authorization appends the assertion query parameter and I had to fix the URL constant to change it from '&' to '?'  See line 116 'AzureHttpServiceFactory.DEFAULT_AUTHORIZATION_URL'

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
psantos1113: yeah it makes sense because the callback is failing to return tokens. the URL returned lists invalid scopes

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomEndpoint.java
psantos1113: The 300 isn't a max length.  It's a best guess for initial capacity.  My test URL was 237 characters long. This is so the StringBuilder allocates a string with enough capacity initially instead of resizing more than once with the append calls.  it will resize dynamically like an ArrayList.  So this just makes it perform a little better. The string underneath is backed by a character array. So the character array will be created with a good enough size to avoid resizing during the append calls.

src/main/js/field/OAuthEndpointButtonField.js
```diff
@@ -0,0 +1,177 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import GeneralButton from 'field/input/GeneralButton';
+import FieldsPopUp from 'field/FieldsPopUp';
+import LabeledField from 'field/LabeledField';
+import * as FieldModelUtilities from 'util/fieldModelUtilities';
+import { createNewConfigurationRequest } from 'util/configurationRequestBuilder';
+import { connect } from 'react-redux';
+import StatusMessage from 'field/StatusMessage';
+import * as HTTErrorUtils from 'util/httpErrorUtilities';
```
bamandel, line 10: This is imported with the wrong name

psantos1113: The 'as' keyword allows the import of all functions and classes in util/httpErrorUtilities to be references via the name provided after the as keyword.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureOauthCallbackController.java
```diff
@@ -0,0 +1,103 @@
...
+        this.azureBoardsCredentialDataStoreFactory = azureBoardsCredentialDataStoreFactory;
+        this.proxyManager = proxyManager;
+        this.configurationAccessor = configurationAccessor;
+        this.configFieldModelConverter = configFieldModelConverter;
+    }
+
+    @GetMapping
+    public ResponseEntity<String> oauthCallback() {
+        try {
+            logger.info("callback method called");
```
gkillough, line 82: This looks like it was added for testing, we may want to log something similar at DEBUG.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/oauth/OAuthCustomEndpoint.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.gson.Gson;
+import com.synopsys.integration.alert.common.action.CustomEndpointManager;
+import com.synopsys.integration.alert.common.descriptor.config.field.endpoint.CustomEndpoint;
+import com.synopsys.integration.alert.common.exception.AlertException;
+import com.synopsys.integration.alert.common.rest.HttpServletContentWrapper;
+import com.synopsys.integration.alert.common.rest.ResponseFactory;
+import com.synopsys.integration.alert.common.rest.model.FieldModel;
+
+public abstract class OAuthCustomEndpoint extends CustomEndpoint<String> {
```
gkillough, line 35: Don't we want this to extend `CustomEndpoint<OAuthEndpointResponse>`?

psantos1113: We would need to refactor CustomEndpoint and CustomEndpointManager which may cause bugs elsewhere.

src/main/js/field/OAuthEndpointButtonField.js
bamandel: Oh sorry, just meant it was spelled wrong. It should be HTTPErrorUtils

PR #1112, opened 2020-08-13
Enhance MessageResult

gkillough:
- Include field statuses in `MessageResult` to allow channels to expose more granular error/warning messages.  
- Implement combined usage of `MessageResult` in `JobConfigActions` for job validation, provider validation, and test messages

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageResult.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageResult.java
@@ -34,6 +34,20 @@ public class MessageResult extends AlertSerializableModel {
     private final String statusMessage;
     private final List<AlertFieldStatus> fieldStatuses;
 
+    public static List<AlertFieldStatus> getFieldStatusesBySeverity(List<AlertFieldStatus> fieldStatuses, FieldStatusSeverity severity) {
+        return fieldStatuses
+                   .stream()
+                   .filter(status -> status.getSeverity().equals(severity))
+                   .collect(Collectors.toList());
+    }
+
+    public static boolean hasFieldStatusBySeverity(List<AlertFieldStatus> fieldStatuses, FieldStatusSeverity severity) {
+        return fieldStatuses
+                   .stream()
+                   .map(AlertFieldStatus::getSeverity)
+                   .anyMatch(severity::equals);
+    }
+
     public MessageResult(String statusMessage) {
         this.statusMessage = statusMessage;
         fieldStatuses = List.of();
@@ -75,17 +89,11 @@ public class MessageResult extends AlertSerializableModel {
     }
 
     public List<AlertFieldStatus> getFieldStatusesBySeverity(FieldStatusSeverity severity) {
-        return fieldStatuses
-                   .stream()
-                   .filter(status -> status.getSeverity().equals(severity))
-                   .collect(Collectors.toList());
+        return getFieldStatusesBySeverity(fieldStatuses, severity);
     }
 
     public boolean hasFieldStatusBySeverity(FieldStatusSeverity severity) {
-        return fieldStatuses
-                   .stream()
-                   .map(AlertFieldStatus::getSeverity)
-                   .anyMatch(severity::equals);
+        return hasFieldStatusBySeverity(fieldStatuses, severity);
     }
 
 }
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageResult.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageResult.java
@@ -33,6 +33,11 @@ public class MessageResult extends AlertSerializableModel {
     private final String statusMessage;
     private final List<AlertFieldStatus> fieldStatuses;
 
+    public static MessageResult singleFieldError(String fieldName, String errorMessage) {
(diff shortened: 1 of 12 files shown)


src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -129,8 +131,13 @@ public void deleteJobById(UUID id) throws AlertException {
     }
 
     public JobFieldModel saveJob(JobFieldModel jobFieldModel) throws AlertException {
-        validateJob(jobFieldModel);
+        MessageResult validationResult = validateJob(jobFieldModel);
+        List<AlertFieldStatus> fieldStatuses = validationResult.getFieldStatuses();
+        if (!fieldStatuses.isEmpty()) {
+            throw new AlertFieldException(fieldStatuses);
+        }
```
bamandel, line 138: This is done in multiple places, we should add a method that performs this check and throws the exception to the MessageResult object.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -152,7 +159,11 @@ public JobFieldModel saveJob(JobFieldModel jobFieldModel) throws AlertException
     }
 
     public JobFieldModel updateJob(UUID id, JobFieldModel jobFieldModel) throws AlertException {
-        validateJob(jobFieldModel);
+        MessageResult validationResult = validateJob(jobFieldModel);
+        List<AlertFieldStatus> fieldStatuses = validationResult.getFieldStatuses();
+        if (!fieldStatuses.isEmpty()) {
+            throw new AlertFieldException(fieldStatuses);
+        }
```
bamandel, line 166: Another example of the exception throwing.

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
```diff
@@ -226,37 +237,41 @@ private boolean filterOutMatchingJobs(@Nullable UUID currentJobId, Configuration
...
+            fieldStatuses.addAll(fieldModelProcessor.validateFieldModel(fieldModel));
         }
 
-        if (!fieldErrors.isEmpty()) {
-            throw new AlertFieldException(fieldErrors);
+        if (!fieldStatuses.isEmpty()) {
+            return new MessageResult("Invalid", fieldStatuses);
         }
-        return "Valid";
+        return new MessageResult("Valid");
```
bamandel, line 249: I think this could be converted into a stream.

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageResult.java
```diff
@@ -22,21 +22,63 @@
...
+import java.util.stream.Collectors;
+
+import com.synopsys.integration.alert.common.descriptor.config.field.errors.AlertFieldStatus;
+import com.synopsys.integration.alert.common.descriptor.config.field.errors.FieldStatusSeverity;
 import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 
 public class MessageResult extends AlertSerializableModel {
-    private String statusMessage;
+    private final String statusMessage;
+    private final List<AlertFieldStatus> fieldStatuses;
```
jamesrichard91, line 34: Should this field be an instance of JobFieldStatuses? and the helper methods moved into the JobFieldStatuses class?

src/main/java/com/synopsys/integration/alert/provider/blackduck/actions/BlackDuckDistributionTestAction.java
```diff
@@ -78,22 +79,22 @@ public MessageResult testConfig(String configId, FieldModel fieldModel, FieldAcc
...
+                    fieldStatuses.add(AlertFieldStatus.error(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME, "User permission failed, cannot read notifications from Black Duck."));
                 }
             }
         } else {
-            throw AlertFieldException.singleFieldError(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME, "A provider configuration is required");
+            fieldStatuses.add(AlertFieldStatus.error(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME, "A provider configuration is required"));
         }
-
-        return new MessageResult("Successfully tested BlackDuck provider fields");
+        return new MessageResult("Successfully tested BlackDuck provider fields", fieldStatuses);
```
jamesrichard91, line 88: This could be confusing to handle further down the line, since the message will say success but there could be errors

src/main/java/com/synopsys/integration/alert/web/config/JobConfigActions.java
gkillough: I was hesitant to do this because I eventually want to get rid of AlertFieldException and just return the results object.

bamandel: I agree with that sentiment, but it doesn't hurt to have the method right now then remove it when it's no longer needed

alert-common/src/main/java/com/synopsys/integration/alert/common/message/model/MessageResult.java
gkillough: JobFieldStatues would be a sub-set of the use case intended for this class.

PR #1116, opened 2020-08-18
Initial Swagger

gkillough:
These changes add auto-generated swagger at `{baseUrl}/swagger-ui.html`. There will need to be several additional improvements needed for this swagger to be useful because of the bad practice in several of our controllers.


src/main/java/com/synopsys/integration/alert/AboutReader.java
```diff
@@ -78,4 +84,9 @@ public String getProductVersion() {
         }
     }
 
+    private String createInternalUrl(String path) {
+        String baseUrl = alertProperties.getServerUrl().orElse("https://localhost:8443/alert");
```
jamesrichard91, line 88: Won't this redirect the user in the browser to this URL? So shouldn't we get the public URL of Alert? or is the idea here to limit it to only users with access to the container?

gkillough: The "localhost" portion is just because the URL is optional. If there is another variable available, I can use that, but I don't know of one.

psantos1113: alert properties will have the public URL if the ALERT_SERVER_HOST environment variable is set. It's the same variable we use to put in the emails for the alert server url

PR #1118, opened 2020-08-19
Refactor Web Packages

gkillough:


gkillough:
> In a PR with the purpose of creating/moving packages, we might want to outline the actual moves in the PR description. It's tough to tell here without pulling the branch.

@bamandel Yeah, I should have done that. It is thoroughly laid out in a Jira ticket, but that doesn't translate well to GitHib with the internal-external system barrier.

PR #1120, opened 2020-08-19
 oauth  request validation

psantos1113:
Create a small in memory cache to manage the OAuth request keys contained in the state as part of validating a request to the AzureOAuthCallbackController.  If it isn't a request that Alert has made then do not process the request further.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/OAuthRequestPurgeTask.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/OAuthRequestPurgeTask.java
@@ -35,7 +35,7 @@ import com.synopsys.integration.alert.common.workflow.task.TaskManager;
 @Component
 public class OAuthRequestPurgeTask extends StartupScheduledTask {
     // every 5 minutes.
-    public static final String CRON_EXPRESSION = "0 0/5 * 1/1 * ?";
+    public static final String CRON_EXPRESSION_EVERY_5_MINUTES = "0 0/5 * 1/1 * ?";
     private final OAuthRequestValidator oAuthRequestValidator;
 
     @Autowired
@@ -46,7 +46,7 @@ public class OAuthRequestPurgeTask extends StartupScheduledTask {
 
     @Override
     public String scheduleCronExpression() {
-        return CRON_EXPRESSION;
+        return CRON_EXPRESSION_EVERY_5_MINUTES;
     }
 
     @Override
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/OAuthRequestPurgeTask.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/OAuthRequestPurgeTask.java
@@ -34,7 +34,8 @@ import com.synopsys.integration.alert.common.workflow.task.TaskManager;
 
 @Component
 public class OAuthRequestPurgeTask extends StartupScheduledTask {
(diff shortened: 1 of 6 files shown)


src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureOAuthCallbackController.java
```diff
@@ -72,73 +74,94 @@
...
+            if (!oAuthRequestValidator.hasRequestKey(state)) {
+                logger.info("OAuth request {} not found.", state);
             } else {
-                if (StringUtils.isBlank(authorizationCode)) {
-                    logger.error("Azure oauth callback: Authorization code isn't valid. Stop processing");
+                logger.info(createOAuthRequestLoggerMessage(state, "Processing..."));
+                oAuthRequestValidator.removeAuthorizationRequest(state);
+                FieldAccessor fieldAccessor = createFieldAccessor();
+                if (fieldAccessor.getFields().isEmpty()) {
+                    logger.error(createOAuthRequestLoggerMessage(state, "Azure oauth callback: Channel global configuration missing"));
```
bamandel, line 116: Could this be improved to an actual response? Or do we want to continue processing?

psantos1113: Azure will be calling this controller since this is the callback URL.  So to send a response here would send it to the Azure server which would not update the Alert UI for the user.  We redirect back to the Alert UI at the end.  We could put query parameters with some details however we don't have a mechanism now to parse the query parameters in the URL to update the UI state at all.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/OAuthRequestPurgeTask.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.workflow.task.StartupScheduledTask;
+import com.synopsys.integration.alert.common.workflow.task.TaskManager;
+
+@Component
+public class OAuthRequestPurgeTask extends StartupScheduledTask {
+    // every 5 minutes.
+    public static final String CRON_EXPRESSION = "0 0/5 * 1/1 * ?";
```
gkillough, line 38: Why not make the variable more self-revealing? `CRON_EXPRESSION_EVERY_5_MINUTES`?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/OAuthRequestValidator.java
```diff
@@ -0,0 +1,62 @@
...
+        return requestMap.containsKey(requestKey);
+    }
+
+    public void removeRequestsOlderThanInstant(Instant instant) {
+        Set<Map.Entry<String, Instant>> entriesToRemove = requestMap.entrySet().stream()
+                                                              .filter(entry -> entry.getValue().isBefore(instant))
+                                                              .collect(Collectors.toSet());
+        entriesToRemove.stream()
+            .map(Map.Entry::getKey)
+            .forEach(this::removeAuthorizationRequest);
```
gkillough, line 60: Why not do this in one stream and just check `isAfter` instead?

gkillough: You can use `.distinct()` on the stream if you want to avoid duplicates.

PR #1122, opened 2020-08-20
Fix Audit Responses

gkillough:


src/main/js/store/actions/distributions.js
```diff
@@ -124,8 +122,7 @@ function fetchAuditInfoForJob(jobConfig) {
                 }
             }).then((response) => {
                 if (response.ok) {
-                    response.json().then((jsonObj) => {
-                        const auditInfo = JSON.parse(jsonObj.message);
+                    response.json().then((auditInfo) => {
                         if (auditInfo != null) {
```
gkillough, line 126: This is the change. The rest is linting.

PR #1123, opened 2020-08-20
Add validator for OAuth 

psantos1113:
Was going to add this in 6.2.0 but the warnings are treated as errors and would prevent saving.  This will allow pushing the save and test button and having the UI display a warning message.

Fixed the cause of a circular dependency in the ConfigurationFieldModelConverter.  The class uses the descriptor keys.  If we use the DescriptorMap it needs all the descriptors which is where the validator is getting used.  So it can't autowire the objects correctly.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureRedirectUtil.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureRedirectUtil.java
@@ -38,6 +38,10 @@ public class AzureRedirectUtil {
         this.alertProperties = alertProperties;
     }
 
+    /**
+     * The OAuth callback controller will redirect back to the Alert UI and is the only place where this method.
+     * @return The url location to redirect to the UI.
+     */
     public String createUIRedirectLocation() {
         StringBuilder locationBuilder = new StringBuilder(200);
         alertProperties.getServerUrl()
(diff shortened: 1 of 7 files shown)


src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureRedirectUtil.java
```diff
@@ -38,6 +38,10 @@ public AzureRedirectUtil(AlertProperties alertProperties) {
         this.alertProperties = alertProperties;
     }
 
+    /**
+     * The OAuth callback controller will redirect back to the Alert UI and is the only place where this method.
```
bamandel, line 42: Incomplete sentence?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/AzureOAuthTokenValidator.java
```diff
@@ -0,0 +1,77 @@
...
+    @Autowired
+    public AzureOAuthTokenValidator(AzureRedirectUtil azureRedirectUtil, AzureBoardsCredentialDataStoreFactory azureBoardsCredentialDataStoreFactory,
+        ConfigurationFieldModelConverter configurationFieldModelConverter, ProxyManager proxyManager) {
+        this.azureRedirectUtil = azureRedirectUtil;
+        this.azureBoardsCredentialDataStoreFactory = azureBoardsCredentialDataStoreFactory;
+        this.configurationFieldModelConverter = configurationFieldModelConverter;
+        this.proxyManager = proxyManager;
+    }
+
+    public ValidationResult validate(FieldValueModel fieldValueModel, FieldModel fieldModel) {
```
bamandel, line 57: is/will this method be used? I'm not really sure what the point of it is.

psantos1113: I can remove it now that I've figured out the circular dependency issue.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureRedirectUtil.java
psantos1113: Thanks I will add that in.

PR #1131, opened 2020-08-24
Fix Certificate Controller

gkillough:
- Fix response statuses 
- Fix return types
- Separate field validation from write operations


src/main/java/com/synopsys/integration/alert/web/api/certificate/CertificateActions.java
```diff
@@ -69,6 +70,43 @@ public CertificateActions(CustomCertificateAccessor certificateAccessor, Certifi
                    .map(this::convertFromDatabaseModel);
     }
 
+    public MessageResult validateCertificate(CertificateModel certificateModel) {
+        CustomCertificateModel convertedModel = convertToDatabaseModel(certificateModel);
```
bamandel, line 74: The class names are confusing to me here. We have a CertificateModel that we run convertToDatabaseModel which changes it to a CustomCertificateModel. I feel like the name CustomCertificateModel is misleading? Or perhaps the convertToDatabaseModel Should be changed?

gkillough: The database API serves a `CustomCertificateModel`. The REST API serves a `CertificateModel`.

bamandel: I guess I'm confused about what a custom certificate is. Is it just a self signed certificate? Or is the user somehow customizing this model?

gkillough: It would probably be better to call it a `UserProvidedCertificate` or something, but this code is actually just moved code from lowing in the class. I could make the change in this PR, but it's slightly out of scope.

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
```diff
@@ -91,6 +91,16 @@ class CertificatesPage extends Component {
...
+
+        const fieldStatuses = fieldErrors.fieldErrors;
+        let fieldErrorMap = {};
+        if (fieldStatuses) {
+            fieldStatuses.forEach((fieldStatus) => {
+                fieldErrorMap[fieldStatus.fieldName] = fieldStatus;
+            });
+        } else {
+            fieldErrorMap = {};
+        }
```
bamandel, line 103: It doesn't look like this else statement is necessary.

src/main/js/dynamic/loaded/certificates/CertificatesPage.js
```diff
@@ -91,6 +91,14 @@ class CertificatesPage extends Component {
     createModalFields() {
         const { certificate } = this.state;
         const { fieldErrors } = this.props;
+
+        const fieldStatuses = fieldErrors.fieldErrors;
```
psantos1113, line 95: This code could most likely be moved into the httpErrorUtilities.js file to be used elsewhere until we convert to using the validation rest endpoint with a better object.

gkillough: Yeah, this was actually the line I was concerned about. Not sure how it was working before, but I think the best solution might be to rename the prop for now so we don't have the weird nesting.

PR #1132, opened 2020-08-24
Feat azure testing

psantos1113:
Fix issues with creating and transitioning issues in Azure.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsIssueHandler.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsIssueHandler.java
@@ -163,8 +163,7 @@ public class AzureBoardsIssueHandler extends IssueHandler<WorkItemResponseModel>
                 JsonElement stateElement = workItemResponse.getFields().get(WorkItemResponseFields.System_State.getFieldName());
                 return transitionName.equals(stateElement.getAsString());
             } catch (HttpServiceException ex) {
-                logger.error("Error transitioning work item {} to {}", issueModel.getId(), transitionName);
-                logger.error("Cause: ", ex);
+                logger.error("Error transitioning work item {} to {}", issueModel.getId(), transitionName, ex);
             }
         }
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsIssueHandler.java
```diff
@@ -160,10 +186,15 @@ protected String getIssueKey(WorkItemResponseModel issueModel) {
     protected IssueTrackerIssueResponseModel createResponseModel(AlertIssueOrigin alertIssueOrigin, String issueTitle, IssueOperation issueOperation, WorkItemResponseModel issueResponse) {
         Integer workItemId = issueResponse.getId();
         Map<String, ReferenceLinkModel> issueLinks = issueResponse.getLinks();
-        ReferenceLinkModel htmlLink = issueLinks.get("html");
-        String uiLink = Optional.ofNullable(htmlLink)
-                            .map(ReferenceLinkModel::getHref)
-                            .orElseGet(this::getIssueTrackerUrl);
+        // the AzureWorkItemReponst does not contain any links when called via the rest API.
```
gkillough, line 189: - Typo in "AzureWorkItemReponst". 
- What does "when called via the rest API" mean? In the response from a POST request? I believe it is there for a GET request.
- We need this link for the callback to Black Duck, so we should add a GET here if necessary.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsIssueHandler.java
```diff
@@ -135,13 +137,37 @@ public AzureBoardsIssueHandler(IssueContentLengthValidator issueContentLengthVal
...
+            try {
+                WorkItemElementOperationModel descriptionField = createUpdateFieldModel(WorkItemResponseFields.System_State, transitionName);
+                requestElementOps.add(descriptionField);
+                WorkItemRequest request = new WorkItemRequest(requestElementOps);
+                WorkItemResponseModel workItemResponse = azureWorkItemService.updateWorkItem(azureBoardsProperties.getOrganizationName(), issueConfig.getProjectName(), issueModel.getId(), request);
+                JsonElement stateElement = workItemResponse.getFields().get(WorkItemResponseFields.System_State.getFieldName());
+                return transitionName.equals(stateElement.getAsString());
+            } catch (HttpServiceException ex) {
+                logger.error("Error transitioning work item {} to {}", issueModel.getId(), transitionName);
+                logger.error("Cause: ", ex);
```
bamandel, line 167: Can we call the error method just once here?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsIssueHandler.java
psantos1113: the retrieveExistingIssues call in AzureBoardsIssueHandler makes an API call that returns a WorkItemResponse that contains null for the links.  When the createWorkItemRequest method is called in AzureBoardsIssueHandler the links aren't null.  That's the issue I was seeing.

PR #1134, opened 2020-08-25
Fix Azure Custom Fields

gkillough:
The hope with the original custom fields was since they are potentially user-visible, to minimize the number of them. Unfortunately, the text fields we used have a size limit of 255, and the other option is not queryable beyond 255 characters, so we needed to add more fields.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
@@ -103,29 +103,6 @@ public class AzureCustomFieldManager {
 
     public void installCustomFields(String projectName, String workItemTypeName) throws AlertException {
         installCustomFields(projectName, workItemTypeName, AZURE_CUSTOM_FIELDS);
-        //        Future<ProjectWorkItemFieldModel> providerKeyFieldCreationResultHolder =
-        //            executorService.submit(() -> findOrCreateAlertCustomProjectField(projectName, ALERT_PROVIDER_KEY_FIELD_NAME, ALERT_PROVIDER_KEY_FIELD_REFERENCE_NAME, ALERT_PROVIDER_KEY_FIELD_DESCRIPTION));
-        //        Future<ProjectWorkItemFieldModel> topicKeyFieldCreationResultHolder =
-        //            executorService.submit(() -> findOrCreateAlertCustomProjectField(projectName, ALERT_TOPIC_KEY_FIELD_NAME, ALERT_TOPIC_KEY_FIELD_REFERENCE_NAME, ALERT_TOPIC_KEY_FIELD_DESCRIPTION));
-        //        Future<ProjectWorkItemFieldModel> componentKeyFieldCreationResultHolder =
-        //            executorService.submit(() -> findOrCreateAlertCustomProjectField(projectName, ALERT_COMPONENT_KEY_FIELD_NAME, ALERT_COMPONENT_KEY_FIELD_REFERENCE_NAME, ALERT_COMPONENT_KEY_FIELD_DESCRIPTION));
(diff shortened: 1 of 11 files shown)


src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
```diff
@@ -46,13 +49,43 @@
...
 public class AzureCustomFieldManager {
-    public static final String ALERT_TOP_LEVEL_KEY_FIELD_NAME = "Alert Top Level Key";
-    public static final String ALERT_TOP_LEVEL_KEY_FIELD_REFERENCE_NAME = "Custom.AlertTopLevelKey";
-    public static final String ALERT_TOP_LEVEL_KEY_FIELD_DESCRIPTION = "A top-level tracking key for Alert";
-
-    public static final String ALERT_COMPONENT_LEVEL_KEY_FIELD_NAME = "Alert Component Level Key";
-    public static final String ALERT_COMPONENT_LEVEL_KEY_FIELD_REFERENCE_NAME = "Custom.AlertComponentLevelKey";
-    public static final String ALERT_COMPONENT_LEVEL_KEY_FIELD_DESCRIPTION = "A component-level tracking key for Alert";
+    public static final String ALERT_PROVIDER_KEY_FIELD_NAME = "Alert Provider Key";
+    public static final String ALERT_PROVIDER_KEY_FIELD_REFERENCE_NAME = "Custom.AlertProviderKey";
```
bamandel, line 53: Do these references name need to be as unique as possible? If so, we might want to add synopsys or something to the name as well so that there are no conflicts.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
```diff
@@ -69,23 +102,31 @@ public AzureCustomFieldManager(String organizationName, AzureProjectService proj
     }
 
     public void installCustomFields(String projectName, String workItemTypeName) throws AlertException {
-        Future<ProjectWorkItemFieldModel> topLevelKeyFieldCreationResultHolder =
-            executorService.submit(() -> findOrCreateAlertCustomProjectField(projectName, ALERT_TOP_LEVEL_KEY_FIELD_NAME, ALERT_TOP_LEVEL_KEY_FIELD_REFERENCE_NAME, ALERT_TOP_LEVEL_KEY_FIELD_DESCRIPTION));
-        Future<ProjectWorkItemFieldModel> componentLevelKeyFieldCreationResultHolder =
-            executorService.submit(() -> findOrCreateAlertCustomProjectField(projectName, ALERT_COMPONENT_LEVEL_KEY_FIELD_NAME, ALERT_COMPONENT_LEVEL_KEY_FIELD_REFERENCE_NAME, ALERT_COMPONENT_LEVEL_KEY_FIELD_DESCRIPTION));
+        installCustomFields(projectName, workItemTypeName, AZURE_CUSTOM_FIELDS);
```
bamandel, line 105: Should we just remove the private method and put the contents in the public one?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
gkillough: Well yes, but actually no. This is a risk we run, but the reference name is just the context plus a dot prefixed to the name of the field without spaces.

gkillough: @jamesrichard91 What are your thoughts? These are user-visible fields, so it will potentially make issues more verbose, but that could be a good thing so that customers aren't confused.

PR #1138, opened 2020-08-27
Add project version into the subject line of the emails

ChomickiM:


src/main/java/com/synopsys/integration/alert/channel/email/EmailChannel.java
```diff
@@ -98,8 +100,16 @@ public void distributeMessage(DistributionEvent event) throws IntegrationExcepti
...
     public void sendMessage(EmailProperties emailProperties, Set<String> emailAddresses, String subjectLine, String formatType, EmailAttachmentFormat attachmentFormat, MessageContentGroup messageContent) throws IntegrationException {
         String topicValue = null;
+        Optional<String> subTopicValue = Optional.empty();
         if (!messageContent.isEmpty()) {
             topicValue = messageContent.getCommonTopic().getValue();
+            subTopicValue = messageContent.getSubContent()
+                                .stream()
+                                .map(ProviderMessageContent::getSubTopic)
+                                .flatMap(Optional::stream)
+                                .filter(linkableItem -> MessageBuilderConstants.LABEL_PROJECT_VERSION_NAME.equals(linkableItem.getName()))
```
gkillough, line 110: I don't think we need to check this. We know that the sub-topic is the project-version for Black Duck and if there is ever another provider, it will be likely that we need to include this for feature parity.

src/main/java/com/synopsys/integration/alert/channel/email/EmailChannel.java
```diff
@@ -98,8 +100,16 @@ public void distributeMessage(DistributionEvent event) throws IntegrationExcepti
...
+                                .map(LinkableItem::getValue)
+                                .findFirst();
```
gkillough, line 112: Because there can be multiple project-versions, we should return a joined list of them.  

```suggestion
                                .collect(Collectors.joining(", "));
```

src/main/java/com/synopsys/integration/alert/channel/email/EmailChannel.java
```diff
@@ -146,9 +156,13 @@ public void sendMessage(EmailProperties emailProperties, Set<String> emailAddres
...
 
-    private String createEnhancedSubjectLine(String originalSubjectLine, String providerProjectName) {
+    private String createEnhancedSubjectLine(String originalSubjectLine, String providerProjectName, Optional<String> providerProjectVersionName) {
         if (StringUtils.isNotBlank(providerProjectName)) {
-            return String.format("%s | For: %s", originalSubjectLine, providerProjectName);
+            String subjectLine = String.format("%s | For: %s", originalSubjectLine, providerProjectName);
+            if (providerProjectVersionName.isPresent()) {
+                subjectLine += String.format(" %s", providerProjectVersionName.get());
+            }
+            return subjectLine;
```
gkillough, line 165: We might want to consider truncating this if it exceeds a certain number of characters.
78 is the suggested number of RCF 2822 https://tools.ietf.org/html/rfc2822#section-2.1.1

src/main/java/com/synopsys/integration/alert/channel/email/EmailChannel.java
ChomickiM: In my testing of this feature I saw that for multiple versions we have multiple emails for each. So "projectA version1" and "projectA version2" would come in as 2 separate emails with those titles. Is there a case where they would be combined?

gkillough: I believe they are supposed to be combined if they get processed in the same batch of notifications.

gkillough: Either way, this will work.

src/main/java/com/synopsys/integration/alert/channel/email/EmailChannel.java
psantos1113: A license limit notification won't include a version unless I'm mistaken. The API of ProviderMessageContent doesn't make it clear that the sub topic is the version so given that API I suggested filtering to be sure you have the project version sub topic.  The initial code just had find first.  But with your suggestion to join the sub topics the filtering may not make sense anymore.  If we remove the filter portion of the stream then I would like to see a comment above that states we're assuming the sub-topic is the version.   Just in case we get a bug later on that the subject isn't correct we have a comment to refresh our memory on the assumption in case it comes into play.

src/main/java/com/synopsys/integration/alert/channel/email/EmailChannel.java
psantos1113: String utils should have the abbreviate method if I remember correctly to do that.

PR #1139, opened 2020-08-27
Ps refresh token fix

psantos1113:
Implement custom classes to be able to construct a refresh token request with the same type of body as the new token request.  The refresh token is used as the assertion in the body rather than the authorization code in the request.

Also updated the AzureTokenResponse class to only include the bare minimum that is needed.  The overridden methods weren't needed the way we were using it.

### DIFF
--- a/azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureAuthorizationCodeFlow.java
+++ b/azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureAuthorizationCodeFlow.java
@@ -69,16 +69,10 @@ public class AzureAuthorizationCodeFlow extends AuthorizationCodeFlow {
             return null;
         }
         AzureCredential.Builder credentialBuilder = new AzureCredential.Builder(credential.getMethod());
-        credentialBuilder.setClientAuthentication(credential.getClientAuthentication());
-        credentialBuilder.setClock(credential.getClock());
-        credentialBuilder.setJsonFactory(credential.getJsonFactory());
-        credentialBuilder.setRefreshListeners(credential.getRefreshListeners());
-        credentialBuilder.setRequestInitializer(credential.getRequestInitializer());
-        credentialBuilder.setTokenServerEncodedUrl(credential.getTokenServerEncodedUrl());
-        credentialBuilder.setTransport(credential.getTransport());
+        credentialBuilder.copyFromExisting(credential);
         credentialBuilder.setRedirectUri(redirectUri);
         credentialBuilder.setClientSecret(clientSecret);
-        credentialBuilder.setCachedRefreshToken(credential.getRefreshToken());
+
         return credentialBuilder.build();
     }
 
--- a/azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureAuthorizationCodeFlow.java
+++ b/azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureAuthorizationCodeFlow.java
@@ -77,6 +77,7 @@ public class AzureAuthorizationCodeFlow extends AuthorizationCodeFlow {
         credentialBuilder.setTokenServerEncodedUrl(credential.getTokenServerEncodedUrl());
         credentialBuilder.setTransport(credential.getTransport());
         credentialBuilder.setRedirectUri(redirectUri);
+        credentialBuilder.setClientSecret(clientSecret);
         credentialBuilder.setCachedRefreshToken(credential.getRefreshToken());
         return credentialBuilder.build();
     }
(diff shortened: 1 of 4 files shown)


azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureAuthorizationCodeFlow.java
```diff
@@ -53,15 +53,35 @@ protected AzureAuthorizationCodeFlow(Builder builder) {
...
+        credentialBuilder.setClientAuthentication(credential.getClientAuthentication());
+        credentialBuilder.setClock(credential.getClock());
+        credentialBuilder.setJsonFactory(credential.getJsonFactory());
+        credentialBuilder.setRefreshListeners(credential.getRefreshListeners());
+        credentialBuilder.setRequestInitializer(credential.getRequestInitializer());
+        credentialBuilder.setTokenServerEncodedUrl(credential.getTokenServerEncodedUrl());
+        credentialBuilder.setTransport(credential.getTransport());
+        credentialBuilder.setRedirectUri(redirectUri);
+        credentialBuilder.setClientSecret(clientSecret);
+        credentialBuilder.setCachedRefreshToken(credential.getRefreshToken());
```
psantos1113, line 81: Can't override getRefreshToken as the method is final.  Since we are using the credential that was stored in the database to create a new credential object, use the getRefreshToken method to save the current refresh token value in the new credential object as a member.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureCredential.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public String getRedirectUri() {
+        return redirectUri;
+    }
+
+    @Override
+    protected TokenResponse executeRefreshToken() throws IOException {
+        String refreshToken = getRefreshToken();
+        if (refreshToken == null) {
```
psantos1113, line 63: getRefreshToken will return null by default if the credential isn't set.  If getRefreshToken is null check the cachedRefreshToken to make sure it isn't null.  If the cached token that was passed to the build is null return a null response.  This has the same behavior of the base class when the refreshToken member is null.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureCredential.java
```diff
@@ -0,0 +1,123 @@
...
+            if (cachedRefreshToken == null) {
+                return null;
+            }
+            refreshToken = cachedRefreshToken;
+        }
+
+        RefreshTokenRequest request = new RefreshTokenRequest(getTransport(), getJsonFactory(), new GenericUrl(getTokenServerEncodedUrl()),
```
psantos1113, line 70: Build the token request with all the parameters Azure needs.  Similar to the new token request.  The refresh token is the assertion value.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureTokenResponse.java
```diff
@@ -38,67 +38,6 @@ public AzureTokenResponse() {
         super();
     }
 
-    @Override
```
psantos1113, line 41: The setters and getters here aren't needed for the  deserialization to work properly.  The annotation is the most important part.  Clean up this class to the bare minimum.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/oauth/AzureAuthorizationCodeFlow.java
```diff
@@ -53,15 +53,35 @@ protected AzureAuthorizationCodeFlow(Builder builder) {
...
+            return null;
+        }
+        AzureCredential.Builder credentialBuilder = new AzureCredential.Builder(credential.getMethod());
+        credentialBuilder.setClientAuthentication(credential.getClientAuthentication());
+        credentialBuilder.setClock(credential.getClock());
+        credentialBuilder.setJsonFactory(credential.getJsonFactory());
+        credentialBuilder.setRefreshListeners(credential.getRefreshListeners());
+        credentialBuilder.setRequestInitializer(credential.getRequestInitializer());
+        credentialBuilder.setTokenServerEncodedUrl(credential.getTokenServerEncodedUrl());
+        credentialBuilder.setTransport(credential.getTransport());
```
gkillough, line 78: Any chance we can add a method to the builder called `copyFromExisting(Credential credential)`?

psantos1113: yeah I can add that.

PR #1140, opened 2020-08-27
using String.format

ekerwin:
Showing an alternative to AzureSpecTemplate.


ekerwin:
String.format is a known solution to this problem and should be preferred to homegrown alternatives.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureApiVersionAppender.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+public class AzureApiVersionAppender {
+    public String appendApiVersion(String spec, String apiVersion) {
+        char queryParamSeparator = '?';
+        if (StringUtils.contains(spec, queryParamSeparator)) {
+            queryParamSeparator = '&';
+        }
+        spec += queryParamSeparator;
```
jamesrichard91, line 35: Instead of modifying the provided String shouldn't we create a new one in this method?

ekerwin: Given that string are immutable, that IS what occurs anyway. But I can make it more explict.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureHttpService.java
```diff
@@ -154,22 +161,19 @@ public GenericUrl constructRequestUrl(String spec) {
...
-            requestUrlBuilder.append(AZURE_API_VERSION);
+            spec = azureApiVersionAppender.appendApiVersion(spec, AZURE_API_VERSION);
         }
+        requestUrlBuilder.append(spec);
 
         return new GenericUrl(requestUrlBuilder.toString());
     }
 
+    public String appendApiVersion(String spec, String apiVersion) {
+        //TODO ejk - is this appropriate or should the classes that need this demand an appender?
```
gkillough, line 173: Probably the latter. Each service that needs to append a version should have an `AzureApiVersionAppender`. The version constants should probably end up in `AzureApiVersionAppender` as well.

The original code here was a misunderstanding of the Azure API (thinking that all endpoints we would use would use the same API version).

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
```diff
@@ -128,8 +128,6 @@ public void installCustomFields(String projectName, String workItemTypeName) thr
     private Optional<ProjectWorkItemFieldModel> getAlertCustomProjectField(String projectName, String fieldReferenceName) {
         try {
             return Optional.of(projectService.getField(organizationName, fieldReferenceName));
-        } catch (IOException e) {
```
jamesrichard91, line 131: Is this not thrown? why are we removing it?

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/service/workitem/AzureWorkItemService.java
```diff
@@ -124,10 +91,8 @@ public WorkItemResponseModel updateWorkItem(String organizationName, String proj
     }
 
     public WorkItemDeletedResponseModel deleteWorkItem(String organizationName, Integer workItemId) throws HttpServiceException {
-        String requestSpec = API_SPEC_ORGANIZATION_PROJECT_WORKITEMS_INDIVIDUAL
-                                 .defineReplacement(PATH_ORGANIZATION_REPLACEMENT, organizationName)
-                                 .defineReplacement(PATH_WORK_ITEM_ID_REPLACEMENT, workItemId.toString())
-                                 .populateSpec();
+        //TODO verify this is a bug
```
gkillough, line 94: Looks like this was supposed to be based on: `API_SPEC_ORGANIZATION_WORKITEMS_INDIVIDUAL`

ChomickiM: I used API_SPEC_ORGANIZATION_PROJECT_WORKITEMS_INDIVIDUAL because the spec template looked to be the same for getting the WorkItemDeleted. I realized it matched the ...WORKITEMS_INDIVIDUAL so I reused it in this spot

gkillough: Right, but that would not have worked right?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureCustomFieldManager.java
ekerwin: It isn't thrown anymore.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/http/AzureHttpService.java
ekerwin: I injected the appender and added some utility methods for appending specific versions.

azure-boards-common/src/main/java/com/synopsys/integration/azure/boards/common/service/workitem/AzureWorkItemService.java
gkillough: @ekerwin This is a bug. Project name is optional in the request, but cannot be null.

ekerwin: Thanks @gkillough !

PR #1143, opened 2020-08-28
Ps authenticate save fix

psantos1113:
Fix issues with clicking save and test while authentication is happening.  Also fixed an issue where test fails after clicking save.

Test would fail because none of the token fields that are saved in the controller are preserved in the new configuration that is saved.  Read those existing fields and save them.

Fixed an issue with JPA batch update errors.  Needed to flush repositories.
Added a validator to see if authentication was in process.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
@@ -325,6 +325,7 @@ public class DefaultConfigurationAccessor implements ConfigurationAccessor {
 
         List<FieldValueEntity> oldValues = fieldValueRepository.findByConfigId(descriptorConfigId);
         fieldValueRepository.deleteAll(oldValues);
+        fieldValueRepository.flush();
 
         ConfigurationModelMutable updatedConfig = createEmptyConfigModel(descriptorConfigEntity.getDescriptorId(), descriptorConfigEntity.getId(),
             descriptorConfigEntity.getCreatedAt(), descriptorConfigEntity.getLastUpdated(), descriptorConfigEntity.getContextId());
@@ -341,6 +342,7 @@ public class DefaultConfigurationAccessor implements ConfigurationAccessor {
                 updatedConfig.put(configFieldModel);
             }
             fieldValueRepository.saveAll(fieldValuesToSave);
+            fieldValueRepository.flush();
         }
         descriptorConfigEntity.setLastUpdated(DateUtils.createCurrentDateTimestamp());
         descriptorConfigsRepository.save(descriptorConfigEntity);
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/actions/AzureBoardsGlobalApiAction.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/actions/AzureBoardsGlobalApiAction.java
@@ -1,0 +1,100 @@
+/**
+ * blackduck-alert
+ *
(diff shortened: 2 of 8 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
```diff
@@ -325,6 +325,7 @@ public ConfigurationModel updateConfiguration(Long descriptorConfigId, Collectio
 
         List<FieldValueEntity> oldValues = fieldValueRepository.findByConfigId(descriptorConfigId);
         fieldValueRepository.deleteAll(oldValues);
+        fieldValueRepository.flush();
```
jamesrichard91, line 328: It seems odd that we need to add these flushes throughout our code.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/actions/AzureBoardsGlobalApiAction.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public FieldModel beforeUpdateAction(FieldModel fieldModel) throws AlertException {
+        FieldModel updatedFieldModel = super.beforeUpdateAction(fieldModel);
+        Optional<ConfigurationModel> existingConfig = configurationAccessor.getConfigurationById(Long.valueOf(fieldModel.getId()));
+        return updateTokenFields(updatedFieldModel, existingConfig);
+    }
+
+    private FieldModel updateTokenFields(FieldModel fieldModel, Optional<ConfigurationModel> existingConfig) {
```
bamandel, line 80: I think this method should take in only a ConfigurationModel and return a Map<String, FieldValueModel> object. Then wherever we generate the Optional, we instead call map and pass this method as a method reference. Then just return the new object like we do on line 93. Does that make sense?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultConfigurationAccessor.java
gkillough: Yeah, seems like a code smell, but it shouldn't cause an issue other than a minor performance hit.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/actions/AzureBoardsGlobalApiAction.java
```diff
@@ -0,0 +1,101 @@
...
+            updateMapWithMissingField(AzureBoardsDescriptor.KEY_OAUTH_USER_EMAIL, existingFields, keyToValues);
+        }
+
+        return new FieldModel(fieldModel.getDescriptorName(), fieldModel.getContext(), fieldModel.getCreatedAt(), fieldModel.getLastUpdated(), keyToValues);
+    }
+
+    private void updateMapWithMissingField(String key, Map<String, FieldValueModel> databaseFields, Map<String, FieldValueModel> fieldModelFields) {
+        if (databaseFields.containsKey(key)) {
+            fieldModelFields.put(key, databaseFields.get(key));
+        }
```
gkillough, line 99: Do we always want to prefer what's in the DB? Or should we use the value of the FieldValueModel if it has one?

psantos1113: The fieldModel on a save will not have the token fields.  They aren't rendered at all in the UI, therefore the FieldModel we are trying to save will be missing those values always.

PR #1144, opened 2020-08-28
Fix UploadEndpointController and Manager

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
```diff
@@ -90,83 +88,74 @@ public void unRegisterTarget(String targetKey) throws AlertException {
         uploadTargets.remove(targetKey);
     }
 
-    public ResponseEntity<String> performUpload(String targetKey, Resource fileResource) {
+    public void performUpload(String targetKey, Resource fileResource) {
         if (!containsTarget(targetKey)) {
-            return new ResponseEntity(NO_UPLOAD_FUNCTIONALITY_REGISTERED, HttpStatus.NOT_IMPLEMENTED);
+            throwNotImplementedException();
         }
```
bamandel, line 94: Can we use the entire if statement in the throwNotImplementedException method? We perform this exact check multiple times. We'd just change the name of the throw method to throwNotImplementedExceptionIfTargetDoesntExist or something not as long

alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
```diff
@@ -90,83 +88,74 @@ public void unRegisterTarget(String targetKey) throws AlertException {
...
 
         UploadTarget target = uploadTargets.get(targetKey);
         if (!authorizationManager.hasUploadWritePermission(target.getContext().name(), target.getDescriptorKey().getUniversalKey())) {
-            return responseFactory.createForbiddenResponse();
+            throw ResponseFactory.createForbiddenException();
         }
```
bamandel, line 99: I think these if statements can also be made into a throw method that is passed a function for which check to perform.

src/main/java/com/synopsys/integration/alert/web/api/upload/UploadEndpointController.java
```diff
@@ -24,54 +24,61 @@
...
-        this.responseFactory = responseFactory;
     }
 
     @GetMapping("/{key}/exists")
-    public ResponseEntity<String> checkUploadedFileExists(@PathVariable String key) {
+    public ExistenceModel checkUploadedFileExists(@PathVariable String key) {
         if (StringUtils.isBlank(key)) {
-            return responseFactory.createBadRequestResponse("", TARGET_KEY_MISSING);
+            throwTargetKeyMissingException();
         }
```
bamandel, line 58: This class also has a number of duplicate if statements that can be put into the private method.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/UploadEndpointManager.java
gkillough: It looks like they're checking different permissions, so I'm gonna leave them.

bamandel: Sure, but I think something like this would look cleaner
`private blah(BiFunction<Boolean, String, String> checkPermission, Target target) {
	if (!checkPermission.apply(target.getContext().name(), target.getDescriptorKey().getUniversalKey())) {
            throw ResponseFactory.createForbiddenException();
        }
}`
then just calling
`blah(authorizationManager::hasUploadWritePermission, target);`

bamandel: wow, didn't keep any formatting

gkillough: Ok, fine. I was just being lazy.

PR #1146, opened 2020-08-31
Fix(Jira): Throw improved exception rather than only logging it

gkillough:


src/main/java/com/synopsys/integration/alert/channel/jira/common/util/JiraIssueHandler.java
```diff
@@ -69,8 +69,8 @@ public JiraIssueHandler(Gson gson, JiraTransitionHandler jiraTransitionHandler,
     public abstract String getIssueCreatorFieldKey();
 
     @Override
-    protected Optional<IssueResponseModel> createIssue(IssueConfig issueConfig, IssueTrackerRequest request)
-        throws IntegrationException {
+    // TODO this does not need to be Optional
```
bamandel, line 72: Would the change be big to resolve this todo right now?

gkillough: It's scope-creep for a code-frozen version. Otherwise I definitely would have made this change.

PR #1151, opened 2020-09-01
Removing incorrect logging from BlackDuckAccumulator

ChomickiM:
While investigating the error message returning next accumulator run time as either 0 or a negative time we discovered that the logging does not reflect the correct results. This feature appears to not have been working as intended or correctly since before 5.3.0. Because future is currently referring to the running instance of the accumulator we are getting a negative run time as the time is actually reflecting the time since it last ran the accumulator. In the cases of it reporting 0 or -1 this is due to a rounding error as the message would be generated in about 500-900ms or occasionally just over 1000ms. When debugging, this time would reflect the amount of time since the accumulator first ran and would therefore be several seconds or minutes depending on how long you stop the execution for.

The next run time is still working successfully in the task view but since it does not operate the way its intended in the BlackDuckAccumulator we have removed that logging as well as the test. The only change we made here was setting future to null when unscheduling the task as there was nothing previously setting it to null.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckAccumulator.java
+++ b/src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckAccumulator.java
@@ -28,7 +28,6 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -99,12 +98,6 @@ public class BlackDuckAccumulator extends ProviderTask {
             saveNextSearchStart(nextSearchStartString);
         } catch (AlertDatabaseConstraintException e) {
             logger.error("Error occurred accumulating data! ", e);
-        } finally {
-            Optional<Long> nextRun = getMillisecondsToNextRun();
-            if (nextRun.isPresent()) {
-                Long seconds = TimeUnit.MILLISECONDS.toSeconds(nextRun.get());
-                logger.debug("Accumulator next run: {} seconds", seconds);
-            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/synopsys/integration/alert/provider/blackduck/tasks/BlackDuckAccumulator.java
```diff
@@ -99,12 +98,6 @@ public void accumulate() {
             saveNextSearchStart(nextSearchStartString);
         } catch (AlertDatabaseConstraintException e) {
             logger.error("Error occurred accumulating data! ", e);
-        } finally {
-            Optional<Long> nextRun = getMillisecondsToNextRun();
```
gkillough, line 103: Wouldn't it be better to fix the `getMillisecondsToNextRun()` method?

bamandel: Sounds like this will need to get fixed regardless of us logging the data. The method name now seems to be misleading.

ChomickiM: This was a discussion I had with Paulo so hopefully I can recall this correctly... The problem here is that when its run from the accumulator the ScheduledFuture object, inside of ScheduledTask, we are passing around is referring to the running accumulator job. This is a problem because when checking to see when the the job is done it will have already been done at the previous start time. We still want to keep that method in place because it is used in other places such as the taskManager which is working correctly to calculate the next run time.

psantos1113: The issue here is that the task has a reference to the future that is a ScheduledFuture.  The ScheduledFuture api has a getDelay  method.  However we are asking the current running task for its delay as to when it will run next.  But the issue is that when we are asking the future for the delay it is returning a negative value because that is the delay when the task was started. It is the current running task so that is no positive delay to return.  For example -923 means this future was started 923 milliseconds ago because it is the current running task. This coincides with the logging message for when the task starts its run.  Negative numbers returned by getDelay means that the task is currently running or has run.  The delay has elapsed which is true.  

The TaskManager component works because the taskScheduler manages updating the future reference to wait to execute again after some time has passed.  The UI and anyone probing the TaskManager are asking for when a task next runs but it is outside the scope of the task.  Where we are logging is within the scope of the task which is running so getDelay will always be a negative millisecond value.  That is why the logging has been removed.  It doesn't actually provide any value at this point.

PR #1152, opened 2020-09-02
Run ESLint

gkillough:


src/main/js/field/input/TableSelectInput.js
```diff
@@ -255,109 +254,140 @@ class TableSelectInput extends Component {
...
+                )
                 : cell;
 
             if (cell) {
-                return <div title={cell.toString()}> {cellContent} </div>;
+                return (
+                    <div title={cell.toString()}>
+                        {' '}
+                        {cellContent}
+                        {' '}
```
bamandel, line 283: These empty spaces are strange, do you know why they were inserted?

gkillough: Looks like they were there in the original, but ESLint wants it to be explicit?

src/main/js/Navigation.js
```diff
@@ -44,7 +44,7 @@ class Navigation extends Component {
         const components = this.createNavItemForDescriptors(DescriptorUtilities.DESCRIPTOR_TYPE.COMPONENT, DescriptorUtilities.CONTEXT_TYPE.GLOBAL, '/alert/components/');
 
         const nav = (
-            <Fragment>
+            <>
```
psantos1113, line 47: Interesting that no tag name is needed here.

gkillough: Yes. I looked into this and the "shorthand notation" is preferred. `<>` is equivalent to `<Fragment>`.

src/main/js/field/input/TableSelectInput.js
psantos1113: The dive has space before an after the cellContent so the linter to preserve those whitespace characters added the `{' '}` code to preserve the spacing

PR #1153, opened 2020-09-02
Ps custom endpoint actions

psantos1113:
First pass at creating a result object for actions.  Use Custom Endpoints as a first implementation to change over to return the ActionResult to provide a small enough set of changes to create the initial API for ActionResult objects.

Fixed a few SonarCloud code smells while fixing code smell issues with the implementation of this class.


alert-common/src/main/java/com/synopsys/integration/alert/common/action/ActionResult.java
```diff
@@ -0,0 +1,84 @@
...
+        this(httpStatus, null, content);
+    }
+
+    public ActionResult(@NotNull HttpStatus httpStatus, @Nullable String message, @Nullable T content) {
+        this.httpStatus = httpStatus;
+        this.content = content;
+        this.message = message;
+    }
+
+    public boolean isOk() {
```
gkillough, line 55: Ok implies 200 IMO, I would prefer we call this `isSuccessful` instead.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/ResponseFactory.java
```diff
@@ -167,4 +168,13 @@ public static ResponseStatusException createNotImplementedException(@Nullable St
         return new ResponseEntity<>(header, HttpStatus.FOUND);
     }
 
+    public <T> ResponseEntity<T> createResponseFromAction(ActionResult<T> actionResult) {
```
bamandel, line 171: Would it be better to give ActionResult a method that would convert itself into a ResponseEntity? At least these basic conversions? Or is that making ActionResult do too much?

src/main/java/com/synopsys/integration/alert/web/api/custom/CustomEndpointController.java
```diff
@@ -66,7 +70,18 @@ public CustomEndpointController(CustomEndpointManager customEndpointManager, Res
...
+                String content = gson.toJson(result.getContent().get());
+                return responseFactory.createOkContentResponse(content);
+            } else {
+                String message = result.getMessage().orElse("");
+                return responseFactory.createOkResponse("", message);
+            }
+        } else {
+            String message = result.getMessage().orElse("");
+            return responseFactory.createResponse(result.getHttpStatus(), null, message);
+        }
```
bamandel, line 85: If we add the conversion to ActionResult, this could all be handled in a methodin ActionResult.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/ValidationActionResult.java
```diff
@@ -0,0 +1,42 @@
...
+package com.synopsys.integration.alert.common.action;
+
+import java.util.List;
+
+import org.springframework.http.HttpStatus;
+
+import com.synopsys.integration.alert.common.descriptor.config.field.errors.AlertFieldStatus;
+
+public class ValidationActionResult extends ActionResult<Void> {
+    private List<AlertFieldStatus> fieldStatusList;
```
gkillough, line 32: Why isn't this the content of `ActionResult`? It would probably be better to update `ValidationResponseModel` and make this `ActionResult<ValidationResponseModel>` instead.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/ResponseFactory.java
psantos1113: I think that would be making ActionResult do too much.  We may want to convert it differently.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/oauth/OAuthCustomEndpoint.java
```diff
@@ -22,31 +22,25 @@
...
         super(fieldKey, customEndpointManager);
-        this.responseFactory = responseFactory;
-        this.gson = gson;
     }
 
     protected abstract OAuthEndpointResponse createOAuthResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper);
 
     @Override
-    public ResponseEntity<String> createResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) {
+    public ActionResult<OAuthEndpointResponse> createResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) {
```
gkillough, line 42: I think `OAuthEndpointResponse` also tracks a status code. We should either get rid of that, or pass it through to be converted to an `HttpStatus`.

src/main/java/com/synopsys/integration/alert/web/api/custom/CustomEndpointController.java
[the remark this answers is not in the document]
psantos1113: This is just the first step and low hanging fruit to start implementing and changing the code.  Ultimately what I would like is similar to what you suggest.  I'm simply starting the process.  I actually need to change the branch to the feature branch because I have a feature branch where these changes will go.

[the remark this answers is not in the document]
gkillough: If the code is completely working, I would say skip the feature branch. I think we could all benefit from having the `ActionResult` object ASAP so that we don't need to add more tech-debt around this. 

Then I think a fast-follow MR fixing custom endpoints would be beneficial.

PR #1155, opened 2020-09-03
Fix Azure queries and remove Work Item Creator Email

gkillough:
- Azure does not handle blank values the same way that Jira does.
- Work Item Creator is handled dynamically in Azure and cannot be changed.


src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsIssueHandler.java
```diff
@@ -208,13 +207,13 @@ protected void logIssueAction(String issueTrackerProjectName, IssueTrackerReques
     }
 
     private WorkItemQueryWhere appendToQueryBuilder(WorkItemQueryWhere queryBuilder, String fieldReferenceName, Optional<String> fieldKey) {
-        if (fieldKey.isPresent()) {
+        if (fieldKey.filter(StringUtils::isNotBlank).isPresent()) {
```
gkillough, line 210: Query fix

jamesrichard91 (approved):

ChomickiM (approved):

PR #1158, opened 2020-09-03
Ps function controllers

psantos1113:
- Rename CustomEndpoint to CustomFunctionAction.
- Create controllers for each custom function api. 
- Create an abstract Controller for the custom functions.
- Remove the different types of CustomEndpoint classes just use CustomEndpoint
- Refactor the existing classes to the new structure.

Use this to give me an initial set of ideas that I've written down to be applied to the other actions and controllers in a different pull request in the future.


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/ResponseFactory.java
```diff
@@ -75,6 +75,18 @@ public static ResponseStatusException createNotImplementedException(@Nullable St
         return new ResponseStatusException(HttpStatus.NOT_IMPLEMENTED, customMessage);
     }
 
+    public static <T> ResponseStatusException createStatusException(ActionResponse<T> actionResponse) {
+        String customMessage = actionResponse.getMessage().orElse("");
```
gkillough, line 79: In `ResponseStatusException`, the message parameter in `@Nullable`. We should use `.orElse(null)` here so that the default message for the `HttpStatus` will be used by the exception.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderSelectCustomFunctionAction.java
```diff
@@ -27,35 +27,39 @@
...
-                   .sorted()
-                   .collect(Collectors.toList());
+    public ActionResponse<List<LabelValueSelectOption>> createActionResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) throws IntegrationException {
+        List<LabelValueSelectOption> content = descriptorMap.getDescriptorByType(DescriptorType.PROVIDER).stream()
+                                                   .map(descriptor -> descriptor.createMetaData(ConfigContextEnum.DISTRIBUTION))
+                                                   .flatMap(Optional::stream)
+                                                   .map(descriptorMetadata -> new LabelValueSelectOption(descriptorMetadata.getLabel(), descriptorMetadata.getName()))
+                                                   .sorted()
+                                                   .collect(Collectors.toList());
+        return new ActionResponse<>(HttpStatus.OK, content);
```
gkillough, line 62: For the sake of consistency, it might be nice to wrap the `List<LabelValueSelectOption>` in an object to make content consistent.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionCustomFunctionAction.java
```diff
@@ -30,34 +30,38 @@
...
+    public ActionResponse<List<ProviderProject>> createActionResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) throws IntegrationException {
         String providerName = fieldModel.getFieldValue(ChannelDistributionUIConfig.KEY_PROVIDER_NAME).orElse("");
         if (StringUtils.isBlank(providerName)) {
             throw new ResponseStatusException(HttpStatus.BAD_REQUEST, MISSING_PROVIDER_ERROR);
         }
 
         String providerConfigName = fieldModel.getFieldValue(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME).orElse("");
-        return providerDataAccessor.getProjectsByProviderConfigName(providerConfigName);
+        List<ProviderProject> content = providerDataAccessor.getProjectsByProviderConfigName(providerConfigName);
+        return new ActionResponse<>(HttpStatus.OK, content);
```
gkillough, line 64: It might be nice to wrap the `List<ProviderProject>`.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderConfigSelectCustomFunctionAction.java
```diff
@@ -27,47 +27,51 @@
...
+            content = configurationModels.stream()
+                          .map(ConfigurationModel::getCopyOfKeyToFieldMap)
+                          .map(FieldAccessor::new)
+                          .map(accessor -> accessor.getString(ProviderDescriptor.KEY_PROVIDER_CONFIG_NAME))
+                          .flatMap(Optional::stream)
+                          .map(LabelValueSelectOption::new)
+                          .collect(Collectors.toList());
         }
-        return List.of();
+        return new ActionResponse<>(HttpStatus.OK, content);
```
gkillough, line 75: I'm working backwards here, but we might want to rethink content being a `List`. It adds one more step to the controller for enforcing the pattern of an API always returning a JSON Object.

src/main/java/com/synopsys/integration/alert/web/api/functions/AbstractFunctionController.java
```diff
@@ -0,0 +1,57 @@
...
+
+public abstract class AbstractFunctionController<T> extends BaseController {
+    public static final String API_FUNCTION_URL = CustomFunctionAction.API_FUNCTION_URL;
+    private final CustomFunctionAction<T> functionAction;
+
+    public AbstractFunctionController(CustomFunctionAction<T> functionAction) {
+        this.functionAction = functionAction;
+    }
+
+    @PostMapping
```
gkillough, line 47: Does this inherit the `@RequestMapping` from the sub-class?

psantos1113: yes it does.  The PostMapping just makes the controller accept post to the path in the RequestMapping defined in the sub-class that implements this abstract class.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderConfigSelectCustomFunctionAction.java
psantos1113: We need to change the corresponding javascript code for those fields to expect an object and rather than an array. I will look into it.

gkillough: Maybe a follow-up PR then, but we could definitely prepare the actions for that change.

psantos1113: I have created a new branch [ps_function_select_options](https://github.com/blackducksoftware/blackduck-alert/tree/ps_function_select_options) for the comments because this PR functions with the UI as is.  I will need to make UI changes to update UI fields to render options properly.

gkillough (approved):

PR #1161, opened 2020-09-08
Fixing NPE issue when trying to create work items while the Azure global config is missing.

jamesrichard91:
Validate the Azure properties before attempting to create an issue in order to avoid a NPE when the global configuration is missing.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsProperties.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsProperties.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.NTCredentials;
@@ -51,6 +52,7 @@ import com.google.gson.Gson;
 import com.synopsys.integration.alert.channel.azure.boards.descriptor.AzureBoardsDescriptor;
 import com.synopsys.integration.alert.channel.azure.boards.oauth.storage.AzureBoardsCredentialDataStoreFactory;
 import com.synopsys.integration.alert.common.channel.issuetracker.config.IssueTrackerServiceConfig;
+import com.synopsys.integration.alert.common.exception.AlertConfigurationException;
 import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.persistence.accessor.FieldAccessor;
 import com.synopsys.integration.azure.boards.common.http.AzureHttpService;
@@ -106,6 +108,15 @@ public class AzureBoardsProperties implements IssueTrackerServiceConfig {
 
     public List<String> getScopes() {
         return scopes;
+    }
+
+    public void validateProperties() throws AlertConfigurationException {
+        if (StringUtils.isBlank(organizationName) || StringUtils.isBlank(clientId) || StringUtils.isBlank(clientSecret)) {
+            throw new AlertConfigurationException("The global configuration for Azure is missing required information.");
+        }
+        if (StringUtils.isBlank(oauthUserId)) {
+            throw new AlertConfigurationException("The Azure connection was not authenticated properly. Please go to the Azure global configuration to authenticate the connection.");
+        }
     }
 
     public AzureHttpService createAzureHttpService(ProxyInfo proxy, Gson gson, String authorizationCode) throws AlertException {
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsRequestDelegator.java
```diff
@@ -63,6 +63,7 @@ public AzureBoardsRequestDelegator(Gson gson, ProxyManager proxyManager, AzureBo
     public IssueTrackerResponse sendRequests(List<IssueTrackerRequest> requests) throws IntegrationException {
         IssueConfig azureIssueConfig = context.getIssueConfig();
         AzureBoardsProperties azureBoardsProperties = context.getIssueTrackerConfig();
+        azureBoardsProperties.validateProperties();
```
jamesrichard91, line 66: Im not sure if this covers the TODO on line 72 for validating the configuration. Was there more to be validated once the connection is established?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsProperties.java
```diff
@@ -108,6 +110,15 @@ public String getOauthUserId() {
         return scopes;
     }
 
+    public void validateProperties() throws AlertConfigurationException {
+        if (StringUtils.isBlank(organizationName) || StringUtils.isBlank(clientId) || StringUtils.isBlank(clientSecret)) {
+            throw new AlertConfigurationException("The global configuration for Azure is missing required information.");
+        }
+        if (StringUtils.isBlank(oauthUserId)) {
```
jamesrichard91, line 117: The `oauthUserId` doesn't get populated until the authentication with Azure occurs, right?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsRequestDelegator.java
gkillough: I think your code handles the TODO. The only other problem I forsee is having the wrong scopes.

gkillough (approved):

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsRequestDelegator.java
gkillough: I guess we might probably want to validate the connection by performing a HEAD request to some endpoint like _projects_.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsProperties.java
psantos1113: The `oauthUserId` get's populated when `fromFieldAccessor` is called.  We don't know the user when performing the handshake so we are always using the value of `DEFAULT_AZURE_OAUTH_USER_ID` constant.  Are you seeing it have a value other than the default?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsRequestDelegator.java
psantos1113: Your code handles the TODO.  Yes if the application is configured with different scopes than we expect then that will cause a failure.

psantos1113 (commented):
Please see the comment on the oauthUserId field.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsProperties.java
jamesrichard91: Ok. I didn't realize that we were setting a default value for that field. I thought it would only get populated if the User has pressed the "Authenticate" button in  the Azure configuration. I assumed it would be blank if the User filled in the fields and only pressed save not authenticate. 
If this field is always going to have a value then should I remove this part of the validation?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsRequestDelegator.java
jamesrichard91: Ya I am only validating the user provided configuration, I am not testing the connection to Azure. Is that something we want to add? If it is I will adjust the TODO statement accordingly

ChomickiM (approved):

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsRequestDelegator.java
psantos1113: I think if we start running into issues with scopes not matching up then we will need to make it configurable in the future.  I think for now we can remove the TODO since you handle the user provided fields.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/service/AzureBoardsProperties.java
psantos1113: yes I would remove it.  The documentation has an example where the email address is passed in via the state query parameter.  We don't capture the user email in the global configuration from the UI to be able to include it in the state query parameter.  Since we have only one configuration for Azure boards our default key is sufficient for now.  If we started having multiple Azure boards tokens then we would need unique keys.

psantos1113 (approved):

PR #1162, opened 2020-09-10
Create a better well defined API for actions and controllers

psantos1113:


alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,182 @@
...
+import com.synopsys.integration.alert.common.action.ValidationActionResponse;
+import com.synopsys.integration.alert.common.enumeration.ConfigContextEnum;
+import com.synopsys.integration.alert.common.exception.AlertException;
+import com.synopsys.integration.alert.common.persistence.accessor.DescriptorAccessor;
+import com.synopsys.integration.alert.common.persistence.model.RegisteredDescriptorModel;
+import com.synopsys.integration.alert.common.rest.model.FieldModel;
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+import com.synopsys.integration.alert.common.security.authorization.AuthorizationManager;
+
+public abstract class AbstractConfigResourceActions implements ResourceActions<FieldModel>, TestAction<FieldModel>, ValidateAction<FieldModel> {
```
psantos1113, line 42: This class checks permissions and validates the content first.  The abstract methods are to be implemented by a concrete class to perform the functionality.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractResourceActions.java
```diff
@@ -0,0 +1,143 @@
...
+import org.springframework.http.HttpStatus;
+
+import com.synopsys.integration.alert.common.action.ActionResponse;
+import com.synopsys.integration.alert.common.action.ValidationActionResponse;
+import com.synopsys.integration.alert.common.descriptor.DescriptorKey;
+import com.synopsys.integration.alert.common.enumeration.ConfigContextEnum;
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+import com.synopsys.integration.alert.common.security.authorization.AuthorizationManager;
+
+public abstract class AbstractResourceActions<T> implements ResourceActions<T>, ValidateAction<T>, TestAction<T> {
```
psantos1113, line 36: This class can be used by classes that deal with resources other than FieldModel or JobFieldModel i.e. CertificateActions, UserActions,RoleActions, and TaskActions.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/ValidationActionResponse.java
```diff
@@ -22,12 +22,32 @@
  */
 package com.synopsys.integration.alert.common.action;
```
gkillough, line 23: I'm thinking that when we further separate Alert into sub-projects, the models for the REST API stuff should be moved into that sub-project as well. That way we can avoid misusing the models and bogging down alert-common.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,182 @@
...
+
+    protected abstract ActionResponse<List<FieldModel>> readAllByContextAndDescriptor(String context, String descriptorName);
+
+    protected abstract Optional<FieldModel> findFieldModel(Long id);
+
+    protected abstract ValidationActionResponse testResource(FieldModel resource);
+
+    protected abstract ActionResponse<FieldModel> updateResource(Long id, FieldModel resource);
+
+    protected abstract ValidationActionResponse validateResource(FieldModel resource);
```
gkillough, line 69: Looking at these names and how they are used below, we might want to distinguish them a little more from each other. Maybe instead of the public `create` and the protected `createResource`, the public method could be called `createWithPermissionCheck` and the protected method could be called `createWithoutPermissionCheck` or something along those lines.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Override
+    public ActionResponse<List<FieldModel>> getAll() {
+        try {
+            Set<String> descriptorNames = descriptorAccessor.getRegisteredDescriptors()
+                                              .stream()
+                                              .map(RegisteredDescriptorModel::getName)
+                                              .collect(Collectors.toSet());
+            if (!authorizationManager.anyReadPermission(List.of(ConfigContextEnum.DISTRIBUTION.name(), ConfigContextEnum.GLOBAL.name()), descriptorNames)) {
```
gkillough, line 97: It might be worthwhile to create an enhancement request for `AuthorizationManager` to have methods that also take collections of enums.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,182 @@
...
+        } catch (AlertException ex) {
+            return new ActionResponse<>(HttpStatus.INTERNAL_SERVER_ERROR, String.format("Error reading configurations: %s", ex.getMessage()));
+        }
+    }
+
+    @Override
+    public ActionResponse<FieldModel> getOne(Long id) {
+        if (null == id) {
+            return new ActionResponse<>(HttpStatus.BAD_REQUEST, AbstractResourceActions.RESOURCE_IDENTIFIER_MISSING);
+        }
```
gkillough, line 110: Based on the way this class is used (i.e. in a controller) is this case actually possible? If the `RequestMapping` requires an id, and an id is missing from the request, then it will either be a call to the "getAll" endpoint (if one exists) or Spring will return a 404.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,211 @@
...
+
+    protected abstract ActionResponse<JobFieldModel> deleteResource(UUID id);
+
+    protected abstract ActionResponse<List<JobFieldModel>> readAllResources();
+
+    protected abstract ValidationActionResponse testResource(JobFieldModel resource);
+
+    protected abstract ActionResponse<JobFieldModel> updateResource(UUID id, JobFieldModel resource);
+
+    protected abstract ValidationActionResponse validateResource(JobFieldModel resource);
```
gkillough, line 72: Same thought as the previous class. We should make it more explicit what the difference between something like `ActionResponse<JobFieldModel> create(JobFieldModel resource)` and `ActionResponse<JobFieldModel> createResource(JobFieldModel resource)` is.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,211 @@
...
+            return new ActionResponse<>(HttpStatus.INTERNAL_SERVER_ERROR, String.format("Error reading job configurations: %s", ex.getMessage()));
+        }
+
+    }
+
+    @Override
+    public ActionResponse<JobFieldModel> getOne(UUID id) {
+        if (null == id) {
+            return new ActionResponse<>(HttpStatus.BAD_REQUEST, AbstractResourceActions.RESOURCE_IDENTIFIER_MISSING);
+        }
```
gkillough, line 118: Same thought as the previous class.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,211 @@
...
+        if (optionalModel.isPresent()) {
+            JobFieldModel fieldModel = optionalModel.get();
+            boolean hasPermissions = hasRequiredPermissions(fieldModel.getFieldModels(), authorizationManager::hasReadPermission);
+            if (!hasPermissions) {
+                return new ActionResponse<>(HttpStatus.FORBIDDEN, AbstractResourceActions.FORBIDDEN_MESSAGE);
+            }
+            return new ActionResponse<>(HttpStatus.OK, fieldModel);
+        }
+
+        return new ActionResponse<>(HttpStatus.NOT_FOUND, null);
```
gkillough, line 131: When the constructor for `ActionResponse` would take a `null` parameter, we should either add a static initializer or a new constructor without the `message` parameter.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/JobResourceActions.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package com.synopsys.integration.alert.common.action.api;
+
+import java.util.List;
+import java.util.UUID;
+
+import com.synopsys.integration.alert.common.action.ActionResponse;
+import com.synopsys.integration.alert.common.rest.model.JobFieldModel;
+
+public interface JobResourceActions {
```
gkillough, line 31: Should this extend `ResourceActions`?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
psantos1113: I would prefer to change the names of the abstract methods because the public methods are displayed in the Swagger API.  To an end user they just want to create, read, update, delete, validate, or test the resources. Whether we perform permission checks or not should not be exposed to them.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/FieldValueModel.java
```diff
@@ -60,7 +60,9 @@ public void setValues(Collection<String> values) {
     }
 
     public void setValue(String value) {
-        setValues(Set.of(value));
+        if (null != value) {
+            setValues(Set.of(value));
+        }
```
gkillough, line 65: Should this call `setValues` with `Set.of()` if the parameter is `null`?

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomFunctionAction.java
```diff
@@ -155,11 +155,13 @@ private OAuthEndpointResponse createErrorResponse(String errorMessage) {
         if (StringUtils.isNotBlank(fieldModel.getId())) {
             if (authorizationManager.hasWritePermission(ConfigContextEnum.GLOBAL.name(), azureBoardsChannelKey.getUniversalKey())) {
                 Long id = Long.parseLong(fieldModel.getId());
-                return Optional.ofNullable(configActions.updateConfig(id, fieldModel));
+                ActionResponse<FieldModel> response = configActions.update(id, fieldModel);
```
gkillough, line 158: Won't `configActions.update` check write permissions? Should the if-statement check if the id is present to determine whether to create or update?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
psantos1113: Yeah I was going to make that change in this set of code changes but I will add it separately.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
psantos1113: Just tested this out Spring throws an exception if it's the wrong type so I will remove that.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
psantos1113: We have a constructor that takes the status only.  I missed some places where that should be used.  I will fix that.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigController.java
```diff
@@ -22,316 +22,83 @@
...
-        }
-
-        try {
-            return runPutConfig(id, restModel);
-        } catch (AlertException e) {
-            logger.error(e.getMessage(), e);
-            return responseFactory.createInternalServerErrorResponse(restModel.getJobId(), e.getMessage());
-        }
+    @GetMapping("/validate")
+    public List<JobFieldStatuses> validateJobs() {
```
gkillough, line 59: Will this clash with the overridden validate method?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/JobResourceActions.java
psantos1113: If I add another Generic to resource actions to define the type for the ids then yes it would be possible.  The id type for Jobs is UUID and other resources is a Long.  Hindsight being 2020 perhaps the ids should have always been a UUID but because we have two different ID types I didn't want to make it extend.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/FieldValueModel.java
psantos1113: yes good catch.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigController.java
psantos1113: no because the overridden validate method is a POST mapping and takes a parameter.  The method signature is different and has a different request mapping associated with it.  They do not conflict.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
gkillough: Aren't the actions invoked by the controllers? The name of the method in the controller should be the name that is displayed, no?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/JobResourceActions.java
gkillough: Ok, I think this could be a future enhancement if we ever migrate to UUIDs across the board.

src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomFunctionAction.java
psantos1113: yes

src/main/java/com/synopsys/integration/alert/web/api/certificate/CertificateActions.java
```diff
@@ -28,77 +28,116 @@
...
+    public ActionResponse<CertificateModel> createResource(CertificateModel resource) {
+        String loggableAlias = escapeUtil.replaceWithUnderscore(resource.getAlias());
         logger.info("Importing certificate with alias {}", loggableAlias);
-        return importCertificate(certificateModel);
+        try {
+            CertificateModel certificateModel = importCertificate(resource);
+            return new ActionResponse<>(HttpStatus.OK, certificateModel);
+        } catch (AlertException ex) {
+            String message = ex.getMessage();
+            logger.error("There was an issue importing the certificate.{}", message);
```
gkillough, line 118: Missing a space before "{}"

src/main/java/com/synopsys/integration/alert/web/api/certificate/CertificateActions.java
```diff
@@ -113,19 +152,22 @@ private CertificateModel importCertificate(CertificateModel certificateModel) th
...
+                certificateUtility.removeCertificate(certificateModel);
+                certificateAccessor.deleteCertificate(id);
+            }
+        } catch (AlertException ex) {
+            logger.error("Error deleting certificate", ex);
+            return new ActionResponse<>(HttpStatus.INTERNAL_SERVER_ERROR, String.format("Error deleting certificate: %s", ex.getMessage()));
         }
-        return false;
+
+        return new ActionResponse<>(HttpStatus.NO_CONTENT, null);
```
gkillough, line 170: `null` passed to the constructor

src/main/java/com/synopsys/integration/alert/web/api/certificate/CertificateActions.java
```diff
@@ -113,19 +152,22 @@ private CertificateModel importCertificate(CertificateModel certificateModel) th
...
-            CustomCertificateModel certificateModel = certificate.get();
-            logger.info("Delete certificate with id: {} and alias: {}", certificateModel.getNullableId(), certificateModel.getAlias());
-            certificateUtility.removeCertificate(certificateModel);
-            certificateAccessor.deleteCertificate(id);
-            return true;
+    @Override
+    public ActionResponse<CertificateModel> deleteResource(Long id) {
+        try {
+            Optional<CustomCertificateModel> certificate = certificateAccessor.getCertificate(id);
+            if (certificate.isPresent()) {
```
gkillough, line 159: If the certificate is not present, we should return a 404

alert-common/src/main/java/com/synopsys/integration/alert/common/action/ValidationActionResponse.java
psantos1113: I think after this PR and the feature branch are merged into master we will no longer need to use AlertFieldException and IntegrationRestException in Alert.  So the static method in this ValidationActionResponse will be able to be removed also.

src/main/java/com/synopsys/integration/alert/web/api/config/ConfigController.java
```diff
@@ -22,241 +22,69 @@
...
-            logger.error("Was not able to find configurations with the context {}, and descriptorName {} to get.", context, descriptorName);
-            return responseFactory.createNotFoundResponse(String.format(EXCEPTION_FORMAT_CONFIGURATIONS_NOT_FOUND_FOR_CONTEXT_AND_DESCRIPTOR, context, descriptorName));
-        }
-
-        if (models.isEmpty()) {
-            return responseFactory.createNotFoundResponse(String.format(EXCEPTION_FORMAT_CONFIGURATIONS_NOT_FOUND_FOR_CONTEXT_AND_DESCRIPTOR, context, descriptorName));
-        }
-
-        return responseFactory.createOkContentResponse(contentConverter.getJsonString(models));
+    public List<FieldModel> getAll(@RequestParam ConfigContextEnum context, @RequestParam(required = false) String descriptorName) {
```
gkillough, line 56: Eventually, we should wrap this list in an object before returning.

src/main/java/com/synopsys/integration/alert/web/api/config/ConfigActions.java
```diff
@@ -31,85 +30,128 @@
...
         this.modelConverter = modelConverter;
+        this.descriptorMap = descriptorMap;
+        this.pkixErrorResponseFactory = pkixErrorResponseFactory;
     }
 
-    public boolean doesConfigExist(String id) throws AlertException {
-        return StringUtils.isNotBlank(id) && doesConfigExist(Long.parseLong(id));
+    @Override
+    protected ActionResponse<List<FieldModel>> readAllResources() {
+        return new ActionResponse<>(HttpStatus.NOT_IMPLEMENTED, null);
```
gkillough, line 90: `null` used in constructor

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -258,53 +324,80 @@ public MessageResult validateJob(JobFieldModel jobFieldModel) {
...
 
-    public Optional<String> checkGlobalConfigExists(String descriptorName) {
-        return globalConfigExistsValidator.validate(descriptorName);
+    public ActionResponse<String> checkGlobalConfigExists(String descriptorName) {
+        Optional<String> configMissingMessage = globalConfigExistsValidator.validate(descriptorName);
+        if (configMissingMessage.isPresent()) {
+            String message = configMissingMessage.get();
+            return new ActionResponse<>(HttpStatus.BAD_REQUEST, message);
+        }
+        return new ActionResponse<>(HttpStatus.NO_CONTENT, null);
```
gkillough, line 400: `null` passed in constructor

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
psantos1113: You're right been looking at this code for too long.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,184 @@
...
+import com.synopsys.integration.alert.common.rest.model.FieldModel;
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+import com.synopsys.integration.alert.common.security.authorization.AuthorizationManager;
+
+public abstract class AbstractConfigResourceActions implements LongResourceActions<FieldModel>, TestAction<FieldModel>, ValidateAction<FieldModel> {
+    private AuthorizationManager authorizationManager;
+    private DescriptorAccessor descriptorAccessor;
+
+    public AbstractConfigResourceActions(AuthorizationManager authorizationManager, DescriptorAccessor descriptorAccessor) {
+        this.authorizationManager = authorizationManager;
```
bamandel, line 47: Do we prefer calling the other constructor ('this' keyword) in cases like this?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    @Override
+    public ActionResponse<FieldModel> getOne(Long id) {
+        Optional<FieldModel> fieldModel = findFieldModel(id);
+        if (fieldModel.isPresent()) {
+            FieldModel model = fieldModel.get();
+            if (!authorizationManager.hasReadPermission(model.getContext(), model.getDescriptorName())) {
+                return new ActionResponse<>(HttpStatus.FORBIDDEN, AbstractResourceActions.FORBIDDEN_MESSAGE);
```
bamandel, line 112: This statement is written a lot. Should we have a private method or static method in ActionResponse to ensure it's always the same message?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,216 @@
...
+            descriptorNames = descriptorAccessor.getRegisteredDescriptors()
+                                  .stream()
+                                  .filter(descriptor -> ALLOWED_JOB_DESCRIPTOR_TYPES.contains(descriptor.getType()))
+                                  .map(RegisteredDescriptorModel::getName)
+                                  .collect(Collectors.toSet());
+        } catch (AlertDatabaseConstraintException ex) {
+            // ignore or add a logger.
+        }
+
+        return descriptorNames;
```
bamandel, line 85: Instead of the Set object, could we return the statement in the try block otherwise return Set.of?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    @Override
+    public ActionResponse<JobFieldModel> create(JobFieldModel resource) {
+        boolean hasPermissions = hasRequiredPermissions(resource.getFieldModels(), authorizationManager::hasCreatePermission);
+        if (!hasPermissions) {
+            return new ActionResponse<>(HttpStatus.FORBIDDEN, AbstractResourceActions.FORBIDDEN_MESSAGE);
```
bamandel, line 92: More of these calls to FORBIDDEN and FORBIDDEN_MESSAGE. Makes me this we should put this call in a static variable somewhere.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractResourceActions.java
```diff
@@ -0,0 +1,158 @@
...
+    protected abstract ActionResponse<T> updateAfterChecks(Long id, T resource);
+
+    protected abstract ValidationActionResponse validateAfterChecks(T resource);
+
+    protected abstract Optional<T> findExisting(Long id);
+
+    @Override
+    public ActionResponse<T> create(T resource) {
+        if (!authorizationManager.hasCreatePermission(context.name(), descriptorKey.getUniversalKey())) {
+            return new ActionResponse<>(HttpStatus.FORBIDDEN, AbstractResourceActions.FORBIDDEN_MESSAGE);
```
bamandel, line 71: This class also has references to the forbidden action response

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractResourceActions.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+        return readAllAfterChecks();
+    }
+
+    @Override
+    public ActionResponse<T> getOne(Long id) {
+        if (null == id) {
+            return new ActionResponse<>(HttpStatus.BAD_REQUEST, RESOURCE_IDENTIFIER_MISSING);
+        }
```
bamandel, line 92: Do we still want this null check or will this be removed?

src/main/java/com/synopsys/integration/alert/web/api/config/ConfigActions.java
```diff
@@ -118,54 +160,103 @@ public void deleteConfig(Long id) throws AlertException {
...
+        ValidationResponseModel responseModel;
+        HttpStatus status = HttpStatus.OK;
+        if (fieldStatuses.isEmpty()) {
+            responseModel = ValidationResponseModel.withoutFieldStatuses("The configuration is valid");
+        } else {
+            status = HttpStatus.BAD_REQUEST;
+            responseModel = ValidationResponseModel.fromStatusCollection("There were problems with the configuration", fieldStatuses);
         }
-        return "Valid";
+        return new ValidationActionResponse(status, responseModel);
```
bamandel, line 222: This may look better if we fo an if check with a return then return Bad request by default if the if statement is skipped.

src/main/java/com/synopsys/integration/alert/web/api/config/ConfigActions.java
```diff
@@ -118,54 +160,103 @@ public void deleteConfig(Long id) throws AlertException {
...
     }
 
-    public String testConfig(FieldModel restModel) throws IntegrationException {
-        validateConfig(restModel, new ArrayList<>());
-        Optional<TestAction> testActionOptional = descriptorProcessor.retrieveTestAction(restModel);
+    @Override
+    protected ValidationActionResponse testAfterChecks(FieldModel resource) {
+        Optional<TestAction> testActionOptional = descriptorProcessor.retrieveTestAction(resource);
+        ValidationResponseModel responseModel;
```
bamandel, line 228: Might be cleaner to remove this declaration and just create the object before returning.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -56,150 +63,183 @@
...
     private final ConfigurationFieldModelConverter modelConverter;
     private final GlobalConfigExistsValidator globalConfigExistsValidator;
+    private final PKIXErrorResponseFactory pkixErrorResponseFactory;
+    private final DescriptorMap descriptorMap;
 
     @Autowired
-    public JobConfigActions(ConfigurationAccessor configurationAccessor, FieldModelProcessor fieldModelProcessor, DescriptorProcessor descriptorProcessor, ConfigurationFieldModelConverter modelConverter,
-        GlobalConfigExistsValidator globalConfigExistsValidator) {
+    public JobConfigActions(AuthorizationManager authorizationManager, DescriptorAccessor descriptorAccessor, ConfigurationAccessor configurationAccessor, FieldModelProcessor fieldModelProcessor, DescriptorProcessor descriptorProcessor,
+        ConfigurationFieldModelConverter modelConverter, GlobalConfigExistsValidator globalConfigExistsValidator, PKIXErrorResponseFactory pkixErrorResponseFactory, DescriptorMap descriptorMap) {
```
bamandel, line 98: A lot of constructor params getting passed. Something to look into, maybe add a TODO to revisit this?

bamandel (approved):

bamandel:
I like the changes but I think we could benefit from some composition here. We have a bunch of inheritance in the ResourceAction classes but never actually use it. If we could change that to some sort of Utility class, I think that would be a good change.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
psantos1113: The other constructor is from an earlier incarnation. I will remove it.

src/main/java/com/synopsys/integration/alert/web/api/config/ConfigActions.java
psantos1113: There's a ticket to add audit logging especially for users and role actions.  To ensure that we are logging consistently we most likely won't be able to use the circuit breaker pattern and return right away if we want to log when an action starts and completes.

bamandel: Won't we still be able to log in all cases? Returning early would act like the if else statement above and you can fit all logs as needed, unless I'm missing something about that logging.

gkillough (approved):

PR #1163, opened 2020-09-11
User controller refactor

ChomickiM:
Refactor of the user actions and controller based on the work and example Gavin made in the certificate controller/actions. I had begun working on moving, improving, and adding to the tests, but if Paulo's changes are adopted I decided to hold off finishing them for now. If his changes are made I will try and revisit the tests and improve them at that point.


src/main/java/com/synopsys/integration/alert/web/api/user/UserController.java
```diff
@@ -51,76 +55,80 @@
...
+            throw ResponseFactory.createInternalServerErrorException("There was an issue with the DB");
+        } catch (AlertFieldException fieldException) {
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, String.format("There were errors with the configuration: %s", fieldException.getFlattenedErrorMessages()));
         }
 
     }
 
     @PutMapping(value = "/{userId}")
-    public ResponseEntity<String> updateUser(@PathVariable Long userId, @RequestBody UserConfig userModel) {
+    public UserConfig updateUser(@PathVariable Long userId, @RequestBody UserConfig userModel) {
```
psantos1113, line 104: Should this return NO_CONTENT here?

src/main/java/com/synopsys/integration/alert/web/api/user/UserActions.java
```diff
@@ -187,11 +199,12 @@ private void validateUserExistsByName(List<AlertFieldStatus> fieldErrors, String
...
-        validateRequiredField(FIELD_KEY_USER_MGMT_USERNAME, fieldErrors, userName);
+    private List<AlertFieldStatus> validateUserExistsById(Long userId, String userName) {
         Optional<UserModel> userModel = userAccessor.getUser(userName);
-        userModel.filter(user -> !user.getId().equals(userId))
-            .ifPresent(user -> fieldErrors.add(AlertFieldStatus.error(FIELD_KEY_USER_MGMT_USERNAME, "A user with that username already exists.")));
+        return userModel.filter(user -> !user.getId().equals(userId))
+                   .map(user -> AlertFieldStatus.error(FIELD_KEY_USER_MGMT_USERNAME, "A user with that username already exists."))
+                   .stream()
+                   .collect(Collectors.toList());
     }
```
bamandel, line 208: This looks wrong. If I am validating a user exists, why is there a list of statuses all saying the same thing being returned? Shouldn't it just be a single item returned?

bamandel (commented):
Will these all need to get changed again to use the new branch Paulo made? If so, we may just want to wait for his changes then modify this PR.

src/main/js/util/configurationRequestBuilder.js
```diff
@@ -65,6 +65,7 @@ export function createUpdateRequest(apiUrl, csrfToken, configurationId, fieldMod
         method: 'PUT',
         body: JSON.stringify(fieldModel),
         headers: {
+            accept: 'application/json',
```
gkillough, line 68: Should usage of "accept" be quoted like the other headers?

ChomickiM: I'm not entirely sure, I was basing this on export functions just above this one where the accept: 'application/json' wasn't in quotes either

ChomickiM: RequestUtilities also seems to have the accept outside of quotes, so I think this may be intended

psantos1113: Our linting rules declare the quotes as unnecessary for the accept header.

psantos1113: We should probably refactor this code to use the Headers object and append the headers. I would probably write up an enhancement ticket for that.

src/main/js/store/actions/users.js
```diff
@@ -95,6 +98,33 @@ function clearFieldErrors() {
     };
 }
 
+function validatingUser() {
```
gkillough, line 101: We also need to make sure we update the deleteUser errors.

gkillough: Take a look at #1165 for an example of this.

psantos1113 (approved):

jamesrichard91 (approved):

src/main/java/com/synopsys/integration/alert/web/api/user/UserController.java
```diff
@@ -51,76 +55,81 @@
...
-        this.contentConverter = contentConverter;
-        this.responseFactory = responseFactory;
         this.authorizationManager = authorizationManager;
         this.userActions = userActions;
         this.descriptorKey = descriptorKey;
     }
 
     @GetMapping
-    public ResponseEntity<String> getAllUsers() {
+    public List<UserConfig> getAllUsers() {
```
gkillough, line 71: Can we wrap this List in an object? 
For exmaple:
```java
public class MultiUserConfigResponseModel {
    private List<UserConfig> users;
    .
    .
    .
}
```

ChomickiM: Sure, something similar to MultiRolePermissionModel.java that just has methods to get the UserConfig list?

gkillough (approved):

PR #1164, opened 2020-09-11
Role Controller Refactor

gkillough:
All functionality is working correctly. This will get another pass once the `ActionResults` changes are available on master.


src/main/js/store/actions/certificates.js
```diff
@@ -173,6 +160,7 @@ export function validateCertificate(certificate) {
 
         const validateRequest = RequestUtilities.createPostRequest(CERTIFICATES_VALIDATION_API_URL, csrfToken, certificate);
         validateRequest.then((response) => {
+            // FIXME no need to check if response is OK
```
gkillough, line 163: I think this is wrong now.

src/main/java/com/synopsys/integration/alert/web/api/role/MultiRolePermissionModel.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class MultiRolePermissionModel extends AlertSerializableModel {
+    private final List<RolePermissionModel> roles;
+
+    MultiRolePermissionModel() {
+        // For serialization
+        this.roles = List.of();
```
bamandel, line 34: Mentioned this in Paulos PR, but do we want to call another constructor here or do we prefer the direct assignment?

bamandel (approved):

PR #1165, opened 2020-09-14
Role Controller Update

gkillough:
Squashed some commits from #1164 to have a more concise commit history.


psantos1113 (approved):

src/main/java/com/synopsys/integration/alert/web/api/role/RoleActions.java
```diff
@@ -86,9 +94,25 @@ public UserRoleModel createRole(RolePermissionModel rolePermissionModel) throws
...
+        Optional<AlertFieldStatus> roleNameMissingError = validateRoleNameFieldRequired(roleName);
+        if (roleNameMissingError.isPresent()) {
+            throw AlertFieldException.singleFieldError(roleNameMissingError.get());
+        }
+
+        UserRoleModel existingRole = getExistingRoleOrThrow404(roleId);
+        boolean targetRoleNameIsUsedByDifferentRole = authorizationUtility.getRoles()
+                                                          .stream()
+                                                          .filter(role -> !role.getId().equals(existingRole.getId()))
+                                                          .anyMatch(role -> role.getName().equalsIgnoreCase(roleName));
```
bamandel, line 108: More of a thought, but would this go through the stream twice? Would it be more efficient to put both checks in either the filter call or the anyMatch call?

bamandel (approved):

ChomickiM (approved):

src/main/java/com/synopsys/integration/alert/web/api/role/RoleActions.java
gkillough: It makes no difference: 
https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html

PR #1166, opened 2020-09-14
Hide Summary processing type when issue tracker channel is selected

jamesrichard91:


src/main/java/com/synopsys/integration/alert/web/custom/ProcessingSelectCustomEndpoint.java
```diff
@@ -0,0 +1,72 @@
...
+    protected List<LabelValueSelectOption> createData(FieldModel fieldModel) throws AlertException {
+        String channelName = fieldModel.getFieldValue(ChannelDistributionUIConfig.KEY_CHANNEL_NAME).orElse("");
+        return Arrays.stream(ProcessingType.values())
+                   .filter(processingType -> this.shouldInclude(processingType, channelName))
+                   .map(processingType -> new LabelValueSelectOption(processingType.getLabel(), processingType.name()))
+                   .collect(Collectors.toList());
+    }
+
+    private boolean shouldInclude(ProcessingType processingType, String channelName) {
+        if (issueTrackerChannelKeys.contains(channelName) && processingType == ProcessingType.SUMMARY) {
```
gkillough, line 66: 1. Why not just make this condition the return statement? 
2. DIGEST doesn't really belong either IMO.

psantos1113 (approved):

src/main/java/com/synopsys/integration/alert/web/custom/ProcessingSelectCustomEndpoint.java
jamesrichard91: I can change it to make the condition the return. I think DIGEST still fits. DEFAULT would let users have a full audit trail of the components reflected in Jira, while DIGEST would be the current items of concern (without opening issues and potentially closing them in the same batch)

PR #1167, opened 2020-09-15
UI: Fix Role Permissions Options

gkillough:
- Only display the contexts relevant to the descriptor selected
- Show no contexts until descriptor selected
- Hide upload permissions for descriptors other than Settings

### DIFF
--- a/src/main/js/dynamic/loaded/users/PermissionTable.js
+++ b/src/main/js/dynamic/loaded/users/PermissionTable.js
@@ -48,7 +48,15 @@ class PermissionTable extends Component {
         const updatedValue = type === 'checkbox' ? checked.toString()
             .toLowerCase() === 'true' : value;
         const trimmedValue = (Array.isArray(updatedValue) && updatedValue.length > 0) ? updatedValue[0] : updatedValue;
-        const newPermissions = Object.assign(permissionsData, { [name]: trimmedValue });
+
+        let newPermissions = Object.assign(permissionsData, { [name]: trimmedValue });
+        if (newPermissions && newPermissions.descriptorName !== 'Settings') {
+            newPermissions = Object.assign(newPermissions, {
+                [PERMISSIONS_TABLE.UPLOAD_READ]: undefined,
+                [PERMISSIONS_TABLE.UPLOAD_WRITE]: undefined,
+                [PERMISSIONS_TABLE.UPLOAD_DELETE]: undefined
+            });
+        }
         this.setState({
             permissionsData: newPermissions
         });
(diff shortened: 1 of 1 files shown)


src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -47,7 +48,15 @@ class PermissionTable extends Component {
...
-        const newPermissions = Object.assign(permissionsData, { [name]: trimmedValue });
+
+        let newPermissions = Object.assign(permissionsData, { [name]: trimmedValue });
+        if (newPermissions && newPermissions.descriptorName !== 'Settings') {
+            newPermissions = Object.assign(newPermissions, {
+                [PERMISSIONS_TABLE.UPLOAD_READ]: undefined,
+                [PERMISSIONS_TABLE.UPLOAD_WRITE]: undefined,
+                [PERMISSIONS_TABLE.UPLOAD_DELETE]: undefined
+            });
+        }
```
gkillough, line 59: If the settings permissions happen to be set, unset them if the descriptor isn't 'Settings'.

src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -147,13 +156,30 @@ class PermissionTable extends Component {
...
+            });
+
+            if (permissionsData.context && !availableContexts.includes(permissionsData.context)) {
+                this.setState({
+                    permissionsData: {
+                        descriptorName: permissionsData.descriptorName
+                    }
+                });
+            }
+        }
```
gkillough, line 177: If the user changes descriptors, clear the context if it is no longer relevant to the new descriptor.

src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -4,6 +4,7 @@ import TableDisplay from 'field/TableDisplay';
 import DynamicSelectInput from 'field/input/DynamicSelect';
 import CheckboxInput from 'field/input/CheckboxInput';
 import { CONTEXT_TYPE } from 'util/descriptorUtilities';
+import StatusMessage from '../../../field/StatusMessage';
```
jamesrichard91, line 7: This import should be absolute not relative

jamesrichard91 (approved):

src/main/js/dynamic/loaded/users/PermissionTable.js
bamandel: Do our lint settings cover this?

bamandel (approved):
I like the helper comments in the PR for JS. Explains strange code well.

src/main/js/dynamic/loaded/users/PermissionTable.js
psantos1113: The linting thinks that the absolute paths are wrong because it cannot resolve the path from the current location where the file is located.  Currently it does not catch this.

src/main/js/dynamic/loaded/users/PermissionTable.js
psantos1113: SAML configuration is in Authentication not Settings.  I think you need to change the keyword from Settings to 'Authentication'

src/main/js/dynamic/loaded/users/PermissionTable.js
```diff
@@ -178,11 +204,44 @@ class PermissionTable extends Component {
     }
 
     createPermissionsModal() {
-        const { permissionsData } = this.state;
+        const { permissionsData, errorMessage } = this.state;
+
+        let uploadInputs;
+        // Currently, there does not seem to be a good way to filter this dynamically.
+        // For now, restrict upload permissions to 'Settings'.
+        if (permissionsData && permissionsData.descriptorName === 'Settings') {
```
psantos1113, line 212: Same as above use 'Authentication' not 'Settings'

psantos1113 (commented):
A few minor tweaks to limit where the permissions should show up.

src/main/js/dynamic/loaded/users/PermissionTable.js
gkillough: Oof, good catch.

psantos1113 (approved):

ChomickiM (approved):

PR #1168, opened 2020-09-16
distribution validation

psantos1113:
Update the distribution jobs to display validation errors when saving or testing configurations.


src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -107,66 +106,73 @@ class DistributionConfiguration extends Component {
...
             const emptyProviderConfig = FieldModelUtilities.createFieldModelWithDefaults(newProvider.fields, newProvider.context, newProvider.name);
-            this.props.checkDescriptorForGlobalConfig(KEY_PROVIDER_NAME, newProvider.name);
+            validateDescriptorForGlobalConfig(KEY_PROVIDER_NAME, newProvider.name);
             this.setState({
                 providerConfig: emptyProviderConfig,
                 currentProvider: newProvider
             });
         }
+
+        if (prevProps.status === 'VALIDATING' && status === 'VALIDATED') {
```
psantos1113, line 128: This ensures the validation happens before saving the data.

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -107,66 +106,73 @@ class DistributionConfiguration extends Component {
...
+            const jsonBody = this.buildJsonBody();
+            if (isUpdatingJob) {
+                updateDistribution(jsonBody);
+            } else {
+                saveDistribution(jsonBody);
+            }
+        }
```
psantos1113, line 135: This ensures validation happens before saving the job.

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -107,66 +106,73 @@ class DistributionConfiguration extends Component {
...
+    handleTestSubmit(event) {
+        event.preventDefault();
+        const { testDistribution } = this.props;
+        const jsonBody = this.buildJsonBody();
+        testDistribution(jsonBody);
     }
 
     buildJsonBody() {
         const { channelConfig, providerConfig } = this.state;
+        const { jobId } = this.props;
```
psantos1113, line 153: Moved the code that checked the job id into this function instead of having the check in multiple places.

src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -6,11 +6,7 @@ import * as DescriptorUtilities from 'util/descriptorUtilities';
 import { OPERATIONS } from 'util/descriptorUtilities';
 import FieldsPanel from 'field/FieldsPanel';
 import {
-    checkDescriptorForGlobalConfig,
-    getDistributionJob,
-    saveDistributionJob,
-    testDistributionJob,
-    updateDistributionJob
+    checkDescriptorForGlobalConfig, getDistributionJob, saveDistributionJob, testDistributionJob, updateDistributionJob, validateDistributionJob
```
psantos1113, line 9: Most of the changes in this class are for linting issues.  I have commented on the changes that actually were made to get distribution job validation to work.

src/main/js/store/actions/distributionConfigs.js
```diff
@@ -206,11 +238,19 @@ export function testDistributionJob(config) {
...
+                    const handler = createErrorHandler(DISTRIBUTION_JOB_TEST_FAILURE, defaultHandler);
+                    if (responseData.errors && !Object.keys(responseData.errors).length) {
                         dispatch(testJobSuccess(responseData.message));
+                    } else if (!response.ok) {
+                        dispatch(handler(response.status));
                     } else {
-                        handleFailureResponse(DISTRIBUTION_JOB_TEST_FAILURE, dispatch, responseData, response.status);
+                        dispatch(handler(400));
                     }
+                    dispatch(handler(response.status));
```
gkillough, line 253: What's the difference among these three calls? It seems like we should be able to drop the else-if and else to just call `dispatch(handler(response.status));`

src/main/js/store/actions/distributionConfigs.js
```diff
@@ -241,3 +281,24 @@ export function checkDescriptorForGlobalConfig(errorFieldName, descriptorName) {
...
+            response.json()
+                .then((responseData) => {
+                    const handler = createErrorHandler(DISTRIBUTION_JOB_VALIDATE_ERROR, () => validateJobError(responseData));
+                    if (responseData.errors && !Object.keys(responseData.errors).length) {
+                        dispatch(validatedJob());
+                    } else if (!response.ok) {
+                        dispatch(handler(response.status));
+                    } else {
+                        dispatch(handler(400));
+                    }
```
gkillough, line 300: When will the else be called, but not the else-if?

gkillough (approved):

src/main/js/store/actions/distributionConfigs.js
```diff
@@ -173,7 +195,12 @@ export function saveDistributionJob(config) {
                     if (response.ok) {
```
ChomickiM, line 195: This should be moved above the promise for responseData

src/main/js/store/actions/distributionConfigs.js
```diff
@@ -241,3 +281,24 @@ export function checkDescriptorForGlobalConfig(errorFieldName, descriptorName) {
...
+    return (dispatch, getState) => {
+        dispatch(validatingJob());
+        const { csrfToken } = getState().session;
+        const request = ConfigRequestBuilder.createValidateRequest(ConfigRequestBuilder.JOB_API_URL, csrfToken, config);
+        request.then((response) => {
+            response.json()
+                .then((responseData) => {
+                    const handler = createErrorHandler(DISTRIBUTION_JOB_VALIDATE_ERROR, () => validateJobError(responseData));
+                    if (responseData.errors && !Object.keys(responseData.errors).length) {
+                        dispatch(validatedJob());
```
jamesrichard91, line 295: It seems like we are checking does the field "errors" have any values and if it does then do the dispatch but we are not passing the errors along. Is this expected? Are the errors that we checked for being ignored/lost?

src/main/js/store/actions/distributionConfigs.js
```diff
@@ -206,11 +238,19 @@ export function testDistributionJob(config) {
...
             response.json()
                 .then((responseData) => {
-                    if (response.ok) {
+                    const defaultHandler = () => jobError({
+                        type: DISTRIBUTION_JOB_TEST_FAILURE,
+                        ...responseData
+                    });
+                    const handler = createErrorHandler(DISTRIBUTION_JOB_TEST_FAILURE, defaultHandler);
+                    if (responseData.errors && !Object.keys(responseData.errors).length) {
                         dispatch(testJobSuccess(responseData.message));
```
jamesrichard91, line 247: It seems odd to me that we are checking for values in the "errors" field and then passing along the "message" field.  Is this check  looking for errors? or is it looking for the absence of errors?

src/main/js/store/actions/distributionConfigs.js
psantos1113: The else will be called when we have validation errors.  The controller has to return an OK response to contain the validation errors.  We check to see if there are errors and  we then handle it like a bad request which updates the UI with the field validation errors.  If we had a 401 or 403 or another unexpected error like 500 then the else if case will be used. Otherwise there are field validation errors and we end up in the else case.

gkillough: Ah yes, that makes sense.

src/main/js/store/actions/distributionConfigs.js
gkillough: Notice the `!` (NOT) operator prefixing the second condition.

src/main/js/store/actions/distributionConfigs.js
psantos1113: the dispatch at the end I need to remove

src/main/js/store/actions/distributionConfigs.js
psantos1113: yes the errors are passed in the else case.  the dispatch of the validatedJob() sets the state that signifies the data is valid. The else called handler(400) meaning bad request so the  default handler will call validateJobError which takes the responseData that contains the message and the errors and set the status to ERROR so a save will never be called.

src/main/js/store/actions/distributionConfigs.js
psantos1113: looking for the absence of errors.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigController.java
```diff
@@ -100,6 +100,7 @@ public ValidationResponseModel validate(JobFieldModel requestBody) {
 
     @Override
     public ValidationResponseModel test(JobFieldModel resource) {
-        return ResponseFactory.createContentResponseFromAction(jobConfigActions.test(resource));
+        ValidationActionResponse response = jobConfigActions.test(resource);
+        return ResponseFactory.createContentResponseFromAction(new ValidationActionResponse(HttpStatus.OK, response.getContent().orElse(null)));
```
bamandel, line 104: If we're ok returning null, then why do we work with an optional as Content?

gkillough: I brought this same issue up in the video code review. It shouldn't ever actually return null, but we should have a static initializer for validation responses anyway.

src/main/js/store/actions/distributionConfigs.js
```diff
@@ -173,7 +195,12 @@ export function saveDistributionJob(config) {
                     if (response.ok) {
                         dispatch(saveJobSuccess(responseData.message));
                     } else {
-                        handleFailureResponse(DISTRIBUTION_JOB_SAVE_ERROR, dispatch, responseData, response.status);
+                        const defaultHandler = () => jobError({
+                            type: DISTRIBUTION_JOB_SAVE_ERROR,
+                            ...responseData
+                        });
+                        const handler = createErrorHandler(DISTRIBUTION_JOB_SAVE_ERROR, defaultHandler);
+                        dispatch(handler(response.status));
```
bamandel, line 203: is handler a function here? We're passing a status to handler but it doesn't look like handler expects anything passed to it.

psantos1113: createErrroHandler creates a function reference like a lambda expression.  We are assigning handler to the function that is returned by createErrorHandler.

bamandel (approved):

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigController.java
psantos1113: ActionResponse has the getContent() method and in some cases there is no content to return it can be null you just want the http status code and/or message with no content. So because content can be null we return an Optional to indicate to other users of the API that the content can be null and avoid NullPointerExceptions.  The Optional allows you to know you need to check if it is present otherwise you would need to have special code to check for null otherwise you would have NullPointerExceptions.  ValidationActionResponse extends ActionResponse. Because we are wrapping the ValidationActionResponse with a new ValidationActionResponse with an OK status to represent validation or test execution, we need to get the content with getContent() which is an Optional.  So to work with the Optional API i have the orElse(null) for the empty optional case.

It's a little beyond the scope of this PR but we might be able to refactor the API to make it a clearer if there's confusion.

PR #1170, opened 2020-09-16
global config validation

ChomickiM:
Updating the front end code for Global Configuration to use our new validation format


psantos1113 (approved):

src/main/js/store/actions/globalConfiguration.js
```diff
@@ -153,18 +157,39 @@ function clearFieldErrors() {
...
+    errorHandlers.push(HTTPErrorUtils.createBadRequestHandler(defaultHandler));
+    errorHandlers.push(HTTPErrorUtils.createPreconditionFailedHandler(defaultHandler));
+    errorHandlers.push(HTTPErrorUtils.createDefaultHandler(defaultHandler));
+
+    return HTTPErrorUtils.createHttpErrorHandler(errorHandlers);
 }
 
-export function refreshConfig(id) {
+export function
+refreshConfig(id) {
```
bamandel, line 192: Do we prefer a multiline method signature?

bamandel (approved):

src/main/js/store/actions/globalConfiguration.js
```diff
@@ -258,17 +304,29 @@ export function updateConfig(config) {
...
+                    response.json()
+                        .then((responseData) => {
+                            const newId = responseData.id;
+                            const updatedConfig = FieldModelUtilities.updateFieldModelSingleValue(config, 'id', newId);
+                            dispatch(configUpdated(updatedConfig));
+                            dispatch(refreshConfig(newId));
+                        });
+                } else {
+                    dispatch(refreshConfig(id));
+                }
```
gkillough, line 318: You could remove the else (for the response.status check) and only need to call refresh once: 
```suggestion
                        });
                }
                dispatch(refreshConfig(id));
```

gkillough (approved):

src/main/js/store/actions/globalConfiguration.js
gkillough: Ah, but you'll have to create an id variable outside of the if-statement and set it for a 201.

src/main/js/store/actions/globalConfiguration.js
jamesrichard91: Why is `export function` and `refreshConfig(id) {` on different lines?

ChomickiM: This was a mistake, I can update it

PR #1172, opened 2020-09-16
Feat action response

psantos1113:
Refactor controllers and actions to return a response object.  This is the initial set of changes to support this design change.


alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomFunctionAction.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final String API_FUNCTION_URL = "/api/function";
+    private AuthorizationManager authorizationManager;
+
+    public CustomFunctionAction(AuthorizationManager authorizationManager) {
+        this.authorizationManager = authorizationManager;
+    }
+
+    public ActionResponse<T> createResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) {
+        try {
+            if (!authorizationManager.hasExecutePermission(fieldModel.getContext(), fieldModel.getDescriptorName())) {
```
jamesrichard91, line 44: So the assumption is that every "function" requires execute permission?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/LabelValueSelectOptions.java
```diff
@@ -0,0 +1,31 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.descriptor.config.field;
+
+import java.util.List;
+
+public class LabelValueSelectOptions extends FieldOptions<LabelValueSelectOption> {
```
jamesrichard91, line 27: Is there a reason we are keeping this around? It seems like we have been changing the endpoints to return their own custom versions of FieldOptions. EX: ProviderProjectOptions.
It seems like the code is split between using this class as a general Option vs custom implementations of FieldOptions. Is the plan to use the class to replace those custom classes? or to replace this by creating more custom classes?

src/main/java/com/synopsys/integration/alert/channel/email/web/EmailAddressSelectOption.java
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.channel.email.web;
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+
+public class EmailAddressSelectOption extends AlertSerializableModel {
+    private String emailAddress;
+    private Boolean optOut;
```
jamesrichard91, line 29: We may want to discuss not exposing the `optOut` field, since it currently does not serve a functional purpose.

src/test/java/com/synopsys/integration/alert/component/settings/PasswordResetServiceTest.java
```diff
@@ -110,7 +110,7 @@ public void resetPasswordValidTestIT() throws AlertException {
         addConfigurationFieldToMap(keyToFieldMap, EmailPropertyKeys.JAVAMAIL_PORT_KEY.getPropertyKey(), testProperties.getProperty(TestPropertyKey.TEST_EMAIL_SMTP_PORT));
 
         String username = "username";
-        UserModel userModel = UserModel.newUser(username, "", "noreply@synopsys.com", AuthenticationType.DATABASE, Set.of(), true);
+        UserModel userModel = UserModel.newUser(username, "", "noreply@blackducksoftware.com", AuthenticationType.DATABASE, Set.of(), true);
```
jamesrichard91, line 113: This seems odd. Why did this get changed to blackducksoftware?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomFunctionAction.java
psantos1113: This is the same permission that we used to check in the CustomEndpointController in the past.

src/main/java/com/synopsys/integration/alert/channel/email/web/EmailAddressSelectOption.java
psantos1113: We can annotate it with `JSONIgnore` so that Jackson will not serialize it or if we aren't going to have optOut then we should remove it.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomFunctionAction.java
gkillough: I think it might be a good idea to create a SPIKE to determine if we should do this on a per-endpoint basis.

gkillough: Perhaps a protected method to perform the permissions check could be added to this class for now?

src/test/java/com/synopsys/integration/alert/component/settings/PasswordResetServiceTest.java
psantos1113: the mail server we use for testing cannot resolve noreply@synopsys.com but it can resolve noreply@blackducksoftware.com since that is the default set in black duck for the sysadmin user.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/ValidationActionResponse.java
```diff
@@ -0,0 +1,58 @@
...
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+import com.synopsys.integration.rest.exception.IntegrationRestException;
+
+public class ValidationActionResponse extends ActionResponse<ValidationResponseModel> {
+
+    public static ValidationActionResponse createResponseFromIntegrationRestException(IntegrationRestException integrationRestException) {
+        String exceptionMessage = integrationRestException.getMessage();
+        String message = exceptionMessage;
+        if (StringUtils.isNotBlank(integrationRestException.getHttpStatusMessage())) {
+            message += " : " + integrationRestException.getHttpStatusMessage();
```
gkillough, line 39: There is a space before the colon.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/LabelValueSelectOptions.java
psantos1113: If the other endpoints are returning FieldOptions<?> then I will change the usage to be consistent with what has been done.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,182 @@
...
+public abstract class AbstractConfigResourceActions implements LongResourceActions<FieldModel>, TestAction<FieldModel>, ValidateAction<FieldModel> {
+    private AuthorizationManager authorizationManager;
+    private DescriptorAccessor descriptorAccessor;
+
+    public AbstractConfigResourceActions(AuthorizationManager authorizationManager, DescriptorAccessor descriptorAccessor) {
+        this.authorizationManager = authorizationManager;
+        this.descriptorAccessor = descriptorAccessor;
+    }
+
+    protected abstract ActionResponse<FieldModel> createAfterChecks(FieldModel resource);
```
gkillough, line 51: I'm still not in love with the naming here. `createAfterChecks` sounds like it will perform checks before creating IMO. `createWithoutChecks` seems like it would be more appropriate.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractConfigResourceActions.java
```diff
@@ -0,0 +1,182 @@
...
+
+    public ActionResponse<List<FieldModel>> getAllByContextAndDescriptor(String context, String descriptorName) {
+        if (!authorizationManager.hasReadPermission(context, descriptorName)) {
+            return ActionResponse.createForbiddenResponse();
+        }
+        return readAllByContextAndDescriptorAfterChecks(context, descriptorName);
+    }
+
+    @Override
+    public ActionResponse<FieldModel> create(FieldModel resource) {
```
gkillough, line 75: Can this method be final? Is there a reason not to make it final here?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,218 @@
...
+    protected abstract ValidationActionResponse testAfterChecks(JobFieldModel resource);
+
+    protected abstract ActionResponse<JobFieldModel> updateAfterChecks(UUID id, JobFieldModel resource);
+
+    protected abstract ValidationActionResponse validateAfterChecks(JobFieldModel resource);
+
+    private Set<String> getDescriptorNames() {
+        Set<String> descriptorNames = Set.of();
+        try {
+            descriptorNames = descriptorAccessor.getRegisteredDescriptors()
```
gkillough, line 76: @bamandel Did you create a ticket to investigate throwing `AlertDatabaseConstraintException`? I think no-args methods like this are definitely a good candidate to remove the exception.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -0,0 +1,218 @@
...
+                                  .collect(Collectors.toSet());
+        } catch (AlertDatabaseConstraintException ex) {
+            // ignore or add a logger.
+        }
+
+        return descriptorNames;
+    }
+
+    @Override
+    public ActionResponse<JobFieldModel> create(JobFieldModel resource) {
```
gkillough, line 89: Same comment about `final` as is in the other ResourceActions abstract class.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/LongResourceActions.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.action.api;
+
+public interface LongResourceActions<T> extends ResourceActions<T, Long> {
```
gkillough, line 25: I feel like `LongIdResourceActions` might be a better name. Thoughts?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/oauth/OAuthEndpointResponse.java
```diff
@@ -27,23 +27,16 @@
 import com.synopsys.integration.util.Stringable;
 
 public class OAuthEndpointResponse extends Stringable implements Serializable {
```
gkillough, line 29: We have a class called `AlertSerializableModel` that `extends Stringable` and `implements Serializable` so that we don't have to extend both like this.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/LabelValueSelectOptions.java
gkillough: Look at most of the custom endpoints. 80% or more use this.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/BaseJobResourceController.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+import com.synopsys.integration.alert.common.rest.model.JobFieldModel;
+
+public interface BaseJobResourceController {
```
gkillough, line 38: Should we do something similar here to abstract the id Type?

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/ValidateController.java
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.rest.api;
+
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+
+public interface ValidateController<S> {
```
gkillough, line 30: This is trivial, but why was a different generic character used here rather than `<T>`?

src/main/java/com/synopsys/integration/alert/channel/jira/cloud/web/JiraCloudCustomEndpoint.java
```diff
@@ -81,20 +78,20 @@ public JiraCloudCustomEndpoint(JiraCloudChannelKey jiraChannelKey, CustomEndpoin
             String accessToken = jiraProperties.getAccessToken();
             Response response = jiraAppService.installMarketplaceCloudApp(JiraConstants.JIRA_APP_KEY, username, accessToken);
             if (BooleanUtils.isTrue(response.isStatusCodeError())) {
-                return responseFactory.createBadRequestResponse("", "The Jira Cloud server responded with error code: " + response.getStatusCode());
+                return new ActionResponse<>(HttpStatus.BAD_REQUEST, "The Jira Cloud server responded with error code: " + response.getStatusCode());
             }
             boolean jiraPluginInstalled = isJiraPluginInstalled(jiraAppService, accessToken, username, JiraConstants.JIRA_APP_KEY);
             if (!jiraPluginInstalled) {
-                return responseFactory.createNotFoundResponse("Was not able to confirm Jira Cloud successfully installed the Jira Cloud plugin. Please verify the installation on you Jira Cloud server.");
+                return new ActionResponse<>(HttpStatus.NOT_FOUND, "Was not able to confirm Jira Cloud successfully installed the Jira Cloud plugin. Please verify the installation on you Jira Cloud server.");
```
gkillough, line 85: I know you didn't make this change, but can we change the wording to be "Unable to confirm..." rather than "Was not able to confirm..."?

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/LabelValueSelectOptions.java
jamesrichard91: Is that the distinction? The custom endpoints will use this class, while the other endpoints will create custom FieldOptions for the swagger?
I just want to figure out what the difference is

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -56,150 +62,183 @@
...
-                processedFieldModels.add(fieldModel);
-            }
-            configurationAccessor.deleteJob(configurationJobModel.getJobId());
-            for (FieldModel fieldModel : processedFieldModels) {
-                fieldModelProcessor.performAfterDeleteAction(fieldModel);
+    @Override
+    protected ActionResponse<JobFieldModel> deleteAfterChecks(UUID id) {
+        try {
+            Optional<ConfigurationJobModel> jobs = configurationAccessor.getJobById(id);
+            if (jobs.isPresent()) {
```
gkillough, line 142: If this is missing, I believe we should return a 404 to conform to the RFC standard.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -56,150 +62,183 @@
...
-        Map<String, FieldModel> descriptorAndContextToPreviousFieldModel = new HashMap<>();
-        for (ConfigurationModel previousJobConfiguration : previousJob.getCopyOfConfigurations()) {
-            FieldModel previousJobFieldModel = modelConverter.convertToFieldModel(previousJobConfiguration);
-            descriptorAndContextToPreviousFieldModel.put(previousJobFieldModel.getDescriptorName() + previousJobFieldModel.getContext(), previousJobFieldModel);
-        }
+    @Override
+    protected ActionResponse<JobFieldModel> updateAfterChecks(UUID id, JobFieldModel resource) {
+        try {
+            ConfigurationJobModel previousJob = configurationAccessor.getJobById(id)
+                                                    .orElseThrow(() -> new IllegalStateException("No previous job present when the only possible valid state for this stage of the method would require it"));
```
gkillough, line 195: I think this should actually be a 404 case.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
bamandel: Yes, there's a Spike in 7.0.0 to investigate whether the exception is necessary.

src/main/js/store/actions/globalConfiguration.js
```diff
@@ -258,17 +303,29 @@ export function updateConfig(config) {
...
+                    response.json()
+                        .then((responseData) => {
+                            const newId = responseData.id;
+                            const updatedConfig = FieldModelUtilities.updateFieldModelSingleValue(config, 'id', newId);
+                            dispatch(configUpdated(updatedConfig));
+                            dispatch(refreshConfig(newId));
+                        });
+                } else {
+                    dispatch(refreshConfig(id));
+                }
```
gkillough, line 317: Would this work? Or is the `response.json()` promise preventing this?
```suggestion
                let newId = id;
                if (response.status === 201) {
                    response.json()
                        .then((responseData) => {
                           newId = responseData.id;
                            const updatedConfig = FieldModelUtilities.updateFieldModelSingleValue(config, 'id', newId);
                            dispatch(configUpdated(updatedConfig));
                        });
                }
                dispatch(refreshConfig(newId));
```

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/ValidateController.java
psantos1113: There was initially in the first version going to be a generic for response R it was going to be ValidateController<S,R> but I didn't need the second type.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/BaseJobResourceController.java
psantos1113: We'd have to abstract more than just the ID we'd have to abstract the request body as well. If that's something we really want to do we can do that in a smaller PR after the fact.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
psantos1113: We already check if the object is present and will return a 404 not found in the delete method of AbstractJobResourceAction.  The check for if present is because the getJobById returns an optional.

src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
psantos1113: in the update method a 404 not found is already returned. If we are executing here and cannot find it it's an issue.

src/main/java/com/synopsys/integration/alert/channel/email/web/EmailAddressSelectOption.java
psantos1113: Alert is already serializing the optOut from the ProviderUserModel class. That's what the custom endpoint did in the past.  Since the custom endpoint dealt with email addresses I created this object to represent the email address data.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/BaseJobResourceController.java
gkillough: Let's leave it out for now.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/LabelValueSelectOptions.java
psantos1113: The reason we have the specific types is because the response would be ActionResponse<FieldOptions<LabelValueSelectOption>> for the method.  In other places where I had a `ActionResponse<List<SomeObject>>` we wanted an object that wraps a list and return something like this `ActionResponse<WrappedListObject>`.

See this PR: 
https://github.com/blackducksoftware/blackduck-alert/pull/1158

Since that was the preference of the team I created the following list wrapper objects:
LabelValueSelectOptions
ProviderProjectOptions
EmailAddressOptions
NotificationFilterModelOptions
ProviderProjectOptions

This was done in this PR: 
https://github.com/blackducksoftware/blackduck-alert/pull/1160

The implementation was the same for each of these objects only the generic type was different.  So I created FieldOptions so I wouldn't write the same code over and over again.

src/main/js/store/actions/globalConfiguration.js
psantos1113: You can't do your proposed change because the parsing of the responData happens in a .then.  It is a promise.  It executes at a different time than the dispatch(refreshConfig(newId) that is proposed on line 316.  It needs to be the way that it is written because of the promise that is processing the response body.

gkillough (approved):

jamesrichard91 (approved):

PR #1173, opened 2020-09-17
Include the context and descriptor in the field model since they are missing

psantos1113:
The context and descriptor name are missing from the global configurations after delete.  Any custom endpoints like the Azure authenticate will not work because the permissions will fail.  There is no context or descriptor name in the field model to use for the permission check of a custom endpoint.  Therefore add the context and descriptor name if they are missing.

### DIFF
--- a/src/main/js/util/fieldModelUtilities.js
+++ b/src/main/js/util/fieldModelUtilities.js
@@ -52,8 +52,8 @@ export function isFieldModelValueSet(fieldModel, key) {
 
 export function areKeyToValuesEmpty(fieldModel) {
     if (fieldModel && fieldModel.keyToValues) {
-        const keyToValues = fieldModel.keyToValues;
-        return Object.keys(keyToValues).some(key => {
+        const { keyToValues } = fieldModel;
+        return Object.keys(keyToValues).some((key) => {
             const value = keyToValues[key];
             return !hasValuesOrIsSet(value);
         });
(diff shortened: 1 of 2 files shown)


src/main/js/util/fieldModelUtilities.js
```diff
@@ -191,9 +191,20 @@ export function createFieldModelErrorKey(fieldKey) {
     return fieldKey.concat('Error');
 }
 
+export function checkContextAndDescriptor(fieldModel, context, descriptorName) {
+    const copy = JSON.parse(JSON.stringify(fieldModel));
```
bamandel, line 195: Don't we have a utility for copying? Even if it does the same thing, it's probably best to reference that function incase we want to change the implementation universally

bamandel (approved):

src/main/js/util/fieldModelUtilities.js
```diff
@@ -191,9 +191,20 @@ export function createFieldModelErrorKey(fieldKey) {
...
+        copy.context = context;
+    }
+    if (!copy.descriptorName) {
+        copy.descriptorName = descriptorName;
+    }
+    return copy;
+}
+
 export function checkModelOrCreateEmpty(fieldModel, fieldKeys) {
     const emptyFieldModel = createEmptyFieldModel(fieldKeys, fieldModel.context, fieldModel.descriptorName);
```
gkillough, line 206: Is this field model truly empty? Could this variable be named better?

gkillough (approved):

ChomickiM (approved):

PR #1176, opened 2020-09-18
 Change the task actions and controller to ActionResponse.

psantos1113:
Changes to use ActionResponse classes and added tests.


jamesrichard91 (approved):

psantos1113 (commented):
I will need to wrap the List<TaskMetaData> into an object to be consistent with the other APIs and updated the UI code.  I will do that before merging.

bamandel (approved):

ChomickiM (approved):

PR #1177, opened 2020-09-21
Jr code quality

jamesrichard91:
We should not have public mutable fields, especially if they are not being used.


psantos1113 (approved):

src/main/java/com/synopsys/integration/alert/web/documentation/SwaggerConfiguration.java
```diff
@@ -41,14 +41,14 @@
...
-        "/v2/api-docs"
-    };
+    //public static final String[] SWAGGER_WEB_RESOURCE_DIRECTORIES = new String[] {
+    //        SWAGGER_DEFAULT_URL,
+    //        "/webjars",
+    //        "/webjars/**",
+    //        "/swagger-resources",
+    //        "/swagger-resources/**",
+    //        "/v2/api-docs"
+    //    };
```
bamandel, line 51: Can we add a comment here as to why this is comented out?

jamesrichard91: Alternatively I can delete this. Having public mutable fields is being reported as a vulnerability and this field was not even being referenced anywhere. So I commented it out to remove the vulnerability until we determine what the purpose of this String array was.

ChomickiM (approved):

src/main/java/com/synopsys/integration/alert/web/documentation/SwaggerConfiguration.java
bamandel: I would delete it. If we need it back we can just get it from the files history.

bamandel (approved):

PR #1180, opened 2020-09-22
audit controller

psantos1113:
Change the AuditController and action class to use ActionResponse.

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
@@ -139,7 +139,7 @@ public class AuditEntryActions {
                     UUID jobConfigId = commonConfig.getJobId();
                     logger.warn("The Distribution Job with Id {} was disabled. This notification could not be sent", jobConfigId);
                     String message = String.format("The Distribution Job is currently disabled. %s", jobConfigId.toString());
-                    return new ActionResponse<>(HttpStatus.GONE, message);
+                    return new ActionResponse<>(HttpStatus.BAD_REQUEST, message);
                 }
             } else {
                 distributionEvents = notificationProcessor.processNotifications(List.of(notificationContent));
@@ -158,7 +158,7 @@ public class AuditEntryActions {
             }
             return get();
         } catch (IntegrationException e) {
-            return new ActionResponse<>(HttpStatus.BAD_REQUEST, e.getMessage());
+            return new ActionResponse<>(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage());
         }
     }
 
--- a/src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
+++ b/src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
@@ -43,10 +43,10 @@ import com.synopsys.integration.alert.common.event.DistributionEvent;
 import com.synopsys.integration.alert.common.persistence.accessor.ConfigurationAccessor;
 import com.synopsys.integration.alert.common.persistence.accessor.NotificationManager;
 import com.synopsys.integration.alert.common.persistence.model.AuditEntryModel;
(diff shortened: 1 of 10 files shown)


src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
```diff
@@ -31,102 +31,135 @@
...
+                UUID commonDistributionId = UUID.fromString(event.getConfigId());
+                Long auditId = auditUtility.findMatchingAuditId(notificationContent.getId(), commonDistributionId).orElse(null);
+                Map<Long, Long> notificationIdToAuditId = new HashMap<>();
+                notificationIdToAuditId.put(notificationContent.getId(), auditId);
+                event.setNotificationIdToAuditId(notificationIdToAuditId);
+                eventManager.sendEvent(event);
+            }
+            return get();
+        } catch (IntegrationException e) {
+            return new ActionResponse<>(HttpStatus.BAD_REQUEST, e.getMessage());
```
gkillough, line 161: Wouldn't this be an INTERNAL_SERVER_ERROR since we return the GONE response if there are issues with the two input fields? Maybe this try-catch is too big and we need two to handle the BAD_REQUEST case and the INTERNAL_SERVER_ERROR case?

gkillough (approved):

src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
```diff
@@ -31,102 +31,135 @@
...
 import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
-import com.synopsys.integration.alert.common.rest.model.AlertPagedModel;
+import com.synopsys.integration.alert.common.security.authorization.AuthorizationManager;
 import com.synopsys.integration.alert.common.workflow.processor.notification.NotificationProcessor;
+import com.synopsys.integration.alert.component.audit.AuditDescriptorKey;
 import com.synopsys.integration.exception.IntegrationException;
 
 @Component
 @Transactional
 public class AuditEntryActions {
```
bamandel, line 57: We can't use the new ConfigAction helper classes to cut down on duplicate code here?

bamandel (approved):

src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
```diff
@@ -31,102 +31,135 @@
...
+                    return new ActionResponse<>(HttpStatus.GONE, message);
+                }
+                ConfigurationJobModel commonConfig = commonDistributionConfig.get();
+                if (commonConfig.isEnabled()) {
+                    distributionEvents = notificationProcessor.processNotifications(commonConfig, List.of(notificationContent));
+                } else {
+                    UUID jobConfigId = commonConfig.getJobId();
+                    logger.warn("The Distribution Job with Id {} was disabled. This notification could not be sent", jobConfigId);
+                    String message = String.format("The Distribution Job is currently disabled. %s", jobConfigId.toString());
+                    return new ActionResponse<>(HttpStatus.GONE, message);
```
jamesrichard91, line 142: If the job is disabled, shouldn't this be a BAD_REQUEST response instead of GONE?

psantos1113: Given that the job is disabled and not deleted then BAD_REQUEST makes sense rather than GONE.

src/main/java/com/synopsys/integration/alert/web/api/audit/AuditEntryActions.java
psantos1113: Unfortunately the classes that are there right now will require more empty implementation code.  Only the 
`get(Long id)` matches one of the method signatures.  Once we have all the Action and controller classes converted to return ActionResponses another pass on the classes we have in case improvements can be made.

ChomickiM (approved):

PR #1181, opened 2020-09-22
Remove Jira Admin Requirements

gkillough:
- Jira Cloud
- Jira Server

### DIFF
--- a/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudRequestDelegator.java
+++ b/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudRequestDelegator.java
@@ -69,15 +69,9 @@ public class JiraCloudRequestDelegator {
         }
         JiraCloudProperties jiraProperties = (JiraCloudProperties) context.getIssueTrackerConfig();
         JiraCloudServiceFactory jiraCloudServiceFactory = jiraProperties.createJiraServicesCloudFactory(logger, gson);
-        PluginManagerService jiraAppService = jiraCloudServiceFactory.createPluginManagerService();
-        logger.debug("Verifying the required application is installed on the Jira Cloud server...");
-        try {
-            boolean missingApp = !jiraAppService.isAppInstalled(jiraProperties.getUsername(), jiraProperties.getAccessToken(), JiraConstants.JIRA_APP_KEY);
-            if (missingApp) {
-                throw new IssueTrackerException("Please configure the Jira Cloud plugin for your server instance via the global Jira Cloud channel settings.");
-            }
-        } catch (IntegrationException ex) {
-            throw new IssueTrackerException("Please configure the Jira Cloud plugin for your server instance via the global Jira Cloud channel settings.", ex);
+
+        if (!jiraProperties.isPluginCheckDisabled()) {
+            checkIfAlertPluginIsInstalled(jiraCloudServiceFactory, jiraProperties);
         }
 
         ProjectService projectService = jiraCloudServiceFactory.createProjectService();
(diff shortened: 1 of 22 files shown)


src/main/java/com/synopsys/integration/alert/channel/jira/cloud/descriptor/JiraCloudGlobalUIConfig.java
```diff
@@ -38,13 +39,16 @@
...
-    public static final String LABEL_ADMIN_API_TOKEN = "Admin API Token";
+    public static final String LABEL_EMAIL_ADDRESS = "Email Address";
+    public static final String LABEL_API_TOKEN = "API Token";
+    public static final String LABEL_DISABLE_PLUGIN_CHECK = "Disable Plugin Check";
     public static final String LABEL_CONFIGURE_PLUGIN = "Configure Jira Cloud plugin";
 
     public static final String DESCRIPTION_URL = "The URL of the Jira Cloud server.";
-    public static final String DESCRIPTION_ADMIN_USER_NAME = "The email address of the admin used to log into the Jira Cloud server that has generated the API token.";
-    public static final String DESCRIPTION_ADMIN_API_TOKEN = "The admin API token used to send API requests to the Jira Cloud server.";
+    public static final String DESCRIPTION_USER_NAME = "The email address of the user Jira Cloud user. Note: Unless 'Disable Plugin Check' is checked, this user must be a Jira admin.";
```
jamesrichard91, line 48: "of the user Jira Cloud user" should be simplified to "of the Jira Cloud user"

jamesrichard91 (approved):

jamesrichard91:
Am I missing the changes to JiraServerGlobalUIConfig? Or have those changes not been made yet?

src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudRequestDelegator.java
```diff
@@ -98,4 +92,17 @@ public IssueTrackerResponse sendRequests(List<IssueTrackerRequest> requests) thr
         return jiraIssueHandler.createOrUpdateIssues(validIssueConfig, requests);
     }
 
+    private void checkIfAlertPluginIsInstalled(JiraCloudServiceFactory jiraCloudServiceFactory, JiraCloudProperties jiraProperties) throws IssueTrackerException {
```
bamandel, line 95: Should this be passed PluginManagerService, username, and accessToken?

gkillough: It's not a public method, so I think it's better to minimize the clutter. Especially since this operation won't always happen.

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomFunctionAction.java
```diff
@@ -85,7 +84,7 @@ public JiraServerCustomFunctionAction(AuthorizationManager authorizationManager,
                 }
                 createBadRequestIntegrationException(e);
             }
-            boolean jiraPluginInstalled = isJiraPluginInstalled(jiraAppService, password, username, JiraConstants.JIRA_APP_KEY);
+            boolean jiraPluginInstalled = JiraPluginCheckUtil.checkIsAppInstalledAndRetryIfNecessary(() -> jiraAppService.isAppInstalled(username, password, JiraConstants.JIRA_APP_KEY));
```
bamandel, line 87: This method is done almost exactly the same as the other of checkIsAppInstalledAndRetryIfNecessary. Can you change the method to take a BiFunction so that all you need here is to pass in username and password then a method reference to jiraAppService::isAppInstalled

bamandel (approved):

src/main/java/com/synopsys/integration/alert/channel/jira/server/web/JiraServerCustomFunctionAction.java
gkillough: I would still need to create a lambda in order to pass the app key as the third param. I don't know if it's worth it at that point.

bamandel: Actually, both server and cloud use the same PluginManagerService. Could you just pass that, username, and password to the static method and run the call yourself?

gkillough: I could. What are you trying to solve here? You don't like the lambda?

bamandel: The lambda just feels a little unnecessary here. I think we could make it clearer with just normal method parameters. Lambdas to me generally mean different functionality, but we're always doing the same check.

bamandel (approved):

ChomickiM (approved):

psantos1113 (approved):

PR #1184, opened 2020-09-23
Fix Global Config Test

gkillough:
No distinction was being made between success and error messages in validation.

### DIFF
--- a/src/main/js/dynamic/GlobalConfiguration.js
+++ b/src/main/js/dynamic/GlobalConfiguration.js
@@ -6,7 +6,11 @@ import FieldsPanel from 'field/FieldsPanel';
 import ConfigurationLabel from 'component/common/ConfigurationLabel';
 
 import {
-    deleteConfig, getConfig, testConfig, updateConfig, validateConfig
+    deleteConfig,
+    getConfig,
+    testConfig,
+    updateConfig,
+    validateConfig
 } from 'store/actions/globalConfiguration';
 import * as FieldModelUtilities from 'util/fieldModelUtilities';
 import * as DescriptorUtilities from 'util/descriptorUtilities';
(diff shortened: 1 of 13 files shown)


src/main/js/dynamic/GlobalConfiguration.js
```diff
@@ -6,7 +6,11 @@ import FieldsPanel from 'field/FieldsPanel';
 import ConfigurationLabel from 'component/common/ConfigurationLabel';
```
gkillough, line 7: The changes here are all linting. I originally thought I had to modify something here, but did not end up needing to. I decided to leave the linting as an incremental improvement.

src/main/js/dynamic/GlobalConfiguration.js
```diff
@@ -24,37 +28,40 @@ class GlobalConfiguration extends React.Component {
...
             currentFields: fields,
             showTest: false
         };
     }
 
     componentDidMount() {
-        const fieldModel = this.state.currentConfig;
-        this.props.getConfig(fieldModel.descriptorName);
+        const { currentConfig } = this.state;
+        const { getConfig: getConfig1 } = this.props;
```
bamandel, line 45: I'm not a fan of this name. Why is it getConfig1? If it's a duplicate item there should be a better name.

src/main/js/dynamic/GlobalConfiguration.js
```diff
@@ -24,37 +28,40 @@ class GlobalConfiguration extends React.Component {
...
+        getConfig1(currentConfig.descriptorName);
     }
 
     componentDidUpdate(prevProps, prevState, snapshot) {
         const { currentFields, currentDescriptor } = this.state;
-        if (this.props.currentConfig !== prevProps.currentConfig && this.props.updateStatus === 'DELETED') {
+        const { updateStatus, updateConfig: updateConfig1, currentConfig } = this.props;
```
bamandel, line 51: This has updateConfig1, there should be a better name for this.

bamandel (approved):

src/main/js/dynamic/GlobalConfiguration.js
bamandel: Just saw this message, the linter made some questionable function names that we will probably want to change.

psantos1113: we just have to make sure that that function names mapDispatchToProps section aren't the same as the function names used in the dispatch functions in the same section.  That is the linting error called shadowing if my memory is correct.

psantos1113 (approved):

ChomickiM (approved):

PR #1186, opened 2020-09-23
list container support

psantos1113:
Add the mechanism to have an object returned by the ReadAll abstract methods to support the MultiUserConfigReponseModel and MultiRolePermissionModel classes.  These changes will unblock the work being done on the user controller.


ChomickiM (approved):

bamandel (approved):

jamesrichard91 (approved):

src/main/js/store/actions/globalConfiguration.js
```diff
@@ -268,8 +269,8 @@ export function getConfig(descriptorName) {
             response.json()
                 .then((responseData) => {
                     if (response.ok) {
-                        if (responseData.length > 0) {
-                            dispatch(configFetched(responseData[0]));
+                        if (responseData.fieldModels && responseData.fieldModels.length > 0) {
```
gkillough, line 272: Do providers use these actions? If not, then we should use `length == 1` here I think

gkillough (approved):

src/main/js/store/actions/globalConfiguration.js
psantos1113: providers do use these actions but not this function specifically.  I will add the change and test.

PR #1188, opened 2020-09-24
 system controller

psantos1113:
Convert SystemActions to use ActionResponse and convert tests to parameterized tests.

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/web/actions/SystemActionsTest.java
+++ b/src/test/java/com/synopsys/integration/alert/web/actions/SystemActionsTest.java
@@ -1,27 +1,28 @@
 package com.synopsys.integration.alert.web.actions;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.time.ZonedDateTime;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-import java.util.Optional;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.mockito.Mockito;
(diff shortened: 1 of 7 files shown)


jamesrichard91 (approved):

gkillough (approved):

src/test/java/com/synopsys/integration/alert/web/actions/SystemActionsTest.java
```diff
@@ -1,157 +1,88 @@
...
-        systemActions.getSystemMessagesBetween("2018-11-13T00:00:00.000Z", "2018-11-13T01:00:00.000Z");
-        Mockito.verify(defaultSystemMessageUtility).findBetween(Mockito.any());
-    }
-
-    @Test
-    public void testGetSystemMessages() {
-        SystemActions systemActions = new SystemActions(defaultSystemStatusUtility, defaultSystemMessageUtility, fieldModelProcessor, settingsUtility);
-        systemActions.getSystemMessages();
-        Mockito.verify(defaultSystemMessageUtility).getSystemMessages();
+    @ParameterizedTest
```
bamandel, line 51: Are we sure we want to use a parameterized test here? It looks like we just want to create a start and end time list and iterate over it in the test. Parameterized tests will just make it harder to debug the test in my opinion.

bamandel (commented):
I think parameterized tests should be used sparingly. If it errors out, it's hard to tell which test failed and it's hard to debug the specific failed test. When we had separate tests, we knew the specific test that failed and can easily debug it.

src/test/java/com/synopsys/integration/alert/web/actions/SystemActionsTest.java
jamesrichard91: This is what parameterized tests are meant for. I don't think we should change it

psantos1113: Parameterized tests are expanded into individual tests where the input is shown as part of the test.

bamandel: Parameterized are meant for stuff like this, but ends up making things harder in my opinion. Why use a parameterized test when you can just pass the list yourself?

bamandel (approved):

ChomickiM (approved):

PR #1189, opened 2020-09-24
Update users to use ActionResponse

ChomickiM:


src/main/java/com/synopsys/integration/alert/web/api/certificate/CertificateActions.java
```diff
@@ -81,7 +81,7 @@ public CertificateActions(CertificatesDescriptorKey descriptorKey, Authorization
     }
 
     @Override
-    public ActionResponse<CertificateModel> readWithoutChecks(Long id) {
+    protected ActionResponse<CertificateModel> readWithoutChecks(Long id) {
```
jamesrichard91, line 84: Are we changing these methods to protected so that only the controller in the same package can call these methods?

ChomickiM: All of the abstract methods in AbstractResourceActions are protected. While working on users we noticed this was a mistake and after checking with Paulo I decided to include the changes in this PR as well.

psantos1113: It was a mistake this wasn't changed initially.

psantos1113 (approved):

gkillough (approved):

bamandel (approved):

PR #1192, opened 2020-09-29
Renaming Utilities to Accessors for consistency

jamesrichard91:
Renaming AuditUtility to AuditAccessor, and AuthorizationUtility to AuthorizationAccessor


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/AuthorizationAccessor.java
```diff
@@ -30,7 +30,7 @@
 import com.synopsys.integration.alert.common.persistence.model.PermissionMatrixModel;
 import com.synopsys.integration.alert.common.persistence.model.UserRoleModel;
 
-public interface AuthorizationUtility {
+public interface AuthorizationAccessor {
```
gkillough, line 33: While we're at it, let's just call it `RoleAccessor`.

gkillough (approved):
I stand by my comment about "RoleAccessor", but other than that, looks good.

PR #1199, opened 2020-09-30
Fix(CSS): Set overflow-y to inherit on expandableContainer

gkillough:
This prevents overriding the page size dropdown from being visible above the container.

### DIFF
--- a/src/main/css/table.scss
+++ b/src/main/css/table.scss
@@ -124,7 +124,7 @@
   border-radius: $border-radius;
   border: 1px solid $border-color;
   background-color: $synopsys-color-section-background;
-  overflow-y: scroll !important;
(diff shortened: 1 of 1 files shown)


src/main/css/table.scss
```diff
@@ -124,7 +124,7 @@
   border-radius: $border-radius;
   border: 1px solid $border-color;
   background-color: $synopsys-color-section-background;
-  overflow-y: scroll !important;
+  overflow-y: inherit !important;
```
jamesrichard91, line 127: What does this change do?

gkillough: >This prevents overriding the page size dropdown from being visible above the container.

psantos1113 (approved):

PR #1200, opened 2020-09-30
Jr auth manager

jamesrichard91:
Moving more of the RoleAccessor usage into the AuthorizationManager.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/security/authorization/AuthorizationManager.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/security/authorization/AuthorizationManager.java
@@ -139,6 +139,23 @@ public class AuthorizationManager {
                    .anyMatch(name -> permissionCache.containsKey(name) && permissionCache.get(name).hasPermissions(permissionKey, operations));
     }
 
+    public Set<UserRoleModel> getRoles() {
+        return roleAccessor.getRoles();
+    }
+
+    public Set<UserRoleModel> getRoles(Collection<Long> roleIds) {
+        return roleAccessor.getRoles(roleIds);
+    }
+
+    public boolean doesRoleNameExist(String name) {
+        return permissionCache.containsKey(name);
+    }
+
+    public void updateRoleName(Long roleId, String roleName) throws AlertDatabaseConstraintException {
+        roleAccessor.updateRoleName(roleId, roleName);
+        loadPermissionsIntoCache();
+    }
+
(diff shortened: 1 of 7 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/security/authorization/AuthorizationManager.java
```diff
@@ -139,6 +139,23 @@ public final boolean hasAllPermissions(String context, String descriptorName, Ac
...
+    public Set<UserRoleModel> getRoles() {
+        return roleAccessor.getRoles();
+    }
+
+    public Set<UserRoleModel> getRoles(Collection<Long> roleIds) {
+        return roleAccessor.getRoles(roleIds);
+    }
+
+    public boolean doesRoleNameExist(String name) {
+        return permissionCache.containsKey(name);
```
jamesrichard91, line 151: Now that the roles are being updated using methods in this class and the cache is being updated in those methods, is it alright to check the cache for the role name? Or should we still using the RoleAccessor to check the database?

ChomickiM (approved):

src/main/java/com/synopsys/integration/alert/web/api/role/RoleActions.java
```diff
@@ -56,19 +55,17 @@
 @Component
 public class RoleActions {
     private static final String FIELD_KEY_ROLE_NAME = "roleName";
-    private final RoleAccessor roleAccessor;
     private final AuthorizationManager authorizationManager;
```
gkillough, line 58: I'm not sure delegating Role retrieval to the AuthorizationManager is the right move. Yes, AuthorizationManager has a RoleAccessor, and yes, the RoleActions would otherwise need both, but I do not believe this is correctly separating concerns. RoleAccessor is the thing that should be depended on to read/write roles to and from the database and AuthorizationManager is the utility that cross-references roles with users.

I'm not entirely sure why AuthorizationManager was even being used in this class before.

jamesrichard91: The AuthorizationManager is being used to make sure the cache is kept up to date.

I was not sure about adding the get's to AuthorizationManager. I would be fine with rejecting this PR, if the team thinks it would be better to keep the role retrieval in RoleAccessor.

gkillough: That's my preference, but I'm also second guessing the permissions cache. I can't remember the motivation for it. It seems to me that, if anything, there should be an object representing just the cache that gets used only when needed.

PR #1201, opened 2020-09-30
helm push fix

psantos1113:
helm chart push use absolute file path.

### DIFF
--- a/buildSrc/deploymentTasks.gradle
+++ b/buildSrc/deploymentTasks.gradle
@@ -62,7 +62,7 @@ task helmPushChart(type: Exec) {
     if (!project.ext.isSnapshot) {
         // if we need to setup helm again to push the charts please run the following command to add the repo
         // helm repo add sig-cloudnative-local https://${System.env.ARTIFACTORY_HOST}/artifactory/sig-cloudnative-local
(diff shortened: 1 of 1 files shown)


buildSrc/deploymentTasks.gradle
```diff
@@ -63,7 +63,7 @@ task helmPushChart(type: Exec) {
         if (!project.ext.isSnapshot) {
             // if we need to setup helm again to push the charts please run the following command to add the repo
             // helm repo add sig-cloudnative-local https://${System.env.ARTIFACTORY_HOST}/artifactory/sig-cloudnative-local
-            commandLine "helm", "push-artifactory", "synopsys-alert*.tgz", "sig-cloudnative-local", "-u", "${project.ext.artifactoryDeployerUsername}", "-p", "${project.ext.artifactoryDeployerPassword}"
+            commandLine "helm", "push-artifactory", "${project.buildDir}/deployment/helm/synopsys-alert/synopsys-alert-${project.version}.tgz", "sig-cloudnative-local", "-u", "${project.ext.artifactoryDeployerUsername}", "-p", "${project.ext.artifactoryDeployerPassword}"
```
jamesrichard91, line 66: We could do something like 
``` java
"${workingDir}/synopsys-alert-${project.version}.tgz"
```
That way we only need to maintain the path in one variable

jamesrichard91 (approved):

PR #1203, opened 2020-10-01
Validate required related fields in custom actions

gkillough:
Currently the UI only displays errors related to EndpointButton fields; however, the request was to fix those specific fields. A future enhancement would be to display errors on all types of Endpoint fields.


alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomFunctionAction.java
```diff
@@ -22,28 +22,53 @@
...
+    private final FieldValidationUtility fieldValidationUtility;
+
+    public CustomFunctionAction(String fieldKey, AuthorizationManager authorizationManager, DescriptorProcessor descriptorProcessor, FieldValidationUtility fieldValidationUtility) {
+        this.fieldKey = fieldKey;
         this.authorizationManager = authorizationManager;
+        this.descriptorProcessor = descriptorProcessor;
+        this.fieldValidationUtility = fieldValidationUtility;
     }
 
     public ActionResponse<T> createResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) {
```
gkillough, line 61: This function is the primary change, most of the other changes are related to updating autowired constructor params.

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validation/FieldValidationUtility.java
```diff
@@ -33,14 +33,13 @@
...
 import com.synopsys.integration.alert.common.descriptor.config.field.errors.AlertFieldStatus;
-import com.synopsys.integration.alert.common.descriptor.config.field.validators.ValidationResult;
 import com.synopsys.integration.alert.common.enumeration.FieldType;
 import com.synopsys.integration.alert.common.rest.model.FieldModel;
 import com.synopsys.integration.alert.common.rest.model.FieldValueModel;
 
 @Component
-public class FieldValidationAction {
-    private final Logger logger = LoggerFactory.getLogger(FieldValidationAction.class);
+public class FieldValidationUtility {
```
gkillough, line 41: This file contains some secondary changes. It is primarily making more methods public and updating the return types instead of modifying input parameters.

jamesrichard91 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomFunctionAction.java
```diff
@@ -22,28 +22,53 @@
...
 
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
 import org.springframework.http.HttpStatus;
 import org.springframework.web.server.ResponseStatusException;
 
+import com.github.jsonldjava.shaded.com.google.common.base.Predicate;
```
psantos1113, line 33: I think you want to use Java's Predicate and not this Predicate object.

psantos1113 (approved):
I think you just need to update the import statement and the changes here look good.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/CustomFunctionAction.java
gkillough: Wow, good catch. Why is this even a suggestion from Intellij haha?

ChomickiM (approved):

PR #1207, opened 2020-10-02
Update roles for action responses

ChomickiM:


src/main/java/com/synopsys/integration/alert/web/api/role/RoleActions.java
```diff
@@ -35,101 +34,139 @@
...
+            String roleName = resource.getRoleName();
+            Optional<UserRoleModel> existingRole = roleAccessor.getRoles(List.of(id))
+                                                       .stream()
+                                                       .findFirst();
+            if (existingRole.isPresent()) {
+                boolean targetRoleNameIsUsedByDifferentRole = roleAccessor.getRoles()
+                                                                  .stream()
+                                                                  .filter(role -> !role.getId().equals(existingRole.get().getId()))
+                                                                  .anyMatch(role -> role.getName().equalsIgnoreCase(roleName));
+                if (targetRoleNameIsUsedByDifferentRole) {
```
psantos1113, line 135: Isn't this handled by the validate method?

src/main/java/com/synopsys/integration/alert/web/api/role/util/PermissionModelUtil.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import com.synopsys.integration.alert.common.enumeration.AccessOperation;
+import com.synopsys.integration.alert.common.persistence.model.PermissionKey;
+import com.synopsys.integration.alert.common.persistence.model.PermissionMatrixModel;
+import com.synopsys.integration.alert.common.util.BitwiseUtil;
+import com.synopsys.integration.alert.web.api.role.PermissionModel;
+
+public class PermissionModelUtil {
```
gkillough, line 35: Didn't we already have something like this?

gkillough (approved):

src/main/java/com/synopsys/integration/alert/web/api/role/util/PermissionModelUtil.java
ChomickiM: convertToPermissionMatrixModel used to be a private method of the RoleActions. We wanted access to it in the test classes so we decided to move it into its own util class that can then be used by both rather than duplicating the code just for tests. If we already had a util class doing this type of work I may be unaware of it.

psantos1113 (approved):

src/main/java/com/synopsys/integration/alert/web/api/role/RoleActions.java
```diff
@@ -35,101 +34,132 @@
...
-                   .map(this::convertUserRoleModel)
-                   .collect(Collectors.toList());
-    }
-
-    public ValidationResponseModel validateRoleFields(RolePermissionModel rolePermissionModel) {
-        return validateRoleNameFieldRequired(rolePermissionModel.getRoleName())
-                   .map(requiredFieldError -> ValidationResponseModel.fromStatusCollection("There were problems with the role configuration", List.of(requiredFieldError)))
-                   .orElseGet(() -> ValidationResponseModel.success("The role name is valid"));
+    @Override
+    protected ActionResponse<RolePermissionModel> createWithoutChecks(RolePermissionModel resource) {
```
jamesrichard91, line 73: I know this change was made a while back but why is every method `...WithoutCheck`?  Is it because the validation is done elsewhere? Why don't we make that more clear without forcing that suffix on every method?

bamandel (approved):

jamesrichard91 (approved):

PR #1208, opened 2020-10-05
Fix Global Config Banner (for 204 No Content response)

gkillough:


jamesrichard91 (approved):

src/main/js/store/reducers/globalConfiguration.js
```diff
@@ -125,7 +125,7 @@ const globalConfiguration = (state = initialState, action) => {
                 fetching: false,
                 updateStatus: 'UPDATED',
                 testing: false,
-                actionMessage: 'Update successful',
+                actionMessage: 'Save successful',
```
gkillough, line 128: The button is called "Save" and that covers both cases rather than mislabeling the POST case.

bamandel (approved):

psantos1113 (approved):

ChomickiM (approved):

PR #1212, opened 2020-10-07
Move Custom Function Actions

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/table/EndpointTableSelectField.java
```diff
@@ -25,17 +25,17 @@
...
-import com.synopsys.integration.alert.common.action.CustomFunctionAction;
 import com.synopsys.integration.alert.common.descriptor.config.field.endpoint.EndpointField;
 import com.synopsys.integration.alert.common.enumeration.FieldType;
+import com.synopsys.integration.alert.common.rest.api.AbstractFunctionController;
 
 public class EndpointTableSelectField extends EndpointField {
     private boolean paged;
     private boolean searchable;
-    private List<TableSelectColumn> columns;
+    private final List<TableSelectColumn> columns;
```
bamandel, line 35: why the use of final here?

bamandel (approved):

psantos1113 (approved):

ChomickiM (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/endpoint/table/EndpointTableSelectField.java
gkillough: Intellij must have added it. Good catch, I'll look into it.

gkillough: Looks like it's because the only operations on `columns` are additive.

PR #1214, opened 2020-10-08
Provider Sub-Project

gkillough:


src/main/java/com/synopsys/integration/alert/workflow/startup/component/LastSearchDataMigration.java
```diff
@@ -57,10 +57,10 @@
...
     // We don't have admin privileges to postgres when alert starts.
 
-    private BlackDuckProviderKey blackDuckProviderKey;
-    private ConfigurationAccessor configurationAccessor;
-    private FilePersistenceUtil filePersistenceUtil;
-    private ProviderTaskPropertiesAccessor providerTaskPropertiesAccessor;
+    private final BlackDuckProviderKey blackDuckProviderKey;
+    private final ConfigurationAccessor configurationAccessor;
+    private final FilePersistenceUtil filePersistenceUtil;
+    private final ProviderTaskPropertiesAccessor providerTaskPropertiesAccessor;
```
bamandel, line 63: I guess I don't understand our uses of final. Do we prefer member objects be final?

bamandel (approved):

src/main/java/com/synopsys/integration/alert/workflow/startup/component/LastSearchDataMigration.java
gkillough: Yes. It's also part of our Integration_OSS_Formatting thing which is why these were automatically added when the file was modified.

psantos1113 (approved):

ChomickiM (approved):

PR #1216, opened 2020-10-09
java web project

psantos1113:
Extracted the web java code into a web project.  In order to do this and not have circular dependencies I need to create additional sub projects.

projects created:
component - everything related to components including authentication and scheduling (similar to channel and provider projects)
workflow - The common classes in the workflow package.


component/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+    }
+}
+
+dependencies {
+    implementation project(':alert-common')
+    implementation project(':alert-database')
+    implementation project(':channel')
+    implementation project(":workflow")
+
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
```
gkillough, line 16: Could we make this an API dependency of channel instead of duplicating it here?

web/src/test/java/com/synopsys/integration/alert/web/api/home/HomeActionsTest.java
```diff
@@ -23,30 +23,30 @@
...
-import com.synopsys.integration.alert.web.security.authentication.saml.SAMLContext;
+import com.synopsys.integration.alert.component.authentication.security.saml.SAMLContext;
 
-public class HomActionsTest {
+public class HomeActionsTest {
     @Test
     public void testVerifyAuthenticationValid() {
+        HttpServletRequest servletRequest = new MockHttpServletRequest();
+        HttpServletResponse servletResponse = new MockHttpServletResponse();
+        HttpSessionCsrfTokenRepository csrfTokenRepository = new HttpSessionCsrfTokenRepository();
```
bamandel, line 33: These may be better created in a beforeEach method? Looks like the majority of these classes are used in most tests

bamandel (approved):

build.gradle
```diff
@@ -87,9 +87,12 @@ dependencies {
     implementation project(':alert-common')
     implementation project(':alert-database')
     implementation project(':channel')
+    implementation project(':component')
     implementation project(':provider')
+    implementation project(':workflow')
     implementation project(':web')
 
+
     // Spring
```
gkillough, line 96: We we able to remove some of the dependencies we copied over to the `web` sub-project?

component/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+
+    implementation 'org.springframework.security:spring-security-core'
+    implementation 'org.springframework.security:spring-security-config'
+    implementation 'org.springframework.security:spring-security-ldap'
+    implementation 'org.springframework.security:spring-security-web'
+
+    implementation 'org.springframework.security.extensions:spring-security-saml2-core:1.0.10.RELEASE'
```
gkillough, line 23: Should some of these be _api_ dependencies that `web` could use?

ChomickiM (approved):

component/build.gradle
psantos1113: Went through all of the build.gradle files and updated the configurations for all the libraries.

build.gradle
psantos1113: Went through all of the build.gradle files and updated the configurations for all the libraries.

component/build.gradle
psantos1113: Went through all of the build.gradle files and updated the configurations for all the libraries.

web/src/test/java/com/synopsys/integration/alert/web/api/home/HomeActionsTest.java
psantos1113: Encapsulated the initialization of the fields.

gkillough (approved):

PR #1217, opened 2020-10-12
Update headers to use Headers object

ChomickiM:


jamesrichard91 (approved):

web/src/main/js/store/actions/audit.js
```diff
@@ -87,11 +87,11 @@ export function getAuditData(pageNumber, pageSize, searchTerm, sortField, sortOr
         const errorHandlers = [];
         errorHandlers.push(HTTPErrorUtils.createUnauthorizedHandler(unauthorized));
         errorHandlers.push(HTTPErrorUtils.createForbiddenHandler(() => auditDataFetchError(HTTPErrorUtils.MESSAGES.FORBIDDEN_READ)));
+        const headers = new Headers();
+        headers.append('X-CSRF-TOKEN', csrfToken);
```
bamandel, line 91: Since headers is now an object, let's add functions to the object that does this for us. Such as

> headers.addXCrsfToken(csrfToken)

gkillough (approved):
It would be nice to have a global function that returned the headers with the CSRF token already appended.

web/src/main/js/util/configurationRequestBuilder.js
```diff
@@ -37,52 +37,52 @@ export function createReadRequest(apiUrl, csrfToken, configurationId = null) {
...
-        }
+        headers
     });
 }
 
 export function createNewConfigurationRequest(apiUrl, csrfToken, fieldModel) {
+    const headers = new Headers();
+    headers.append('Accept', 'application/json');
+    headers.append('Content-Type', 'application/json');
+    headers.append('X-CSRF-TOKEN', csrfToken);
```
bamandel, line 52: You could also create constant header objects in the header class that will return this grouping of headers so you don't have to repeat it every time.

bamandel: The function would need to take in a csrfToken though.

bamandel (approved):
I think the header object could do a bit more. But it looks like a good start

psantos1113 (approved):
I agree with the changes suggested here.  Encapsulate the setting of the headers into a utility to avoid any spelling mistakes.

psantos1113 (approved):

PR #1220, opened 2020-10-12
 saml init disable

psantos1113:
Disable SAML if the configuration is found to be invalid on startup.

Update the test functionality to check the entity id as well.


component/src/main/java/com/synopsys/integration/alert/component/authentication/security/saml/SAMLContext.java
```diff
@@ -69,6 +70,24 @@ public boolean isSAMLEnabled() {
...
+            ConfigurationModel configurationModel = getCurrentConfiguration();
+            Map<String, ConfigurationFieldModel> fields = configurationModel.getCopyOfKeyToFieldMap();
+            ConfigurationFieldModel enabledField = fields.get(AuthenticationDescriptor.KEY_SAML_ENABLED);
+            if (null != enabledField) {
+                enabledField.setFieldValue(String.valueOf(false));
+            } else {
+                enabledField = ConfigurationFieldModel.create(AuthenticationDescriptor.KEY_SAML_ENABLED);
+                enabledField.setFieldValue(String.valueOf(false));
+                fields.put(AuthenticationDescriptor.KEY_SAML_ENABLED, enabledField);
+            }
```
gkillough, line 84: This could be simplified to:
```suggestion
            if (null == enabledField) {
                enabledField = ConfigurationFieldModel.create(AuthenticationDescriptor.KEY_SAML_ENABLED);
                fields.put(AuthenticationDescriptor.KEY_SAML_ENABLED, enabledField);
            }
            enabledField.setFieldValue(String.valueOf(false));
```
If it feels weird to modify the reference after the `fields.put(...)` call, you could call `fields.put(...)` after the last line without any significant performance impact.

component/src/main/java/com/synopsys/integration/alert/component/authentication/security/saml/SAMLManager.java
```diff
@@ -166,4 +173,33 @@ private ExtendedMetadataDelegate createDelegate(MetadataProvider provider) {
...
+        if (!entityIdValid) {
+            logger.error("Validated SAML entity id missing.");
+        }
+        try {
+            Optional<ExtendedMetadataDelegate> provider = createHttpProvider(metaDataUrl);
+            if (provider.isPresent()) {
+                provider.get().initialize();
+                httpProviderValid = true;
+            }
+        } catch (Exception ex) {
```
bamandel, line 190: Should we be handling such broad exceptions here?

component/src/main/java/com/synopsys/integration/alert/component/authentication/security/saml/SAMLManager.java
```diff
@@ -166,4 +173,33 @@ private ExtendedMetadataDelegate createDelegate(MetadataProvider provider) {
...
+            logger.error("Validating SAML Metadata URL error: ", ex);
+        }
+
+        try {
+            Optional<ExtendedMetadataDelegate> provider = createFileProvider();
+            if (provider.isPresent()) {
+                provider.get().initialize();
+                fileProviderValid = true;
+            }
+        } catch (Exception ex) {
```
bamandel, line 200: Another very broad expcetion handling

bamandel (approved):

gkillough (approved):

ChomickiM (approved):

component/src/main/java/com/synopsys/integration/alert/component/authentication/security/saml/SAMLManager.java
psantos1113: Yes, because if there's a runtime exception on initialize of the metadata provider then you won't be able to login with SAML, LDAP or the database users.  The only exception that is advertised by the initialize method is a MetadataProviderException, but if the XML Metadata file isn't formatted correctly you can get a SAXParser exception which isn't wrapped in a MetaDataProviderException or advertised as a thrown exception.  We don't want the exception to bubble all the way up because it could cause the server to crash.

Here we are validating as best we can that the SAML configuration is good.  If it is, then configure SAML.  If it isn't valid, then we want to be able to disable the SAML configuration.  SAML has two parts to it.  The database configuration to enable SAML authentication in Alert, which has already happened before this point, and the SAML library portion.  We need to be able to recover from an exception in order to disable SAML configuration in the database. This is the most cautious way to be able to recover.

component/src/main/java/com/synopsys/integration/alert/component/authentication/security/saml/SAMLManager.java
psantos1113: see the previous comment

component/src/main/java/com/synopsys/integration/alert/component/authentication/security/saml/SAMLContext.java
psantos1113: That's not a problem.  I'll implement the change.

PR #1222, opened 2020-10-13
expressing an alternative to action contracts

ekerwin:


alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/MultiResponseModel.java
```diff
@@ -0,0 +1,23 @@
...
+    MultiResponseModel() {
+        // For serialization
+        this(List.of());
+    }
+
+    public MultiResponseModel(List<T> models) {
+        this.models = models;
+    }
+
+    public List<T> getModels() {
```
psantos1113, line 19: You'll need to annotate this method with `JsonIgnore` so that the JSON object created doesn't add a models attribute to the resulting object.

ekerwin: Could I alternatively (or in addition) make the method protected?

jamesrichard91 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/MultiResponseModel.java
psantos1113: I honestly not sure what Jackson would do.  From what I understand it uses the getters and setters which are usually public but I don't know if protected getter would be serialized or hidden.  I assume it would be hidden.

psantos1113: https://github.com/FasterXML/jackson-docs

psantos1113 (approved):

component/src/main/java/com/synopsys/integration/alert/component/users/web/role/RoleActions.java
```diff
@@ -112,6 +112,7 @@ public RoleActions(UserManagementDescriptorKey userManagementDescriptorKey, Role
         return new ActionResponse<>(HttpStatus.NOT_FOUND);
     }
 
+
```
gkillough, line 115: Do you not have SaveActions enabled? Two lines of whitespace here.

channel/src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureBoardsCustomFunctionAction.java
```diff
@@ -75,7 +76,7 @@
     private final ProxyManager proxyManager;
     private final OAuthRequestValidator oAuthRequestValidator;
     // TODO create and use intermediate class for this
```
gkillough, line 78: Can this TODO be removed now that `CompositeResourceActions` exists?

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/MultiResponseModel.java
gkillough: Regardless of Jackson, I think `getModels()` should be protected to better express intent.

PR #1225, opened 2020-10-15
Update to jira common 2.0.0

bamandel:
The biggest change is the issue creation returns a new object that more accurately represents the data returned. Use this returned data to then retrieve all the data about the newly created response object.

### DIFF
--- a/channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/util/JiraCloudIssueHandler.java
+++ b/channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/util/JiraCloudIssueHandler.java
@@ -27,11 +27,11 @@ import java.util.List;
 
 import com.google.gson.Gson;
 import com.synopsys.integration.alert.channel.jira.cloud.JiraCloudProperties;
-import com.synopsys.integration.alert.channel.jira.common.util.JiraIssueHandler;
 import com.synopsys.integration.alert.channel.jira.cloud.descriptor.JiraCloudDescriptor;
 import com.synopsys.integration.alert.channel.jira.common.JiraIssueSearchProperties;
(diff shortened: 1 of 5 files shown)


channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/util/JiraCloudIssueHandler.java
```diff
@@ -60,7 +61,8 @@ public JiraCloudIssueHandler(IssueService issueService, JiraCloudProperties jira
 
     @Override
     public IssueResponseModel createIssue(String issueCreator, String issueType, String projectName, IssueRequestModelFieldsMapBuilder fieldsBuilder) throws IntegrationException {
-        return issueService.createIssue(new IssueCreationRequestModel(issueCreator, issueType, projectName, fieldsBuilder, Collections.emptyList()));
+        IssueCreationResponseModel issueCreationResponseModel = issueService.createIssue(new IssueCreationRequestModel(issueCreator, issueType, projectName, fieldsBuilder, Collections.emptyList()));
+        return issueService.getIssue(issueCreationResponseModel.getKey());
```
gkillough, line 65: How did we get an `IssueResponseModel` from the service before this? Was it calling `getIssue` after creation internally?

gkillough (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/util/JiraCloudIssueHandler.java
bamandel: IssueCretionResponseModel is newly added. We were originally returning IssueResponseModel but pretty much all the values were empty. except 3-4 of them. So I created this new model to reflect that actual content.

jamesrichard91 (approved):

ChomickiM (approved):

PR #1226, opened 2020-10-15
removing read...WithoutChecks

ekerwin:
There was a duplicated call to findExisting - this removes that duplication and eliminates the need for both readWithout and readAllWithout.


jamesrichard91 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -41,13 +41,14 @@
...
 import com.synopsys.integration.alert.common.persistence.model.RegisteredDescriptorModel;
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 import com.synopsys.integration.alert.common.rest.model.FieldModel;
 import com.synopsys.integration.alert.common.rest.model.JobFieldModel;
 import com.synopsys.integration.alert.common.rest.model.MultiJobFieldModel;
 import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
 import com.synopsys.integration.alert.common.security.authorization.AuthorizationManager;
 
-public abstract class AbstractJobResourceActions implements JobResourceActions, ReadAllAction<MultiJobFieldModel>, ValidateAction<JobFieldModel>, TestAction<JobFieldModel> {
+public abstract class AbstractJobResourceActions implements CompositeResourceActions<JobFieldModel, UUID> {
```
gkillough, line 51: FWIW, this controller will not have a "read all" type action when I'm done with implementing paging, so these changes will be instantly reverted. I'm thinking that CompositeResourceActions should not include readAll, because more and more things will soon incorporate paging.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/MultiFieldModel.java
```diff
@@ -24,14 +24,13 @@
 
 import java.util.List;
 
-public class MultiFieldModel extends AlertSerializableModel {
-    private List<FieldModel> fieldModels;
-
-    public MultiFieldModel(List<FieldModel> fieldModels) {
-        this.fieldModels = fieldModels;
+public class MultiFieldModel extends MultiResponseModel<FieldModel> {
+    public MultiFieldModel(final List<FieldModel> models) {
```
gkillough, line 28: I'm shocked to see a `final` here.

gkillough: Why constructors, but not elsewhere? That seems inconsistent with the libraries which definitely modify things like headers and requestBodies in constructors.

component/src/main/java/com/synopsys/integration/alert/component/users/web/role/RoleActions.java
```diff
@@ -113,21 +112,21 @@ public RoleActions(UserManagementDescriptorKey userManagementDescriptorKey, Role
...
-            return new ActionResponse<>(HttpStatus.OK, role.get());
-        }
-        //This is covered by the findExistingCheck in AbstractResourceActions. TODO for 6.4.0
-        return new ActionResponse<>(HttpStatus.NOT_FOUND, String.format("Role with id:%d not found.", id));
+    protected MultiRolePermissionModel createMultiResponseModel(final List<RolePermissionModel> roles) {
+        return new MultiRolePermissionModel(roles);
+    }
+
+    @Override
+    protected RolePermissionModel convertDatabaseModel(UserRoleModel userRoleModel) {
```
gkillough, line 125: I think it would be more clear if the method name included the result. e.g. `convertDatabaseModelToRestModel` or something along those lines.

component/src/main/java/com/synopsys/integration/alert/component/certificates/web/CertificateActions.java
```diff
@@ -69,25 +69,23 @@ public CertificateActions(CertificatesDescriptorKey descriptorKey, Authorization
...
-        if (model.isPresent()) {
-            return new ActionResponse<>(HttpStatus.OK, model.get());
-        }
+    protected MultiCertificateModel createMultiResponseModel(final List<CertificateModel> certificates) {
+        return new MultiCertificateModel(certificates);
+    }
 
-        return new ActionResponse<>(HttpStatus.NOT_FOUND, String.format("Certificate with id:%d not found.", id));
+    @Override
+    protected CertificateModel convertDatabaseModel(CustomCertificateModel databaseCertificateModel) {
```
gkillough, line 86: Again, I feel like this method name only chooses to tell one half of the story. Why not `convertToRestModel`? It's also true, but incomplete.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractResourceActions.java
```diff
@@ -49,16 +54,17 @@ public AbstractResourceActions(DescriptorKey descriptorKey, ConfigContextEnum co
...
         this.authorizationManager = authorizationManager;
-
     }
 
     protected abstract ActionResponse<T> createWithoutChecks(T resource);
 
     protected abstract ActionResponse<T> deleteWithoutChecks(Long id);
 
-    protected abstract ActionResponse<M> readAllWithoutChecks();
+    protected abstract List<D> getDatabaseModels();
```
gkillough, line 63: Can we use the word "retrieve" or "read" rather than "get"? The "get" method-prefix in Java is a ubiquitous convention for accessing _fields_ of a class, so it's nice to make a distinction.

gkillough (changes requested):
I think these changes have the right motivation, but do not take into account the upcoming paging changes that we are all aware of. For now, we should remove "getAll" from the contract of `CompositeResourceActions` and keep it in its own interface.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/MultiFieldModel.java
ekerwin: I can address the final, but I don't understand the comment about the constructors, I'll follow up with you.

component/src/main/java/com/synopsys/integration/alert/component/users/web/role/RoleActions.java
ekerwin: Righto.

component/src/main/java/com/synopsys/integration/alert/component/certificates/web/CertificateActions.java
ekerwin: It might have been easier to critique the method name in the base class so you only had to do it once.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractResourceActions.java
ekerwin: Do you want to apply the same level of precision to the accessors? I went off of their method convention of using 'get'.

ekerwin: I'd argue that consistency is the better part of valor here - so if you object to 'get' as a prefix for non-trivial methods, we should alter the api of the accessors, or just leave the 'get' prefix alone.

gkillough: I was thinking about this with the accessors while modifying them recently and I agree that their prefixes should also change. We can lay one good brick with the actions right now and when the accessors aren't in such a high state of flux, we can refactor those as well.

component/src/main/java/com/synopsys/integration/alert/component/certificates/web/CertificateActions.java
gkillough: My mistake.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractResourceActions.java
ekerwin: Well phrased. :)

PR #1229, opened 2020-10-20
style: Update "Url" to "URL" in field models and some tests

ChomickiM:
Updating some of the usages of "Url" I found to use capital "URL".

### DIFF
--- a/channel/src/test/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudRequestDelegatorTest.java
+++ b/channel/src/test/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudRequestDelegatorTest.java
@@ -240,7 +240,7 @@ public class JiraCloudRequestDelegatorTest {
 
     private JiraIssueSearchProperties createSearchProperties() {
         return new JiraIssueSearchProperties("provider",
-            "providerUrl",
+            "providerURL",
             "topicName",
             "topicValue",
             "subTopicName",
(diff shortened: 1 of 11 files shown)


channel/src/test/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudRequestDelegatorTest.java
```diff
@@ -240,7 +240,7 @@ private JiraCloudProperties createMockServerConfig() throws IssueTrackerExceptio
 
     private JiraIssueSearchProperties createSearchProperties() {
         return new JiraIssueSearchProperties("provider",
-            "providerUrl",
```
gkillough, line 243: This would be a breaking change to the Alert Issue Property Indexer plugin, so we should leave this alone for now.

channel/src/test/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudTestActionTest.java
```diff
@@ -149,7 +149,7 @@ private JiraCloudProperties createMockServerConfig() throws IssueTrackerExceptio
 
     private JiraIssueSearchProperties createSearchProperties() {
         return new JiraIssueSearchProperties("provider",
-            "providerUrl",
+            "providerURL",
```
gkillough, line 152: This would be a breaking change to the Alert Issue Property Indexer plugin, so we should leave this alone for now.

bamandel (approved):

gkillough (approved):
Disregard my previous comments. I thought it was modifying a separate file.

psantos1113 (approved):

PR #1231, opened 2020-10-27
Add static jobs tables to liquibase

gkillough:


ChomickiM (approved):

alert-database/src/main/resources/liquibase/6.4.0/static-distribution-jobs.xml
```diff
@@ -0,0 +1,399 @@
...
+            <column name="job_id" type="UUID">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="add_comments" type="BOOLEAN" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+            <column name="issue_creator_username" type="VARCHAR">
+                <constraints nullable="true"/>
+            </column>
+            <column name="project_name_or_key" type="VARCHAR">
```
bamandel, line 314: Can/Should we decide to use either name of key here instead of either?

bamandel (approved):

alert-database/src/main/resources/liquibase/6.4.0/static-distribution-jobs.xml
gkillough: Good question. Either is valid for Jira, so we're sticking with the or-name.

PR #1233, opened 2020-10-27
Add static job java models

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/job/blackduck/BlackDuckJobDetailsEntity.java
```diff
@@ -0,0 +1,119 @@
...
+    @Column(name = "global_config_id")
+    private Long globalConfigId;
+
+    @Column(name = "filter_by_project")
+    private Boolean filterByProject;
+
+    @Column(name = "project_name_pattern")
+    private String projectNamePattern;
+
+    @OneToMany(mappedBy = "blackDuckJobDetails")
```
bamandel, line 33: Should we have this be blackduck instead of blackDuck? There are more instances if we need to change this.

bamandel (approved):
Are our Blackduck classes supposed to be Blackduck or BlackDuck? I thought it was decided to be the former. Additionally, all variables referncing these classes also have this captilization.

alert-database/src/main/java/com/synopsys/integration/alert/database/job/blackduck/BlackDuckJobDetailsEntity.java
gkillough: ugh lol, not sure

gkillough: I think for capitalized usages and camelCase, we want it to be BlackDuck and blackDuck respectively. With snake case, all characters should be lowercase, but it should still be one word.

jamesrichard91 (approved):

PR #1234, opened 2020-10-28
Add an additional audit logger for Alert

ChomickiM:
The focus of this branch was to create additional logging around users and roles due to issues in the past with customers changing user/role permissions and encountering issues that support ends up missing. With this change we are introducing a new AlertLoggerFactory that creates a new AlertCompositeLogger which is 2 loggers working in tandem but outputting to different locations. The messages created by the "alertAudit" logger will always be logged at the trace level and be output into a new alertAudit.log file under the log directory. This way the user will still be able to preserve the blackduck-alert.log and console logs to the levels they have provided in the application.properties.

This should have no impact on the existing base logger that spring is using and will only be used by classes using the AlertLoggerFactory rather than LoggerFactory.


alert-common/src/main/java/com/synopsys/integration/alert/common/logging/AlertLoggerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.common.logging;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AlertLoggerFactory {
+
```
gkillough, line 29: This blank line can be removed.

src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    <appender name="ALERT_AUDIT" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <encoder>
+			<pattern>${FILE_LOG_PATTERN}</pattern>
+		</encoder>
+		<file>log/alertAudit.log</file>
+		<rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+			<cleanHistoryOnStart>${LOG_FILE_CLEAN_HISTORY_ON_START:-false}</cleanHistoryOnStart>
+			<fileNamePattern>${ROLLING_FILE_NAME_PATTERN:-log/alertAudit.log.%d{yyyy-MM-dd}.%i.gz}</fileNamePattern>
+			<maxFileSize>${LOG_FILE_MAX_SIZE:-10MB}</maxFileSize>
```
gkillough, line 14: This seems a little small. What made you choose this default value? I realized that it's a rolling log file, but it seems like this size could produce a lot of files. I would expect somewhere between 50MB and 250MB.

src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,24 @@
...
+			<maxHistory>${LOG_FILE_MAX_HISTORY:-7}</maxHistory>
+			<totalSizeCap>${LOG_FILE_TOTAL_SIZE_CAP:-0}</totalSizeCap>
```
gkillough, line 16: Will zero here prevent us from logging at all or is it "unlimited"?

src/main/resources/logback-spring.xml
ChomickiM: The appender here is actually a copy of the default file appender used by spring to generate our current blackduck-alert.log file. The only difference here is the path that it is being logged to. The default values here are already the same ones we are using in the alert logs.

alert-common/src/main/java/com/synopsys/integration/alert/common/logging/AlertLoggerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    public static final String ALERT_AUDIT_LOGGER = "alertAudit";
```
bamandel, line 30: This is the value that will appear in the logs, correct? If so, I don't think it should be camel case. Perhaps match the class format?

src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <include resource="org/springframework/boot/logging/logback/base.xml"/>
+
+    <appender name="ALERT_AUDIT" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <encoder>
+			<pattern>${FILE_LOG_PATTERN}</pattern>
+		</encoder>
+		<file>log/alertAudit.log</file>
```
bamandel, line 10: Is the format strange here? I'm not sure if it's github or if it's actually offset

src/main/resources/logback-spring.xml
ChomickiM: I'm looking for an answer still to this specific property to see if a value of 0 is considered unlimited. But regardless this logging is still successfully writing to the alertAudit file

src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,24 @@
...
+		<rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+			<cleanHistoryOnStart>${LOG_FILE_CLEAN_HISTORY_ON_START:-false}</cleanHistoryOnStart>
+			<fileNamePattern>${ROLLING_FILE_NAME_PATTERN:-log/alertAudit.log.%d{yyyy-MM-dd}.%i.gz}</fileNamePattern>
+			<maxFileSize>${LOG_FILE_MAX_SIZE:-10MB}</maxFileSize>
+			<maxHistory>${LOG_FILE_MAX_HISTORY:-7}</maxHistory>
+			<totalSizeCap>${LOG_FILE_TOTAL_SIZE_CAP:-0}</totalSizeCap>
+		</rollingPolicy>
+    </appender>
+
+	<logger name="alertAudit" level="trace" additivity="false">
```
bamandel, line 20: I think this would need to change if we change the logger name in my other comment?

bamandel (approved):

src/main/resources/logback-spring.xml
ChomickiM: I believe its Github, it looks normal in the IDE.

src/main/resources/logback-spring.xml
ChomickiM: Yes, whatever the logger name is here is what we will need to use in the AlertLoggerFactory.

gkillough (approved):

src/main/resources/logback-spring.xml
jamesrichard91: I agree with Gavin here, the default for both should be increased to 50 - 250MB.

jamesrichard91 (approved):

src/main/resources/logback-spring.xml
ChomickiM: This value can also be controlled for both logs by setting logging.file.max-size to the desired size, might be easier to set it there depending on the customer and how big the logs they produce are

psantos1113: The property logging.file.max-size would be a property in the application.properties file which isn't exposed to the customer.  I would suggest adding that property to the application.properties file and also creating an environment variable so it can be controlled by support, devops, or the customer.
Something like this.
`logging.file.max-size=${ALERT_LOG_FILE_SIZE:50MB}`

psantos1113 (approved):

PR #1237, opened 2020-11-02
Jr improvements

jamesrichard91:
Making changes based on the 6.2.0 customer patch I have been working on.
Changing/adding logging to make debugging easier.
Updating the scheduled tasks so each task can define its cron schedule.
Updating the notification processing so we can re-use the stateful provider across multiple jobs using the same provider.
Updating the BlackDuckProperties to clean up the unnecessary Optional return types.
Changing the BlackDuckDataSyncTask to run once a day.


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/notification/NotificationProcessor.java
```diff
@@ -71,18 +76,42 @@ public NotificationProcessor(ConfigurationAccessor configurationAccessor, JobAcc
...
+        for (Map.Entry<NotificationFilterModel, Set<AlertNotificationModel>> entry : notificationFilterMap.entrySet()) {
+            NotificationFilterModel notificationFilterModel = entry.getKey();
+            List<ConfigurationJobModel> matchingJobs = jobAccessor.getMatchingEnabledJobs(notificationFilterModel.getProviderConfigId(), notificationFilterModel.getNotificationType());
+
+            List<AlertNotificationModel> matchingNotifications = new ArrayList<>(entry.getValue());
+            if (!matchingNotifications.isEmpty() && !matchingJobs.isEmpty()) {
+                events.addAll(processNotificationsThatMatchFilter(notificationFilterModel, matchingJobs, matchingNotifications));
+            }
+        }
+        return events;
```
bamandel, line 110: this method is pretty much a duplicate of 79. I think you can create a private method that takes in a bifunction that performs the getMatchingEnabledJobs functionality.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/notification/NotificationProcessor.java
```diff
@@ -152,13 +225,32 @@ public NotificationProcessor(ConfigurationAccessor configurationAccessor, JobAcc
...
-            logger.error("Could not retrieve the provider config for job: {}", job.getName(), e);
+            logger.error("Could not retrieve the provider config for ID: {}", providerConfigID);
+            logger.debug(e.getMessage(), e);
             return Optional.empty();
         }
     }
 
+    private SetMap<NotificationFilterModel, AlertNotificationModel> extractNotificationInformation(List<AlertNotificationModel> notifications) {
+        SetMap<NotificationFilterModel, AlertNotificationModel> notificationFilterMap = SetMap.createDefault();
+        for (AlertNotificationModel alertNotificationModel : notifications) {
```
bamandel, line 240: Would we prefer streams over loops? We could convert this to a stream, map, then collect.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultJobAccessor.java
```diff
@@ -70,6 +71,28 @@ public DefaultJobAccessor(ConfigGroupRepository configGroupRepository, DefaultCo
...
+    @Override
+    public List<ConfigurationJobModel> getMatchingEnabledJobs(Long providerConfigId, NotificationType notificationType) {
+        //TODO change this to return a page of results
+        List<UUID> matchingJobIds = configGroupRepository.findMatchingEnabledJobIds(String.valueOf(providerConfigId), notificationType.name());
+        if (matchingJobIds.isEmpty()) {
+            return List.of();
+        }
+        List<ConfigGroupEntity> jobEntities = configGroupRepository.findByJobIds(matchingJobIds);
+        return convertToJobModels(jobEntities);
+    }
```
bamandel, line 94: I believe these can also use a common private method that is passed a bifunction. You would pass the findMatchingEnabledJobIds function in.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -168,6 +169,24 @@ public StaticJobAccessor(
...
+    }
+
+    @Override
+    public List<ConfigurationJobModel> getMatchingEnabledJobs(Long providerConfigId, NotificationType notificationType) {
+        // TODO change this to return a page of jobs
+        return distributionJobRepository.findMatchingEnabledJob(providerConfigId, notificationType.name())
+                   .stream()
+                   .map(this::convertToConfigurationJobModel)
+                   .collect(Collectors.toList());
+    }
```
bamandel, line 188: This should be able to use a shared private method

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/BlackDuckDistributionTestAction.java
```diff
@@ -72,12 +74,15 @@ public MessageResult testConfig(String configId, FieldModel fieldModel, FieldUti
...
-                                                                            .map(blackDuckProvider::createStatefulProvider)
-                                                                            .map(statefulProvider -> (BlackDuckProperties) statefulProvider.getProperties());
-            if (optionalBlackDuckProperties.isPresent()) {
-                BlackDuckProperties blackDuckProperties = optionalBlackDuckProperties.get();
+            BlackDuckProperties blackDuckProperties = null;
+            Optional<ConfigurationModel> providerConfigurationOptional = configurationAccessor.getConfigurationById(providerConfigId);
+            if (providerConfigurationOptional.isPresent()) {
+                ConfigurationModel providerConfiguration = providerConfigurationOptional.get();
+                StatefulProvider statefulProvider = blackDuckProvider.createStatefulProvider(providerConfiguration);
+                blackDuckProperties = (BlackDuckProperties) statefulProvider.getProperties();
```
bamandel, line 82: Can this be changed to optional mapping? It would remove this if statement and you could orElse(null) at the end.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
```diff
@@ -42,7 +42,9 @@
 public abstract class ScheduledTask implements Runnable {
     public static final String FORMAT_PATTERN = "MM/dd/yyy hh:mm a";
     public static final String STOP_SCHEDULE_EXPRESSION = "";
+    //Spring Cron documentation  https://riptutorial.com/spring/example/21209/cron-expression
     public static final String EVERY_MINUTE_CRON_EXPRESSION = "0 0/1 * 1/1 * *";
+    public static final String ONCE_DAILY_CRON_EXPRESSION = "0 0 0 * * *";
```
ChomickiM, line 47: I realize this is just for the patch, so this may not be an issue, but would it be a problem for customers in other timezones?

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -64,9 +64,11 @@ public void handleEvent(NotificationEvent event) {
                 logger.warn("Can not process a notification event without notification Id's.");
                 return;
             }
+            logger.debug("Event {}", event);
             logger.info("Processing event for {} notifications.", event.getNotificationIds().size());
             List<AlertNotificationModel> notifications = notificationAccessor.findByIds(event.getNotificationIds());
             List<DistributionEvent> distributionEvents = notificationProcessor.processNotifications(FrequencyType.REAL_TIME, notifications);
+            logger.info("Sending {} events for notifications.", distributionEvents.size());
```
bamandel, line 71: Do you think this could be too verbose for info? We already mention processing events. I would prefer debug

workflow/src/main/java/com/synopsys/integration/alert/workflow/scheduled/frequency/ProcessingTask.java
```diff
@@ -72,7 +72,9 @@ public DateRange getDateRange() {
     public void runTask() {
         DateRange dateRange = getDateRange();
         List<AlertNotificationModel> notificationList = read(dateRange);
+        logger.info("Processing {} notifications.", notificationList.size());
         List<DistributionEvent> distributionEvents = notificationProcessor.processNotifications(getDigestType(), notificationList);
+        logger.info("Sending {} events for notifications.", distributionEvents.size());
```
bamandel, line 77: Should one of these be debug as well? We don't want to spam too much

bamandel (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/notification/NotificationProcessor.java
psantos1113: We could use streams streams aren't always as performant as regular for loops.  But if there are optimizations for streams in subsequent releases of Java the code won't automatically get those performance benefits.  I would say it depends on running the performance tests that were written before changing over to streams.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
psantos1113: We are always using UTC timezone.  So it would run based on the UTC timezone.

psantos1113 (approved):

JakeMathews (approved):

ChomickiM (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/notification/NotificationProcessor.java
jamesrichard91: I would be curious how you would use a Stream here? Because we use the `AlertNotificationModel` to create a `NotificationFilterModel` and then put them both in the SetMap. Unless you are just suggesting I turn this for loop into ``` java
notifications.stream()
            .forEach( alertNotificationModel -> {
                NotificationFilterModel notificationFilterModel = extractNotificationInformation(alertNotificationModel);
                notificationFilterMap.add(notificationFilterModel, alertNotificationModel);
            });
```
Is this preffered?

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/BlackDuckDistributionTestAction.java
jamesrichard91: `blackDuckProvider.createStatefulProvider(providerConfiguration)` throws an exception so we can't change this if block to use the Optional.map function.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
jamesrichard91: Also this is not for the patch, this PR is for the current version of Alert.

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
jamesrichard91: This is very helpful when debugging with the performance issues lately. And the logs serve different purposes. The log on line 68 says how many notifications are being processed and line 71  says how many events were created as a result.

workflow/src/main/java/com/synopsys/integration/alert/workflow/scheduled/frequency/ProcessingTask.java
jamesrichard91: Same as above.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/notification/NotificationProcessor.java
bamandel: Oh you're right, it adds it to a completely separate object. Might as well leave it for now.

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/processor/notification/NotificationProcessor.java
jamesrichard91: I will add an if statement to the method that takes the FrequencyType parameter and just have the other method call that. I can't use a Bifunction as you described because the `getMatchingEnabledJobs` in each method have a different number of parameters.

PR #1238, opened 2020-11-02
Fix all field error messages not displaying on distribution test configuration error

ChomickiM:
In 6.3.0 we split up the validation step and had a case where if the name was missing it wouldn't get the field status messages for the other fields. This changes the name validation to also return a fieldstatus and correctly display all of the errors in the UI.

### DIFF
--- a/web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
+++ b/web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
@@ -22,6 +22,7 @@
  */
 package com.synopsys.integration.alert.web.api.job;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -230,18 +231,18 @@ public class JobConfigActions extends AbstractJobResourceActions {
         }
     }
 
-    private ValidationResponseModel validateJobNameUnique(@Nullable UUID currentJobId, JobFieldModel jobFieldModel) {
-        ValidationResponseModel responseModel = ValidationResponseModel.success("Valid name");
+    private List<AlertFieldStatus> validateJobNameUnique(@Nullable UUID currentJobId, JobFieldModel jobFieldModel) {
+        List<AlertFieldStatus> fieldStatuses = new ArrayList<>();
         for (FieldModel fieldModel : jobFieldModel.getFieldModels()) {
-            responseModel = validateJobNameUnique(currentJobId, fieldModel);
-            if (responseModel.hasErrors()) {
-                return responseModel;
+            Optional<AlertFieldStatus> fieldStatus = validateJobNameUnique(currentJobId, fieldModel);
+            if (fieldStatus.isPresent()) {
+                fieldStatuses.add(fieldStatus.get());
             }
         }
-        return responseModel;
+        return fieldStatuses;
     }
 
-    private ValidationResponseModel validateJobNameUnique(@Nullable UUID currentJobId, FieldModel fieldModel) {
+    private Optional<AlertFieldStatus> validateJobNameUnique(@Nullable UUID currentJobId, FieldModel fieldModel) {
         Optional<FieldValueModel> jobNameFieldOptional = fieldModel.getFieldValueModel(ChannelDistributionUIConfig.KEY_NAME);
         String error = "";
         if (jobNameFieldOptional.isPresent()) {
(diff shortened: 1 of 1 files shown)


web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -230,18 +231,18 @@ public JobConfigActions(AuthorizationManager authorizationManager, DescriptorAcc
         }
     }
 
-    private ValidationResponseModel validateJobNameUnique(@Nullable UUID currentJobId, JobFieldModel jobFieldModel) {
-        ValidationResponseModel responseModel = ValidationResponseModel.success("Valid name");
+    private List<AlertFieldStatus> validateJobNameUnique(@Nullable UUID currentJobId, JobFieldModel jobFieldModel) {
+        List<AlertFieldStatus> fieldStatuses = new ArrayList<>();
```
jamesrichard91, line 235: There should only be 1 name per job, so why are we returning a list for this method?

ChomickiM: I think your right, I saw the the for loop and incorrectly thought it should have been returned as a list because it looped through all of the fieldModels. I can change this to return a single AlertFieldStatus.

web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -230,18 +231,17 @@ public JobConfigActions(AuthorizationManager authorizationManager, DescriptorAcc
...
     }
 
-    private ValidationResponseModel validateJobNameUnique(@Nullable UUID currentJobId, JobFieldModel jobFieldModel) {
-        ValidationResponseModel responseModel = ValidationResponseModel.success("Valid name");
+    private Optional<AlertFieldStatus> validateJobNameUnique(@Nullable UUID currentJobId, JobFieldModel jobFieldModel) {
         for (FieldModel fieldModel : jobFieldModel.getFieldModels()) {
-            responseModel = validateJobNameUnique(currentJobId, fieldModel);
-            if (responseModel.hasErrors()) {
-                return responseModel;
+            Optional<AlertFieldStatus> fieldStatus = validateJobNameUnique(currentJobId, fieldModel);
```
psantos1113, line 236: Could you use a stream here to remove the for loop?  You would have to change the method signature of validateJobNameUnique to take a FieldValueModel.

```
Optional<AlertFieldStatus> fieldStatus = jobFieldModel.getFieldModels().stream()
     .filter(fieldModel ->                                            
                     fieldModel.getFieldValueModel(ChannelDistributionUIConfig.KEY_NAME).isPresent())
     .map(fieldModel.getFieldValueModel(ChannelDistributionUIConfig.KEY_NAME))
     .map(fieldValueModel -> validateJobNameUnique(currentJobId, fieldValueModel)
     .orElse(Optional.empty());
```

jamesrichard91 (approved):

web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -230,43 +231,38 @@ public JobConfigActions(AuthorizationManager authorizationManager, DescriptorAcc
...
-                                                 .filter(configurationFieldModel -> (null != configurationFieldModel) && configurationFieldModel.getFieldValue().isPresent())
-                                                 .anyMatch(configurationFieldModel -> jobName.equals(configurationFieldModel.getFieldValue().get()));
-                if (foundDuplicateName) {
-                    error = "A distribution configuration with this name already exists.";
-                }
-            } else {
-                error = "Name cannot be blank.";
+        String jobName = fieldValueModel.getValue().orElse(null);
+        if (StringUtils.isNotBlank(jobName)) {
+            List<ConfigurationJobModel> jobs = configurationAccessor.getAllJobs();
```
gkillough, line 247: This is for 6.3.0 right? This method should not be used in 6.4.0+

ChomickiM: Yes, its for 6.3.0, this is coming from Line 250 in the old code

gkillough: Gotcha. Just make sure when you merge 6.3.0 -> master that you don't use `getAllJobs`.

gkillough (approved):

jamesrichard91 (approved):

PR #1241, opened 2020-11-05
Proposal: descriptor-api sub-project

gkillough:
Not settled on what should live in this sub-project besides channel keys, but right now that's all that it contains. Open to suggestions.


jamesrichard91 (approved):
This looks fine to me, can you run the licenseFormatMain task to update the file headers so it is more clear which files moved?

jamesrichard91 (approved):

psantos1113 (approved):

ChomickiM (approved):

PR #1244, opened 2020-11-06
Changing the DescriptorKey so they can be serializable.

jamesrichard91:
One of the reasons we can't change FieldModel is because the descriptor keys are not serializable. This change makes it so that they can be serialized.


bamandel (commented):
I'm surprised we didn't have it this way initially

bamandel (approved):

gkillough (approved):

psantos1113 (approved):

ChomickiM (approved):

PR #1246, opened 2020-11-09
Fix distribution test configuration to display warnings as successes

ChomickiM:
Due to our changes around fieldStatuses, warnings were coming up in the UI as errors but containing a success message and a successful response. This change makes it so that we will check for a severity of ERROR first to determine if we should throw an error status in the UI and allow warnings to continue as a success status.

As a side note, my save actions were messing with the formatting removing spaces in several areas. I tried to play around with this and fix it but the save actions would overwrite it each time. If anyone thinks its incorrect, please let me know how you modified it in your save actions so I can try to fix it.

### DIFF
--- a/ui/src/main/js/dynamic/DistributionConfiguration.js
+++ b/ui/src/main/js/dynamic/DistributionConfiguration.js
@@ -1,13 +1,15 @@
-import React, {Component} from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import {connect} from 'react-redux';
+import { connect } from 'react-redux';
 import * as FieldModelUtilities from 'util/fieldModelUtilities';
 import * as DescriptorUtilities from 'util/descriptorUtilities';
-import {OPERATIONS} from 'util/descriptorUtilities';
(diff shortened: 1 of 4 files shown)


ui/src/main/js/dynamic/DistributionConfiguration.js
```diff
@@ -91,15 +89,15 @@ class DistributionConfiguration extends Component {
             }
         }
 
-        const { channelConfig, currentChannel, currentProvider } = this.state;
+        const {channelConfig, currentChannel, currentProvider} = this.state;
```
gkillough, line 92: It looks like your JS formatting settings are different than the rest of ours.

gkillough: Nothing in this file seems to have changed other than whitespace.

gkillough (approved):
This seems fine, but let's ensure that we're all using the same formatting and remove the whitespace changes from this PR.

jamesrichard91 (approved):

psantos1113 (approved):

PR #1249, opened 2020-11-09
Remove default value handling from DefaultNotificationAccessor

JakeMathews:
# Remove default value handling from DefaultNotificationAccessor

Fixes IALERT-2008

In the method `DefaultNotificationAccessor::getPageRequestForNotifications` the paging parameters are `@Nullable`, this should not be the case and default value handling should happen at a higher level.

- Move default value handling higher into the API
- Change default page size from `Integer.MAX_VALUE` to `10`. Uses the constant defined in `AuditEntryPageModel`.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
@@ -24,11 +24,13 @@ package com.synopsys.integration.alert.common.rest.model;
 
 import java.util.List;
 
+import com.synopsys.integration.alert.common.rest.api.ReadPageController;
+
 import net.minidev.json.annotate.JsonIgnore;
 
 public class AlertPagedModel<M extends AlertSerializableModel> extends AlertSerializableModel {
-    public static final Integer DEFAULT_PAGE_NUMBER = 0;
-    public static final Integer DEFAULT_PAGE_SIZE = 10;
+    public static final Integer DEFAULT_PAGE_NUMBER = Integer.valueOf(ReadPageController.DEFAULT_PAGE_NUMBER);
+    public static final Integer DEFAULT_PAGE_SIZE = Integer.valueOf(ReadPageController.DEFAULT_PAGE_SIZE);
 
     // FIXME we should use terminology based on "offset" and "limit" which are standard REST API paging terms
     private final int totalPages;
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
@@ -29,8 +29,6 @@ import net.minidev.json.annotate.JsonIgnore;
 public class AlertPagedModel<M extends AlertSerializableModel> extends AlertSerializableModel {
     public static final Integer DEFAULT_PAGE_NUMBER = 0;
     public static final Integer DEFAULT_PAGE_SIZE = 10;
-    public static final String DEFAULT_PAGE_NUMBER_STRING = "0";
-    public static final String DEFAULT_PAGE_SIZE_STRING = "10";
 
     // FIXME we should use terminology based on "offset" and "limit" which are standard REST API paging terms
     private final int totalPages;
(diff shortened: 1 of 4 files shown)


component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActions.java
```diff
@@ -91,7 +92,9 @@ public AuditEntryActions(
         if (!authorizationManager.hasReadPermission(ConfigContextEnum.GLOBAL, descriptorKey)) {
             return new ActionResponse<>(HttpStatus.FORBIDDEN, ActionResponse.FORBIDDEN_MESSAGE);
         }
-        AuditEntryPageModel pagedRestModel = auditAccessor.getPageOfAuditEntries(pageNumber, pageSize, searchTerm, sortField, sortOrder, onlyShowSentNotifications, auditAccessor::convertToAuditEntryModelFromNotification);
+        Integer page = ObjectUtils.defaultIfNull(pageNumber, 0);
+        Integer size = ObjectUtils.defaultIfNull(pageSize, 100);
```
gkillough, line 96: I think the class that `AuditEntryPageModel` extends might have constants that could be used for the default values.

gkillough (approved):

jamesrichard91 (approved):

component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActions.java
bamandel: If those constants don't exist we should at least make some in this class

bamandel (approved):

ChomickiM (approved):

psantos1113 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
```diff
@@ -27,8 +27,10 @@
 import net.minidev.json.annotate.JsonIgnore;
 
 public class AlertPagedModel<M extends AlertSerializableModel> extends AlertSerializableModel {
-    public static final String DEFAULT_PAGE_NUMBER = "0";
-    public static final String DEFAULT_PAGE_SIZE = "10";
+    public static final Integer DEFAULT_PAGE_NUMBER = 0;
+    public static final Integer DEFAULT_PAGE_SIZE = 10;
+    public static final String DEFAULT_PAGE_NUMBER_STRING = "0";
+    public static final String DEFAULT_PAGE_SIZE_STRING = "10";
```
JakeMathews, line 33: Spring requires a String constant in `ReadPageController`. We want to use the same defaults, but as an Integer. Should we convert the String to an Integer at runtime? Do what I've done here? Something else?

component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActions.java
JakeMathews: I am now using the constants in AlertPagedModel. But that introduces a new issue https://github.com/blackducksoftware/blackduck-alert/pull/1249#discussion_r520829142

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
gkillough:
```java
    public static final Integer DEFAULT_PAGE_NUMBER = 0;
    public static final Integer DEFAULT_PAGE_SIZE = 10;
    public static final String DEFAULT_PAGE_NUMBER_STRING = DEFAULT_PAGE_NUMBER.toString();
    public static final String DEFAULT_PAGE_SIZE_STRING = DEFAULT_PAGE_SIZE.toString();
```
based on https://stackoverflow.com/a/654068/6921621

JakeMathews: Then it is no longer a constant. 
Screen Shot 2020-11-10 at 3 17 12 PM

gkillough: Alright, in that case maybe move the string constants into that interface and leave them as literals. Leave the integers in the paged model.

jamesrichard91: You could reverse it, leave the static Strings as literals and have the static integers based on those>
```
public static final Integer DEFAULT_PAGE_NUMBER = Integer.valueOf(ReadPageController. DEFAULT_PAGE_NUMBER);
public static final Integer DEFAULT_PAGE_SIZE =  Integer.valueOf(ReadPageController. DEFAULT_PAGE_SIZE);
```

JakeMathews: I was hesitant to do the String parsing at runtime, but I think that's better than declaring the constants twice.

PR #1250, opened 2020-11-09
On demand data

jamesrichard91:
- Retrieve the Black Duck data on demand, instead of syncing the Black Duck data with the Alert database

* Leave the old DefaultProviderDataAccessor in place in case we need to switch back to it
* Add a new BlackDuckProviderDataAccessor to replace the DefaultProviderDataAccessor
* The BlackDuckProviderDataAccessor will retrieve the Project and User data from Black Duck as needed

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
@@ -41,7 +41,6 @@ import org.springframework.transaction.annotation.Transactional;
 
 import com.synopsys.integration.alert.common.exception.AlertDatabaseConstraintException;
 import com.synopsys.integration.alert.common.persistence.accessor.ConfigurationAccessor;
-import com.synopsys.integration.alert.common.persistence.accessor.ProviderDataAccessor;
 import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
 import com.synopsys.integration.alert.common.persistence.model.ProviderProject;
 import com.synopsys.integration.alert.common.persistence.model.ProviderUserModel;
@@ -54,7 +53,8 @@ import com.synopsys.integration.alert.database.provider.user.ProviderUserReposit
 
 @Component
 @Transactional
-public class DefaultProviderDataAccessor implements ProviderDataAccessor {
+public class DefaultProviderDataAccessor {
+    //implements ProviderDataAccessor {
     public static final int MAX_DESCRIPTION_LENGTH = 250;
     public static final int MAX_PROJECT_NAME_LENGTH = 507;
 
@@ -73,7 +73,7 @@ public class DefaultProviderDataAccessor implements ProviderDataAccessor {
         this.configurationAccessor = configurationAccessor;
     }
 
-    @Override
+    //    @Override
     @Transactional(readOnly = true, isolation = Isolation.READ_COMMITTED)
     public List<ProviderProject> getProjectsByProviderConfigName(String providerConfigName) {
         try {
@@ -91,7 +91,7 @@ public class DefaultProviderDataAccessor implements ProviderDataAccessor {
         return List.of();
     }
 
-    @Override
+    //    @Override
     @Transactional(readOnly = true, isolation = Isolation.READ_COMMITTED)
     public List<ProviderProject> getProjectsByProviderConfigId(Long providerConfigId) {
         return providerProjectRepository.findByProviderConfigId(providerConfigId)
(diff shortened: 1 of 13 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -54,7 +53,8 @@
 
 @Component
 @Transactional
-public class DefaultProviderDataAccessor implements ProviderDataAccessor {
+public class DefaultProviderDataAccessor {
```
psantos1113, line 56: Can this class be deleted since the BlackDuckProviderDataAccessor now exists.  And in the past DefaultProviderDataAccessor was the only ProviderDataAccessor implementation once we removed support for Polaris.

channel/src/main/java/com/synopsys/integration/alert/channel/email/EmailAddressHandler.java
```diff
@@ -93,7 +93,7 @@ public FieldUtility updateEmailAddresses(Long providerConfigId, MessageContentGr
         return new FieldUtility(fieldMap);
     }
 
-    public Set<String> getEmailAddressesForProject(ProviderProject project, boolean projectOwnerOnly) {
+    public Set<String> getEmailAddressesForProject(Long providerConfigID, ProviderProject project, boolean projectOwnerOnly) {
```
gkillough, line 96: `providerConfigID` and `providerConfigId` are both used in this class. We should stick to one capitalization strategy.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
```diff
@@ -73,7 +73,7 @@ public DefaultProviderDataAccessor(ProviderProjectRepository providerProjectRepo
         this.configurationAccessor = configurationAccessor;
     }
 
-    @Override
+    //    @Override
```
gkillough, line 76: We should delete all of these commented annotations and worry about re-adding them if we ever want to implement the interface again.

gkillough (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
jamesrichard91: We want to keep this class around along with the provider data tables. The changes I made to this class for now should render this class unused until we decide to delete it.

gkillough: I know, I was just pointing this out because Intellij can re-add those annotations in ~1 second.

psantos1113 (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProviderDataAccessor.java
gkillough: I do think we want to add `@Deprecated` here just to make sure people don't try to use this whether in tests or production.

provider/src/main/java/com/synopsys/integration/alert/database/api/BlackDuckProviderDataAccessor.java
```diff
@@ -0,0 +1,220 @@
...
+        BlackDuckHttpClient blackDuckHttpClient = properties.createBlackDuckHttpClient(logger);
+        BlackDuckServicesFactory blackDuckServicesFactory = properties.createBlackDuckServicesFactory(blackDuckHttpClient, logger);
+        ProjectService projectService = blackDuckServicesFactory.createProjectService();
+        List<ProjectView> allProjects = projectService.getAllProjects();
+        return convertBlackDuckProjects(allProjects, blackDuckServicesFactory.getBlackDuckService());
+    }
+
+    @Override
+    public void deleteProjects(Collection<ProviderProject> providerProjects) {
+        //ignored since we are not using the database
```
bamandel, line 109: For API's we have a response NOT_IMPLEMENTED. is there some similar runtime exception we could throw here?

provider/src/main/java/com/synopsys/integration/alert/database/api/BlackDuckProviderDataAccessor.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public Set<String> getEmailAddressesForProjectHref(Long providerConfigId, String projectHref) {
```
bamandel, line 113: All other methods return a List. Why is this one specifically a set? I feel we should return all lists or all sets to be consistent

provider/src/main/java/com/synopsys/integration/alert/database/api/BlackDuckProviderDataAccessor.java
jamesrichard91: We don't want this to throw an exception. We want this class to be used in place of the DefaultProviderDataAccessor so that it will have minimal impact

bamandel (approved):

ChomickiM (approved):

PR #1253, opened 2020-11-11
Fix Swagger

gkillough:


web/src/main/java/com/synopsys/integration/alert/web/documentation/SwaggerConfiguration.java
```diff
@@ -56,6 +58,14 @@ public Docket api() {
                    .apiInfo(apiEndpointInfo());
     }
 
+    @Bean
+    public UiConfiguration alertSwaggerUiConfiguration() {
+        return UiConfigurationBuilder
+                   .builder()
+                   .supportedSubmitMethods(new String[] { "get" })
```
gkillough, line 65: This only works if it is lower-case.

bamandel: Can we turn it into a constant with a comment above it stating this fact?

bamandel (approved):

jamesrichard91 (approved):

ChomickiM (approved):

PR #1254, opened 2020-11-11
Updating the job validation to make sure the fields are properly validated

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/field/validation/FieldValidationUtility.java
```diff
@@ -47,13 +47,13 @@
     }
 
     public List<AlertFieldStatus> validateConfig(Map<String, ConfigField> fieldKeyToConfigField, Collection<FieldModel> fieldModels) {
-        logger.debug("Begin validating fields in configuration field model.");
+        logger.trace("Begin validating fields in configuration field model.");
```
jamesrichard91, line 50: Changed these logs to `trace` level because they have not been useful for debugging and create a lot of noise at `debug` level

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/config/ui/ProviderDistributionUIConfig.java
```diff
@@ -163,6 +163,17 @@ private ValidationResult validateConfigExists(FieldValueModel fieldToValidate, F
         }
     }
 
+    private ValidationResult validateFilterByProject(FieldValueModel fieldToValidate, FieldModel fieldModel) {
```
jamesrichard91, line 166: This makes sure that if the User has checked `filterByProject` that they have to specify either a project name pattern or select a project.

jamesrichard91: There is an odd difference between validation of new jobs and existing jobs. The UI will send empty fields to the back end and they get validated but editing a job does not do the same. If the project name pattern or configured projects have no values, they are not included in the payload for the validate request.

ChomickiM (approved):

gkillough (approved):

PR #1258, opened 2020-11-13
Jira proxy usage, updating int-jira-common

jamesrichard91:
Making sure the Jira connections are using the proxy settings. 
Updating the int-jira-common version to the latest version.

### DIFF
--- a/channel/build.gradle
+++ b/channel/build.gradle
@@ -3,6 +3,6 @@ ext.moduleName = 'com.synopsys.integration.alert.channel'
 dependencies {
     implementation project(':alert-common')
     implementation project(':azure-boards-common')
-    implementation 'com.synopsys.integration:int-jira-common:1.1.0'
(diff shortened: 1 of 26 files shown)


channel/build.gradle
```diff
@@ -3,6 +3,6 @@ ext.moduleName = 'com.synopsys.integration.alert.channel'
 dependencies {
     implementation project(':alert-common')
     implementation project(':azure-boards-common')
-    implementation 'com.synopsys.integration:int-jira-common:1.1.0'
+    implementation 'com.synopsys.integration:int-jira-common:2.0.1-SNAPSHOT'
```
jamesrichard91, line 6: In the process of releasing int-jira-common with the ProxyInfo change to the Builders. This will be updated as soon as the version is released.

bamandel (approved):

gkillough (approved):

ChomickiM (approved):

PR #1259, opened 2020-11-13
BlackDuck Project Paging

gkillough:
Serve paged responses to the UI for projects in table-select.

### DIFF
--- a/ui/src/main/js/field/input/TableSelectInput.js
+++ b/ui/src/main/js/field/input/TableSelectInput.js
@@ -86,64 +86,18 @@ class TableSelectInput extends Component {
         };
     }
 
-    // FIXME componentWillMount is deprecated, use the constructor or componentDidMount to initialize this
-    // componentDidMount() {
-    //     const { currentPage, currentPageSize } = this.state;
-    //     const { value } = this.props;
-    //     if (value && value.length > 0) {
-    //         this.retrieveTableData(currentPage, currentPageSize).then(() => {
-    //             this.updateSelectedValues();
-    //         });
-    //     }
-    // }
-
-    // componentDidUpdate(prevProps, prevState, snapshot) {
-    //     const { value } = this.props;
-    //     const didPrevHaveValues = prevProps.value && prevProps.value.length === 0;
-    //     const doesCurrentHaveValues = value && value.length > 0;
-    //     const isAnythingSelected = this.state.selectedData.length === 0;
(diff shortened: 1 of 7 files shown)


web/src/main/java/com/synopsys/integration/alert/web/api/provider/project/ProviderProjectCustomFunctionAction.java
```diff
@@ -60,23 +67,48 @@ public ProviderProjectCustomFunctionAction(AuthorizationManager authorizationMan
 
     @Override
     public ActionResponse<ProviderProjectOptions> createActionResponse(FieldModel fieldModel, HttpServletContentWrapper servletContentWrapper) {
+        HttpServletRequest httpRequest = servletContentWrapper.getHttpRequest();
+        Map<String, String[]> parameterMap = httpRequest.getParameterMap();
+
+        int pageNumber = extractPagingParam(parameterMap, "pageNumber", 0);
+        int pageSize = extractPagingParam(parameterMap, "pageSize", 10);
```
jamesrichard91, line 74: We don't have to change this in this version of Alert but I think this is a good argument for getting rid of the dynamic endpoints and just create controllers/endpoints as we need. Having to extract the parameters this way is not appealing.

gkillough: Agreed. I think it will be part of a larger effort to rewrite the Jobs UI to have tightly coupled channel endpoints.

ui/src/main/js/field/input/TableSelectInput.js
```diff
@@ -72,51 +75,30 @@ class TableSelectInput extends Component {
...
         selectedData.push(...value);
-        const keyColumnHeader = this.props.columns.find((column) => column.isKey).header;
         const convertedValues = selectedData.map((selected) => {
-            const columnContainsValue = data.map((dataValue) => dataValue[keyColumnHeader])
-                .includes(selected);
             return {
                 label: selected,
                 value: selected,
-                missing: !columnContainsValue
+                missing: false
```
gkillough, line 101: We cannot support "missing" data with paging because the data could be in a different page that has yet to be retrieved.

jamesrichard91 (approved):

psantos1113 (approved):

PR #1260, opened 2020-11-13
Fix character missing from Azure workitem type POST request

ChomickiM:


gkillough (approved):
Nice find. Looks like this character was dropped when someone converted from the old "spec templates" to `String.format`

jamesrichard91 (approved):

PR #1264, opened 2020-11-16
Change ConfigurationFieldModelConverter to throw a Runtime exception

ChomickiM:
Per IALERT-2011, we want to change ConfigurationFieldModelConverter.getDescriptorName to throw a runtime exception rather than an AlertDatabaseConstraintException. We do not want this exception because if we trace it up to where the ConfigurationModel that is being passed in comes from, it will come from a database read where an option check has already been performed (in DefaultDescriptorGlobalConfigUtility).

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/ConfigurationFieldModelConverter.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/ConfigurationFieldModelConverter.java
@@ -168,16 +168,14 @@ public class ConfigurationFieldModelConverter {
     }
 
     private String getDescriptorName(ConfigurationModel configurationModel) {
-        String descriptorName;
         try {
-            descriptorName = descriptorAccessor.getRegisteredDescriptorById(configurationModel.getDescriptorId())
-                                 .map(RegisteredDescriptorModel::getName)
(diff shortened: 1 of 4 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/util/ConfigurationFieldModelConverter.java
```diff
@@ -161,11 +167,17 @@ private void populateAndSecureFields(ConfigurationFieldModel fieldModel, Map<Str
...
 
-    private String getDescriptorName(ConfigurationModel configurationModel) throws AlertDatabaseConstraintException {
-        return descriptorAccessor.getRegisteredDescriptorById(configurationModel.getDescriptorId())
-                   .map(RegisteredDescriptorModel::getName)
-                   // FIXME this is the wrong type of exception to throw here
-                   .orElseThrow(() -> new AlertDatabaseConstraintException("Expected to find registered descriptor but none was found."));
+    private String getDescriptorName(ConfigurationModel configurationModel) {
+        String descriptorName;
+        try {
+            descriptorName = descriptorAccessor.getRegisteredDescriptorById(configurationModel.getDescriptorId())
```
gkillough, line 173: I think we might just be able to return from here.

jamesrichard91 (approved):

gkillough (approved):

psantos1113 (approved):

PR #1265, opened 2020-11-16
Additional Email Address Paging

gkillough:


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
```diff
@@ -195,26 +205,35 @@ public void updateProjectAndUserData(Long providerConfigId, Map<ProviderProject,
...
+        Predicate<UserView> searchFilter = userView -> StringUtils.isNotBlank(userView.getEmail());
+        if (StringUtils.isNotBlank(searchTerm)) {
+            searchFilter = searchFilter.and(userView -> StringUtils.containsIgnoreCase(userView.getEmail(), searchTerm));
+        }
+
+        BlackDuckPageResponse<UserView> pageOfViews = retrieveBlackDuckPageResponse(blackDuckServicesFactory, ApiDiscovery.USERS_LINK_RESPONSE, pageNumber, pageSize, searchFilter);
+
+        List<ProviderUserModel> foundProjects = pageOfViews.getItems()
+                                                    .stream()
+                                                    .map(UserViewV4::getEmail)
```
jamesrichard91, line 233: Can we not use UserView here because it doesn't declare the getEmail method?

gkillough: This was a typo that I have now fixed. It still worked though because they both deserialized to the same thing.

ChomickiM (approved):

PR #1268, opened 2020-11-18
test-common sub-project

gkillough:
A sub-project to commonize test utilities and classes so that other sub-projects can leverage them for their tests.


test-common/build.gradle
```diff
@@ -0,0 +1,25 @@
+plugins {
+    id 'java'
+}
+
+group 'com.synopsys.integration'
+version '6.4.0-SNAPSHOT'
```
jamesrichard91, line 6: Are these needed here? Shouldn't the group and version be set by the main project, just like the other sub-projects?

gkillough: Yep. This was generated; I'm gonna clean it up right now.

psantos1113 (approved):

jamesrichard91 (approved):

PR #1274, opened 2020-11-24
Request all job-audit data at once

gkillough:
This is to minimize network latency and remove the odd-looking loading of one row at a time.


component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActions.java
```diff
@@ -123,6 +125,22 @@ public AuditEntryActions(
...
+            return new ActionResponse<>(HttpStatus.FORBIDDEN, ActionResponse.FORBIDDEN_MESSAGE);
+        }
+
+        List<UUID> jobIds = queryRequestModel.getJobIds();
+        for (UUID jobId : jobIds) {
+            if (null == jobId) {
+                return new ActionResponse<>(HttpStatus.BAD_REQUEST, "The field 'jobIds' cannot contain null values");
+            }
+        }
+
```
jamesrichard91, line 139: You could do this slightly differently but it shouldn't make a difference.
```
Optional<UUID> nullJobId = jobIds.stream()
 .filter(Objects::isNull)
 .findFirst();
if(nullJobId.isPresent()){
	return new ActionResponse<>(HttpStatus.BAD_REQUEST, "The field 'jobIds' cannot contain null values");
}
```

jamesrichard91 (approved):

PR #1275, opened 2020-11-24
Jm job accessor refactor

JakeMathews:
#### Create new `JobAccessor` to aid in the transition from `ConfigurationJobModel` to `DistributionJobModel`

* Create new `JobAccessor` interface: `JobAccessorV2`
* Create new `JobAccessorV2` implementation: `StaticJobAccessorV2`
* Don't carry over `JobAccessor::getAllJobs` or `JobAccessor::getJobsByFrequency`

Classes that currently use `JobAccessor` will be incrementally updated in future PR's to utilize these new classes.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/job/DistributionJobRequestModel.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/job/DistributionJobRequestModel.java
@@ -1,0 +1,112 @@
+package com.synopsys.integration.alert.common.persistence.model.job;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.jetbrains.annotations.Nullable;
+
+import com.synopsys.integration.alert.common.enumeration.FrequencyType;
+import com.synopsys.integration.alert.common.enumeration.ProcessingType;
+import com.synopsys.integration.alert.common.persistence.model.job.details.DistributionJobDetailsModel;
+
+public class DistributionJobRequestModel {
+    private final boolean enabled;
+    private final String name;
+    private final FrequencyType distributionFrequency;
+    private final ProcessingType processingType;
+    private final String channelDescriptorName;
+
+    // Black Duck fields will be common as long as it is the only provider
+    private final Long blackDuckGlobalConfigId;
+    private final boolean filterByProject;
+    @Nullable
+    private final String projectNamePattern;
+    private final List<String> notificationTypes;
+    private final List<String> projectFilterProjectNames;
+    private final List<String> policyFilterPolicyNames;
+    private final List<String> vulnerabilityFilterSeverityNames;
+
+    private final DistributionJobDetailsModel distributionJobDetails;
+
+    public DistributionJobRequestModel(
+        boolean enabled,
+        String name,
+        FrequencyType distributionFrequency,
+        ProcessingType processingType,
+        String channelDescriptorName,
+        Long blackDuckGlobalConfigId,
+        boolean filterByProject,
+        @Nullable String projectNamePattern,
+        List<String> notificationTypes,
+        List<String> projectFilterProjectNames,
+        List<String> policyFilterPolicyNames,
+        List<String> vulnerabilityFilterSeverityNames,
+        DistributionJobDetailsModel distributionJobDetails
+    ) {
+        this.enabled = enabled;
+        this.name = name;
+        this.distributionFrequency = distributionFrequency;
+        this.processingType = processingType;
+        this.channelDescriptorName = channelDescriptorName;
+        this.blackDuckGlobalConfigId = blackDuckGlobalConfigId;
+        this.filterByProject = filterByProject;
+        this.projectNamePattern = projectNamePattern;
+        this.notificationTypes = notificationTypes;
+        this.projectFilterProjectNames = projectFilterProjectNames;
+        this.policyFilterPolicyNames = policyFilterPolicyNames;
+        this.vulnerabilityFilterSeverityNames = vulnerabilityFilterSeverityNames;
+        this.distributionJobDetails = distributionJobDetails;
+    }
+
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public FrequencyType getDistributionFrequency() {
+        return distributionFrequency;
+    }
+
+    public ProcessingType getProcessingType() {
+        return processingType;
+    }
+
+    public String getChannelDescriptorName() {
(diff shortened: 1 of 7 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -0,0 +1,225 @@
...
+
+    @Override
+    @Transactional
+    public void deleteJob(UUID jobId) {
+        distributionJobRepository.deleteById(jobId);
+    }
+
+    private DistributionJobModel convertToDistributionJobModel(DistributionJobEntity jobEntity) {
+        ConfigurationJobModel configurationJobModel = convertToConfigurationJobModel(jobEntity);
+        Map<String, ConfigurationFieldModel> fields = configurationJobModel.getFieldUtility().getFields();
```
gkillough, line 183: We should be able to cut out these two steps and go more directly to the `DistributionJobModel`.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -0,0 +1,225 @@
...
+        return JobConfigurationModelFieldExtractorUtils.convertToDistributionJobModel(configurationJobModel.getJobId(), fields, jobEntity.getCreatedAt(), jobEntity.getLastUpdated());
+    }
+
+    private ConfigurationJobModel convertToConfigurationJobModel(DistributionJobEntity jobEntity) {
```
jamesrichard91, line 187: This is only used in the `convertToDistributionJobModel(DistributionJobEntity jobEntity)` method, and it looks like the only pieces we use from ConfigurationJobModel are the Job Id and the fields map. So why not change this method to just return the fields map, since the Job Id can be taken from the DistributionJobEntity?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -0,0 +1,225 @@
...
+    @Override
+    @Transactional
+    public DistributionJobModel updateJob(UUID jobId, Collection<String> descriptorNames, Collection<ConfigurationFieldModel> configuredFields) throws AlertDatabaseConstraintException {
+        DistributionJobEntity jobEntity = distributionJobRepository.findById(jobId)
+                                              .orElseThrow(() -> new AlertDatabaseConstraintException(String.format("No job exists with the id [%s]", jobId.toString())));
+        OffsetDateTime createdAt = jobEntity.getCreatedAt();
+
+        deleteJob(jobId);
+        return createJobWithId(jobId, configuredFields, createdAt, DateUtils.createCurrentDateTimestamp());
+    }
```
gkillough, line 173: I think we want to take in a different model rather than the "old" way of `Collection<String> descriptorNames` and `Collection<ConfigurationFieldModel> configuredFields`. Something along the lines of `DistributionJobRequestModel` that has a subset of the fields of `DistributionJobModel`. This model would not have `jobId`, `createdAt` or `lastUpdated` fields.

gkillough (changes requested):
I think the create and update methods should require more strongly-typed arguments.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
JakeMathews: This method is now removed.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/job/DistributionJobRequestModel.java
```diff
@@ -0,0 +1,112 @@
+package com.synopsys.integration.alert.common.persistence.model.job;
```
gkillough, line 1: Don't forget to generate the license headers before merging.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -0,0 +1,316 @@
...
+
+    private List<DistributionJobModel> getMatchingEnabledJobs(Supplier<List<DistributionJobEntity>> getJobs) {
+        // TODO change this to return a page of jobs
+        List<DistributionJobEntity> matchingEnabledJob = getJobs.get();
+        return matchingEnabledJob
+                   .stream()
+                   .map(this::convertToDistributionJobModel)
+                   .collect(Collectors.toList());
+    }
```
gkillough, line 181: Let's bump this private method down to below the last public method.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -0,0 +1,316 @@
...
+
+    @Override
+    public List<DistributionJobModel> getMatchingEnabledJobs(FrequencyType frequency, Long providerConfigId, NotificationType notificationType) {
+        // TODO change this to return a page of jobs
+        return getMatchingEnabledJobs(() -> distributionJobRepository.findMatchingEnabledJob(frequency.name(), providerConfigId, notificationType.name()));
+    }
+
+    @Override
+    public List<DistributionJobModel> getMatchingEnabledJobs(Long providerConfigId, NotificationType notificationType) {
```
gkillough, line 169: We can add the `@Transactional(readOnly = true)` to these methods.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/job/DistributionJobRequestModel.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import org.jetbrains.annotations.Nullable;
+
+import com.synopsys.integration.alert.common.enumeration.FrequencyType;
+import com.synopsys.integration.alert.common.enumeration.ProcessingType;
+import com.synopsys.integration.alert.common.persistence.model.job.details.DistributionJobDetailsModel;
+
+public class DistributionJobRequestModel {
```
gkillough, line 34: I hate to do this, but I forgot to mention it might be nice to have an abstract class that this and `DistributionJobModel` both extend so we can abstract some of the common fields.

gkillough (approved):

psantos1113 (approved):

PR #1283, opened 2020-12-02
Remove project-owner email field

gkillough:


ChomickiM (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/JobProviderProjectFieldModel.java
```diff
@@ -25,15 +25,13 @@
 public class JobProviderProjectFieldModel extends AlertSerializableModel {
     private String name;
     private String href;
-    private String projectOwnerEmail;
 
     public JobProviderProjectFieldModel() {
     }
 
-    public JobProviderProjectFieldModel(String name, String href, String projectOwnerEmail) {
+    public JobProviderProjectFieldModel(String name, String href) {
```
bamandel, line 32: I know this isn't your change, but all other models with these 2 fields are prefixed with Blackduck while this one is JobProvider. Might be something we want to look into making more clear.

bamandel (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/JobProviderProjectFieldModel.java
gkillough: Yeah, the whole Job REST API is being redone in 6.5.0, so this is really just a placeholder for now.

gkillough: The other thing to keep in mind is that this is the model displayed in Swagger. So it will be consistent with our current provider-centric API.

JakeMathews (approved):

PR #1284, opened 2020-12-03
Truncate provider project descriptions

gkillough:
We used to do this when the projects were synced. Long descriptions bloat the table in the UI.

### DIFF
--- a/provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
+++ b/provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
@@ -32,6 +32,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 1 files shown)


jamesrichard91 (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
```diff
@@ -314,4 +319,12 @@ private int computeTotalCount(BlackDuckPageResponse<?> blackDuckPageResponse, in
         return blackDuckResponsesTransformer.getSomeMatchingResponses(pagedRequest, responseClass, searchFilter, pagedRequest.getLimit());
     }
 
+    private String truncateDescription(@Nullable String originalDescription) {
+        String trimmedDescription = StringUtils.trimToEmpty(originalDescription);
+        if (trimmedDescription.length() > PROJECT_DESCRIPTION_MAX_CHARS) {
+            return StringUtils.truncate(trimmedDescription, PROJECT_DESCRIPTION_MAX_CHARS) + ". . .";
```
bamandel, line 325: Would it be beneficial to truncate to a few under 256 then add the new content? Not sure the most efficient way of storing data in our current DB.

bamandel (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
gkillough: We're really just trying to minimize UI bloat here whereas before we were writing to the DB, so it needed to be more precise.

PR #1285, opened 2020-12-03
Refactored PhoneHomeTask to use new JobAccessor

JakeMathews:


src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeTask.java
```diff
@@ -182,26 +181,24 @@ private PhoneHomeService createPhoneHomeService(ExecutorService phoneHomeExecuto
...
-
-                if (StringUtils.isBlank(channelName) || StringUtils.isBlank(providerName)) {
-                    // We want to specifically get the channel configuration here and the only way to determine that is if it has these fields.
-                    continue;
-                }
+        for (DistributionJobModel job : jobs) {
+            String channelName = job.getChannelDescriptorName();
+            String providerName = new BlackDuckProviderKey().getUniversalKey();
+
+            if (StringUtils.isBlank(channelName)) {
```
JakeMathews, line 191: I was unsure if this is still necessary.

gkillough: The database does not allow `null` values for the channel descriptor name, so I'm inclined to think that this being blank would be a bug (if we are correctly validating the data that gets saved). So I think this can be removed.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeTask.java
```diff
@@ -182,26 +181,24 @@ private PhoneHomeService createPhoneHomeService(ExecutorService phoneHomeExecuto
...
-                String channelName = configuration.getField(ChannelDistributionUIConfig.KEY_CHANNEL_NAME).flatMap(ConfigurationFieldModel::getFieldValue).orElse("");
-                String providerName = configuration.getField(ChannelDistributionUIConfig.KEY_PROVIDER_NAME).flatMap(ConfigurationFieldModel::getFieldValue).orElse("");
-
-                if (StringUtils.isBlank(channelName) || StringUtils.isBlank(providerName)) {
-                    // We want to specifically get the channel configuration here and the only way to determine that is if it has these fields.
-                    continue;
-                }
+        for (DistributionJobModel job : jobs) {
+            String channelName = job.getChannelDescriptorName();
+            String providerName = new BlackDuckProviderKey().getUniversalKey();
```
gkillough, line 189: Since this class is a component, you can autowire `BlackDuckProviderKey`.

JakeMathews: I went vague an autowired a `ProviderKey` if that's alright.

gkillough: Yeah, that's fine. If we ever do need to refactor to multiple providers, it will be with `ProviderKey`.

src/main/java/com/synopsys/integration/alert/workflow/scheduled/PhoneHomeTask.java
```diff
@@ -182,26 +182,19 @@ private PhoneHomeService createPhoneHomeService(ExecutorService phoneHomeExecuto
         return channelMetadata.toArray(String[]::new);
     }
 
-    private Set<String> retrieveChannelMetadataForJobs(List<ConfigurationJobModel> jobs) {
+    private Set<String> retrieveChannelMetadataForJobs(List<DistributionJobModel> jobs) {
         Map<String, Integer> createdDistributions = new HashMap<>();
         String successKeyPart = "::Successes";
```
gkillough, line 187: While you're in here, can you make this a constant?

gkillough (approved):

PR #1286, opened 2020-12-04
Update NotificationProcessor and DistributionEvents to use new Accessor

gkillough:
- `DistribuionEvents` are used in a lot of places
- Our previous API was terrible because rather than asking for the fields it required to run, all fields were stuffed into a `FieldUtility` and then passed around


alert-common/src/main/java/com/synopsys/integration/alert/common/event/DistributionEvent.java
```diff
@@ -24,30 +24,43 @@
...
 import com.synopsys.integration.alert.common.message.model.MessageContentGroup;
-import com.synopsys.integration.alert.common.persistence.accessor.FieldUtility;
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
+import com.synopsys.integration.alert.common.persistence.model.job.DistributionJobModel;
 
 public class DistributionEvent extends ContentEvent {
-    private static final long serialVersionUID = -2067819359358348281L;
-    private final FieldUtility fieldUtility;
-    private final String configId;
+    private static final long serialVersionUID = -7858733753649257748L;
```
gkillough, line 38: I would love to know if this is required, or if we can remove it.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraTestIssueRequestCreator.java
```diff
@@ -51,20 +52,29 @@
 
 public class JiraTestIssueRequestCreator implements TestIssueRequestCreator {
     private final Logger logger = LoggerFactory.getLogger(JiraTestIssueRequestCreator.class);
-    private final FieldUtility fieldUtility;
+
+    private final String customTopic;
+    private final String customMessage;
     private final JiraMessageParser jiraMessageParser;
 
     public JiraTestIssueRequestCreator(FieldUtility fieldUtility, JiraMessageParser jiraMessageParser) {
```
JakeMathews, line 60: Do we prefer a constructor for this over a static method?

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudContextBuilder.java
```diff
@@ -78,4 +83,23 @@ public JiraCloudContext build(FieldUtility fieldUtility) {
         return new JiraCloudContext(jiraCloudPropertiesFactory.createJiraProperties(fieldUtility), createIssueConfig(fieldUtility));
     }
 
+    @Override
+    public JiraCloudContext build(ConfigurationModel globalConfig, DistributionJobModel jobModel) {
+        FieldUtility globalFieldUtility = new FieldUtility(globalConfig.getCopyOfKeyToFieldMap());
+        JiraCloudProperties jiraProperties = jiraCloudPropertiesFactory.createJiraProperties(globalFieldUtility);
```
JakeMathews, line 89: Is including the properties from `FieldUtlity` for temporary compatibility, or are there fields the context needs that aren't in the `JiraCloudJobDetailsModel`?

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
```diff
@@ -55,8 +57,9 @@ public JiraServerChannel(Gson gson, JiraServerChannelKey descriptorKey, AuditAcc
 
     @Override
     protected IssueTrackerContext getIssueTrackerContext(DistributionEvent event) {
-        FieldUtility fieldUtility = event.getFieldUtility();
-        return jiraServerContextBuilder.build(fieldUtility);
+        ConfigurationModel globalConfig = event.getChannelGlobalConfig()
+                                              .orElseThrow(() -> new AlertRuntimeException(new AlertConfigurationException("Missing Jira Server global configuration")));
```
jamesrichard91, line 61: Why do we have the 2 exceptions here, rather than just the AlertRuntimeException?

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudContextBuilder.java
gkillough: The former. The Jira channels share several common interfaces and I was trying to simplify for compatibility.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraTestIssueRequestCreator.java
gkillough: No. This is definitely an area that needs improvement, but changing this existing constructor would have only added to the already-excessive scope-creep.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
gkillough: I didn't want to modify the method signature, but a missing global configuration is an `AlertConfigurationException`. It will be wrapped, but will make it more clear what the problem is.

src/test/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActionsTest.java
```diff
@@ -152,20 +157,26 @@ public void testPagedRequest() throws AlertDatabaseConstraintException {
 
         NotificationContentRepository notificationRepository = Mockito.mock(NotificationContentRepository.class);
         AuditNotificationRepository auditNotificationRepository = Mockito.mock(AuditNotificationRepository.class);
-        JobAccessor jobAccessor = Mockito.mock(JobAccessor.class);
+        JobAccessor oldJobAccessor = Mockito.mock(JobAccessor.class);
+        JobAccessorV2 jobAccessor = Mockito.mock(JobAccessorV2.class);
+        Mockito.when(jobAccessor.getJobById(Mockito.any())).thenReturn(null);
```
JakeMathews, line 162: It looks like you are missing the `Mockito.when` for the `oldJobAccessor`.

JakeMathews: I see you do this further down. So is line 162 still necessary?
```
Mockito.doReturn(Optional.of(configuration)).when(oldJobAccessor).getJobById(Mockito.any());
        Mockito.doReturn(Optional.of(distributionJob)).when(jobAccessor).getJobById(Mockito.any());
```

gkillough: I updated `AuditEntryActions` because it used `NotificationProcessor`, but I did not need to update `DefaultAuditAccessor` for this PR.

JakeMathews (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/email/actions/EmailDistributionTestAction.java
```diff
@@ -62,10 +62,12 @@ public MessageResult testConfig(
     private DistributionJobModel creatUpdatedJobModelWithEmailAddresses(DistributionJobModel originalJobModel, @Nullable String destination) throws IntegrationException {
         Set<String> updateEmailAddresses = emailActionHelper.createUpdatedEmailAddresses(originalJobModel, destination);
         EmailJobDetailsModel originalEmailJobDetails = originalJobModel.getDistributionJobDetails().getAsEmailJobDetails();
+
+        // For testing configuration, just use additional email addresses field
```
jamesrichard91, line 67: This does not seem correct, why are we only using the additional email addresses? Part of the email distribution test is to verify the email recipients. So we should be using the configuration as specified in the test.

gkillough: `EmailTestActionHelper` will get all of the correct recipients for a Job. 

When testing configuration, the event is not tied to a real notification, so when the event gets to the channel, the `EmailAddressHandler` will not be able to look up the "Project Owner" or "project members" correctly. Before my changes, we would shove an extra field into `FieldUtility` with the results of `EmailTestActionHelper`; now that there is no `FieldUtility` we append the results of `EmailTestActionHelper` as additional email addresses instead and only use those emails for testing.

channel/src/main/java/com/synopsys/integration/alert/channel/email/actions/EmailTestActionHelper.java
```diff
@@ -24,72 +24,68 @@
...
+
+        Long providerConfigId = distributionJobModel.getBlackDuckGlobalConfigId();
+        boolean onlyAdditionalEmails = emailJobDetails.isAdditionalEmailAddressesOnly();
 
         if (null != providerConfigId && !onlyAdditionalEmails) {
-            Set<ProviderProject> providerProjects = retrieveProviderProjects(fieldUtility, filterByProject, providerConfigId);
-            if (null != providerProjects) {
-                Set<String> providerEmailAddresses = addEmailAddresses(providerConfigId, providerProjects, fieldUtility);
+            Set<ProviderProject> providerProjects = retrieveProviderProjects(distributionJobModel, providerConfigId);
+            if (null != providerProjects && !providerProjects.isEmpty()) {
```
JakeMathews, line 69:
```suggestion
            if (CollectionUtils.isNotEmpty(providerProjects)) {
```

src/test/java/com/synopsys/integration/alert/channel/ChannelDescriptorTestIT.java
```diff
@@ -53,43 +63,54 @@
...
         channelEvent = createChannelEvent();
-        destinationName = getDestinationName();
+        distributionJobModel = saveDistributionJob();
     }
 
     @AfterEach
     public void cleanupTest() {
-        if (null != global_config && global_config.isPresent()) {
-            configurationAccessor.deleteConfiguration(global_config.get());
+        if (null != optionalChannelGlobalConfig && optionalChannelGlobalConfig.isPresent()) {
```
JakeMathews, line 100: Why would the optional be null?

JakeMathews (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/job/JobConfigurationModelFieldExtractorUtils.java
```diff
@@ -66,10 +81,12 @@ public static DistributionJobModel convertToDistributionJobModel(UUID jobId, Map
                                                   .policyFilterPolicyNames(extractFieldValues("blackduck.policy.notification.filter", configuredFieldsMap))
                                                   .vulnerabilityFilterSeverityNames(extractFieldValues("blackduck.vulnerability.notification.filter", configuredFieldsMap));
 
-        List<BlackDuckProjectDetailsModel> blackDuckProjectDetails = extractFieldValues("channel.common.configured.project", configuredFieldsMap)
-                                                                         .stream()
-                                                                         .map(projectName -> new BlackDuckProjectDetailsModel(projectName, projectName))
-                                                                         .collect(Collectors.toList());
+        List<BlackDuckProjectDetailsModel> blackDuckProjectDetails = Optional.ofNullable(projectFilterDetails)
```
bamandel, line 84: Isn't this an example of a way to **not** use an optional? Instead of being passed a null list, we should be passing an empty list which we then check for content and modify/use as needed.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudChannel.java
```diff
@@ -55,8 +57,9 @@ public JiraCloudChannel(JiraCloudChannelKey jiraChannelKey, Gson gson, AuditAcce
 
     @Override
     protected IssueTrackerContext getIssueTrackerContext(DistributionEvent event) {
-        FieldUtility fieldUtility = event.getFieldUtility();
-        return jiraCloudContextBuilder.build(fieldUtility);
+        ConfigurationModel globalConfig = event.getChannelGlobalConfig()
+                                              .orElseThrow(() -> new AlertRuntimeException(new AlertConfigurationException("Missing Jira Cloud global configuration")));
```
bamandel, line 61: Instead of nesting exceptions, we should make a runtime configuration exception.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
bamandel: Perhaps we can introduce a new exception in a separate PR?

channel/src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsChannel.java
```diff
@@ -59,8 +63,11 @@ public MsTeamsChannel(MsTeamsKey msTeamsKey, Gson gson, AuditAccessor auditAcces
 
     @Override
     public void distributeMessage(DistributionEvent event) throws IntegrationException {
-        FieldUtility fields = event.getFieldUtility();
-        String webhook = fields.getString(MsTeamsDescriptor.KEY_WEBHOOK)
+        DistributionJobModel distributionJobModel = event.getDistributionJobModel();
+        DistributionJobDetailsModel distributionJobDetails = distributionJobModel.getDistributionJobDetails();
+        MSTeamsJobDetailsModel asMSTeamsJobDetails = distributionJobDetails.getAsMSTeamsJobDetails();
+        String webhook = Optional.ofNullable(asMSTeamsJobDetails.getWebhook())
```
bamandel, line 69: We shouldn't be creating this inline optional object to act as our **if** statement. I believe an if statement here would be much clearer

channel/src/main/java/com/synopsys/integration/alert/channel/slack/parser/SlackChannelEventParser.java
```diff
@@ -76,7 +81,9 @@ public SlackChannelEventParser(SlackChannelMessageParser slackChannelMessagePars
             throw new AlertFieldException(fieldErrors);
         }
 
-        String channelUsername = fields.getString(SlackDescriptor.KEY_CHANNEL_USERNAME).orElse(SLACK_DEFAULT_USERNAME);
+        String channelUsername = Optional.ofNullable(slackJobDetails.getChannelUsername())
+                                     .filter(StringUtils::isNotBlank)
+                                     .orElse(SLACK_DEFAULT_USERNAME);
```
bamandel, line 86: This optional should be a ternary operator

web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -324,33 +325,56 @@ protected ValidationActionResponse validateWithoutChecks(JobFieldModel resource)
     @Override
     protected ValidationActionResponse testWithoutChecks(JobFieldModel resource) {
         ValidationResponseModel responseModel;
-        String id = resource.getJobId();
+        String jobIdString = resource.getJobId();
+        UUID jobId = Optional.ofNullable(jobIdString)
+                         .filter(StringUtils::isNotBlank)
+                         .map(UUID::fromString)
+                         .orElse(null);
```
bamandel, line 332: This should be an **if** statement

web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
```diff
@@ -324,33 +325,56 @@ protected ValidationActionResponse validateWithoutChecks(JobFieldModel resource)
...
                     }
 
-                    MessageResult testActionResult = testAction.testConfig(jobId, channelFieldModel, fieldUtility);
+                    // Not all channels have a global config
+                    ConfigurationModel nullableChannelGlobalConfig = configurationAccessor.getConfigurationsByDescriptorNameAndContext(channelFieldModel.getDescriptorName(), ConfigContextEnum.GLOBAL)
+                                                                         .stream()
+                                                                         .findFirst()
+                                                                         .orElse(null);
+
+                    List<BlackDuckProjectDetailsModel> projectFilterDetails = Optional.ofNullable(resource.getConfiguredProviderProjects()).orElse(List.of())
```
bamandel, line 365: Resource should return an empty list instead of null so we can process without checking for anything.

bamandel (commented):
Some uses of optional were unexpected. Do we want to create an optional object instead of using if statements now? I always thought our goal was to use optionals as types to signify the content may not exist.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -194,6 +194,8 @@ public final ValidationActionResponse test(JobFieldModel resource) {
             return ValidationActionResponse.createOKResponseWithContent(validationResponse);
         }
         ValidationActionResponse response = testWithoutChecks(resource);
+        // FIXME this should only return OK if validation COMPLETED successfully (even
+        //  if there are field errors), not if there were other types of errors unrelated to validation
```
jamesrichard91, line 198: Im not sure what this comment means, I found an issue with 6.3.0 where the test discovers a certificate error but it does not make it to the UI because this is changing the response to an OK status. Please refer to https://github.com/blackducksoftware/blackduck-alert/pull/1289 to see the changes I would like to make in this area.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/job/DistributionJobModelBuilder.java
```diff
@@ -53,6 +53,7 @@
     private DistributionJobDetailsModel distributionJobDetails;
 
     public DistributionJobModel build() {
+        // TODO validate required fields are present
```
jamesrichard91, line 56: Is this going to be done in this PR? or is there a Jira ticket to track this?

channel/src/main/java/com/synopsys/integration/alert/channel/email/actions/EmailTestActionHelper.java
```diff
@@ -24,72 +24,68 @@
...
-            String projectNamePattern = fieldUtility.getStringOrEmpty(ProviderDistributionUIConfig.KEY_PROJECT_NAME_PATTERN);
+        if (distributionJobModel.isFilterByProject()) {
+            Set<String> configuredProjects = distributionJobModel.getProjectFilterDetails()
+                                                 .stream()
+                                                 .map(BlackDuckProjectDetailsModel::getName)
+                                                 .collect(Collectors.toSet());
+            String projectNamePattern = distributionJobModel.getProjectNamePattern().orElse("");
             return providerProjects
                        .stream()
                        .filter(databaseEntity -> doesProjectMatchConfiguration(databaseEntity.getName(), projectNamePattern, configuredProjects))
```
jamesrichard91, line 91: I don't think the ProviderProject's are database entities anymore, this should be updated to reflect that.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
gkillough: Since we do this in a few places, we should try to figure out why this exception can be thrown in a place that does not allow checked exceptions before we assume that a new exception is needed.

gkillough: Jira issue created.

src/test/java/com/synopsys/integration/alert/channel/ChannelDescriptorTestIT.java
gkillough: I believe this code was here before and I just renamed the variable. I'll clean this up though.

channel/src/main/java/com/synopsys/integration/alert/channel/msteams/MsTeamsChannel.java
gkillough: I thought we as a team decided this was the preferred way to handle these cases?

web/src/main/java/com/synopsys/integration/alert/web/api/job/JobConfigActions.java
gkillough: `resource` is a REST API request-body, so this can absolutely be null.

alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
gkillough: I created a 6.5.0 Jira issue for this as well. I'll take a look at your PR.

gkillough: Your issue is different. This is addressing the fact that field-errors produce a 2xx response (which they should because the validation _completed_ successfully), but errors during validation also result in a 2xx which they should not.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/job/DistributionJobModelBuilder.java
gkillough: It will not be done in this PR because this is meant to have parity with the previous model which had no checks for completeness. I will create a Jira issue.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudChannel.java
psantos1113: Why not just throw the AlertRuntimeException with the error message?  The nested exception doesn't buy you anything.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
psantos1113: We change the method signature in the future to throw an AlertConfigurationException since that appears to be the correct action.  That should be done in a smaller PR.

psantos1113 (approved):
A few things others have brought up. Once those are settled I have no issues with the changes.

alert-common/src/main/java/com/synopsys/integration/alert/common/event/DistributionEvent.java
gkillough: Are they compiler warnings or IDE warnings? I could have sworn we determined that we should no longer be generating `serialVersionUID` on `Serializable` classes.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudChannel.java
gkillough: I created a Jira issue to resolve this. One reason for wrapping it is because the `AlertConfigurationException` is the correct exception to throw, so when someone comes to resolve this, they will have the correct alternative available.

bamandel (approved):

PR #1288, opened 2020-12-07
Update the ProvidersMissingTask validation

JakeMathews:
The desire was to have `ProvidersMissingTask` validate on startup and only validate every hour instead of every minute.
`ProvidersMissingTask` already extends `StartupScheduledTask` so the `StartupScheduledTask::postTaskStartup` needed to be implemented.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
@@ -44,6 +44,7 @@ public abstract class ScheduledTask implements Runnable {
     public static final String STOP_SCHEDULE_EXPRESSION = "";
     //Spring Cron documentation  https://riptutorial.com/spring/example/21209/cron-expression
     public static final String EVERY_MINUTE_CRON_EXPRESSION = "0 0/1 * 1/1 * *";
+    public static final String EVERY_HOUR_CRON_EXPRESSION = "0 * * * * *";
     public static final String ONCE_DAILY_CRON_EXPRESSION = "0 0 0 * * *";
     public static final Long EVERY_MINUTE_SECONDS = 60L;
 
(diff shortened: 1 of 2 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
```diff
@@ -44,6 +44,7 @@
     public static final String STOP_SCHEDULE_EXPRESSION = "";
     //Spring Cron documentation  https://riptutorial.com/spring/example/21209/cron-expression
     public static final String EVERY_MINUTE_CRON_EXPRESSION = "0 0/1 * 1/1 * *";
+    public static final String EVERY_HOUR_CRON_EXPRESSION = "0 * * * * *";
```
jamesrichard91, line 47: I don't think this is every hour, this looks like every minute. Have you tested this? If you start Alert you can go to the task management section and see the next run time to confirm if this is working as expected

JakeMathews: What cron standard are you using?

jamesrichard91: The comment on line 45 references the Spring cron documentation that is used

alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/task/ScheduledTask.java
```diff
@@ -44,6 +44,7 @@
     public static final String STOP_SCHEDULE_EXPRESSION = "";
     //Spring Cron documentation  https://riptutorial.com/spring/example/21209/cron-expression
     public static final String EVERY_MINUTE_CRON_EXPRESSION = "0 0/1 * 1/1 * *";
+    public static final String EVERY_HOUR_CRON_EXPRESSION = "0 0 0/1 * * *";
```
jamesrichard91, line 47: This looks fine

jamesrichard91 (approved):

gkillough (approved):

PR #1289, opened 2020-12-07
Fixing the way we handle testing configurations vs validating configurations

jamesrichard91:


ui/src/main/js/field/MessageFormatter.js
```diff
@@ -53,11 +52,13 @@ class MessageFormatter extends Component {
 
 MessageFormatter.propTypes = {
     id: PropTypes.string,
+    isDetailed: PropTypes.bool,
     message: PropTypes.string
 };
 
 MessageFormatter.defaultProps = {
     id: 'messageFormatterId',
+    isDetailed: false,
```
gkillough, line 61: Might be nice to use the same `errorIsDetailed` variable name here as in `GlobalConfiguration` so it's obvious that they represent the same thing.

ui/src/main/js/util/httpErrorUtilities.js
```diff
@@ -15,11 +16,13 @@ export function createErrorObject(errorResponse) {
         return createEmptyErrorObject();
     }
     const responseMessage = errorResponse.message;
+    const { isDetailed } = errorResponse;
     const responseErrors = errorResponse.errors;
```
gkillough, line 20: Why not destructure all of these and then just map them to the correct name in the return?

gkillough (approved):

ui/src/main/js/util/httpErrorUtilities.js
psantos1113: message is defined as a const below on line 19 so we cant' destructure without a variable conflict.  In order to destructure but we would have to use a different variable name and syntax to set the message property in the response object.

`'message': responseMessage`

psantos1113 (approved):

ChomickiM (approved):

PR #1294, opened 2020-12-10
Jira channel-side custom field support

gkillough:


channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudContextBuilder.java
```diff
@@ -91,13 +52,18 @@ public JiraCloudContext build(ConfigurationModel globalConfig, DistributionJobMo
...
         JiraCloudJobDetailsModel jiraCouldJobDetails = distributionJobDetails.getAsJiraCouldJobDetails();
 
-        IssueConfig issueConfig = new IssueConfig();
-        issueConfig.setProjectName(jiraCouldJobDetails.getProjectNameOrKey());
-        issueConfig.setIssueCreator(jiraCouldJobDetails.getIssueCreatorEmail());
-        issueConfig.setIssueType(jiraCouldJobDetails.getIssueType());
-        issueConfig.setCommentOnIssues(jiraCouldJobDetails.isAddComments());
-        issueConfig.setResolveTransition(jiraCouldJobDetails.getResolveTransition());
-        issueConfig.setOpenTransition(jiraCouldJobDetails.getReopenTransition());
+        // FIXME add custom fields
```
gkillough, line 55: This will be done in the next PR.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerContextBuilder.java
```diff
@@ -91,14 +52,18 @@ public JiraServerContext build(ConfigurationModel globalConfig, DistributionJobM
...
 
-        IssueConfig issueConfig = new IssueConfig();
-        issueConfig.setProjectName(jiraServerJobDetails.getProjectNameOrKey());
-        issueConfig.setIssueCreator(jiraServerJobDetails.getIssueCreatorUsername());
-        issueConfig.setIssueType(jiraServerJobDetails.getIssueType());
-        issueConfig.setCommentOnIssues(jiraServerJobDetails.isAddComments());
-        issueConfig.setResolveTransition(jiraServerJobDetails.getResolveTransition());
-        issueConfig.setOpenTransition(jiraServerJobDetails.getReopenTransition());
-
+        // FIXME add custom fields
```
gkillough, line 55: This will be done in the next PR.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudContextBuilder.java
```diff
@@ -91,13 +52,18 @@ public JiraCloudContext build(ConfigurationModel globalConfig, DistributionJobMo
...
+        JiraIssueConfig issueConfig = new JiraIssueConfig(
+            jiraCouldJobDetails.getProjectNameOrKey(),
+            jiraCouldJobDetails.getProjectNameOrKey(),
```
jamesrichard91, line 58: Are we passing in the projectNameOrKey twice on purpose? or was this an accident?

channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraCustomFieldResolver.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    private void initializeCache() throws IntegrationException {
+        List<CustomFieldCreationResponseModel> userVisibleFields = retrieveAvailableFields.get();
+        for (CustomFieldCreationResponseModel fieldModel : userVisibleFields) {
+            fieldCache.put(fieldModel.getName(), fieldModel);
+        }
+        isCachePopulated = true;
+    }
```
jamesrichard91, line 110: How long are we keeping this cache around for?

channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/JiraCloudContextBuilder.java
gkillough: I'm not entirely sure we use all of these fields and it's something that needs a closer look, but this object has "project name" and "project key" fields and I'm passing it into both just in case. In Jira APIs you can use both interchangeably (unlike the project id).

channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraCustomFieldResolver.java
gkillough: For the duration of processing for a single event.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/model/JiraCustomFieldConfig.java
```diff
@@ -0,0 +1,42 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.channel.jira.common.model;
+
+public class JiraCustomFieldConfig {
+    private final String fieldName;
+    private final String fieldValue;
```
jamesrichard91, line 27: Do we also need the Jira field type? Should this class be serializable?

gkillough: We get the field type when we look up the field to get its ID. This class is transient in the way that it is used, so I don't think it needs to be serializable.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/util/JiraServerCustomFieldResolver.java
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.
+ */
+package com.synopsys.integration.alert.channel.jira.server.util;
+
+import com.synopsys.integration.alert.channel.jira.common.JiraCustomFieldResolver;
+import com.synopsys.integration.jira.common.server.service.FieldService;
+
+public class JiraServerCustomFieldResolver extends JiraCustomFieldResolver {
+    public JiraServerCustomFieldResolver(FieldService fieldService) {
+        super(fieldService::getUserVisibleFields);
```
bamandel, line 30: Instead of using an entire class for this function, you can open up the method that uses this function to accept the function as a parameter. Then wherever you make the distinction between JiraServerCustomFieldResolver and JiraCloudCustomFieldResolver (Couldn't find it in the PR) you instead set a variable function equal to what is being passed here... does that make sense? I think it would be clearer than having 2 separate basically empty classes.

bamandel (approved):

psantos1113 (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/util/JiraServerCustomFieldResolver.java
gkillough: Yeah, I was thinking about this. I didn't know if it would be the most expressive API to ask for a `Supplier` in the constructor, but I think you're right that it's better than the duplication.

PR #1295, opened 2020-12-10
Notification reciever improvements

ChomickiM:
We are reworking the NotificationReceiver to now process the and send notifications after they are saved by the BlackDuckAccumulator. It will now receive a wakeup call every time a new notification comes in rather than a list of notifications. There exists a current issue where notifications that come in may not be sent out until the NotificationReceiver finishes processing all of the messages. It is possible that if a constant stream of notifications is being sent to Alert that this loop will continue until we stop at 100 loops (10,000 Notifications). Once we find a solution to the threading problem we can remove this upper limit as it will no longer impact or cause issues to this.


alert-database/src/main/resources/liquibase/6.4.0/notification-content-processing-column.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="martinch" id="2020-12-01-14-04-07-258">
```
psantos1113, line 5: The id's should have a unique text describing the change now.  Try something like add_processed_column_to_notification_table

jamesrichard91: I thought we were still using the time as the ID, I gave Martin bad information.

psantos1113: We started changing back in September to more descriptive IDs

alert-common/src/main/java/com/synopsys/integration/alert/common/event/NotificationEvent.java
```diff
@@ -22,18 +22,10 @@
...
 package com.synopsys.integration.alert.common.event;
 
-import java.util.List;
-
 public class NotificationEvent extends AlertEvent {
     public static final String NOTIFICATION_EVENT_TYPE = "notification_event";
-    private final List<Long> notificationIds;
 
-    public NotificationEvent(final List<Long> notificationIds) {
+    public NotificationEvent() {
```
gkillough, line 28: It might be more appropriate to call this "NotificationsReceivedEvent" now.

alert-database/src/main/java/com/synopsys/integration/alert/database/notification/NotificationContentRepository.java
```diff
@@ -84,4 +85,6 @@
                        + ")")
     Page<NotificationEntity> findMatchingSentNotification(@Param("searchTerm") String searchTerm, Pageable pageable);
 
+    @Query(value = "SELECT entity FROM NotificationEntity entity WHERE entity.processed = false")
+    Page<NotificationEntity> findNotProcessedNotifications(Pageable pageable);
```
gkillough, line 89: Should we also order by date received?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationAccessor.java
```diff
@@ -185,6 +193,25 @@ public PageRequest getPageRequestForNotifications(Integer pageNumber, Integer pa
...
     }
 
+    //TODO this needs unit tests
+    @Override
+    public Page<AlertNotificationModel> findNotificationsNotProcessed() {
+        Sort.Order sortingOrder = Sort.Order.asc("providerCreationTime");
+        PageRequest pageRequest = PageRequest.of(0, 100, Sort.by(sortingOrder));
+        return notificationContentRepository.findNotProcessedNotifications(pageRequest)
+                   .map(this::toModel);
+    }
```
gkillough, line 203: The sort and page size should be reflected in the method name. Also, I believe we don't want to expose `Page` (use `AlertPagedModel` instead?) outside of the DB API, but we may not have done the best job of that up until this point.

src/test/java/com/synopsys/integration/alert/performance/ScalingPerformanceTest.java
```diff
@@ -69,7 +68,7 @@ public static void initTest() {
 
     @Test
     @Ignore
-    @Disabled
+    //@Disabled //TODO remove this when testing
```
jamesrichard91, line 71: Make sure this does not get merged into master.

jamesrichard91 (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationAccessor.java
```diff
@@ -185,6 +174,27 @@ public PageRequest getPageRequestForNotifications(Integer pageNumber, Integer pa
         return PageRequest.of(pageNumber, pageSize, Sort.by(sortingOrder));
     }
 
+    @Override
+    public AlertPagedModel<AlertNotificationModel> getFirstPageOfNotificationsNotProcessed() {
```
gkillough, line 178: It seems like we should be passing the `pageSize` as a parameter. This class shouldn't be the one that decides on that size.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationAccessor.java
```diff
@@ -215,7 +225,8 @@ private AlertNotificationModel toModel(NotificationEntity entity) {
                                      .flatMap(ConfigurationFieldModel::getFieldValue)
                                      .orElse(providerConfigName);
         }
-        return new AlertNotificationModel(entity.getId(), providerConfigId, entity.getProvider(), providerConfigName, entity.getNotificationType(), entity.getContent(), entity.getCreatedAt(), entity.getProviderCreationTime());
+        return new AlertNotificationModel(entity.getId(), providerConfigId, entity.getProvider(), providerConfigName, entity.getNotificationType(), entity.getContent(), entity.getCreatedAt(), entity.getProviderCreationTime(),
+            entity.getProcessed());
```
gkillough, line 229: Can we clean up this line-wrapping?

alert-database/src/main/java/com/synopsys/integration/alert/database/notification/NotificationContentRepository.java
```diff
@@ -84,4 +82,6 @@
...
```
gkillough, line 86: We might be able to do this without a custom query: 
https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods.query-creation

```suggestion
    Page<NotificationEntity> findByProcessedFalse(Pageable pageable);
```

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -51,32 +54,44 @@
...
 
     @Override
-    public void handleEvent(NotificationEvent event) {
-        if (NotificationEvent.NOTIFICATION_EVENT_TYPE.equals(event.getDestination())) {
-            if (null == event.getNotificationIds() || event.getNotificationIds().isEmpty()) {
-                logger.warn("Can not process a notification event without notification Id's.");
-                return;
-            }
+    public void handleEvent(NotificationReceivedEvent event) {
+        if (NotificationReceivedEvent.NOTIFICATION_RECEIVED_EVENT_TYPE.equals(event.getDestination())) {
```
gkillough, line 65: Rather than nesting all of the logic inside of this if-statement, can we negate this condition and use the code from the else here? Then `return` to short-circuit.

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -51,32 +54,44 @@
...
+                List<AlertNotificationModel> notifications = pageOfAlertNotificationModels.getModels();
+                logger.info("Sending {} notifications.", notifications.size());
+                List<DistributionEvent> distributionEvents = notificationProcessor.processNotifications(FrequencyType.REAL_TIME, notifications);
+                logger.info("Sending {} events for notifications.", distributionEvents.size());
+                eventManager.sendEvents(distributionEvents);
+                notificationAccessor.setNotificationsProcessed(notifications);
+                numPagesProcessed++;
+                pageOfAlertNotificationModels = notificationAccessor.getFirstPageOfNotificationsNotProcessed();
+                logger.trace("Processing Page: {}. New pages found: {}", numPagesProcessed,
+                    pageOfAlertNotificationModels.getTotalPages());
```
gkillough, line 83: Can we clean up this line-wrapping?

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -51,32 +54,46 @@
...
+        }
+        logger.debug("Event {}", event);
+        logger.info("Processing event for notifications.");
+
+        int numPagesProcessed = 0;
+        int pageSize = 100;
+
+        AlertPagedModel<AlertNotificationModel> pageOfAlertNotificationModels = notificationAccessor.getFirstPageOfNotificationsNotProcessed(pageSize);
+        //TODO: Once we create a way of handling channel events in parallel, we can remove the MAX_NUMBER_PAGES_PROCESSED.
+        while (!CollectionUtils.isEmpty(pageOfAlertNotificationModels.getModels()) && numPagesProcessed < MAX_NUMBER_PAGES_PROCESSED) {
```
bamandel, line 77: Why are we getting a paged result and verifying that we process a specific separate number? Wouldn't we want to get the paged result and simply process all of them?

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -51,32 +54,46 @@
...
-                return;
-            }
-            logger.debug("Event {}", event);
-            logger.info("Processing event for {} notifications.", event.getNotificationIds().size());
-            List<AlertNotificationModel> notifications = notificationAccessor.findByIds(event.getNotificationIds());
+    public void handleEvent(NotificationReceivedEvent event) {
+        if (!NotificationReceivedEvent.NOTIFICATION_RECEIVED_EVENT_TYPE.equals(event.getDestination())) {
+            logger.warn("Received an event of type '{}', but this listener is for type '{}'.", event.getDestination(), NotificationReceivedEvent.NOTIFICATION_RECEIVED_EVENT_TYPE);
+            return;
+        }
```
bamandel, line 68: Is this a valid concern? Shouldn't this receiver be setup to only run when this event type is being processed?

alert-database/src/main/java/com/synopsys/integration/alert/database/notification/NotificationEntity.java
```diff
@@ -121,6 +125,14 @@ public OffsetDateTime getProviderCreationTime() {
         return providerCreationTime;
     }
 
+    public boolean getProcessed() {
+        return processed;
+    }
+
+    public void setProcessed() {
```
bamandel, line 132: This method name confused me in another file, we should change the name as setProcessed implies a simple setter where as we are explicitly setting this to true. Something as simple as setProcessedToTrue() would be easier to understand when reading the method elsewhere.

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
ChomickiM: There is a possibility where a customer can have a constant stream of notifications. We have a limitation right now Paulo and I discovered getting addressed in IALERT-2116. Right now, we must wait until it finishes any pages that are in the queue before it actually sends the event to the channels. Because of this we are putting in a hard upper limit so that at some point, even if it takes a long time, it will still send the messages to the channels.

bamandel: I guess my next question is the page is 100 items and the max number is 100. Why are these the same?

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
ChomickiM: This was refactored as part of Gavin's earlier suggestion. Looking into it I'm not entirely sure if it would be possible in this setup for the message to go to the wrong place since its the only consumer of NotificationReceivedEvent. In the event there were multiple consumers I believe then it would be a concern but I'm not too sure, I would need to investigate how JMS would handle that. If others think its no longer relevant then I can remove this.

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
ChomickiM: Admittedly the numbers are fairly arbitrary, this setup would allow us to do 10,000 notifications. So each page can have up to 100 notifications, and we would process 100 pages before we would kick out of the loop and allow the notifications to be sent to the channels. With the setup once the next notification is picked up by the accumulator, the receiver would wake up again and continue processing where it left off.

psantos1113 (approved):

bamandel (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationAccessor.java
```diff
@@ -65,16 +64,14 @@
     private final AuditEntryRepository auditEntryRepository;
     private final AuditNotificationRepository auditNotificationRepository;
     private final ConfigurationAccessor configurationAccessor;
-    private final EventManager eventManager;
 
     @Autowired
     public DefaultNotificationAccessor(NotificationContentRepository notificationContentRepository, AuditEntryRepository auditEntryRepository, AuditNotificationRepository auditNotificationRepository,
-        ConfigurationAccessor configurationAccessor, EventManager eventManager) {
+        ConfigurationAccessor configurationAccessor) {
```
gkillough, line 70: This wrapping is also a little funky.

gkillough (approved):

PR #1296, opened 2020-12-10
Adding support for Upgrade Guidance for vulnerable components

jamesrichard91:


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/util/VulnerabilityUtil.java
```diff
@@ -158,6 +151,24 @@ public static LinkableItem createVulnerabilityAttributeItem(String severityValue
...
+    }
+
+    private static Optional<LinkableItem> createLongTermGuidance(Supplier<ComponentVersionUpgradeGuidanceLongTermView> getLongTermOption, String remediationLabel) {
+        ComponentVersionUpgradeGuidanceLongTermView longTermGuidance = getLongTermOption.get();
+        if (null != longTermGuidance) {
+            String versionText = createLongTermGuidanceVersionText(longTermGuidance);
+            return Optional.of(new LinkableItem(remediationLabel, versionText, longTermGuidance.getVersion()));
+        }
+        return Optional.empty();
+    }
```
bamandel, line 170: There's a lot of duplication going on in this class. Since you're working with two different types that don't share a parent, then we will probably want to create a wrapper class around ComponentVersionUpgradeGuidanceLongTermView and ComponentVersionUpgradeGuidanceLongTermView. The class would contain all the relevant shared fields with a constructor for each type where you map the share fields appropriately. Then we can have a lot of this common code in a single place.

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/util/VulnerabilityUtil.java
```diff
@@ -169,6 +180,60 @@ private static String createRemediationVersionText(RemediatingVersionView remedi
...
+                stringBuilder.append(" (Vulnerabilities: ");
+                boolean previousVulnerabilites = false;
+                previousVulnerabilites = addVulnerabilityCount(stringBuilder, vulnerabilityRisk.getCritical(), " Critical: ", previousVulnerabilites);
+                previousVulnerabilites = addVulnerabilityCount(stringBuilder, vulnerabilityRisk.getHigh(), " High: ", previousVulnerabilites);
+                previousVulnerabilites = addVulnerabilityCount(stringBuilder, vulnerabilityRisk.getMedium(), " Medium: ", previousVulnerabilites);
+                addVulnerabilityCount(stringBuilder, vulnerabilityRisk.getLow(), " Low: ", previousVulnerabilites);
+                stringBuilder.append(")");
+            }
+        }
+        return stringBuilder.toString();
```
bamandel, line 218: These methods can also be handled by the wrapper class mentioned previously

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/util/VulnerabilityUtil.java
jamesrichard91: What do you think of the  updated method with the new models?

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/model/UpgradeGuidanceRiskView.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+import com.synopsys.integration.blackduck.api.generated.component.ComponentVersionUpgradeGuidanceLongTermVulnerabilityRiskView;
+import com.synopsys.integration.blackduck.api.generated.component.ComponentVersionUpgradeGuidanceShortTermVulnerabilityRiskView;
+
+public class UpgradeGuidanceRiskView extends AlertSerializableModel {
+    private BigDecimal critical;
+    private BigDecimal high;
+    private BigDecimal medium;
+    private BigDecimal low;
```
gkillough, line 13: Is the number of vulnerabilities not discrete? I assume using `BigDecimal` is not a mistake, but it seems so odd.

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/util/VulnerabilityUtil.java
```diff
@@ -169,6 +171,37 @@ private static String createRemediationVersionText(RemediatingVersionView remedi
...
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append(upgradeGuidanceView.getVersionName());
+        UpgradeGuidanceRiskView vulnerabilityRisk = upgradeGuidanceView.getVulnerabilityRisk();
+        if (vulnerabilityRisk.hasVulnerabilities()) {
+            stringBuilder.append(" (Vulnerabilities:");
+            boolean previousVulnerabilites = false;
+            previousVulnerabilites = addVulnerabilityCount(stringBuilder, " Critical: ", previousVulnerabilites, vulnerabilityRisk::hasCriticalVulnerabilities, vulnerabilityRisk::getCritical);
+            previousVulnerabilites = addVulnerabilityCount(stringBuilder, " High: ", previousVulnerabilites, vulnerabilityRisk::hasHighVulnerabilities, vulnerabilityRisk::getHigh);
+            previousVulnerabilites = addVulnerabilityCount(stringBuilder, " Medium: ", previousVulnerabilites, vulnerabilityRisk::hasMediumVulnerabilities, vulnerabilityRisk::getMedium);
+            addVulnerabilityCount(stringBuilder, " Low: ", previousVulnerabilites, vulnerabilityRisk::hasLowVulnerabilities, vulnerabilityRisk::getLow);
```
gkillough, line 184: Why not pass the true label (e.g. "Critical") and append the spaces and colon inside of `addVulnerabilityCount`?

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/model/UpgradeGuidanceRiskView.java
jamesrichard91: I use BigDecimal here because that is what we get in the response from Black Duck. I can probably change this to be integer

gkillough: No need to change it. It's just odd that BlackDuck's response is not discrete.

gkillough (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/collector/builder/model/UpgradeGuidanceRiskView.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class UpgradeGuidanceRiskView extends AlertSerializableModel {
+    private Integer critical;
+    private Integer high;
+    private Integer medium;
+    private Integer low;
+
+    public static UpgradeGuidanceRiskView fromShortTermVulnerabilityRiskView(ComponentVersionUpgradeGuidanceShortTermVulnerabilityRiskView shortTermVulnerabilityRiskView) {
+        if (null == shortTermVulnerabilityRiskView) {
+            return new UpgradeGuidanceRiskView(null, null, null, null);
```
bamandel, line 18: You could add a constructor with no params that sets everything to null

bamandel (approved):

PR #1301, opened 2020-12-15
revise implementation to use updated blackduck-common

ekerwin:
***NOTE***
This does not compile. All I'm trying to show here is a different implementation based on an issue created by James related to this class. It adds a clearer dependency on ProjectVersionView, which should be possible to acquire and pass in. Once that is done, there may be other side effects of upgrading to 51.0.0.

### DIFF
--- a/provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
+++ b/provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
@@ -26,101 +26,79 @@ import java.time.Instant;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
-import java.util.function.BiFunction;
-
-import org.apache.commons.lang3.StringUtils;
-import org.apache.http.HttpHeaders;
 
 import com.google.gson.Gson;
-import com.synopsys.integration.blackduck.api.generated.view.IssueView;
+import com.synopsys.integration.blackduck.api.generated.view.ProjectVersionIssuesView;
+import com.synopsys.integration.blackduck.api.generated.view.ProjectVersionView;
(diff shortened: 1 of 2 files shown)


ekerwin:
Also, this doesn't make use of the new IssueService, which you may want to consider as well.

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
```diff
@@ -26,101 +26,79 @@
...
-        HttpUrl issueLookupHttpUrl = new HttpUrl(issueLookupUrl);
-        BlackDuckRequestBuilder requestBuilder = blackDuckRequestFactory.createCommonGetRequestBuilder(issueLookupHttpUrl)
-                                                     .addHeader(HttpHeaders.ACCEPT, ISSUE_ENDPOINT_MEDIA_TYPE_V6);
+        performRequest(requestUri, httpMethod, issueRequestModel);
+    }
 
-        // This is really a List<BomComponentIssueView>, but BomComponentIssueView is not considered a BlackDuckResponse.
-        List<IssueView> bomComponentIssues = blackDuckApiClient.getAllResponses(requestBuilder, IssueView.class);
+    private Optional<ProjectVersionIssuesView> retrieveExistingIssue(ProjectVersionView projectVersionView, String issueKey) throws IntegrationException {
+        List<ProjectVersionIssuesView> bomComponentIssues = blackDuckApiClient.getAllResponses(projectVersionView, ProjectVersionView.ISSUES_LINK_RESPONSE);
```
jamesrichard91, line 78: This seems like a much better approach to retrieve existing issues

jamesrichard91:
Applying these suggestions in https://github.com/blackducksoftware/blackduck-alert/pull/1312

PR #1302, opened 2020-12-15
Caching update

jamesrichard91:
Removing the bucket service use, updating to blackduck-common 51.0.0, using the new cache client in the notification processor.


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProperties.java
```diff
@@ -118,6 +118,11 @@ public BlackDuckHttpClient createBlackDuckHttpClient(IntLogger intLogger) throws
...
     }
 
+    public BlackDuckHttpClient createBlackDuckCacheClient(IntLogger intLogger) throws AlertException {
+        BlackDuckServerConfig blackDuckServerConfig = createBlackDuckServerConfig(intLogger);
+        return blackDuckServerConfig.createCacheHttpClient(intLogger);
+    }
+
     public BlackDuckHttpClient createBlackDuckHttpClient(IntLogger intLogger, BlackDuckServerConfig blackDuckServerConfig) {
         return blackDuckServerConfig.createBlackDuckHttpClient(intLogger);
     }
```
gkillough, line 128: Will we always want to use the cache client? If so, we should add a TODO for the second method here or deprecate it.

jamesrichard91: No I don't think it makes sense to always use the cache client. We want to use it in the notification processing because we know there is a large of amount of duplicate requests but in the rest of the code it is likely to only add a memory cost with little to no benefit.

gkillough (approved):

bamandel (approved):

ChomickiM (approved):

PR #1303, opened 2020-12-15
refactor(job-accessor): Added StaticJobAccessorV2Test.

JakeMathews:
Added test for StaticJobAccessor.

This will allow us to safely remove the DefaultJobAccessor.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
@@ -193,10 +193,6 @@ public class StaticJobAccessorV2 implements JobAccessorV2 {
     @Override
     @Transactional
     public DistributionJobModel updateJob(UUID jobId, DistributionJobRequestModel requestModel) throws AlertConfigurationException {
-        if (jobId == null) {
-            throw new AlertConfigurationException("Cannot update job with null id.");
-        }
-
         DistributionJobEntity jobEntity = distributionJobRepository.findById(jobId)
                                               .orElseThrow(() -> new AlertConfigurationException(String.format("No job exists with the id [%s]", jobId.toString())));
         OffsetDateTime createdAt = jobEntity.getCreatedAt();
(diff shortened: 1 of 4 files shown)


alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -193,17 +207,12 @@ public DistributionJobModel updateJob(UUID jobId, DistributionJobRequestModel re
 
     @Override
     @Transactional
-    public void deleteJob(UUID jobId) {
-        distributionJobRepository.deleteById(jobId);
-    }
+    public void deleteJob(UUID jobId) throws AlertConfigurationException {
+        if (jobId == null) {
+            throw new AlertConfigurationException("Cannot delete job with null id.");
+        }
```
gkillough, line 213: This may have gotten lost in translation, but I think we should skip the null check. Accessors are "internal" APIs, so if the values they receive are null, then it's a bug. I think we decided than anything without a `@Nullable` annotation should be assumed not to be null except when it comes to objects received in REST requests.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
```diff
@@ -182,9 +192,13 @@ public DistributionJobModel createJob(DistributionJobRequestModel requestModel)
 
     @Override
     @Transactional
-    public DistributionJobModel updateJob(UUID jobId, DistributionJobRequestModel requestModel) throws AlertDatabaseConstraintException {
+    public DistributionJobModel updateJob(UUID jobId, DistributionJobRequestModel requestModel) throws AlertConfigurationException {
+        if (jobId == null) {
+            throw new AlertConfigurationException("Cannot update job with null id.");
+        }
```
gkillough, line 198: We can remove this check for the same reason as below.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessorV2.java
JakeMathews: Sorry! I apparently read the opposite of what you said. I've removed the null checks and the associated tests.

gkillough (approved):

ChomickiM (approved):

jamesrichard91 (approved):

bamandel (approved):

PR #1314, opened 2020-12-18
Throw JsonParseException is no suitable sub-class can be found for deserialization

gkillough:


ekerwin (approved):
Looks perfect - I love the changes to the test instead of to the 'real' code. :)

PR #1315, opened 2020-12-18
Job UI - Missing Projects 

gkillough:
The performance impact of doing this is significant. We should definitely consider a different approach if one exists.


jamesrichard91:
The existence check only occurs if the User uses the UI to select the projects or the endpoint to retrieve the list of projects right? I think it would be reasonable for the User to expect a delay in those cases.

jamesrichard91 (approved):

PR #1323, opened 2020-12-22
JIRA: Add support for 'priority' and 'user' fields

gkillough:


channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/JiraCustomFieldResolver.java
```diff
@@ -93,9 +95,16 @@ protected Object convertValueToRequestObject(CustomFieldDefinitionModel fieldDef
...
-                return jsonObject;
+                return createJsonObject("value", innerFieldValue);
+            case CUSTOM_FIELD_TYPE_PRIORITY_VALUE:
+                return createJsonObject("name", innerFieldValue);
+            case CUSTOM_FIELD_TYPE_USER_VALUE:
+                // "name" is used for Jira Server (ignored on Jira Cloud)
+                JsonObject createUserObject = createJsonObject("name", innerFieldValue);
+                // "accountId" is used for Jira Cloud (ignored on Jira Server)
+                createUserObject.addProperty("accountId", innerFieldValue);
+                // TODO consider separating this functionality depending on which Jira channel is being used
```
gkillough, line 106: I don't think it's worth doing right now since this works and we will potentially be rewriting a good chunk of issue-tracking functionality in the coming versions.

jamesrichard91: Does adding both here work for Jira Cloud and Jira Server? Or does it error because we provided additional fields?

gkillough: I tested both with this change and it simply ignores the unknown field.

jamesrichard91 (approved):

PR #1327, opened 2021-01-04
NotificationReciever timeout

ChomickiM:
Per this mornings discussion, I've updated NotificationReciever to timeout after 2 hours of running. Also I've updated the time-to-live for the jms messages to now be 4 hours long.

I will add a new ticket to the 6.5.0 changes to change this timeout back to 2 minutes.

### DIFF
--- a/workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
+++ b/workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
@@ -22,6 +22,7 @@
  */
 package com.synopsys.integration.alert.workflow.message;
 
+import java.time.OffsetDateTime;
 import java.util.List;
 
 import org.slf4j.Logger;
@@ -39,6 +40,7 @@ import com.synopsys.integration.alert.common.event.NotificationReceivedEvent;
 import com.synopsys.integration.alert.common.persistence.accessor.NotificationAccessor;
 import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
 import com.synopsys.integration.alert.common.rest.model.AlertPagedModel;
+import com.synopsys.integration.alert.common.util.DateUtils;
 import com.synopsys.integration.alert.common.workflow.MessageReceiver;
 import com.synopsys.integration.alert.common.workflow.processor.notification.NotificationProcessor;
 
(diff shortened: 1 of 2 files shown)


workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -82,10 +88,14 @@ public void handleEvent(NotificationReceivedEvent event) {
             logger.trace("Processing Page: {}. New pages found: {}",
                 numPagesProcessed,
                 pageOfAlertNotificationModels.getTotalPages());
+            hasNotTimedOut = DateUtils.createCurrentDateTimestamp().isBefore(timeLimit);
         }
         if (numPagesProcessed == MAX_NUMBER_PAGES_PROCESSED) {
             logger.warn("Receiver reached upper page limit of pages processed: {}, exiting.", MAX_NUMBER_PAGES_PROCESSED);
         }
+        if (!hasNotTimedOut) {
```
gkillough, line 96: This double negative is a bit confusing. Can we rename the variable in the positive so the logic surrounding it is easier to follow?

jamesrichard91: If we make the change you are suggesting then the while condition will have to be negated. So I think this is easier to follow.

jamesrichard91: Or are you saying, that would be easier to follow because the "Not" will be removed from the variable name?

ChomickiM: I think I agree with Gavin. If we change it to hasTimedOut=false, we would negate in the while loop to read as "has not timed out". And it would be easier to read in the if statement as "has timed out"

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -82,10 +88,14 @@ public void handleEvent(NotificationReceivedEvent event) {
...
+            logger.warn("Receiver has timed out after 2 hours.");
```
JakeMathews, line 97: Is there any way to make this dynamic or is that not important?

JakeMathews (approved):

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
ChomickiM: Right now we are arbitrarily adding 2 hours in the timeLimit, so I think this should be okay unless someone else has an issue with it and I can save that value to a variable. This should hopefully be resolved in the changes in 6.5.0 and will no longer need this timeout. If anything I can put in an extra comment here that this should be removed once those changes are made.

jamesrichard91: This will hopefully become unnecessary in 6.5.0 and removed.

gkillough (approved):

jamesrichard91 (approved):

PR #1328, opened 2021-01-06
Utils Tests

gkillough:


alert-common/src/test/java/com/synopsys/integration/alert/common/util/UrlUtilsTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void argSurroundedByWhitespaceTest() {
+        String url = "https://awebsite.com/";
+        String formattedUrl = UrlUtils.appendTrailingSlashIfNoneExists("  " + url + "     ");
+        assertEquals(url, formattedUrl);
+    }
+
+    @Test
+    public void nullArgTest() {
+        String formattedUrl = UrlUtils.appendTrailingSlashIfNoneExists(null);
+        assertEquals("", formattedUrl);
```
bamandel, line 32: Do we want to use parameterized tests for this stuff or no? I remember a previous PR where parameterized tests were advocated for in this case.

bamandel (approved):

ChomickiM (approved):

alert-common/src/test/java/com/synopsys/integration/alert/common/util/UrlUtilsTest.java
gkillough: I wanted to use them, but we don't seem to have access to that dependency here for some reason. Definitely worth more investigation.

JakeMathews (approved):

jamesrichard91 (approved):

PR #1336, opened 2021-01-08
Job details abstraction

bamandel:
Abstracted the large amount of if statements away when converting FieldModels into JobDetailModels. Slightly modified what AutoActionable registers.


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/DescriptorProcessor.java
```diff
@@ -60,16 +61,20 @@ public DescriptorProcessor(DescriptorMap descriptorMap, ConfigurationAccessor co
         for (AutoActionable autoActionable : autoActionables) {
             DistributionChannel channel = autoActionable.getChannel();
             ChannelKey channelKey = autoActionable.getChannelKey();
+            ConfigurationAction configurationAction = allConfigurationActions.get(channelKey.getUniversalKey());
```
gkillough, line 64: Does this work? I thought the point of `AutoActionable` was to create this for classes that don't declare it?

bamandel: Every channel will need a ConfigurationAction now as they all need to do the mapping for the JobDetailModels. However, not all of these ConfigurationActions need custom testing functionality. So this is still here to retrieve the existing ConfigurationActions and add a basic TestAction.

jamesrichard91 (approved):

gkillough (approved):
My only suggestion would be to change the word "Processor" to "Extractor" or "Populator" since we have a concept of "Processing" already.

PR #1337, opened 2021-01-08
6.4.0 Test Coverage Improvements

ChomickiM:
Improve test coverage around the StaticJobAccessor, BlackDuckJobDetailsAccessor, and EmailJobDetailsAccessor.

### DIFF
--- a/alert-database/src/test/java/com/synopsys/integration/alert/database/api/StaticJobAccessorTest.java
+++ b/alert-database/src/test/java/com/synopsys/integration/alert/database/api/StaticJobAccessorTest.java
@@ -20,13 +20,29 @@ import com.synopsys.integration.alert.common.enumeration.ProcessingType;
 import com.synopsys.integration.alert.common.persistence.accessor.JobAccessor;
 import com.synopsys.integration.alert.common.persistence.model.job.DistributionJobModel;
 import com.synopsys.integration.alert.common.persistence.model.job.DistributionJobRequestModel;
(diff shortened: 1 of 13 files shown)


alert-database/src/test/java/com/synopsys/integration/alert/database/api/StaticJobAccessorTest.java
```diff
@@ -315,6 +321,154 @@ void getMatchingEnabledJobsWithFrequencyTest() {
         assertEquals(jobName, distributionJobModel.getName());
     }
 
+    @Test
```
bamandel, line 324: It looks like all of these new tests have a lot of duplication between them, we can probably cut down on that duplication.

gkillough (approved):

bamandel (approved):

alert-database/src/test/java/com/synopsys/integration/alert/database/api/StaticJobAccessorTest.java
ChomickiM: the process for each is pretty similar but each one of them is using its channel specific model's and entities. Theres a few places I think that can be consolidated like 2 of the Mockito.when's or the assertEquals. Would you prefer those moved out into other private methods?

bamandel: If we're trying to avoid using inheritance in tests, then this should be fine and is expected

jamesrichard91 (approved):

PR #1339, opened 2021-01-08
Fix map unwrapping

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/provider/state/StatefulProvider.java
```diff
@@ -54,8 +53,13 @@ public static StatefulProvider create(
         ProviderMessageContentCollector messageContentCollector
     ) {
         Map<String, ConfigurationFieldModel> keyToFieldMap = configurationModel.getCopyOfKeyToFieldMap();
-        Boolean configEnabled = MapUtils.getBoolean(keyToFieldMap, ProviderDescriptor.KEY_PROVIDER_CONFIG_ENABLED, false);
```
gkillough, line 57: This map's values are Objects 🤦

jamesrichard91 (approved):

bamandel (approved):

PR #1342, opened 2021-01-15
Processing Interfaces

gkillough:


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+        this.distributor = distributor;
+    }
+
+    public abstract void processNotifications(List<AlertNotificationModel> notifications);
+
+    public abstract void processNotifications(FrequencyType frequency, List<AlertNotificationModel> notifications);
+
+    protected abstract Map<DistributionJobModel, List<FilterableNotificationWrapper<?>>> mapJobsToNotificationsOrSomething(List<? extends FilterableNotificationWrapper<?>> filterableNotifications);
+
+    // TODO include frequency?
```
bamandel, line 55: Frequency should be passed until filtering. The filter will filter contents based off of which frequency is being used. After that, frequency is no longer needed in this processing.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationPreProcessor.java
```diff
@@ -0,0 +1,9 @@
+package com.synopsys.integration.alert.processor.api;
+
+import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
+import com.synopsys.integration.alert.processor.api.filter.FilterableNotificationWrapper;
+
+public interface NotificationPreProcessor {
```
gkillough, line 6: Any naming suggestions here? I was gonna call it `NotificationWrapper`, but "Wrapper" is already used as a noun for the returned object.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessor.java
gkillough: Frequency is required when notifications come from the accumulator, but not when they come from resending an Audit (because we even want it resent to DAILY configs in that case. Any ideas about how to separate those concerns?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationPreProcessor.java
bamandel: This is still an extractor in my opinion so I'd call this one also extractor but use the model it extracts in the name.

jamesrichard91: NotificationExtractor?

jamesrichard91: FilterableNotificationExtractor?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessor.java
bamandel: Resending audits goes through the processor again? Regardless, we shouldn't even be running the processor unless it's explicitly told to run (i.e. Daily task, real time task, or audit runs it). This would mean we already fulfill the timing aspect of frequency, so now we just need to filter the jobs and be finished with frequency. So Frequency will always need to reach filtering imo

bamandel: To actually answer your question... Audit can kick off the processor twice passing each frequency

gkillough: Hmm. Not very efficient though.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/model/CombinableModel.java
```diff
@@ -0,0 +1,70 @@
...
+    List<T> combine(T otherModel);
+
+    static <T extends CombinableModel<T>> List<T> combine(List<T> lhs, List<T> rhs) {
+        List<T> unified = ListUtils.union(lhs, rhs);
+        return combine(unified);
+    }
+
+    static <T extends CombinableModel<T>> List<T> combine(List<T> models) {
+        LinkedList<T> copyOfModels = new LinkedList<>(models);
+        int i = 0;
```
bamandel, line 40: Can we name these so that it's easier for newcomers to understand?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/model/project/BomComponentDetails.java
```diff
@@ -0,0 +1,101 @@
...
+
+public class BomComponentDetails extends AlertSerializableModel implements CombinableModel<BomComponentDetails> {
+    private final LinkableItem component;
+    private final LinkableItem componentVersion;
+    private final List<ComponentConcern> componentConcerns;
+
+    private final LinkableItem license;
+    private final String usage;
+
+    // TODO add callback url
```
bamandel, line 39: Should we add this now or is it not too important?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessor.java
```diff
@@ -0,0 +1,121 @@
...
+                                                              .stream()
+                                                              .map(providerMessageExtractor::extract)
+                                                              .reduce(ProviderMessageHolder::reduce)
+                                                              .orElse(ProviderMessageHolder.empty());
+
+        ProviderMessageHolder processedProviderMessages = processExtractedNotifications(job, extractedProviderMessages);
+        providerMessageDistributor.distribute(job, processedProviderMessages);
+    }
+
+    protected abstract Map<DistributionJobModel, List<FilterableNotificationWrapper<?>>> mapJobsToNotifications(List<? extends FilterableNotificationWrapper<?>> filterableNotifications, @Nullable FrequencyType frequency);
```
bamandel, line 101: Since NotificationProcessor appears to be per notification type, can we have a generic passed to this class and that type would be the type of FilterableNotificationWrapper or something similar?

gkillough: It isn't per notification type. I think it is called with a batch of arbitrary notifications over a specific date range.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationExtractor.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Map;
+
+import org.jetbrains.annotations.Nullable;
+
+import com.synopsys.integration.alert.common.enumeration.FrequencyType;
+import com.synopsys.integration.alert.common.persistence.model.job.DistributionJobModel;
+import com.synopsys.integration.alert.processor.api.filter.model.FilterableNotificationWrapper;
+
+public interface JobNotificationExtractor {
+    Map<DistributionJobModel, List<FilterableNotificationWrapper<?>>> mapJobsToNotifications(List<? extends FilterableNotificationWrapper<?>> filterableNotifications, @Nullable FrequencyType frequency);
```
bamandel, line 35: Would we want to use SetMap here instead of Map?

bamandel: Not sure what benefit we get from a list over set here.

gkillough: I think it more clearly expresses intent: The notifications _must_ maintain their order. If we are inserting duplicate notifications into these lists I believe that would be a bug.

bamandel (approved):

ChomickiM (approved):

JakeMathews (approved):

PR #1345, opened 2021-01-20
Channel Interfaces

gkillough:


jamesrichard91 (approved):

channel-api/src/main/java/com/synopys/integration/alert/channel/api/ChannelReceiver.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package com.synopys.integration.alert.channel.api;
+
+import com.google.gson.Gson;
+import com.synopsys.integration.alert.common.persistence.accessor.JobDetailsAccessor;
+import com.synopsys.integration.alert.common.persistence.model.job.details.DistributionJobDetailsModel;
+import com.synopsys.integration.alert.common.workflow.MessageReceiver;
+import com.synopsys.integration.alert.processor.api.distribute.DistributionEventV2;
+
+public abstract class ChannelReceiver<D extends DistributionJobDetailsModel, T, U> extends MessageReceiver<DistributionEventV2> {
```
bamandel, line 31: I may be blind, but I don't think U is used anywhere

gkillough: You're right. It's left-over from a different idea.

PR #1349, opened 2021-01-22
Implement ProjectMessage summarizer

gkillough:
Two questions that need to be answered before this can be merged:
- Do we get any value from having a separate interface for this particular class?
- What do we want to do about the "enum coupling" TODO comments?


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/summarize/DefaultProjectMessageSummarizer.java
```diff
@@ -0,0 +1,174 @@
...
+    private List<LinkableItem> constructMessageDetails(ProjectMessage projectMessage) {
+        List<LinkableItem> details = new LinkedList<>();
+        details.add(projectMessage.getProvider());
+        details.add(projectMessage.getProject());
+        projectMessage.getProjectVersion().ifPresent(details::add);
+
+        Map<Triple<ComponentConcernType, ItemOperation, ComponentConcernSeverity>, Integer> concernCounts = new LinkedHashMap<>();
+        for (BomComponentDetails bomComponent : projectMessage.getBomComponents()) {
+            for (ComponentConcern concern : bomComponent.getComponentConcerns()) {
+                Triple<ComponentConcernType, ItemOperation, ComponentConcernSeverity> concernKey = Triple.of(concern.getType(), concern.getOperation(), concern.getSeverity());
```
JakeMathews, line 115: Does it make sense to have a data class here? I find the `Triple` makes the code more difficult to read. It seems rather contained though.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/summarize/DefaultProjectMessageSummarizer.java
```diff
@@ -0,0 +1,160 @@
...
+        );
+    }
+
+    private List<LinkableItem> constructMessageDetails(ProjectMessage projectMessage) {
+        List<LinkableItem> details = new LinkedList<>();
+        details.add(projectMessage.getProvider());
+        details.add(projectMessage.getProject());
+        projectMessage.getProjectVersion().ifPresent(details::add);
+
+        Map<Triple<ComponentConcernType, ItemOperation, ComponentConcernSeverity>, Integer> concernCounts = new LinkedHashMap<>();
```
bamandel, line 112: I think we should try avoiding Pair and Triple. When we make a class for these fields, we give them more of a purpose and can add functionality within the class to help with processing

JakeMathews (approved):

bamandel (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/summarize/DefaultProjectMessageSummarizer.java
gkillough: I was going back and forth on it since the data class would not be exposed as part of any API, but I think you're right that readability is somewhat degraded.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/summarize/DefaultProjectMessageSummarizer.java
gkillough: You and Jake make good points. I was concerned about having data classes that don't get exposed as part of an API, but I think it would help readability.

bamandel: Is this the class for a private inner class?

gkillough: Yes, but I'm also concerned about bloating the code.

ChomickiM (approved):

psantos1113 (approved):

PR #1351, opened 2021-01-22
Docker swarm deployment fix

ChomickiM:
These changes will fix deployment for both docker swarm and docker-swarm using an external database. For an external DB you now need to provide admin username/password as secrets. I have updated the README to reflect those new requirements and cleaned up some of the install procedure that we discovered was incorrect.


gkillough (approved):

docker-entrypoint.sh
```diff
@@ -49,6 +49,9 @@ then
   echo "Alert Database user secret set; using value from secret."
   alertDatabaseUser=$(cat $dockerSecretDir/ALERT_DB_USERNAME)
   export ALERT_DB_USERNAME=$alertDatabaseUser;
+
+  alertDatabaseAdminUser=$alertDatabaseUser;
```
bamandel, line 53: Is this set to the appropriate value? Shouldn't it use the Admin user?

docker-entrypoint.sh
```diff
@@ -57,6 +60,9 @@ then
   echo "Alert Database password secret set; using value from secret."
   alertDatabasePassword=$(cat $dockerSecretDir/ALERT_DB_PASSWORD)
   export ALERT_DB_PASSWORD=$alertDatabasePassword;
+
+  alertDatabaseAdminPassword=$alertDatabasePassword;
```
bamandel, line 64: Is this set to the appropriate value? Shouldn't it use the Admin user?

docker-entrypoint.sh
psantos1113: The entrypoint has to support the exernal database cases, the helm case which requires a separate password for the admin user due to the database image used,  and the case where we are using the internal database with docker swarm.  So for the internal database in docker swarm the database user and admin user are the same.  For helm with the internal database the admin user name is postgres but the password can be set.  For the external database case you have to set both the user name and password.  

The placement of where the environment variables are set is important to handle those cases.

docker-entrypoint.sh
psantos1113: see the comment above

psantos1113 (approved):

PR #1352, opened 2021-01-22
Bm filtering

bamandel:
Implemented the job mapping which also covers the job filtering for notifications. There can definitely be some improvements to the query and collapsing before queries, but it seems to work now.


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessorV2.java
```diff
@@ -89,25 +86,26 @@ public final void processNotifications(List<AlertNotificationModel> notification
                                                                                        .stream()
                                                                                        .map(filterableNotificationExtractor::wrapNotification)
                                                                                        .collect(Collectors.toList());
-        Map<DistributionJobModel, List<FilterableNotificationWrapper<?>>> jobsToNotifications = jobNotificationExtractor.mapJobsToNotifications(filterableNotifications, frequency);
-        for (Map.Entry<DistributionJobModel, List<FilterableNotificationWrapper<?>>> jobToNotifications : jobsToNotifications.entrySet()) {
-            processJobNotifications(jobToNotifications.getKey(), jobToNotifications.getValue());
+        Map<FilteredDistributionJobResponseModel, List<FilterableNotificationWrapper<?>>> jobsToNotifications = jobNotificationExtractor.mapJobsToNotifications(filterableNotifications, frequency);
+        for (FilteredDistributionJobResponseModel filteredDistributionJobResponseModel : jobsToNotifications.keySet()) {
```
bamandel, line 90: Would it be better to loop over the entry set, or leave it as is?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -183,40 +182,40 @@ public StaticJobAccessor(
...
         } else if (!vulnerabilitySeverities.isEmpty()) {
-            distributionJobEntities = distributionJobRepository.findMatchingEnabledJobsWithVulnerabilitySeverities(frequencyType.name(), notificationType.name(), projectName, vulnerabilitySeverities);
+            distributionJobEntities = distributionJobRepository.findMatchingEnabledJobsWithVulnerabilitySeverities(frequencyTypes, notificationType.name(), projectName, vulnerabilitySeverities);
         } else {
-            distributionJobEntities = distributionJobRepository.findMatchingEnabledJobs(frequencyType.name(), notificationType.name(), projectName);
+            distributionJobEntities = distributionJobRepository.findMatchingEnabledJobs(frequencyTypes, notificationType.name(), projectName);
         }
 
-        // TODO running project name pattern checks in java code, try to do this in SQL instead
+        // TODO running project name pattern checks in java code, try to do this in SQL instead (Won't need to return DistributionJobEntity anymore if this happens
```
bamandel, line 208: Doing this in SQL requires us to implement some JPA repositories, leaving it as an improvement in the future

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessorV2.java
gkillough: Considering you call `jobsToNotifications.get(filteredDistributionJobResponseModel));` later on, it would be better to iterate over `.entrySet()`. That would be the most performant anyway while making the least assumptions about the Map implementation.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/DefaultJobNotificationExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+@Component
+public class DefaultJobNotificationExtractor implements JobNotificationExtractor {
+    private JobAccessor jobAccessor;
+
+    @Autowired
+    public DefaultJobNotificationExtractor(JobAccessor jobAccessor) {
+        this.jobAccessor = jobAccessor;
+    }
+
+    /*
```
gkillough, line 53: Should we make this a JavaDoc comment so IDEs expose it?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
jamesrichard91: I don't think the project name pattern can matched in SQL. Java and SQL use different pattern formats that are not compatible.

gkillough: If we use a native query we could potentially pull it off: 
https://www.postgresql.org/docs/12/functions-matching.html#FUNCTIONS-SIMILARTO-REGEXP

bamandel: Ya the SIMILAR TO would be the functionality i'd be adding when implementing. If that doesn't work then I'll have to continue using the java filtering

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/DefaultJobNotificationExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+            for (FilteredDistributionJobResponseModel filteredDistributionJobResponseModel : filteredDistributionJobResponseModels) {
+                List<FilterableNotificationWrapper<?>> set = groupedFilterableNotifications.computeIfAbsent(filteredDistributionJobResponseModel, ignoredKey -> new LinkedList<>());
+                set.add(filterableNotificationWrapper);
+            }
+        }
+        return groupedFilterableNotifications;
+    }
+
+    private List<FilteredDistributionJobResponseModel> retrieveMatchingJobs(FilterableNotificationWrapper filterableNotificationWrapper, @Nullable FrequencyType frequencyType) {
+        // If a frequency is null, we assume they want all frequency types as a frequency is required.
```
gkillough, line 79: We should just change the method signature to accept a list of frequencies now.

gkillough: If it's empty then there's nothing to do.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
bamandel: Gavin has pionted out that POSIX is needed over SIMILAR TO, but it still looks possible

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/DefaultJobNotificationExtractor.java
bamandel: I'm thinking of bubbling this all the way up to the NotificationProcessorV2 instead of the processNotifications methods

psantos1113 (approved):

PR #1353, opened 2021-01-25
Email Channel v2

gkillough:


channel-api/src/main/java/com/synopys/integration/alert/channel/api/DistributionEventReceiver.java
```diff
@@ -22,26 +22,53 @@
...
 import com.google.gson.Gson;
+import com.synopsys.integration.alert.common.descriptor.accessor.AuditAccessor;
+import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.persistence.accessor.JobDetailsAccessor;
 import com.synopsys.integration.alert.common.persistence.model.job.details.DistributionJobDetailsModel;
 import com.synopsys.integration.alert.common.workflow.MessageReceiver;
 import com.synopsys.integration.alert.processor.api.distribute.DistributionEventV2;
 
 public abstract class DistributionEventReceiver<D extends DistributionJobDetailsModel> extends MessageReceiver<DistributionEventV2> {
+    private final AuditAccessor auditAccessor;
```
gkillough, line 36: Audit handling is tracked in a different internal issue and will be resolved separately.

channel/src/main/java/com/synopsys/integration/alert/channel/email2/util/ProjectMessageToMessageContentGroupConversionUtils.java
```diff
@@ -0,0 +1,145 @@
...
+import com.synopsys.integration.alert.common.message.model.MessageContentGroup;
+import com.synopsys.integration.alert.common.message.model.ProviderMessageContent;
+import com.synopsys.integration.alert.processor.api.extract.model.project.BomComponentDetails;
+import com.synopsys.integration.alert.processor.api.extract.model.project.ComponentConcern;
+import com.synopsys.integration.alert.processor.api.extract.model.project.ComponentConcernType;
+import com.synopsys.integration.alert.processor.api.extract.model.project.ProjectMessage;
+import com.synopsys.integration.alert.processor.api.extract.model.project.ProjectOperation;
+
+// TODO remove this in 7.0.0
+public final class ProjectMessageToMessageContentGroupConversionUtils {
```
gkillough, line 44: This class is just to convert `ProjectMessages` into `MessageContentGroups` for backward-compatibility.

channel/src/main/java/com/synopsys/integration/alert/channel/email2/EmailChannelMessageSender.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    public MessageResult sendMessages(EmailJobDetailsModel emailJobDetails, List<EmailChannelMessageModel> emailMessages) throws AlertException {
+        EmailProperties emailProperties = new EmailProperties(retrieveGlobalEmailConfig());
+        EmailMessagingService emailMessagingService = new EmailMessagingService(emailProperties, freemarkerTemplatingService);
+
+        EmailAttachmentFormat attachmentFormat = EmailAttachmentFormat.getValueSafely(emailJobDetails.getAttachmentFileType());
+
+        for (EmailChannelMessageModel message : emailMessages) {
+            // FIXME this does not currently account for project-summary messages which do not include a source
```
gkillough, line 93: This is tracked internally and will be resolved in a subsequent PR.

ChomickiM (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -101,7 +101,7 @@ public StaticJobAccessor(
         DistributionJobRepository distributionJobRepository,
         BlackDuckJobDetailsAccessor blackDuckJobDetailsAccessor,
         AzureBoardsJobDetailsAccessor azureBoardsJobDetailsAccessor,
-        EmailJobDetailsAccessor emailJobDetailsAccessor,
+        DefaultEmailJobDetailsAccessor emailJobDetailsAccessor,
```
bamandel, line 104: Any reason we are injecting the implementation instead of the interface? I imagine we would want to inject the most basic form of the parameter

channel/src/main/java/com/synopsys/integration/alert/channel/email2/EmailChannelMessageSender.java
bamandel: Can this be resolved now with the source field?

JakeMathews (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -86,7 +86,7 @@
     private final DistributionJobRepository distributionJobRepository;
     private final BlackDuckJobDetailsAccessor blackDuckJobDetailsAccessor;
     private final AzureBoardsJobDetailsAccessor azureBoardsJobDetailsAccessor;
-    private final EmailJobDetailsAccessor emailJobDetailsAccessor;
+    private final DefaultEmailJobDetailsAccessor emailJobDetailsAccessor;
```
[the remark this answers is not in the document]
JakeMathews: Brian always beats me to it ☹️

channel/src/main/java/com/synopsys/integration/alert/channel/email2/EmailChannelMessageSender.java
```diff
@@ -0,0 +1,155 @@
...
+            // We trust that the file was created correctly, so the path should be correct.
+            emailTarget.setAttachmentFilePath(attachmentFile.getPath());
+        }
+        return optionalAttachmentFile;
+    }
+
+    private String getImagePath(String imageFileName) throws AlertException {
+        String imagesDirectory = alertProperties.getAlertImagesDir();
+        if (StringUtils.isNotBlank(imagesDirectory)) {
+            return imagesDirectory + "/" + imageFileName;
```
bamandel, line 150: Should we be using some sort of path builder to build these paths instead of strings? I think you can use Paths to merge these together.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
gkillough: The interface is creating a contract that `DistributionEventReceiver` requires whereas `DefaultEmailJobDetailsAccessor` is the "whole" accessor. I suspect this will be an area that would be more clear if/when we separate things into microservice-like sub-projects.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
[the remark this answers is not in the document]
gkillough: Replied to Brain's

bamandel (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/email2/EmailChannelMessageSender.java
gkillough: Let's add it as an improvement ticket internally. I'm trying to stay as true as possible (i.e. copy/paste) to the same functionality as before in this PR.

channel/src/main/java/com/synopsys/integration/alert/channel/email2/EmailChannelMessageSender.java
gkillough: Yes, once that PR is merged I will add it.

channel/src/main/java/com/synopsys/integration/alert/channel/email2/util/EmailAddressGatherer.java
```diff
@@ -0,0 +1,124 @@
...
+        emailAddresses.addAll(emailJobDetails.getAdditionalEmailAddresses());
+        if (emailJobDetails.isAdditionalEmailAddressesOnly()) {
+            return emailAddresses;
+        }
+
+        Optional<Long> optionalBlackDuckGlobalConfigId = jobAccessor.getJobById(emailJobDetails.getJobId())
+                                                             .map(DistributionJobModel::getBlackDuckGlobalConfigId);
+        if (optionalBlackDuckGlobalConfigId.isPresent()) {
+            Set<String> providerEmailAddresses = gatherProviderEmailAddresses(emailJobDetails.isProjectOwnerOnly(), projectHrefs, optionalBlackDuckGlobalConfigId.get());
+            emailAddresses.addAll(providerEmailAddresses);
```
jamesrichard91, line 68: According to the description of the Project Owner Only field, this does not seem correct. I think this is a reasonable change but the Project Owner Only field description should be updated to mention the additional email addresses will be included as well. We should also add a release note and make sure the documentation is updated to match.

jamesrichard91 (changes requested):

channel/src/main/java/com/synopsys/integration/alert/channel/email2/util/EmailAddressGatherer.java
gkillough: Our API constrains it so that these cannot both be true, but I think you raising this point indicates that we shouldn't make that assumption here. I don't think we need to change the feature, I'll just update the logic to be more explicit about mutual exclusion.

jamesrichard91 (approved):

PR #1356, opened 2021-01-26
logging path fixes

ChomickiM:
Since 6.0.0 we were logging incorrectly to the root /log/ directory of the alert containers. This change will set that default path explicitly to the location used previously in Alert 5.x and below. 

Additionally we've exposed ALERT_LOG_FILE_PATH as an environment variable in the application.properties but are not sharing this currently with customers. Likewise, for local deployments of Alert through the IDE and through tests I've updated it to use the log directory so it continues to the project's log directory.

### DIFF
--- a/src/main/resources/logback-spring.xml
+++ b/src/main/resources/logback-spring.xml
@@ -7,10 +7,10 @@
         <encoder>
 			<pattern>${FILE_LOG_PATTERN}</pattern>
 		</encoder>
-		<file>log/alert_audit.log</file>
+		<file>${LOG_PATH:-log}/alert_audit.log</file>
 		<rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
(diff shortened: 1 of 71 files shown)


gkillough (approved):

src/main/resources/logback-spring.xml
```diff
@@ -7,10 +7,10 @@
         <encoder>
 			<pattern>${FILE_LOG_PATTERN}</pattern>
 		</encoder>
-		<file>log/alert_audit.log</file>
+		<file>${LOG_PATH:-log}/alert_audit.log</file>
```
jamesrichard91, line 10: Where is LOG_PATH defined? Is that a logback variable?

bamandel (approved):

src/main/resources/logback-spring.xml
ChomickiM: LOG_PATH is the environment variable for logging.file.path in the application.properties thats used by spring

JakeMathews (approved):

jamesrichard91 (approved):

PR #1359, opened 2021-01-27
Bm noti extractor

bamandel:
Added only vulnerability extractor as a first step for supporting all existing notification types.


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/DefaultFilterableNotificationExtractor.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        logger.warn("Did not find extractor for notification type: {}", notificationTypeString);
+        return Optional.empty();
+    }
+
+    private Optional<NotificationExtractor> getNotificationExtractor(NotificationType notificationType) {
+        return notificationExtractors.stream()
+                   .filter(notificationExtractor -> notificationExtractor.getNotificationType() == notificationType)
+                   .findFirst();
```
bamandel, line 74: Would people prefer i create a map for this so that I don't iterate over the extractors for every notification?

jamesrichard91: That would be a great improvement!

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/FilterableNotificationExtractor.java
```diff
@@ -22,10 +22,12 @@
  */
 package com.synopsys.integration.alert.processor.api.filter;
 
+import java.util.Optional;
+
 import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
 import com.synopsys.integration.alert.processor.api.filter.model.FilterableNotificationWrapper;
 
 public interface FilterableNotificationExtractor {
```
gkillough, line 30: I think we decided we don't need an interface when there will only be one implementation, so this class can be deleted.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/DefaultFilterableNotificationExtractor.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
+import com.synopsys.integration.alert.processor.api.filter.extractor.NotificationExtractor;
+import com.synopsys.integration.alert.processor.api.filter.model.FilterableNotificationWrapper;
+import com.synopsys.integration.blackduck.api.manual.enumeration.NotificationType;
+
+@Component
+public class DefaultFilterableNotificationExtractor implements FilterableNotificationExtractor {
```
gkillough, line 39: Once the interface is deleted, let's rename this class to just `FilterableNotificationExtractor `

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/NotificationExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.processor.api.filter.extractor;
+
+import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
+import com.synopsys.integration.alert.processor.api.filter.model.FilterableNotificationWrapper;
+import com.synopsys.integration.blackduck.api.manual.enumeration.NotificationType;
+
+public abstract class NotificationExtractor {
```
gkillough, line 29: Can we add `<T extends NotificationContent>` to this class, then return `FilterableNotificationWrapper<T>` from the abstract method?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+    public VulnerabilityNotificationExtractor(Gson gson) {
+        super(NotificationType.VULNERABILITY);
+        this.gson = gson;
+    }
+
+    @Override
+    public FilterableNotificationWrapper convertToFilterableNotificationWrapper(AlertNotificationModel alertNotificationModel) {
+        VulnerabilityNotificationContent vulnerabilityNotificationContent = gson.fromJson(alertNotificationModel.getContent(), VulnerabilityNotificationContent.class);
+
+        List<String> affectedProjectNames = vulnerabilityNotificationContent.getAffectedProjectVersions()
```
gkillough, line 57: I wonder if this notification being related to multiple projects will allow other projects that are not configured to slip through processing.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/NotificationExtractor.java
bamandel: Yes I think i'm going to slightly modify this class to use generics and do a little more in the parent class

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
```diff
@@ -0,0 +1,94 @@
...
+                                                .collect(Collectors.toList());
+        List<String> applicableSeverityTypes = getApplicableSeverityTypes(vulnerabilityNotificationContent)
+                                                   .stream()
+                                                   .map(VulnerabilitySeverityType::name)
+                                                   .collect(Collectors.toList());
+
+        return FilterableNotificationWrapper.vulnerability(alertNotificationModel, vulnerabilityNotificationContent, affectedProjectNames, applicableSeverityTypes);
+    }
+
+    private Set<VulnerabilitySeverityType> getApplicableSeverityTypes(VulnerabilityNotificationContent vulnerabilityNotificationViewContent) {
```
gkillough, line 69: Why not have this return the names so you don't have to have the extra streaming on the `Set` this method returns?

bamandel: I actually need to convert this to a list later on, so I will need to stream it afterwards regardless. We could maybe convert the List<String> for policyNames and vulnSeverities to Set<String> As it's used for filtering and we don't want duplication anyway

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
bamandel: Next PR will address the issue with vuln notifications having multiple projects

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
gkillough: You don't need to stream. You can use a constructor to create the list out of it. Or you could call `Stream.unique()` and then collect to a list.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/model/FilterableNotificationWrapper.java
```diff
@@ -29,45 +29,45 @@
...
-        return new FilterableNotificationWrapper<>(notificationModel, notificationContent, projectName, List.of(), List.of());
+        return new FilterableNotificationWrapper<>(notificationModel, notificationContent, List.of(projectName), List.of(), List.of());
     }
 
     public static FilterableNotificationWrapper<? extends NotificationContentComponent> projectless(AlertNotificationModel notificationModel, NotificationContentComponent notificationContent) {
         return new FilterableNotificationWrapper<>(notificationModel, notificationContent, null, List.of(), List.of());
     }
 
-    public FilterableNotificationWrapper(AlertNotificationModel alertNotificationModel, T notificationContent, String projectName, List<String> policyNames, List<String> vulnerabilitySeverities) {
+    public FilterableNotificationWrapper(AlertNotificationModel alertNotificationModel, T notificationContent, List<String> projectNames, List<String> policyNames, List<String> vulnerabilitySeverities) {
```
psantos1113, line 62: If the static methods are the only methods that are used to create this object then the constructor could be private.

ChomickiM (approved):

gkillough (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/model/FilterableNotificationWrapper.java
bamandel: I will make this change in my next PR as there may be some further modifications to how these work

PR #1360, opened 2021-01-27
Jm comment limit handling

JakeMathews:
Fixes IALERT-2139

This fixes a few issues
- `MessageSplitter` could return a `splitIndex` equal to 0. This causes a `StackOverflowError`
- `MessageSplitter` could return a `splitIndex` greater than the limit provided (likely source of the issue)
- The description generated by `IssueTrackerMessageParser` was not split when its contents were already longer than the limit


psantos1113 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/message/IssueTrackerMessageParser.java
```diff
@@ -40,13 +42,20 @@
     private final int titleSizeLimit;
     private final int messageSizeLimit;
 
+    // TODO: What if description length differs from comment length? Should description length be accepted here?
```
gkillough, line 45: Let's create a Spike for this. This will become relevant when we begin implementing issue-trackers in the new channel framework.

jamesrichard91 (approved):

gkillough (approved):

alert-common/src/test/java/com/synopsys/integration/alert/common/channel/message/MessageSplitterTest.java
```diff
@@ -0,0 +1,32 @@
...
+            "category (ADD) - component-value \n",
+            // Can produce a StackOverflowError if the split index is 0
+            " testing_of_splitting_index_0 ",
+            "",
+            " "
+        ));
+
+        for (String splitMessage : splitMessages) {
+            System.out.println("'" + splitMessage + "'");
+            Assertions.assertTrue(splitMessage.length() <= limit,
```
ChomickiM, line 26: I believe we usually will import the assertions we need like assertTrue and not the entire Assertions package

ChomickiM (approved):

bamandel (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/issuetracker/message/IssueTrackerMessageParser.java
JakeMathews: Upon looking at this further. It appears the message boards use the same limit for both description and comments. I think this should be addressed if it ever becomes an issue. I'm going to remove the TODO.

PR #1361, opened 2021-01-28
Revert Old Message Field Name Changes

gkillough:
These classes will be replaced and will only be used for backward-compatibility with email attachments.


bamandel (approved):
It looks good, but why do we need to revert this? Does email use some sort of serialization that requires these names?

jamesrichard91 (approved):

gkillough:
I'm gonna hold off on merging this until #1360 is merged because I expect some merge conflicts.

ChomickiM (approved):

PR #1367, opened 2021-01-29
ProjectNotificationExtractor

gkillough:
- Fixed generics
- Implemented extraction for ProjectNotificationContent


jamesrichard91 (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/ProjectNotificationExtractor.java
```diff
@@ -0,0 +1,46 @@
...
+@Component
+public class ProjectNotificationExtractor extends NotificationExtractor<ProjectNotificationContent> {
+    @Autowired
+    public ProjectNotificationExtractor(Gson gson) {
+        super(NotificationType.PROJECT, ProjectNotificationContent.class, gson);
+    }
+
+    @Override
+    protected FilterableNotificationWrapper convertToFilterableNotificationWrapper(AlertNotificationModel alertNotificationModel, ProjectNotificationContent projectNotificationContent) {
+        return FilterableNotificationWrapper.project(alertNotificationModel, projectNotificationContent, projectNotificationContent.getProjectName());
```
bamandel, line 43: Can we modify FilterableNotificationWrapper further so that we don't need to pass AlertNotificationmodel and its subset of ProjectNotificationContent? I don't think these extractors will ever need AlertNotificationmodel

gkillough: I think we will need it for instantiating the `StatefulProvider` (or new equivalent) eventually.

bamandel (approved):

PR #1368, opened 2021-01-29
Bm vuln parsing

bamandel:
Instead of accepting that vulnerability notifications apply to multiple projects, we now break up the notification into individual notifications per project. This should help with querying the DB and simplifies some models.


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
```diff
@@ -49,14 +48,27 @@ public VulnerabilityNotificationExtractor(Gson gson) {
...
-                                                .stream()
-                                                .map(AffectedProjectVersion::getProjectName)
-                                                .collect(Collectors.toList());
+    public List<FilterableNotificationWrapper> convertToFilterableNotificationWrapper(AlertNotificationModel alertNotificationModel, VulnerabilityNotificationContent vulnerabilityNotificationContent) {
         List<String> applicableSeverityTypes = new LinkedList<>(getApplicableSeverityTypes(vulnerabilityNotificationContent));
 
-        return FilterableNotificationWrapper.vulnerability(alertNotificationModel, vulnerabilityNotificationContent, affectedProjectNames, applicableSeverityTypes);
+        return breakDownIntoSeparateNotifications(vulnerabilityNotificationContent).stream()
+                   .map(vulnerabilitySingleProjectNotificationContent -> FilterableNotificationWrapper.vulnerability(
+                       // This leaves the AlertNotificationModel as the original but modifies the NotificationContent field
```
bamandel, line 56: Is this what we would want? Or would people expect the content returned here to exactly mimic the contents of AlertNotificationModel?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilitySingleProjectNotificationContent.java
```diff
@@ -0,0 +1,107 @@
...
+package com.synopsys.integration.alert.processor.api.filter.extractor;
+
+import java.util.List;
+
+import com.synopsys.integration.blackduck.api.manual.component.AffectedProjectVersion;
+import com.synopsys.integration.blackduck.api.manual.component.NotificationContentComponent;
+import com.synopsys.integration.blackduck.api.manual.component.VulnerabilityNotificationContent;
+import com.synopsys.integration.blackduck.api.manual.component.VulnerabilitySourceQualifiedId;
+
+public class VulnerabilitySingleProjectNotificationContent extends NotificationContentComponent {
```
gkillough, line 32: This becomes risky if the API of the original notification class changes.

bamandel: True but the only constructor requires you pass in the existing object. So if something changes there, it should break the build when mapping.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
```diff
@@ -49,14 +48,27 @@ public VulnerabilityNotificationExtractor(Gson gson) {
...
-        List<String> affectedProjectNames = vulnerabilityNotificationContent.getAffectedProjectVersions()
-                                                .stream()
-                                                .map(AffectedProjectVersion::getProjectName)
-                                                .collect(Collectors.toList());
+    public List<FilterableNotificationWrapper> convertToFilterableNotificationWrapper(AlertNotificationModel alertNotificationModel, VulnerabilityNotificationContent vulnerabilityNotificationContent) {
         List<String> applicableSeverityTypes = new LinkedList<>(getApplicableSeverityTypes(vulnerabilityNotificationContent));
 
-        return FilterableNotificationWrapper.vulnerability(alertNotificationModel, vulnerabilityNotificationContent, affectedProjectNames, applicableSeverityTypes);
+        return breakDownIntoSeparateNotifications(vulnerabilityNotificationContent).stream()
+                   .map(vulnerabilitySingleProjectNotificationContent -> FilterableNotificationWrapper.vulnerability(
```
gkillough, line 55: This seems like it should be done in the original loop.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
```diff
@@ -49,14 +48,27 @@ public VulnerabilityNotificationExtractor(Gson gson) {
...
+                   ))
+                   .collect(Collectors.toList());
+    }
+
+    // Separating this notification to be per project should fix a bug with alerts being sent about unrelated projects
+    private List<VulnerabilitySingleProjectNotificationContent> breakDownIntoSeparateNotifications(VulnerabilityNotificationContent vulnerabilityNotificationContent) {
+        return vulnerabilityNotificationContent.getAffectedProjectVersions()
+                   .stream()
+                   .map(affectedProjectVersion -> new VulnerabilitySingleProjectNotificationContent(vulnerabilityNotificationContent, affectedProjectVersion))
+                   .distinct()
```
gkillough, line 70: This shouldn't be necessary.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
bamandel: Wow don't know how I missed that, good catch

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
gkillough: We shouldn't need to use the `String content` from the `AlertNotificationModel` since the strongly-typed content is included.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/extractor/VulnerabilityNotificationExtractor.java
```diff
@@ -49,14 +48,21 @@ public VulnerabilityNotificationExtractor(Gson gson) {
...
         List<String> applicableSeverityTypes = new LinkedList<>(getApplicableSeverityTypes(vulnerabilityNotificationContent));
 
-        return FilterableNotificationWrapper.vulnerability(alertNotificationModel, vulnerabilityNotificationContent, affectedProjectNames, applicableSeverityTypes);
+        // Separating this notification to be per project should fix a bug with alerts being sent about unrelated projects
+        return vulnerabilityNotificationContent.getAffectedProjectVersions()
+                   .stream()
+                   .map(affectedProjectVersion -> new VulnerabilitySingleProjectNotificationContent(vulnerabilityNotificationContent, affectedProjectVersion))
+                   .map(vulnerabilitySingleProjectNotificationContent -> FilterableNotificationWrapper.vulnerability(
+                       // This leaves the AlertNotificationModel as the original but modifies the NotificationContent field
+                       alertNotificationModel,
```
jamesrichard91, line 60: So every FilterableNotificationWrapper has a copy of the AlertNotificationModel? Couldn't this become a memory problem if the notification content is large?

bamandel: This is a good point, perhaps we should pass just an ID

gkillough: The consumers of this class need more than an id, but do not need the original `content` string. Whatever we decide should get its own PR though.

JakeMathews (approved):

PR #1369, opened 2021-01-29
ProviderMessageExtractor

gkillough:
- Delegator
- (Abstract) Extractor
- `ProjectNotificationMessageExtractor`

### DIFF
--- a/processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
+++ b/processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
@@ -24,7 +24,6 @@ package com.synopsys.integration.alert.processor.api.extract;
 
 import java.util.EnumMap;
 import java.util.List;
-import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -49,18 +48,18 @@ public final class ProviderMessageExtractionDelegator {
         this.notificationTypeToExtractor = initializeExtractorMap(providerMessageExtractors);
     }
 
-    public Optional<ProviderMessageHolder> extract(FilterableNotificationWrapper filteredNotification) {
+    public ProviderMessageHolder extract(FilterableNotificationWrapper filteredNotification) {
         String notificationTypeString = filteredNotification.extractNotificationType();
         NotificationType filteredNotificationType = EnumUtils.getEnum(NotificationType.class, notificationTypeString);
         if (null == filteredNotificationType) {
             logger.warn("Notification did not match any existing notification type: {}", notificationTypeString);
-            return Optional.empty();
+            return ProviderMessageHolder.empty();
         }
 
         ProviderMessageExtractor providerMessageExtractor = notificationTypeToExtractor.get(filteredNotificationType);
         if (null == providerMessageExtractor) {
             logger.warn("No matching extractor for notification type: {}", notificationTypeString);
-            return Optional.empty();
+            return ProviderMessageHolder.empty();
         }
 
         return providerMessageExtractor.extract(filteredNotification);
--- a/processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
+++ b/processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
@@ -24,8 +24,6 @@ package com.synopsys.integration.alert.processor.api.extract;
 
 import java.util.EnumMap;
 import java.util.List;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.synopsys.integration.alert.processor.api.detail.ProviderMessageHolder;
+import com.synopsys.integration.alert.processor.api.filter.model.FilterableNotificationWrapper;
+import com.synopsys.integration.blackduck.api.manual.enumeration.NotificationType;
+
+@Component
+public final class ProviderMessageExtractionDelegator {
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private final EnumMap<NotificationType, ProviderMessageExtractor> notificationTypeToExtractor;
```
bamandel, line 44: I'm thinking we may want a class that enforces that a NotificationExtractor and a ProviderMessageExtractor both exist. (Unless the mappings won't be 1 to 1)

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
```diff
@@ -0,0 +1,74 @@
...
+        ProviderMessageExtractor providerMessageExtractor = notificationTypeToExtractor.get(filteredNotificationType);
+        if (null == providerMessageExtractor) {
+            logger.warn("No matching extractor for notification type: {}", notificationTypeString);
+            return ProviderMessageHolder.empty();
+        }
+
+        return providerMessageExtractor.extract(filteredNotification);
+    }
+
+    private EnumMap<NotificationType, ProviderMessageExtractor> initializeExtractorMap(List<ProviderMessageExtractor> providerMessageExtractors) {
```
bamandel, line 68: Can we add this to the DataStructureUtils class? Will be convenient for future use.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractor.java
```diff
@@ -22,10 +22,52 @@
...
+                   .orElse(ProviderMessageHolder.empty());
+    }
+
+    protected abstract ProviderMessageHolder extract(FilterableNotificationWrapper filteredNotification, T notificationContent);
 
-public interface ProviderMessageExtractor {
-    ProviderMessageHolder extract(FilterableNotificationWrapper filteredNotification);
+    private Optional<T> castToStrongType(NotificationContentComponent notificationContent) {
+        try {
+            T stronglyTypedContent = notificationContentClass.cast(notificationContent);
```
bamandel, line 65: A mapping between extractor and this implementation will ensure these match up

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
gkillough: If we end up using the changes in #1368, then they won't be the same constraints.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractor.java
```diff
@@ -22,10 +22,52 @@
...
+            return Optional.of(stronglyTypedContent);
+        } catch (ClassCastException e) {
+            logger.error("Failed to cast the notification content class from {} to {}", notificationContent.getClass().getSimpleName(), notificationContentClass.getSimpleName(), e);
+            return Optional.empty();
+        }
```
bamandel, line 70: I think this is a case of us wanting to fail fast, if we have code that doesn't match expected inputs we should know about it immediately.

bamandel: Hiding it in a log message may mean it sneaks into production

gkillough: You're right, I think this would be a bug. I'll get rid of the exception handling.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/ProviderMessageExtractionDelegator.java
gkillough: Also, I'm not sure we could pull it off without separate `NotificationProcessors` for each different notification type.

bamandel (approved):

jamesrichard91 (approved):

PR #1377, opened 2021-02-02
Rule Violation Details Extractor

gkillough:


jamesrichard91 (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/model/DetailedNotificationContent.java
```diff
@@ -47,9 +48,9 @@ public static DetailedNotificationContent policy(
         AlertNotificationModel notificationModel,
         NotificationContentComponent notificationContent,
         String projectName,
-        List<String> policyNames
+        String policyName
```
bamandel, line 51: This might not necessarily break the DB queries, but we may want to modify the queries to now use = instead of IN

bamandel (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/model/DetailedNotificationContent.java
gkillough: Yeah, for now I just changed this static initializer. Eventually we can update the filtering. I want to get everything working first.

PR #1381, opened 2021-02-03
LicenseLimitNotificationDetailExtractor Test

gkillough:


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationMapper.java
```diff
@@ -88,9 +88,9 @@ public JobNotificationMapper(JobAccessor jobAccessor) {
         FilteredDistributionJobRequestModel filteredDistributionJobRequestModel = new FilteredDistributionJobRequestModel(
             frequencyTypes,
             EnumUtils.getEnum(NotificationType.class, contentWrapper.extractNotificationType()),
-            detailedNotificationContent.getProjectName(),
+            detailedNotificationContent.getProjectName().orElse(null),
             detailedNotificationContent.getVulnerabilitySeverities(),
-            detailedNotificationContent.getPolicyNames()
+            detailedNotificationContent.getPolicyName().map(List::of).orElse(List.of())
```
gkillough, line 93: This is a different API that we can update separately.

jamesrichard91: Just to be clear, DetailedNotificationContent is for one policy and FilteredDistributionJobRequestModel is for a list of policies?
Is that what you are referring to that we can update separately?

gkillough: Yes. The database API requests a `List`, but the processing API requests a `String`.

PR #1391, opened 2021-02-05
Fixing a bug sonarqube found

bamandel:


alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -210,7 +211,10 @@ public StaticJobAccessor(
                    .filter(distributionJobEntity -> !(distributionJobEntity.getBlackDuckJobDetails().getFilterByProject() &&
                                                           distributionJobEntity.getBlackDuckJobDetails().getProjectNamePattern() != null &&
                                                           !Pattern.matches(distributionJobEntity.getBlackDuckJobDetails().getProjectNamePattern(), projectName) &&
-                                                          !distributionJobEntity.getBlackDuckJobDetails().getBlackDuckJobProjects().contains(projectName))
+                                                          !distributionJobEntity.getBlackDuckJobDetails().getBlackDuckJobProjects()
+                                                               .stream()
+                                                               .map(BlackDuckJobProjectEntity::getProjectName)
+                                                               .anyMatch(projectName::equals))
```
jamesrichard91, line 217: What are you checking for here? it seems you want the opposite of the `anyMatch()` result. So why not use the `noneMatch()` method?

ChomickiM (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -210,7 +211,10 @@ public StaticJobAccessor(
                    .filter(distributionJobEntity -> !(distributionJobEntity.getBlackDuckJobDetails().getFilterByProject() &&
                                                           distributionJobEntity.getBlackDuckJobDetails().getProjectNamePattern() != null &&
                                                           !Pattern.matches(distributionJobEntity.getBlackDuckJobDetails().getProjectNamePattern(), projectName) &&
-                                                          !distributionJobEntity.getBlackDuckJobDetails().getBlackDuckJobProjects().contains(projectName))
+                                                          distributionJobEntity.getBlackDuckJobDetails().getBlackDuckJobProjects()
+                                                              .stream()
+                                                              .map(BlackDuckJobProjectEntity::getProjectName)
+                                                              .noneMatch(projectName::equals))
                    )
```
gkillough, line 218: We call `distributionJobEntity.getBlackDuckJobDetails()` times within the same filter. This is the perfect candidate for a helper method.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -207,15 +208,23 @@ public StaticJobAccessor(
...
+    private boolean filterByProjects(DistributionJobEntity distributionJobEntity, String projectName) {
+        BlackDuckJobDetailsEntity blackDuckJobDetails = distributionJobEntity.getBlackDuckJobDetails();
+        String projectNamePattern = blackDuckJobDetails.getProjectNamePattern();
+        return !(blackDuckJobDetails.getFilterByProject() &&
+                     projectNamePattern != null &&
+                     !Pattern.matches(projectNamePattern, projectName) &&
+                     blackDuckJobDetails.getBlackDuckJobProjects()
+                         .stream()
+                         .map(BlackDuckJobProjectEntity::getProjectName)
+                         .noneMatch(projectName::equals));
```
jamesrichard91, line 225: This is very confusing. You are negating the entire condition? Why not change the conditions to what you would expect them to be?
EX:
``` java
!blackDuckJobDetails.getFilterByProject() || 
(projectNamePattern != null && Pattern.matches(projectNamePattern, projectName)) ||
blackDuckJobDetails.getBlackDuckJobProjects()
                         .stream()
                         .map(BlackDuckJobProjectEntity::getProjectName)
                         .anyMatch(projectName::equals))
```

jamesrichard91 (approved):

alert-database/src/main/java/com/synopsys/integration/alert/database/api/StaticJobAccessor.java
```diff
@@ -207,15 +208,22 @@ public StaticJobAccessor(
...
 
+    private boolean filterByProjects(DistributionJobEntity distributionJobEntity, String projectName) {
+        BlackDuckJobDetailsEntity blackDuckJobDetails = distributionJobEntity.getBlackDuckJobDetails();
+        String projectNamePattern = blackDuckJobDetails.getProjectNamePattern();
+        return !blackDuckJobDetails.getFilterByProject() ||
+                   (projectNamePattern != null && Pattern.matches(projectNamePattern, projectName)) ||
+                   blackDuckJobDetails.getBlackDuckJobProjects()
+                       .stream()
+                       .map(BlackDuckJobProjectEntity::getProjectName)
+                       .anyMatch(projectName::equals);
```
gkillough, line 224: I still find the readability here pretty low. It might be beneficial to introduce if-statements and short-circuiting.

bamandel: In reality, this should be replaced by a repository implementation that uses postgres keywords to query the DB using projectNamePattern. The query will be much clearer as a lot of this content is what's required to filter the response data a second time. I'd be happy to add short circuits if we think it's necessary

gkillough: I hear you, but this is the implementation we're stuck with for now, so I think it's best to make it as readable as possible.

gkillough (approved):

PR #1392, opened 2021-02-05
VulnerabilityNotificationMessageExtractor

gkillough:

PR #1393, opened 2021-02-05
BomEditNotificationMessageExtractor

gkillough:
Piggy-backed off of #1392 . Not to be merged before that.

### DIFF
--- a/provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/detail/BomEditNotificationDetailExtractor.java
+++ b/provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/detail/BomEditNotificationDetailExtractor.java
@@ -35,12 +35,14 @@ import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
 import com.synopsys.integration.alert.processor.api.detail.DetailedNotificationContent;
 import com.synopsys.integration.alert.processor.api.detail.NotificationDetailExtractor;
 import com.synopsys.integration.alert.provider.blackduck.processor.NotificationExtractorBlackDuckServicesFactoryCache;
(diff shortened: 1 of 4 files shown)


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/UpgradeGuidanceRiskModel.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
+import com.synopsys.integration.blackduck.api.generated.component.ComponentVersionUpgradeGuidanceLongTermVulnerabilityRiskView;
+import com.synopsys.integration.blackduck.api.generated.component.ComponentVersionUpgradeGuidanceShortTermVulnerabilityRiskView;
+
+public class UpgradeGuidanceRiskModel extends AlertSerializableModel {
+    private final Integer critical;
+    private final Integer high;
+    private final Integer medium;
+    private final Integer low;
```
bamandel, line 38: Why did we decide to use Integer? Can we get away with the use of int and defaulting everything to 0? Do we actually benefit from a null value being returned as opposed to a 0?

bamandel (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/UpgradeGuidanceRiskModel.java
gkillough: This code was mostly copied/adapted from some old code handling this. I was trying to keep feature parity. These fields can be null in the response we get from BlackDuck, so I'm guessing we were trying to give a similar experience.

ChomickiM (approved):

PR #1394, opened 2021-02-08
Issue Tracker Framework

gkillough:


channel-api/src/main/java/com/synopys/integration/alert/channel/api/issue/model/IssueTransitionType.java
```diff
@@ -0,0 +1,29 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopys.integration.alert.channel.api.issue.model;
+
+public enum IssueTransitionType {
+    REOPEN,
```
bamandel, line 26: I think OPEN would make more sense here. If we're thinking in terms of a general Issue manager, we open and resolve Issues. REOPEN to me implies we don't OPEN in the first place

bamandel (approved):

psantos1113 (approved):

channel-api/src/main/java/com/synopys/integration/alert/channel/api/issue/model/IssueTransitionType.java
gkillough: FWIW, this only represents the operations we would perform on _existing_ issues, not new issues.

PR #1395, opened 2021-02-08
adding a layer for log messages

ekerwin:
# Pull Request template

*

*

*
*
*


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
```diff
@@ -358,4 +358,22 @@ private String truncateDescription(@Nullable String originalDescription) {
         return trimmedDescription;
     }
 
+    //ejk - the way I read these messages, they could be a little more consistent - is that, along
+    // with pleasing sonar, enough of a value add to add this layer?
+    private String PROJECT_NOT_FOUND(String providerConfigName, String message) {
```
gkillough, line 363: I know with `...static final` variables we use all caps, but in this case the private method is not static (then again, it also doesn't really have any interaction with the state of the object). Is this a convention?

PR #1399, opened 2021-02-11
Use cache client in test action

gkillough:


psantos1113 (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/task/AddUserToProjectsRunnable.java
```diff
@@ -55,7 +55,7 @@ public AddUserToProjectsRunnable(BlackDuckProperties blackDuckProperties, Collec
     public void run() {
         try {
             Slf4jIntLogger intLogger = new Slf4jIntLogger(logger);
-            BlackDuckHttpClient blackDuckHttpClient = blackDuckProperties.createBlackDuckHttpClient(intLogger);
+            BlackDuckHttpClient blackDuckHttpClient = blackDuckProperties.createBlackDuckCacheClient(intLogger);
```
jamesrichard91, line 58: This is fine, but will not provide much benefit. We may want to create a cache client outside of this class that lives for about 2 minutes and can be used more than once. That would allow for real performance improvements.

PR #1408, opened 2021-02-19
Register JmsListenerContainerFactories to Listeners to send distribution channel messages concurrently

ChomickiM:
Currently in Alert, sending messages will only happen after the notificationReceiver finishes processing all pages of notifications, timing out, or hitting an upper limit of 10,000 messages. The distribution events would be sent but never received as Alert ran with a concurrency of 1.

This change exposes the JmsListenerContainerFactories which allow us to concurrently run the NotificationReceiver and distribute message on separate threads.

Additionally we created new interfaces AlertChannelEventListener and AlertDefaultEventListener to differentiate between different types of AlertEventListeners and allows us to set different JmsListenerContainerFactories with different levels of concurrency. Currently, for non-channel listeners, we only want to have a concurrency of 1 while running so there aren't cases of multiple NotificationReceivers running and sending out of order messages when reading pages of notifications.


alert-common/src/main/java/com/synopsys/integration/alert/common/channel/DistributionChannel.java
```diff
@@ -27,15 +27,15 @@
...
+import com.synopsys.integration.alert.common.event.AlertChannelEventListener;
 import com.synopsys.integration.alert.common.event.DistributionEvent;
 import com.synopsys.integration.alert.common.exception.AlertException;
 import com.synopsys.integration.alert.common.message.model.MessageResult;
 import com.synopsys.integration.alert.common.workflow.MessageReceiver;
 import com.synopsys.integration.exception.IntegrationException;
 import com.synopsys.integration.rest.exception.IntegrationRestException;
 
-public abstract class DistributionChannel extends MessageReceiver<DistributionEvent> implements AlertEventListener {
+public abstract class DistributionChannel extends MessageReceiver<DistributionEvent> implements AlertChannelEventListener {
```
gkillough, line 38: `DistributionChannelV2` should probably also get this change.

alert-common/src/main/java/com/synopsys/integration/alert/common/event/EventListenerConfigurer.java
```diff
@@ -29,38 +29,51 @@
...
-                endpoint.setDestination(destinationName);
-                endpoint.setMessageListener(listener);
-                registrar.registerEndpoint(endpoint);
-            }
+        alertDefaultEventListeners.forEach(listener -> registerListenerEndpoint(registrar, listener, defaultJmsListenerContainerFactory));
+        alertChannelEventListeners.forEach(listener -> registerListenerEndpoint(registrar, listener, distributionChannelJmsListenerContainerFactory));
+    }
+
+    private void registerListenerEndpoint(JmsListenerEndpointRegistrar registrar, AlertEventListener listener, DefaultJmsListenerContainerFactory jmsListenerContainerFactory) {
+        if (listener != null) {
```
gkillough, line 64: Why would this ever be null?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/distribute/ProviderMessageDistributor.java
```diff
@@ -53,6 +53,8 @@ public ProviderMessageDistributor(AuditAccessor auditAccessor, EventManager even
 
     public void distribute(ProcessedNotificationDetails processedNotificationDetails, ProviderMessageHolder providerMessageHolder) {
         UUID jobId = processedNotificationDetails.getJobId();
+        //FIXME need to look for existing audit entries and relate the notificationIds to that entry if needed
+        // In the case of audit resending this is an issue because it will create a new audit entry rather than update the existing one
         Long auditId = auditAccessor.createAuditEntryForJob(jobId, processedNotificationDetails.getNotificationIds());
```
gkillough, line 58: Is there an internal ticket for this?

ChomickiM: I believe it is IALERT-2252

src/main/java/com/synopsys/integration/alert/ApplicationConfiguration.java
```diff
@@ -136,4 +138,23 @@ public AuthorizationManager authorizationManager(RoleAccessor roleAccessor) {
         return new AuthorizationManager(roleAccessor);
     }
```
gkillough, line 140: Can we register these beans in their own configuration class? e.g. something called `AlertJmsConfiguration`.

src/main/java/com/synopsys/integration/alert/ApplicationConfiguration.java
```diff
@@ -136,4 +138,23 @@ public AuthorizationManager authorizationManager(RoleAccessor roleAccessor) {
...
+        return factory;
+    }
+
+    @Bean
+    public DefaultJmsListenerContainerFactory distributionChannelJmsListenerContainerFactory(CachingConnectionFactory connectionFactory) {
+        DefaultJmsListenerContainerFactory factory = new DefaultJmsListenerContainerFactory();
+        factory.setConnectionFactory(connectionFactory);
+        int minConcurrency = 1;
+        int maxThreadCount = Runtime.getRuntime().availableProcessors();
+        int maxConcurrency = maxThreadCount < 10 ? maxThreadCount : 10;
```
gkillough, line 154: Once these beans are in separate classes, can we create a constant for this? `MAX_ALLOWED_THREAD_COUNT`. Then we should also include a comment on why this number was chosen.

alert-common/src/main/java/com/synopsys/integration/alert/common/event/EventListenerConfigurer.java
ChomickiM: This is a copy-paste from the previous implementation in configureJmsListeners. But you may be correct, I can't think of a case where the listener would come back as null

src/test/java/com/synopsys/integration/alert/workflow/message/JmsNotificationReceiverTestIT.java
```diff
@@ -0,0 +1,183 @@
...
+        defaultNotificationAccessor.deleteNotificationList(savedModels);
+        staticJobAccessor.deleteJob(distributionJobModel.getJobId());
+    }
+
+    @Test
+    @Ignore
+    @Disabled
+    public void testJms() throws InterruptedException {
+        // Set breakpoints throughout this test, there is nothing to assert against here. Suggestions for breakpoints:
+        //      Registering listeners: EvenListenerConfigurer
```
gkillough, line 135: Missing a 'T' in 'EvenListenerConfigurer'.

src/main/java/com/synopsys/integration/alert/ApplicationConfiguration.java
```diff
@@ -136,4 +138,23 @@ public AuthorizationManager authorizationManager(RoleAccessor roleAccessor) {
...
     }
 
+    @Bean
+    public DefaultJmsListenerContainerFactory defaultJmsListenerContainerFactory(CachingConnectionFactory connectionFactory) {
+        DefaultJmsListenerContainerFactory factory = new DefaultJmsListenerContainerFactory();
+        factory.setConnectionFactory(connectionFactory);
+        return factory;
+    }
+
+    @Bean
```
gkillough, line 148: Can we register these beans in their own configuration class? e.g. something called `AlertJmsConfiguration`.

alert-common/src/main/java/com/synopsys/integration/alert/common/event/EventListenerConfigurer.java
gkillough: Let's figure it out and (hopefully) get rid of the null-check.

bamandel (approved):

src/main/java/com/synopsys/integration/alert/AlertJmsConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+        return factory;
+    }
+
+    @Bean
+    public DefaultJmsListenerContainerFactory distributionChannelJmsListenerContainerFactory(CachingConnectionFactory connectionFactory) {
+        DefaultJmsListenerContainerFactory factory = new DefaultJmsListenerContainerFactory();
+        factory.setConnectionFactory(connectionFactory);
+        int minConcurrency = 1;
+        int maxThreadCount = Runtime.getRuntime().availableProcessors();
+        int maxConcurrency = maxThreadCount < MAX_ALLOWED_THREAD_COUNT ? maxThreadCount : MAX_ALLOWED_THREAD_COUNT;
```
JakeMathews, line 48:
```suggestion
        int maxConcurrency = Math.min(maxThreadCount, MAX_ALLOWED_THREAD_COUNT)
```

psantos1113 (approved):

JakeMathews (approved):

PR #1409, opened 2021-02-22
Include provider config id in ProviderMessage

gkillough:


channel-api/src/main/java/com/synopsys/integration/alert/channel/api/issue/IssueTrackerSearcher.java
```diff
@@ -33,13 +33,15 @@
...
+import com.synopsys.integration.alert.processor.api.extract.model.ProviderDetails;
 import com.synopsys.integration.alert.processor.api.extract.model.project.BomComponentDetails;
 import com.synopsys.integration.alert.processor.api.extract.model.project.MessageReason;
 import com.synopsys.integration.alert.processor.api.extract.model.project.ProjectMessage;
 
 public abstract class IssueTrackerSearcher<T extends Serializable> {
     public final List<IssueSearchResult<T>> findIssues(ProjectMessage projectMessage) throws AlertException {
-        LinkableItem provider = projectMessage.getProvider();
+        ProviderDetails providerDetails = projectMessage.getProviderDetails();
+        LinkableItem provider = providerDetails.getProvider();
```
bamandel, line 44: I see you calling getProvider after getProviderDetails a number of times. We should add a method to ProviderDetails that directly gets the LinkableItem instead of having to do it twice everywhere

bamandel (approved):

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/issue/IssueTrackerSearcher.java
gkillough: I was definitely tempted to create a delegate, but struggle with pulling the trigger lol.

JakeMathews (approved):

ChomickiM (approved):

PR #1410, opened 2021-02-23
Summarizer Constants

gkillough:
SonarCloud code smell clean-up.


ChomickiM (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/summarize/ProjectMessageSummarizer.java
```diff
@@ -45,6 +45,12 @@
 
 @Component
 public class ProjectMessageSummarizer {
+    public static final String OP_PARTICIPLE_ADDED = "added";
```
bamandel, line 48: Is OP supposed to mean operation?

gkillough: Yes

PR #1412, opened 2021-02-23
Issue Tracker BOM Model

gkillough:


channel-api/src/main/java/com/synopsys/integration/alert/channel/api/convert/BomComponentDetailConverter.java
```diff
@@ -48,27 +49,35 @@ public BomComponentDetailConverter(ChannelMessageFormatter formatter) {
     public List<String> gatherBomComponentPieces(BomComponentDetails bomComponent) {
         List<String> bomComponentSectionPieces = new LinkedList<>();
 
+        List<String> preConcernSectionPieces = gatherPreConcernSectionPieces(bomComponent);
```
jamesrichard91, line 52: What is a preConcern?

gkillough: The concern section is related to "component concerns" (i.e. policy violations and vulnerabilities). The pre-concern section is the section before the component concerns.

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/issue/convert/IssueVulnerabilityDetailsConverter.java
```diff
@@ -0,0 +1,100 @@
...
+        vulnDetailsSectionPieces.add(formatter.encode(SECTION_LABEL_VULNERABILITIES));
+
+        List<String> addedSectionPieces = createIssueVulnerabilityCollectionSectionPieces(OP_PARTICIPLE_ADDED, vulnerabilityDetails.getVulnerabilitiesAdded());
+        vulnDetailsSectionPieces.addAll(addedSectionPieces);
+
+        List<String> updatedSectionPieces = createIssueVulnerabilityCollectionSectionPieces(OP_PARTICIPLE_UPDATED, vulnerabilityDetails.getVulnerabilitiesUpdated());
+        vulnDetailsSectionPieces.addAll(updatedSectionPieces);
+
+        List<String> deletedSectionPieces = createIssueVulnerabilityCollectionSectionPieces(OP_PARTICIPLE_DELETED, vulnerabilityDetails.getVulnerabilitiesDeleted());
+        vulnDetailsSectionPieces.addAll(deletedSectionPieces);
```
bamandel, line 61: If you pass vulnDetailsSectionPieces into  createIssueVulnerabilityCollectionSectionPieces you could change all of this into 3 lines of code and perform the addAll in the private method

bamandel (approved):

JakeMathews (approved):

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/issue/convert/IssueVulnerabilityDetailsConverter.java
gkillough: I originally did this, but I don't feel that it is as expressive.

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/convert/BomComponentDetailConverter.java
```diff
@@ -48,27 +49,35 @@ public BomComponentDetailConverter(ChannelMessageFormatter formatter) {
...
+        bomComponentSectionPieces.addAll(preConcernSectionPieces);
+
+        List<String> componentConcernSectionPieces = createComponentConcernSectionPieces(bomComponent);
+        bomComponentSectionPieces.addAll(componentConcernSectionPieces);
+
+        return bomComponentSectionPieces;
+    }
+
+    public List<String> gatherPreConcernSectionPieces(AbstractBomComponentDetails bomComponent) {
```
psantos1113, line 61: Should this be public?  Is this called elsewhere?

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/convert/BomComponentDetailConverter.java
psantos1113: Is a pre-concern specific to the BomComponent conversion at this point?   Or is this something in other converters as well?  From looking at the code here in this class we're defining the component data which includes the component name and version, as well as, component attributes.  The variable name with preconcern had me asking the same question.  What's a preconcern?

If the variable name containing preconcern is confusing, then could we name it componentSectionPieces?  The componentSectionPieces contains the component header and the component attributes and the component concerns section is exactly that the component concerns.  The naming of variables is the hard part.  If renaming the variable would help avoid confusion.

Or if the intent isn't clear then could we split gatherPreConcernSectionPieces into two methods:
`List<String> gatherComponentHeaderPieces(AbstractBomComponentDetails bomComponent)`
That method executes the code existing in the 64 -73 to get the component name and version.
Then have another method:
`List<String> gatherComponentAttributePieces(AbstractBomComponentDetails bomComponent)`
That executes lines 75-79 to create all the component attribute data.

The last step would be to then change the method to execute this:
```
  List<String> componentHeaderPieces = gatherComponentHeaderPieces(bomComponent);
  List<String> componentAttributesPieces = gatherComponentAttributePieces(bomComponent);
  List<String> componentConcernSectionPieces = createComponentConcernSectionPieces(bomComponent);
  bomComponentSectionPieces.addAll(componentHeaderPieces);
  bomComponentSectionPieces.addAll(componentAttributesPieces);
  bomComponentSectionPieces.addAll(componentConcernSectionPieces);
```

psantos1113: That's my suggestion if the preconcern variable is causing confusion.

gkillough: Yeah, when I wrote it, I was reading it as "the section before the concern-section" not "the section related to 'preconcerns'". I see that it is causing some confusion, so I'll try to make the variables a little more clear.

psantos1113 (approved):

PR #1413, opened 2021-02-23
Exclude snakeyaml

jamesrichard91:
We are not using snakeyaml and the version that is being dragged in transitively is vulnerable, so we should exclude it. Excluding it does not seem to have any impact at all. I have started Alert using the runServer task and everything seems to be working perfectly fine, there do not seem to be any issues with liquibase when excluding this.


psantos1113:
I would try installing it with Docker Swarm because we extract liquibase and run the command line version on the startup of Alert in the entry point.

PR #1415, opened 2021-02-25
Jm ms teams improvements

JakeMathews:
An `MSTeamsChannelMessageModel` now contains message sections and titles. This allows the entire message to be constructed in `MSTeamsChannelMessageConverter` rather than partially in the `MSTeamsChannelMessageSender`.

This PR also removes the need to complete IALERT-2247 (Add provider field to ProviderMessageHolder) as the `ProviderDetails` are no longer needed in `MSTeamsChannelMessageSender`.

### DIFF
--- a/channel/src/main/java/com/synopsys/integration/alert/channel/msteams2/MSTeamsChannelMessageSender.java
+++ b/channel/src/main/java/com/synopsys/integration/alert/channel/msteams2/MSTeamsChannelMessageSender.java
@@ -25,6 +25,7 @@ package com.synopsys.integration.alert.channel.msteams2;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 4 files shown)


jamesrichard91 (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/msteams2/MSTeamsChannelMessageSender.java
```diff
@@ -62,43 +61,34 @@ public MessageResult sendMessages(MSTeamsJobDetailsModel msTeamsJobDetailsModel,
...
-                   .findFirst()
-                   .map(MSTeamsChannelMessageModel::getProviderDetails)
-                   .map(ProviderDetails::getProvider)
-                   .map(LinkableItem::getValue)
-                   .orElse("BlackDuck");
-    }
-
-    private Request createRequestsForMessage(String webhook, String messageTitle, List<MSTeamsChannelMessageModel> messages, Map<String, String> requestHeaders) {
-        String jsonString = createJsonString(messageTitle, messages);
+    private Request createRequestsForMessage(String webhook, MSTeamsChannelMessageModel messageModel, Map<String, String> requestHeaders) {
```
jamesrichard91, line 73: Should this method be changed from plural to singular? Since it is only returning a single request?

gkillough (approved):

PR #1418, opened 2021-02-26
Jm jira server commenter

JakeMathews:
Abstracted common Jira issue commenting logic from JiraCloudIssueCommenter into JiraIssueCommenter


channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/JiraIssueCommenter.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.synopsys.integration.alert.channel.api.issue.model.ProjectIssueModel;
+import com.synopsys.integration.alert.channel.api.issue.search.ExistingIssueDetails;
+import com.synopsys.integration.alert.channel.api.issue.send.IssueTrackerIssueCommenter;
+import com.synopsys.integration.alert.channel.api.issue.send.IssueTrackerIssueResponseCreator;
+import com.synopsys.integration.alert.common.exception.AlertException;
+import com.synopsys.integration.exception.IntegrationException;
+import com.synopsys.integration.jira.common.cloud.service.IssueService;
```
jamesrichard91, line 36: I don't think you can use the Jira Cloud IssueService for Jira Server, so this common class would be specific to Jira Cloud not both.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/server/delegate/JiraServerIssueCommenter.java
```diff
@@ -0,0 +1,42 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.synopsys.integration.alert.channel.jira2.server.delegate;
+
+import com.synopsys.integration.alert.channel.api.issue.send.IssueTrackerIssueResponseCreator;
+import com.synopsys.integration.alert.channel.jira2.common.JiraIssueCommenter;
+import com.synopsys.integration.alert.common.persistence.model.job.details.JiraServerJobDetailsModel;
+import com.synopsys.integration.jira.common.cloud.service.IssueService;
```
jamesrichard91, line 28: I don't think you can use the Jira Cloud IssueService for Jira Server.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/JiraIssueCommenter.java
JakeMathews: I didn't see that. Good catch

JakeMathews:
Changes will need to be made to int-jira-common

PR #1422, opened 2021-03-01
NotificationReceiverV2

ChomickiM:
To implement the new changes the team has made to the NotificationProcessorV2 while still maintaining old functionality for now I have created a NotificationReceiverV2. The main changes here is that we removed the list of Distribution events being created and sent in the Receiver and instead are sent as soon as they are done being processed. To this end we had to move around some other methods such as setting the notifcation events in the database to processed inside of the NotificationProcessorV2 which required the changes you see below to the NotificationAccessor.

Also, in order to run and test these changes, due to how we wire in the JMS listeners we are required to create new AlertEvents as well as a new ChannelKey for the V2 distributions. Without doing so Alert will fail on start up due to duplicate listeners being registered with JMS. Once we fully replace the old MessageReceiver, implement the new DistributionEventV2, and replace the old event, we should be able to delete the temporary V2 files and ChannelKeys. These have been marked with FIXMEs


alert-database/src/test/java/com/synopsys/integration/alert/database/api/DefaultNotificationAccessorTest.java
```diff
@@ -325,6 +326,24 @@ public void setNotificationsProcessedTest() {
...
+        NotificationContentRepository notificationContentRepository = Mockito.mock(NotificationContentRepository.class);
+        ConfigurationAccessor configurationAccessor = Mockito.mock(ConfigurationAccessor.class);
+
+        Mockito.when(notificationContentRepository.findAllById(Mockito.any())).thenReturn(List.of(notificationEntity));
+        Mockito.when(configurationAccessor.getConfigurationById(Mockito.any())).thenReturn(Optional.of(configurationModel));
+
+        DefaultNotificationAccessor notificationManager = new DefaultNotificationAccessor(notificationContentRepository, null, null, configurationAccessor);
+        notificationManager.setNotificationsProcessedById(notificationIds);
+
+        Mockito.verify(notificationContentRepository).saveAll(Mockito.any());
```
jamesrichard91, line 344: There is no check to see if the "processed" value was changed?

ChomickiM: Because its the noificationContentRepository is mocked we don't have a way of checking the processed value similar to the other test above. I created the IT tests in order to check if the value is set to processed or not.

descriptor-api/src/main/java/com/synopsys/integration/alert/descriptor/api/model/ChannelKeys.java
```diff
@@ -42,6 +45,8 @@ private static void putKey(ChannelKey channelKey) {
         ChannelKeys.putKey(ChannelKeys.JIRA_SERVER);
         ChannelKeys.putKey(ChannelKeys.MS_TEAMS);
         ChannelKeys.putKey(ChannelKeys.SLACK);
+        //FIXME: these ChannelKeys are used for testing the NotificationReceiverV2. These must be removed before releasing 6.5.0.
+        ChannelKeys.putKey(ChannelKeys.SLACK_V2);
```
gkillough, line 49: What if we remove the `@Component` annotation from the old `NotificationReceiver`? We're not using it anymore right?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultNotificationAccessor.java
```diff
@@ -176,6 +178,12 @@ public void setNotificationsProcessed(List<AlertNotificationModel> notifications
         notificationContentRepository.saveAll(notificationEntities);
     }
 
+    @Override
+    public void setNotificationsProcessedById(Set<Long> notificationIds) {
+        List<AlertNotificationModel> notificationModels = findByIds(new ArrayList<>(notificationIds));
+        setNotificationsProcessed(notificationModels);
```
gkillough, line 184: It seems to me that we should be doing the opposite here. Instead of looking up notifications and calling the old method, we should make the old method extract a set of notificationIds and pass them to a new method that makes one call to the DB to set notifications by id.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessorV2.java
```diff
@@ -51,15 +53,16 @@ protected NotificationProcessorV2(
         ProjectMessageDigester projectMessageDigester,
         ProjectMessageSummarizer projectMessageSummarizer,
         ProviderMessageDistributor providerMessageDistributor,
-        List<NotificationProcessingLifecycleCache> lifecycleCaches
-    ) {
+        List<NotificationProcessingLifecycleCache> lifecycleCaches,
+        NotificationAccessor notificationAccessor) {
```
gkillough, line 57: Looks like Intellij caused the `) {` to not be on a new line. Can you fix that?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessorV2.java
```diff
@@ -92,6 +95,7 @@ private void processAndDistribute(List<AlertNotificationModel> notifications, Co
             ProviderMessageHolder providerMessageHolder = processJobNotifications(jobNotificationWrapper.getProcessingType(), filteredNotifications);
 
             providerMessageDistributor.distribute(processedNotificationDetails, providerMessageHolder);
+            notificationAccessor.setNotificationsProcessedById(notificationIds);
```
gkillough, line 98: It might be better to return the notificationIds of the notifications processed (and provide a JavaDoc comment about the return). Otherwise we would need this to be `Transactional` which would lock the DB for the entire duration of notification processing (this seems wrong).

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiverV2.java
```diff
@@ -0,0 +1,69 @@
...
+        this.notificationProcessor = notificationProcessor;
+    }
+
+    @Override
+    public void handleEvent(NotificationReceivedEventV2 event) {
+        logger.debug("Event {}", event);
+        logger.info("Processing event for notifications.");
+
+        int numPagesProcessed = 0;
+        int pageSize = 100;
```
gkillough, line 49: It seems like this should be a constant.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/IssueTrackerChannel.java
```diff
@@ -58,7 +58,8 @@ public final MessageResult sendMessage(DistributionEvent event) throws Integrati
 
     @Override
     public final String getDestinationName() {
-        return channelKey.getUniversalKey();
+        // Required to avoid conflicts when registering new JMS Listeners
```
gkillough, line 61: I believe this class will be deleted anyway, but I'd prefer a FIXME to flag this so it's easy to find when we're ready to drop support.

alert-common/src/main/java/com/synopsys/integration/alert/common/channel/NamedDistributionChannel.java
```diff
@@ -33,7 +33,8 @@ public MessageResult sendMessage(DistributionEvent event) throws IntegrationExce
 
     @Override
     public String getDestinationName() {
-        return channelKey.getUniversalKey();
+        // Required to avoid conflicts when registering new JMS Listeners
```
gkillough, line 36: Same comment about FIXME.

src/test/java/com/synopsys/integration/alert/NotificationAccessorTestITNonTransactional.java
```diff
@@ -0,0 +1,128 @@
...
+import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
+import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
+import com.synopsys.integration.alert.common.util.DateUtils;
+import com.synopsys.integration.alert.database.DatabaseDataSource;
+import com.synopsys.integration.alert.database.api.DefaultNotificationAccessor;
+import com.synopsys.integration.alert.descriptor.api.BlackDuckProviderKey;
+import com.synopsys.integration.alert.provider.blackduck.descriptor.BlackDuckDescriptor;
+import com.synopsys.integration.alert.test.common.TestTags;
+import com.synopsys.integration.alert.util.DescriptorMocker;
+
```
gkillough, line 46: Let's throw a TODO a about `@AlertIntegrationTest` and removing `@Transactional` so that we can remove this duplication of annotations in the future.

jamesrichard91 (approved):

gkillough (approved):

workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiverV2.java
```diff
@@ -0,0 +1,70 @@
...
+        logger.debug("Event {}", event);
+        logger.info("Processing event for notifications.");
+
+        int numPagesProcessed = 0;
+
+        AlertPagedModel<AlertNotificationModel> pageOfAlertNotificationModels = notificationAccessor.getFirstPageOfNotificationsNotProcessed(PAGE_SIZE);
+        while (!CollectionUtils.isEmpty(pageOfAlertNotificationModels.getModels())) {
+            List<AlertNotificationModel> notifications = pageOfAlertNotificationModels.getModels();
+            logger.info("Sending {} notifications.", notifications.size());
+            notificationProcessor.processNotifications(notifications, List.of(FrequencyType.REAL_TIME));
```
bamandel, line 56: Is there another Receiver that will process Daily Frequencies?

ChomickiM: I believe that is done by the DailyTask but I am unfamiliar with how exactly we run the daily task. The original NotificationReceiver also used a REAL_TIME frequency type which is what this was based on.

bamandel (approved):

JakeMathews (approved):

PR #1423, opened 2021-03-01
Jm jira server commentor

JakeMathews:
Abstracted common Jira issue commenting logic from `JiraCloudIssueCommenter` into `JiraIssueCommenter`.

`JiraIssueCommenter` no longer depends on the cloud specific version of IssueService and now accepts a `ThrowingConsumer<IssueCommentRequestModel, IntegrationException>`. This allows each Jira type to have it's own unrelated `IssueService`, while abstracting away common functionality and, without any updates to int-jira-common.


channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/JiraIssueCommenter.java
```diff
@@ -0,0 +1,79 @@
...
+
+        for (String comment : comments) {
+            IssueCommentRequestModel issueCommentRequestModel = new IssueCommentRequestModel(issueKey, comment);
+            try {
+                addCommentConsumer.accept(issueCommentRequestModel);
+            } catch (IntegrationException e) {
+                throw new AlertException(String.format("Failed to add a comment in Jira. Issue Key: %s", issueKey), e);
+            }
+        }
+    }
```
gkillough, line 67: I believe I was able to add abstraction such that these methods are no longer needed. Double check and see if you can get away with using `createIssueTrackerIssue(...)` where these are currently being used.

jamesrichard91 (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/JiraIssueCommenter.java
JakeMathews: Yeah these methods seem redundant. They have been removed.

PR #1425, opened 2021-03-02
Fix NULL href

gkillough:


psantos1113 (approved):

jamesrichard91:
Can we create a database with a project that has a null "href" to test this?

gkillough:
> Can we create a database with a project that has a null "href" to test this?

@jamesrichard91 I'm currently doing that

PR #1428, opened 2021-03-03
Azure Boards Channel

gkillough:


JakeMathews (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureRedirectUtil.java
```diff
@@ -17,6 +17,7 @@
 import com.synopsys.integration.alert.common.rest.AlertRestConstants;
 
 @Component
+// TODO rename this class
```
ChomickiM, line 20: Is this something that is being saved for a future task or can it be done in this PR?

gkillough: It's a bit of scope creep for this PR, but it is something we will want to do eventually. I think it will happen when we move/remove old channel classes.

ChomickiM (approved):

bamandel (approved):

jamesrichard91 (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/azure/boards/AzureRedirectUtil.java
gkillough: My opinion is that "Util" is not a suffix we ever want to use. "Utils" is used for classes containing common, stateless (i.e static) methods (e.g. StringUtils). Anything that is a Spring component or a stateful method should not be called "Util" or "Utils".

PR #1429, opened 2021-03-03
Simplify Provider Callback Handing

gkillough:
This is a step in preparation for removing `AlertIssueOrigin`. Mostly refactoring with ~9 classes that need a closer look.


channel/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/util/JiraCloudIssueHandler.java
```diff
@@ -79,7 +82,17 @@ protected String getIssueKey(IssueResponseModel issueModel) {
...
-        return new IssueTrackerIssueResponseModel(alertIssueOrigin, issueResponse.getKey(), uiLink, issueTitle, issueOperation);
+        return new IssueTrackerIssueResponseModel(
+            issueResponse.getKey(),
+            uiLink,
+            issueTitle,
+            issueOperation,
+            new IssueTrackerCallbackInfo(
+                alertIssueOrigin.getProviderContentKey().getProviderConfigId(),
+                alertIssueOrigin.getComponentItem().flatMap(ComponentItem::getCallbackInfo).map(ComponentItemCallbackInfo::getCallbackUrl).orElse(null),
+                alertIssueOrigin.getProviderContentKey().getSubTopicValue() // FIXME this is wrong, but this class will be deleted before 6.5.0 is released
```
jamesrichard91, line 93: Is this for the  old Jira Cloud implementation, is that why this class is going to be removed?

gkillough: It's for the old implementation. There is a ticket to remove this class.

jamesrichard91 (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
```diff
@@ -61,17 +58,14 @@ public void createOrUpdateBlackDuckIssue(String bomComponentVersionIssuesUrl, Bl
...
-            return bomComponentIssues
-                       .stream()
-                       .filter(issue -> issue.getIssueId().equals(issueKey))
-                       .findAny();
-        }
-        return Optional.empty();
+    private Optional<ProjectVersionIssuesView> retrieveExistingIssue(String projectVersionUrl, String blackDuckIssueId) throws IntegrationException {
+        HttpUrl projectVersionHttpUrl = new HttpUrl(projectVersionUrl);
+        ProjectVersionView projectVersion = blackDuckApiClient.getResponse(projectVersionHttpUrl, ProjectVersionView.class);
+        List<ProjectVersionIssuesView> bomComponentIssues = blackDuckApiClient.getAllResponses(projectVersion, ProjectVersionView.ISSUES_LINK_RESPONSE);
```
bamandel, line 64: Is there no servive available that retrieves a list of ProjectVersionIssuesView?

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckIssueTrackerCallbackHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        return new BlackDuckProviderIssueModel(event.getIssueKey(), blackDuckIssueStatus, event.getIssueSummary(), event.getIssueUrl());
+    }
+
+    private String mapOperationToAlertStatus(IssueOperation issueOperation) {
+        switch (issueOperation) {
+            case OPEN:
+            case UPDATE:
+                return "Created by Alert";
+            case RESOLVE:
+                return "Resolved by Alert";
```
bamandel, line 102: We may benefit by adding the string associated here to the IssueOperation class. That way we can just call the specific method from the Enum instead of a switch

bamandel (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
gkillough: Ah, there is now. I don't know if there was one when we originally created the class.

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckIssueTrackerCallbackHandler.java
gkillough: I went back and forth on this, but I think this logic does not belong in `IssueOperation` because it is only related to the callback handling.

ChomickiM (approved):

JakeMathews (approved):

PR #1430, opened 2021-03-04
Jm jira server issue creator

JakeMathews:
Adds an abstraction to Jira issue creation.
Added classes for Jira Server V2 issue creation.

Sonar is complaining about the number of constructor arguments in `JiraServerIssueCreator` and `JiraCloudIssueCreator`. 
Do we want a holder class for some of the common dependencies?


channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/delegate/JiraIssueCommenter.java
```diff
@@ -17,7 +17,7 @@
 import com.synopsys.integration.jira.common.model.request.IssueCommentRequestModel;
 
 public abstract class JiraIssueCommenter extends IssueTrackerIssueCommenter<String> {
-    private final ThrowingConsumer<IssueCommentRequestModel, IntegrationException> addCommentConsumer;
+    private final ThrowingConsumer<IssueCommentRequestModel, IntegrationException> addCommentConsumer; // TODO: Just expose another abstract method instead of this
```
gkillough, line 20: Will this be a follow-up PR?

gkillough: Alternatively, does this class need to be abstract? Could the necessary fields be passed through the constructor such that no subclasses would be needed?

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/cloud/JiraIssueAlertPropertiesManager.java
```diff
@@ -15,11 +15,11 @@
 import com.synopsys.integration.jira.common.model.response.IssuePropertyResponseModel;
 import com.synopsys.integration.jira.common.rest.service.IssuePropertyService;
```
gkillough, line 16: Ok, so this class can be used with both Jira Cloud and Jira Server?

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/cloud/delegate/JiraCloudIssueCreator.java
```diff
@@ -8,93 +8,61 @@
...
     }
 
     @Override
-    protected void assignAlertSearchProperties(ExistingIssueDetails<String> createdIssueDetails, ProjectIssueModel alertIssueSource) throws AlertException {
-        JiraIssueSearchProperties searchProperties = createSearchProperties(alertIssueSource);
-        issuePropertiesManager.assignIssueProperties(createdIssueDetails.getIssueKey(), searchProperties);
-    }
-
-    private IssueCreationRequestModel createIssueCreationRequest(IssueCreationModel alertIssueCreationModel) throws AlertException {
+    public IssueCreationRequestModel createIssueCreationRequest(IssueCreationModel alertIssueCreationModel) throws AlertException {
```
gkillough, line 65: Should this method be `public`? It seems like it should remain `protected`.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/server/delegate/JiraServerIssueCreator.java
```diff
@@ -0,0 +1,115 @@
...
+            JiraServerDescriptor.KEY_ISSUE_CREATOR
+        );
+        this.distributionDetails = distributionDetails;
+        this.issueService = issueService;
+        this.projectService = projectService;
+        this.jiraIssueCreationRequestCreator = jiraIssueCreationRequestCreator;
+    }
+
+    @Override
+    public IssueCreationRequestModel createIssueCreationRequest(IssueCreationModel alertIssueCreationModel) throws AlertException {
```
gkillough, line 64: Should this method be `public`? It seems like it should remain `protected`.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/server/delegate/JiraServerIssueCreator.java
```diff
@@ -0,0 +1,115 @@
...
+        ProjectComponent jiraProject = retrieveProjectComponent();
+        JiraCustomFieldReplacementValues replacementValues = alertIssueCreationModel.getSource()
+                                                                 .map(this::createCustomFieldReplacementValues)
+                                                                 .orElse(JiraCustomFieldReplacementValues.trivial(alertIssueCreationModel.getProvider()));
+        IssueRequestModelFieldsMapBuilder fieldsBuilder = jiraIssueCreationRequestCreator.createIssueRequestModel(
```
gkillough, line 69: It seems like this class is common. Any chance you could make this the argument of this method and move the above logic into the superclass?

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/delegate/JiraIssueCreator.java
```diff
@@ -0,0 +1,129 @@
...
+        issuePropertiesManager.assignIssueProperties(createdIssueDetails.getIssueKey(), searchProperties);
+    }
+
+    protected abstract T createIssueCreationRequest(IssueCreationModel alertIssueCreationModel) throws AlertException;
+
+    protected abstract IssueCreationResponseModel createIssue(T alertIssueCreationModel) throws IntegrationException;
+
+    protected abstract IssueResponseModel fetchIssue(String createdIssueKey) throws IntegrationException;
+
+    protected abstract String getReporter(T creationRequest);
```
gkillough, line 78: Can we call this `extractReporter`? I feel like `get` makes this seem like it could be performing a GET request.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/cloud/JiraIssueAlertPropertiesManager.java
JakeMathews: That was my impression from reading the old code.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/delegate/JiraIssueCreator.java
JakeMathews: I made `fetchIssue` over the concern that `get` might not imply a rest request. So `getIssue` and `extractReporter`?

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/delegate/JiraIssueCommenter.java
JakeMathews: This will be a follow-up PR. As the code is now, the class probably doesn't need to be abstract. But I prefer adding an abstract method instead of the field being passed into the constructor. We can discuss in the next PR.

channel/src/main/java/com/synopsys/integration/alert/channel/jira2/common/delegate/JiraIssueCreator.java
gkillough: I actually prefer `fetch`/`retrieve` instead of `get` for I/O (network, database, filesystem). I would like `get` to only be used for "getters". The only reason I mentioned a GET request in this case is because `getReporter` takes an argument that has the word "request" in it.

Sorry for making it overly complicated 😅 

To summarize:
- `getIssue` 👎🏻 
- `fetchIssue` 👍🏻 
- `getReporter` 👎🏻 
- `extractReporter` 👍🏻

gkillough:
I saw your comments about Sonar. Is there repeated functionality that we could abstract into its own class? For example, `AzureBoardsIssueTransitioner` has a field for `AzureBoardsWorkItemTypeStateRetriever` to serve a similar purpose.

jamesrichard91 (approved):

gkillough (approved):

JakeMathews:
> I saw your comments about Sonar. Is there repeated functionality that we could abstract into its own class? For example, `AzureBoardsIssueTransitioner` has a field for `AzureBoardsWorkItemTypeStateRetriever` to serve a similar purpose.

I took a look and it's not immediately obvious to me how that would work here in a way that wouldn't harm readability.

JakeMathews:
> > I saw your comments about Sonar. Is there repeated functionality that we could abstract into its own class? For example, `AzureBoardsIssueTransitioner` has a field for `AzureBoardsWorkItemTypeStateRetriever` to serve a similar purpose.
> 
> @gkillough I took a look and it's not immediately obvious to me how that would work here in a way that wouldn't harm readability.

I bound the blackduck-alert SonarCloud project to my IDE and updated the SonarLint plugin. I no longer get warned about the number of arguments. It seems we have raised the limit at some point

ChomickiM (approved):

bamandel (approved):

PR #1438, opened 2021-03-09
Fix inconsistent search properties

gkillough:


channel/src/main/java/com/synopsys/integration/alert/channel/jira2/cloud/JiraCloudSearcher.java
```diff
@@ -178,8 +178,11 @@ public JiraCloudSearcher(String jiraProjectKey, IssueSearchService issueSearchSe
 
     private ExistingIssueDetails<String> createExistingIssueDetails(IssueResponseModel issue) {
         String issueCallbackLink = JiraCallbackUtils.createUILink(issue);
-        IssueFieldsComponent issueFields = issue.getFields();
-        return new ExistingIssueDetails<>(issue.getId(), issue.getKey(), issueFields.getSummary(), issueCallbackLink);
+        IssueFieldsComponent nullableIssueFields = issue.getFields();
+        String existingIssueSummary = Optional.ofNullable(nullableIssueFields)
+                                          .map(IssueFieldsComponent::getSummary)
+                                          .orElse("Not included");
```
gkillough, line 184: Not sure what the best solution for this is. I don't think we want to make an additional request for the full issue for every search result. As far as I can tell, the issue fields should be present based on Jira's documentation and the way we call the API, but they are not.

JakeMathews (approved):

ChomickiM (approved):

PR #1440, opened 2021-03-09
Remove old NotificationReceiver

gkillough:
This also fixes the NPE in the new `...DistributionEventReceiver...`


workflow/src/main/java/com/synopsys/integration/alert/workflow/message/NotificationReceiver.java
```diff
@@ -17,34 +16,30 @@
...
 import com.synopsys.integration.alert.common.persistence.accessor.NotificationAccessor;
 import com.synopsys.integration.alert.common.rest.model.AlertNotificationModel;
 import com.synopsys.integration.alert.common.rest.model.AlertPagedModel;
-import com.synopsys.integration.alert.common.util.DateUtils;
 import com.synopsys.integration.alert.common.workflow.MessageReceiver;
-import com.synopsys.integration.alert.common.workflow.processor.notification.NotificationProcessor;
+import com.synopsys.integration.alert.processor.api.NotificationProcessorV2;
 
 @Component(value = NotificationReceiver.COMPONENT_NAME)
 public class NotificationReceiver extends MessageReceiver<NotificationReceivedEvent> implements AlertDefaultEventListener {
```
gkillough, line 29: The old `NotificationReceiver` was deleted, then `NotificationReceiverV2` was renamed to drop 'V2', but the git diff just shows it as a refactoring of this class.

bamandel (approved):

PR #1447, opened 2021-03-11
Delete old channel event handling

gkillough:


jamesrichard91 (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
```diff
@@ -54,7 +53,7 @@ protected IssueTrackerContext getIssueTrackerContext(DistributionEvent event) th
 
     @Override
     public IssueTrackerResponse sendRequests(IssueTrackerContext context, List<IssueTrackerRequest> requests) throws IntegrationException {
-        JiraServerRequestDelegator jiraServerService = new JiraServerRequestDelegator(getGson(), context);
+        JiraServerRequestDelegator jiraServerService = new JiraServerRequestDelegator(new Gson(), context);
```
bamandel, line 56: Everything else passes in a gson object but this one creates its own, is that intended?

bamandel (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerChannel.java
gkillough: This class will be deleted. I just needed it to compile safely for now.

PR #1448, opened 2021-03-12
Deploy Alert on helm using kubernetes secrets

ChomickiM:
This is an alternative for users who wish to pass their database credentials using secrets rather than having to create an environment variable on their system and modifying their helm install procedure, or hardcoding the values into the values.yaml file.


deployment/helm/synopsys-alert/values.yaml
```diff
@@ -54,6 +54,8 @@ postgres:
   databaseName: alertdb
   adminUserName: postgres
   adminPassword: ""
+  dbCredentialSecretName: "" #Secret that contains both the database users username & password
+  dbAdminCredentialSecretName: "" #Secret that contains both the database admin username & password
```
bamandel, line 58: Is there a specific format the username and password should be in?

ChomickiM: The README has a section for creating secrets that explains how to set them up. The db*SecretName parameters here are just for the name of the secret that has the actual credentials

bamandel (approved):

deployment/helm/synopsys-alert/README.md
```diff
@@ -196,6 +199,68 @@ Alternatively, a YAML file that specifies the values for the above parameters ca
...
+  ```bash 
+   $ kubectl create secret generic <SECRET_NAME> -n <ALERT_NAMESPACE> \
+   --from-file=ALERT_DB_USERNAME=<USER_USERNAME> \
+   --from-file=ALERT_DB_PASSWORD=<USER_PASSWORD>
+  ```
+  - Replace `<SECRET_NAME>` with the desired name for the secret.
+  - Replace `<ALERT_NAMESPACE>` with the namespace being used for Alert.
+  - Replace `<USER_USERNAME>` with the database user owning the Alert database.
+  - Replace `<USER_PASSWORD>` with the database user's password.
+  - Note: Using this secret will take priority over setting userUserName and userPassword in the Configuration Parameters in values.yaml.
```
gkillough, line 216: Is `userUserName` a typo?

ChomickiM: no, that is correct. In the configuration file we have both userUserName referring the db user, and another adminUserName for the postgres admin.

gkillough (approved):

jamesrichard91 (approved):

deployment/helm/synopsys-alert/templates/alert.yaml
```diff
@@ -94,6 +98,22 @@ spec:
...
         {{- end }}
+        {{- if .Values.postgres.dbCredentialSecretName }}
+        - mountPath: /tmp/secrets/ALERT_DB_USERNAME
+          name: db-credentials
+          subPath: ALERT_DB_USERNAME
+        - mountPath: /tmp/secrets/ALERT_DB_PASSWORD
+          name: db-credentials
+          subPath: ALERT_DB_PASSWORD
+        {{- end }}
+        {{- if .Values.postgres.dbAdminCredentialSecretName }}
```
psantos1113, line 109: In the postgres.yaml file we don't appear to need to mount the secrets this way.  We just need the secret key references so would that be all that is needed for this alert.yaml file and not mounting the secrets?  If we don't need it then we don't need lines 109-116 and 156-167.

JakeMathews (approved):

gkillough (approved):

psantos1113 (approved):

jamesrichard91 (approved):

PR #1457, opened 2021-03-16
Jm Jira Server Channel

JakeMathews:
Added JiraServerChannelV2 and JiraServerProcessorFactory.

I believe these are the final components of the new Jira Server implementation.


gkillough (changes requested):
Can you also update the `JiraServerDistributionTestAction` class now that the channel is available?

ChomickiM (approved):

PR #1463, opened 2021-03-17
Bm static slack global UI

bamandel:
Converted the empty Slack global configuration page to be static instead of using the dynamic code. Inserted this into the list of dynamically generated content in place of slacks dynamic version.

Removed the last updated field from this page as that should not be shown for pages that have no data.


ui/src/main/js/MainPage.js
```diff
@@ -79,12 +80,14 @@ class MainPage extends Component {
             );
         }
 
+        const renderComponent = (name === 'channel_slack') ? <SlackGlobalConfiguration /> : <GlobalConfiguration key={name} descriptor={component} />;
```
gkillough, line 83: Can we create a map of _name_ to static component and then loop over that?

psantos1113: We could perhaps use DESCRIPTOR_NAME.CHANNEL_SLACK from the descriptorUtilites.js file. That defines the constants today.  We can either refactor that class to only contain the descriptor names and operations definitions.  Or create a new js file to start from scratch.  Have a look at descriptorUtilities.js and decide from there.

bamandel: I introduced a new model For Slack info that would hold this data as I want to separate as much of the implementation from the descriptors as possible.

bamandel: As to your recommendation Gavin, I did debate doing it that way. I'll push a change for your viewing pleasure

ui/src/main/js/MainPage.js
```diff
@@ -12,6 +12,8 @@ import { getDescriptors } from 'store/actions/descriptors';
 import DescriptorContentLoader from 'dynamic/loaded/DescriptorContentLoader';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import ProviderTable from 'providers/ProviderTable';
+import SlackGlobalConfiguration from 'channels/SlackGlobalConfiguration';
+import { SLACK_INFO } from 'channels/SlackModels';
```
gkillough, line 16: It might also be nice to make the channel files more organized with a /slack directory.

ui/src/main/js/MainPage.js
bamandel: Upon further investigation we may want to stick with this approach for now. The other configurations will require props to be passed when they're implemented (we'll need the if statement anyway to see what props to pass) and this is also temporary until all the dynamic items are removed.

JakeMathews (approved):

gkillough (approved):

PR #1464, opened 2021-03-17
Add ID to IssueTrackerIssueResponseModel

gkillough:


bamandel (approved):

JakeMathews (approved):

JakeMathews:
Is the `issueId` ever not a `String`?

gkillough:
> Is the `issueId` ever not a `String`?

@JakeMathews In Jira it is a `String`, but in Azure it is a `Long`. One issue tracker that was SPIKED had a key similar to Jira, but used UUIDs for its API IDs.

PR #1468, opened 2021-03-18
Azure Boards Test Action

gkillough:


channel/src/main/java/com/synopsys/integration/alert/channel/azure/boards/distribution/AzureBoardsMessageSenderFactory.java
```diff
@@ -0,0 +1,100 @@
...
+        this.callbackInfoCreator = callbackInfoCreator;
+        this.channelKey = channelKey;
+        this.azureBoardsPropertiesFactory = azureBoardsPropertiesFactory;
+        this.proxyManager = proxyManager;
+    }
+
+    @Override
+    public IssueTrackerMessageSender<Integer> createMessageSender(AzureBoardsJobDetailsModel distributionDetails) throws AlertException {
+        AzureBoardsProperties azureBoardsProperties = azureBoardsPropertiesFactory.createAzureBoardsProperties();
+        String organizationName = azureBoardsProperties.getOrganizationName();
```
bamandel, line 60: Is there a reason we call this so early? Since it's a simple getter I imagine we should just call it in the return statement

bamandel (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/azure/boards/distribution/AzureBoardsMessageSenderFactory.java
gkillough: Oh yeah, this is some copy-pasta from the previous factory.

JakeMathews (approved):

PR #1469, opened 2021-03-19
Jr npm retry

jamesrichard91:


gkillough:
@jamesrichard91 would you mind providing a description of how this works? If I run `./gradlew clean` will it also trigger an NPM cache clean? 

Also, would you mind squashing your commits for this one?

jamesrichard91:
Do the commits cause some kind of problem? I had to try several different things and push them because this seems to only be reproducible in our Jenkins build environment.

npmCacheClean will run if the npmInstall task is called. npmCacheClean will only execute the commandline invocation if the npmInstall returned a non zero exit code. 
npmInstallRetry will run if the npmCacheClean task is called. npmInstallRetry will only execute the commandline invocation if the npmInstall returned a non zero exit code. 

The idea was not to change how the npm build process worked but to add a few extra steps in between in case the npm install failed due to a JSON parse error that can occur due to a number of different reasons.

gkillough:
For the squashing of commits, it's your call, but it's much easier to follow what changes were grouped together if there are fewer commits (in case someone other than you is trying maintain this code in the future). Just a personal preference because I often use commit history when debugging.

psantos1113 (approved):

bamandel (approved):

PR #1476, opened 2021-03-22
Constrain MessageReceiver to AlertEvents

JakeMathews:
IALERT-2218
Constrained `MessageReceiver` to accept `AlertEvent`s to allow the EventID to be logged.
All usages of `MessageReceiver` currently pass in `AlertEvent`s.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageReceiver.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageReceiver.java
@@ -15,13 +15,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.gson.Gson;
+import com.synopsys.integration.alert.common.event.AlertEvent;
 
-public abstract class MessageReceiver<T> implements MessageListener {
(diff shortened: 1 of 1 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/workflow/MessageReceiver.java
```diff
@@ -39,6 +40,7 @@ public void onMessage(Message message) {
                 T event = gson.fromJson(textMessage.getText(), eventClass);
                 logger.trace("{} event {}", receiverClassName, event);
                 handleEvent(event);
+                logger.debug("Received Event ID: {}", event.getEventId());
```
gkillough, line 43: I think it might make more sense to log this before handling the event. If a `RuntimeException` is thrown, we'll still know the event ID at _debug_.

JakeMathews: Good idea

gkillough (approved):

psantos1113 (approved):

ChomickiM (approved):

PR #1477, opened 2021-03-22
Handle BOM component 404

gkillough:


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/extract/model/project/AbstractBomComponentDetails.java
```diff
@@ -61,6 +61,7 @@ public String getUsage() {
         return additionalAttributes;
     }
 
+    // TODO make optional
```
gkillough, line 64: For a different PR.

jamesrichard91 (approved):

ChomickiM (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/BomComponent404Handler.java
```diff
@@ -0,0 +1,29 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.rest.exception.IntegrationRestException;
+
+@Component
+public class BomComponent404Handler {
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public void logIf404OrThrow(IntegrationRestException restException, String componentName, @Nullable String componentVersionName) throws IntegrationRestException {
```
bamandel, line 21: This may be more useful as a static method as we won't need to inject this in to classes and no constructor is currently necessary

psantos1113 (approved):

bamandel (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/BomComponent404Handler.java
gkillough: Yeah, it's 50/50, but much easier to add components to the constructor if need be rather than updating every static usage IMO.

JakeMathews (approved):

PR #1478, opened 2021-03-24
Fix Email Channel

gkillough:


channel/src/main/java/com/synopsys/integration/alert/channel/email/distribution/EmailChannelMessageConverter.java
```diff
@@ -54,11 +54,11 @@ public EmailChannelMessageConverter(EmailChannelMessageFormatter emailChannelMes
     @Override
     protected List<EmailChannelMessageModel> convertProjectMessageToChannelMessages(EmailJobDetailsModel distributionDetails, ProjectMessage projectMessage, List<String> messageChunks) {
         String subjectLine = createSubjectLine(distributionDetails, projectMessage);
-        String messageContent = StringUtils.join(messageChunks);
+        String messageContent = StringUtils.join(messageChunks, "");
```
jamesrichard91, line 57: Without the  `, ""` I assume this was resulting in null, is that why this is needed?

gkillough: `StringUtils.join()` with a single argument takes an array, so `messageChunks` was a being read in as an array of lists. That caused the output to be wrapped in `[ . . . ]`.

Example: 
```
[Test from Alert
---------------------------------------
This is a test message from Alert.
---------------------------------------
Test Provider: Test Provider Config
] 
```

bamandel (approved):

jamesrichard91 (approved):

PR #1482, opened 2021-03-25
Job Test Configuration - Provider Config Errors

gkillough:


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/web/BlackDuckProjectCustomFunctionAction.java
```diff
@@ -31,17 +31,28 @@
...
 import com.synopsys.integration.alert.common.security.authorization.AuthorizationManager;
+import com.synopsys.integration.alert.provider.blackduck.BlackDuckProperties;
+import com.synopsys.integration.alert.provider.blackduck.factory.BlackDuckPropertiesFactory;
+import com.synopsys.integration.alert.provider.blackduck.validator.BlackDuckApiTokenValidator;
 
 @Component
-public class ProviderProjectCustomFunctionAction extends PagedCustomFunctionAction<ProviderProjectOptions> {
-    private static final String MISSING_PROVIDER_ERROR = "Provider name is required to retrieve projects.";
+public class BlackDuckProjectCustomFunctionAction extends PagedCustomFunctionAction<ProviderProjectOptions> {
+    private static final String MISSING_PROVIDER_ERROR = "A provider configuration is required to retrieve projects.";
```
bamandel, line 40: Can we reference BlackDuck instead of provider now?

bamandel (approved):

JakeMathews (approved):

jamesrichard91 (approved):

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/web/BlackDuckProjectCustomFunctionAction.java
gkillough: We could, but right now the field is also called "Provider Configuration", so because this is a missing field-value, I think the error message can say provider for now. If we change the field we can update it to reference the new field name.

PR #1484, opened 2021-03-25
Improve generic usage of configuration test actions

gkillough:
## Motivations: 
 - In 6.4.0, the introduction of concrete classes for Distribution Jobs created an opportunity for improving "TestActions". Up until this point, all TestActions took a FieldModel and FieldUtility as parameters, but now they could take in a DistributionJobModel. A separate class was created for Distribution TestActions (i.e. TestActions that distributed a message based on a Distribution Job's fields). 

 - The class ConfigurationAction used to manage all actions of a single "descriptor" and it did this with TestActions by putting them into a Map from Context to TestAction. With the introduction of a new class for Channel DistributionTestActions, the Map was no longer useful.

## Discoveries:
 - In making these changes, it became apparent that these classes existed primarily to serve the dynamic UI.

 - The class ConfigurationAction is abstract and implementations of it autowire TestActions to pass to it, then those implementations are autowired into a class called DescriptorProcessor, which is autowired into a class called FieldModelProcessor, which is finally autowired into the places we actually want to get the originally autowired TestActions.

 - These classes were autowired in more places (four places) than they were actually called (which was litetally one place)

 - This process, besides being incredibly convoluted, is totally unnecessary with because we can simply autowire the TestActions in the place we actually want them. The only problem is that they won't be in a Map.

 ## Solution:
 - Create a class that takes in a list of test actions and provides "lookup" functionality based on the channel name (like a map).

 - Remove all references to the new TestActions in ConfigurationAction, DescriptorProcessor, and FieldModelProcessor

 - Create an integration test that verifies whether a channel has declared all necessary Spring components required to work properly

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionMap.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionMap.java
@@ -1,0 +1,53 @@
+/*
+ * alert-common
+ *
+ * Copyright (c) 2021 Synopsys, Inc.
+ *
+ * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All rights reserved worldwide.
+ */
+package com.synopsys.integration.alert.common.descriptor.action;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import com.synopsys.integration.alert.common.exception.AlertRuntimeException;
+import com.synopsys.integration.alert.descriptor.api.model.DescriptorKey;
+
+public final class DescriptorActionMap<T extends DescriptorAction> {
+    private final Map<String, T> universalKeyToAction;
+
+    public DescriptorActionMap(Collection<T> descriptorActions) {
+        this.universalKeyToAction = initializeMap(descriptorActions);
+    }
+
+    public final T findRequiredAction(DescriptorKey descriptorKey) throws AlertRuntimeException {
+        return findRequiredAction(descriptorKey.getUniversalKey());
+    }
+
+    public final T findRequiredAction(String descriptorUniversalKey) throws AlertRuntimeException {
+        T foundAction = universalKeyToAction.get(descriptorUniversalKey);
+        if (null == foundAction) {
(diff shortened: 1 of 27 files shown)


src/test/java/com/synopsys/integration/alert/channel/Channel_ComprehensiveRequirements_ITTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import com.synopsys.integration.alert.channel.api.DistributionEventReceiver;
+import com.synopsys.integration.alert.channel.api.issue.action.IssueTrackerTestAction;
+import com.synopsys.integration.alert.common.channel.DistributionChannelTestAction;
+import com.synopsys.integration.alert.descriptor.api.model.ChannelKey;
+import com.synopsys.integration.alert.descriptor.api.model.IssueTrackerChannelKey;
+import com.synopsys.integration.alert.util.AlertIntegrationTest;
+
+@AlertIntegrationTest
+public class Channel_ComprehensiveRequirements_ITTest {
```
gkillough, line 21: This class is important for channel bindings in the future.

Side note: I need to rename this to "TestIT" instead of "ITTest".

psantos1113 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/action/DescriptorActionMap.java
```diff
@@ -0,0 +1,53 @@
...
+        return findRequiredAction(descriptorKey.getUniversalKey());
+    }
+
+    public final T findRequiredAction(String descriptorUniversalKey) throws AlertRuntimeException {
+        T foundAction = universalKeyToAction.get(descriptorUniversalKey);
+        if (null == foundAction) {
+            throw new AlertRuntimeException(String.format("Missing required action for descriptor: %s", descriptorUniversalKey));
+        }
+        return foundAction;
+    }
```
JakeMathews, line 36:
```suggestion
    public final T findRequiredAction(String descriptorUniversalKey) throws AlertRuntimeException {
        return findOptionalAction(descriptorUniversalKey)
            .orElseThrow(() -> new AlertRuntimeException(String.format("Missing required action for descriptor: %s", descriptorUniversalKey)));
    }
```
Formatting is probably off in the suggestion

ChomickiM (approved):

PR #1489, opened 2021-04-02
Bm static email UI

bamandel:
There are some small issues in this PR like the status message not always popping up with a message if you perform a successful task multiple times in a row, but this should be resolved once everything moves to static and we can fix the status message component


ui/src/main/js/Navigation.js
```diff
@@ -56,6 +60,16 @@ class Navigation extends Component {
         );
     }
 
+    createEmailNavItem(uriPrefix) {
```
gkillough, line 63: Can this and `createSlackNavItem` be abstracted into one method that also takes an "info" param?

ui/src/main/js/global/GlobalTestModal.js
```diff
@@ -0,0 +1,38 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Modal } from 'react-bootstrap';
+
+const GlobalTestModal = ({
+    children, showTestModal, handleTest, handleCancel
+}) => (
+    <Modal show={showTestModal} onHide={handleCancel}>
+        <Modal.Header closeButton>
+            <Modal.Title>Test Your Configuration</Modal.Title>
```
gkillough, line 10: It might be nice to have a default prop for this.

ui/src/main/js/global/GlobalTestModal.js
```diff
@@ -0,0 +1,38 @@
...
+            {children}
+        </Modal.Body>
+        <Modal.Footer>
+            <button
+                id="testSend"
+                type="submit"
+                className="btn btn-primary"
+                onClick={handleTest}
+            >
+                Send Test Message
```
gkillough, line 22: default prop as well

ui/src/main/js/global/GlobalTestModal.js
bamandel: I'd rather not have these as props until we actually use the prop functionality. Trying to stick to as static as possible

ui/src/main/js/global/GlobalTestModal.js
bamandel: same as above

ui/src/main/js/global/GlobalTestModal.js
gkillough: Static doesn't mean "not-reusable" it just means we won't be generating this based on a request-response with the Alert server. So like with "Send Test Message" below, that is still unique to email. Testing Global Jira is totally different.

bamandel: What is the message on the globa jira test? I don't see it use a global test modal at all

gkillough: My mistake, Jira doesn't use a modal, but it does a different _type_ of test than email. I still think the point about reusability is different than being dynamically generated though.

bamandel: I agree that we want to make it reusable which is why I created the new component as a separate item instead of built into the GlobalChannel component. Using as few props as possible makes it easier to read and understand in my opinion as you'll have to check multiple places to figure out what the resulting value is. If we end up needing to change any hardcoded value it's very simple to add a prop as you suggested but until then I say we keep the content as simple as possible.

ui/src/main/js/global/channels/email/EmailGlobalConfiguration.js
```diff
@@ -0,0 +1,600 @@
...
+            <CommonGlobalConfigurationForm
+                csrfToken={csrfToken}
+                formData={fieldModel}
+                setErrors={(errors) => setFieldErrors(errors)}
+                setFormData={(content) => setFieldModel(content)}
+                testFields={testField}
+                testFormData={testFieldData}
+                setTestFormData={(values) => setTestFieldData(values)}
+            >
+                <TextInput
```
gkillough, line 59: Should we include ids in these fields?

gkillough (approved):

jamesrichard91 (approved):

ui/src/main/js/global/channels/email/EmailGlobalConfiguration.js
bamandel: Upon further inspection, all the input fields look like they could use a shared key/ID value. Aparna has already opened a ticket for adding ID's to these items and I will make this change in the PR, adding a comment to that issue now.

PR #1493, opened 2021-04-07
Added the empty MS Teams global config page

bamandel:


ui/src/main/js/global/channels/msteams/MSTeamsGlobalConfiguration.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import EmptyGlobalConfiguration from 'global/channels/EmptyGlobalConfiguration';
+import { MSTEAMS_INFO } from 'global/channels/msteams/MSTeamsModel';
+
+const MSTeamsGlobalConfiguration = () => (
+    <EmptyGlobalConfiguration label={MSTEAMS_INFO.label} description="Configure MS Teams for Alert." />
```
gkillough, line 6: Should we be saying, "No global configuration needed" for these types of channels?

gkillough (approved):

ui/src/main/js/global/channels/msteams/MSTeamsGlobalConfiguration.js
bamandel: Yup, that's what the EmptyGlobalConfiguration component ensures. It should look exactly the same as before.

PR #1503, opened 2021-04-14
Jr paged model

jamesrichard91:


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedModel.java
```diff
@@ -7,48 +7,19 @@
...
-        return pageSize;
+        super(totalPages, currentPage, pageSize, models);
     }
 
+    @Override
     @JsonIgnore
     public List<M> getModels() {
-        return models;
+        return super.getModels();
     }
```
jamesrichard91, line 24: Overriding this and calling the super method here in order to add the JsonIgnore annotation so that we ignore the getModels() field when serializing this class.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessorV2.java
```diff
@@ -69,13 +69,13 @@ private void processAndDistribute(List<AlertNotificationModel> notifications, Co
...
-        StatefulAlertPagedModel<FilteredJobNotificationWrapper> mappedNotifications = jobNotificationMapper.mapJobsToNotifications(filterableNotifications, frequencies);
+        StatefulAlertPage<FilteredJobNotificationWrapper> mappedNotifications = jobNotificationMapper.mapJobsToNotifications(filterableNotifications, frequencies);
         do {
-            for (FilteredJobNotificationWrapper jobNotificationWrapper : mappedNotifications.getCurrentModels()) {
+            for (FilteredJobNotificationWrapper jobNotificationWrapper : mappedNotifications.getCurrentPage().getModels()) {
                 processAndDistribute(jobNotificationWrapper);
             }
             mappedNotifications = mappedNotifications.retrieveNextPage();
-        } while (mappedNotifications.hasNextPage());
+        } while (mappedNotifications.hasNextPage() || !mappedNotifications.isEmpty());
```
ChomickiM, line 78: Mentioned this earlier but the logic here will be somewhat changed after I implement the JobNotificationMapper correctly in the future.

ChomickiM (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedDetails.java
```diff
@@ -0,0 +1,52 @@
...
+ * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All rights reserved worldwide.
+ */
+package com.synopsys.integration.alert.common.rest.model;
+
+import java.util.List;
+
+import com.synopsys.integration.alert.common.rest.api.ReadPageController;
+import com.synopsys.integration.util.Stringable;
+
+public class AlertPagedDetails<M extends Stringable> {
```
gkillough, line 15: I would honestly drop the `Stringable`. It's nice to have the constraint, but I think we want this class to be as flexible as possible.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedDetails.java
```diff
@@ -0,0 +1,52 @@
...
+    public static final Integer DEFAULT_PAGE_NUMBER = Integer.valueOf(ReadPageController.DEFAULT_PAGE_NUMBER);
+    public static final Integer DEFAULT_PAGE_SIZE = Integer.valueOf(ReadPageController.DEFAULT_PAGE_SIZE);
```
gkillough, line 17: I think I prefer keeping these in `AlertPagedModel` because that is the model we use in our API, but I could be convinced otherwise.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/StatefulAlertPage.java
```diff
@@ -0,0 +1,39 @@
...
+        this.nextPageRetriever = nextPageRetriever;
+    }
+
+    public boolean isEmpty() {
+        return alertPagedDetails.getModels().isEmpty();
+    }
+
+    public StatefulAlertPage<T> retrieveNextPage() throws IntegrationException {
+        AlertPagedDetails<T> nextPage = nextPageRetriever.retrieveNextPage(alertPagedDetails.getCurrentPage(), alertPagedDetails.getPageSize());
+        return new StatefulAlertPage<>(nextPage, nextPageRetriever);
```
gkillough, line 29: I think an if-statement checking `hasNextPage()` would be useful for efficiency here. Return an empty page if not.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/StatefulAlertPage.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public boolean hasNextPage() {
+        return alertPagedDetails.getCurrentPage() < alertPagedDetails.getTotalPages() - 1;
+    }
+
+    public AlertPagedDetails<T> getCurrentPage() {
+        return alertPagedDetails;
+    }
```
gkillough, line 38: I would prefer to only expose the List of models rather than the details here. It keeps the purpose of this class very constrained.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/StatefulAlertPage.java
```diff
@@ -0,0 +1,39 @@
...
+    public StatefulAlertPage(AlertPagedDetails alertPagedDetails, NextPageRetriever<T> nextPageRetriever) {
+        this.alertPagedDetails = alertPagedDetails;
+        this.nextPageRetriever = nextPageRetriever;
+    }
+
+    public boolean isEmpty() {
+        return alertPagedDetails.getModels().isEmpty();
+    }
+
+    public StatefulAlertPage<T> retrieveNextPage() throws IntegrationException {
```
gkillough, line 27: We could parameterize the  exception this throws so that some classes could throw a `RuntimeException` instead.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationMapper.java
```diff
@@ -46,17 +48,12 @@ public JobNotificationMapper(ProcessingJobAccessor processingJobAccessor) {
      * Filter by Policy name (From notification if applicable)
      * @param detailedContents List of notifications that will be iterated over and applied to jobs that are found
      * @param frequencies      an Additional filter to specify when querying data from the DB
-     * @return a {@code StatefulAlertPagedModel} where a page of distributions job is used to map to a list of notifications that were passed in.
+     * @return a {@code StatefulAlertPage} where a page of distributions job is used to map to a list of notifications that were passed in.
      */
-    public StatefulAlertPagedModel<FilteredJobNotificationWrapper> mapJobsToNotifications(List<DetailedNotificationContent> detailedContents, List<FrequencyType> frequencies) {
+    public StatefulAlertPage<FilteredJobNotificationWrapper> mapJobsToNotifications(List<DetailedNotificationContent> detailedContents, List<FrequencyType> frequencies) {
         AlertPagedModel<FilteredJobNotificationWrapper> pageOfJobsToNotification = mapPageOfJobsToNotification(detailedContents, frequencies, INITIAL_PAGE_NUMBER, PAGE_SIZE);
```
ChomickiM, line 54: Doesn't this also need to change if we are trying to avoid using the AlertPagedModel for FileredJobNoificationWrapper?

gkillough (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/model/AlertPagedDetails.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class AlertPagedDetails<M> {
+
+    // FIXME we should use terminology based on "offset" and "limit" which are standard REST API paging terms
+    private final int totalPages;
+    private final int currentPage;
+    private final int pageSize;
+    private final List<M> models;
+
+    public static <M> AlertPagedDetails<M> EMPTY_PAGE() {
```
gkillough, line 20: I don't think static methods need to be all caps. Think `List.of`, `Collections.singletonList(...)`, etc.

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/task/accumulator/BlackDuckNotificationRetriever.java
```diff
@@ -76,52 +79,21 @@ private String toDateString(SimpleDateFormat sdf, OffsetDateTime offsetDateTime)
...
         }
 
-        private boolean hasNextPage() {
-            return currentTotal == currentLimit;
+        @Override
+        public AlertPagedDetails<NotificationView> retrieveNextPage(int currentOffset, int currentLimit) throws IntegrationException {
+            int newOffset = currentOffset + currentLimit;
+            BlackDuckPageDefinition pageDefinition = new BlackDuckPageDefinition(currentLimit, newOffset);
+            BlackDuckPageResponse<NotificationView> notificationPage = retrievePageOfFilteredNotifications(blackDuckRequestBuilder, pageDefinition);
+            return new AlertPagedDetails(newOffset, currentLimit, notificationPage.getTotalCount(), notificationPage.getItems());
```
gkillough, line 95: Seeing this implementation makes me think the interface should be called not be a "next" page retriever, but simply a page retriever. Then the increment logic can be in `StatefulAlertPage`. Otherwise every implementation will repeat the step of `int newOffset = currentOffset + currentLimit;` which is the thing I believe we are trying to abstract away from the end-user of `StatefulAlertPage`.

PR #1506, opened 2021-04-15
Convert the input field components to use hooks

psantos1113:


ui/src/main/js/field/FieldMappingField.js
```diff
@@ -1,135 +1,126 @@
...
-            modalError: null,
-            modalTitle: props.newMappingTitle
-        });
-    }
+        setTableId(currenttableId);
+        setTableData(fieldMappings);
+    }, []);
 
-    handleChange({ target }) {
+    const handleChange = ({ target }) => {
```
bamandel, line 48: This function looks similar to FieldModelUtilities.handleTestChange. Maybe we can use that here as well?

bamandel: If so, we might want to change the shared function name

ui/src/main/js/field/MessageFormatter.js
```diff
@@ -1,54 +1,46 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { NavLink } from 'react-router-dom';
 
-class MessageFormatter extends Component {
-    constructor(props) {
-        super(props);
-    }
+const MessageFormatter = (props) => {
```
bamandel, line 5: Can we expand the props here?

psantos1113: yes I missed this one

psantos1113: Implemented the change.

ui/src/main/js/field/input/SelectInput.js
```diff
@@ -1,24 +1,47 @@
...
-    render() {
-        const {
-            onChange, id, inputClass, options, isSearchable, placeholder, value, removeSelected, hasMultipleValues, components, selectSpacingClass, readOnly, clearable
-        } = this.props;
-        const selectClasses = `${selectSpacingClass} d-inline-flex p-2`;
-        let closeOnSelect = true;
-        if (hasMultipleValues) {
-            closeOnSelect = false;
-        }
+const SelectInput = ({
```
bamandel, line 10: Is this still used? I know a lot of our code changed to using the dynamic select component.

ui/src/main/js/global/channels/jira/server/JiraServerGlobalConfiguration.js
```diff
@@ -93,6 +93,7 @@ const JiraServerGlobalConfiguration = ({ csrfToken, readonly }) => {
                     onChange={FieldModelUtilities.handleChange(formData, setFormData)}
                     errorName={FieldModelUtilities.createFieldModelErrorKey(JIRA_SERVER_GLOBAL_FIELD_KEYS.configurePlugin)}
                     errorValue={errors[JIRA_SERVER_GLOBAL_FIELD_KEYS.configurePlugin]}
+                    successBox
```
bamandel, line 96: Does this make the successbox appear? Is that how we originally had Jira server and cloud?

ui/src/main/js/util/fieldMapping.js
```diff
@@ -233,22 +225,27 @@ function buildOAuthEndpointField(items, field) {
...
-    const { leftSide, rightSide, mappingTitle, key } = field
-    return <FieldMappingField
-        storedMappings={value}
-        leftSideMapping={leftSide}
-        rightSideMapping={rightSide}
-        newMappingTitle={mappingTitle}
-        fieldMappingKey={key}
-        {...items}
-    />
+function buildFieldMappingField(items, field, csrfToken) {
```
bamandel, line 234: csrfToken is expected but doesn't seem to be passed

ui/src/main/js/field/input/SelectInput.js
psantos1113: No I will remove it.

ui/src/main/js/global/channels/jira/server/JiraServerGlobalConfiguration.js
psantos1113: I recall we used to show a checkbox after it installed the plugin.  Nothing in the dynamic side of things is setting the successBox in the java code so I will remove it.

ui/src/main/js/field/FieldMappingField.js
psantos1113: It is similar but the contents of it aren't like a field model field.  It is a single table row that will eventually be added to a field model.  For now this will need to be the same but perhaps we can make a more abstract handler as we implement more UI changes.

PR #1508, opened 2021-04-16
Notification to job mapping by pages

ChomickiM:
This PR is for paging the mapping of notifications to applicable jobs. In the new process, a List of notifications is provided into the NotificationProcessorV2 which will request a page of 100 jobs that match against these notifications.
The current approach is as follows:
1. Extract all notification content that we use to produce a FilteredDistributionJobRequestModel using the project names, frequency types, notification types, vulnerability severities, and policies. 
2. We pass the request model into the processingJobAccessor and are returned a page of FilteredDistributionJobResponseModels.
3. Using these jobs, we loop through all notifications saved in memory to determine which jobs apply and add them to a new Map of jobs to notifications.
4. We convert this map into a list of FilteredJobNotificationWrapper which are used by NotificationProcessor to distribute the messages to the channels.

As an additional note, the page object that is being returned is still currently being worked on in a separate PR and will likely somewhat change from this code.


processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationMapper.java
```diff
@@ -45,16 +46,37 @@ public JobNotificationMapper(JobAccessor jobAccessor) {
...
-            List<FilteredDistributionJobResponseModel> filteredDistributionJobResponseModels = retrieveMatchingJobs(detailedNotificationContent, frequencies);
-            for (FilteredDistributionJobResponseModel filteredDistributionJobResponseModel : filteredDistributionJobResponseModels) {
-                List<NotificationContentWrapper> applicableNotifications = groupedFilterableNotifications.computeIfAbsent(filteredDistributionJobResponseModel, ignoredKey -> new LinkedList<>());
-                applicableNotifications.add(detailedNotificationContent.getNotificationContentWrapper());
+            detailedNotificationContent.getProjectName().ifPresent(filteredDistributionJobRequestModel::addProjectName);
+            filteredDistributionJobRequestModel.addNotificationType(detailedNotificationContent.getNotificationContentWrapper().extractNotificationType());
+            filteredDistributionJobRequestModel.addVulnerabilitySeverities(detailedNotificationContent.getVulnerabilitySeverities());
+            detailedNotificationContent.getPolicyName().ifPresent(filteredDistributionJobRequestModel::addPolicyName);
+        }
```
gkillough, line 71: Any chance we could make this its own method?

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProcessingJobAccessor.java
```diff
@@ -0,0 +1,96 @@
...
+        Set<String> vulnerabilitySeverities = filteredDistributionJobRequestModel.getVulnerabilitySeverities().isEmpty() ? null : filteredDistributionJobRequestModel.getVulnerabilitySeverities();
+        
+        PageRequest pageRequest = PageRequest.of(pageNumber, pageLimit);
+        Page<DistributionJobEntity> pageOfDistributionJobEntities;
+
+        pageOfDistributionJobEntities = distributionJobRepository.findMatchingEnabledJobsByFilteredNotifications(frequencyTypes, notificationTypes, projectNames, policyNames, vulnerabilitySeverities, pageRequest);
+
+        List<FilteredDistributionJobResponseModel> distributionJobResponseModels = pageOfDistributionJobEntities.getContent()
+                                                                                       .stream()
+                                                                                       .map(this::convertToFilteredDistributionJobResponseModel)
```
gkillough, line 63: `.map` should be on a separate line

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationMapper.java
```diff
@@ -45,16 +46,31 @@ public JobNotificationMapper(JobAccessor jobAccessor) {
      * Filter by Policy name (From notification if applicable)
      * @param detailedContents List of notifications that will be iterated over and applied to jobs that are found
      * @param frequencies      an Additional filter to specify when querying data from the DB
-     * @return a {@code Map} where the distribution job is used to map to a list of notifications that were passed in.
+     * @return a {@code StatefulAlertPagedModel} where a page of distributions job is used to map to a list of notifications that were passed in.
```
bamandel, line 49: Small thing, but I think we want this as 'a page of distribution jobs'

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationMapper.java
```diff
@@ -45,16 +46,31 @@ public JobNotificationMapper(JobAccessor jobAccessor) {
...
+
+        AlertPagedModel<FilteredDistributionJobResponseModel> jobs = processingJobAccessor.getMatchingEnabledJobsByFilteredNotifications(filteredDistributionJobRequestModel, pageNumber, pageSize);
         for (DetailedNotificationContent detailedNotificationContent : detailedContents) {
-            List<FilteredDistributionJobResponseModel> filteredDistributionJobResponseModels = retrieveMatchingJobs(detailedNotificationContent, frequencies);
-            for (FilteredDistributionJobResponseModel filteredDistributionJobResponseModel : filteredDistributionJobResponseModels) {
-                List<NotificationContentWrapper> applicableNotifications = groupedFilterableNotifications.computeIfAbsent(filteredDistributionJobResponseModel, ignoredKey -> new LinkedList<>());
-                applicableNotifications.add(detailedNotificationContent.getNotificationContentWrapper());
+            for (FilteredDistributionJobResponseModel filteredDistributionJobResponseModel : jobs.getModels()) {
+                if (JobNotificationFilterUtils.doesNotificationApplyToJob(filteredDistributionJobResponseModel, detailedNotificationContent)) {
```
bamandel, line 70: Do we need this check? Shouldn't the jobs already be filtered by the notification content? What kind of case will return false with the given notificationContent?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/FilteredJobNotificationWrapper.java
```diff
@@ -11,10 +11,12 @@
 import java.util.UUID;
 
 import com.synopsys.integration.alert.common.enumeration.ProcessingType;
-import com.synopsys.integration.util.Stringable;
+import com.synopsys.integration.alert.common.rest.model.AlertSerializableModel;
 
 // NotificationContentWrapper is not serializable, so this class cannot be serializable (and doesn't need to be)
-public class FilteredJobNotificationWrapper extends Stringable {
+//FIXME: This can't extend AlertSerializableModel
```
bamandel, line 17: What are we trying to achieve by making this AlertSerializableModel? It's kind of confusing that we made this serializable when we have comments saying we shouldn't and a new fixme that says we can't do this.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultProcessingJobAccessor.java
```diff
@@ -0,0 +1,96 @@
...
+        Set<String> notificationTypes = filteredDistributionJobRequestModel.getNotificationTypes();
+
+        // If no policies and/or vulnerabilitySeverities exist the repository query expects a null to be passed
+        Set<String> policyNames = filteredDistributionJobRequestModel.getPolicyNames().isEmpty() ? null : filteredDistributionJobRequestModel.getPolicyNames();
+        Set<String> vulnerabilitySeverities = filteredDistributionJobRequestModel.getVulnerabilitySeverities().isEmpty() ? null : filteredDistributionJobRequestModel.getVulnerabilitySeverities();
+        
+        PageRequest pageRequest = PageRequest.of(pageNumber, pageLimit);
+        Page<DistributionJobEntity> pageOfDistributionJobEntities;
+
+        pageOfDistributionJobEntities = distributionJobRepository.findMatchingEnabledJobsByFilteredNotifications(frequencyTypes, notificationTypes, projectNames, policyNames, vulnerabilitySeverities, pageRequest);
```
bamandel, line 59: Should this initialization and declaration be on separate lines?

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/FilteredJobNotificationWrapper.java
ChomickiM: This is being addressed in a followup PR by James when we fix paging. Right now this is required because we must return an AlertPagedModel which requires that the object its passing extends AlertSerializableModel. The plan is to address this still within 6.5.0.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationMapper.java
ChomickiM: Yes this check is required because it will be using the JobNotificationFilterUtils to determine if the job applies to the notification in the list of notifications that it's looping through. When we get the jobs from line 67 we are getting a page of ALL jobs that match our pre-filtered list, but not all of these jobs match to every notification. We then need to loop through each notification and then through each job and run this check to determine if the job matches the notification. Once we have this we run the compileIfAbsent to add it to the groupedFilterableNotifications where the job is then mapped to all of its notifications. 

As an example, imagine you have 2 jobs, one for a HIGH severity vuln and another for a LOW severity vuln, and then 2 notifications for a HIGH and another for a LOW vuln. severity are produced. On line 67 we get a page containing 2 jobs, one HIGH and one LOW. Then we need to loop through the notifications but based on these jobs only the relevant notification gets mapped to each job.

bamandel (approved):

jamesrichard91 (approved):

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/StatefulAlertPage.java
```diff
@@ -0,0 +1,42 @@
...
+    public StatefulAlertPage<T, E> retrieveNextPage() throws E {
+        if (hasNextPage()) {
+            AlertPagedDetails<T> nextPage = pageRetriever.retrieveNextPage(alertPagedDetails.getCurrentPage(), alertPagedDetails.getPageSize());
+            return new StatefulAlertPage<>(nextPage, pageRetriever);
+        }
+        return new StatefulAlertPage<>(AlertPagedDetails.emptyPage(), pageRetriever);
+    }
+
+    public boolean hasNextPage() {
+        return alertPagedDetails.getCurrentPage() < alertPagedDetails.getTotalPages() - 1;
```
gkillough, line 36: This check could be wrong for different paging schemes. I believe it should be: 
```java 
return !isEmpty() && alertPagedDetails.getModels().size() == alertPagedDetails.getPageSize();
```

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
```diff
@@ -0,0 +1,93 @@
...
+        if (!doesProjectApplyToJob(filteredDistributionJobResponseModel, projectName)) {
+            return false;
+        }
+        switch (notificationTypeEnum) {
+            case VULNERABILITY:
+                List<String> notificationSeverities = detailedNotificationContent.getVulnerabilitySeverities();
+                if (!doVulnerabilitySeveritiesApplyToJob(filteredDistributionJobResponseModel, notificationSeverities)) {
+                    return false;
+                }
+                break;
```
gkillough, line 38: Can we return `doVulnerabilitySeveritiesApplyToJob(filteredDistributionJobResponseModel, notificationSeverities)`? Then we wouldn't need the negation or the `break`.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
```diff
@@ -0,0 +1,93 @@
...
+            case POLICY_OVERRIDE:
+            case RULE_VIOLATION:
+            case RULE_VIOLATION_CLEARED:
+                String policyName = detailedNotificationContent.getPolicyName().orElse("");
+                if (!doesPolicyApplyToJob(filteredDistributionJobResponseModel, policyName)) {
+                    return false;
+                }
+                break;
```
gkillough, line 46: Can we return `doesPolicyApplyToJob(filteredDistributionJobResponseModel, policyName)`?

gkillough (approved):

PR #1515, opened 2021-04-19
Bm update router dependency AND fix to readonly field

bamandel:
Updated the react router to the current version to allow the use of hooks which simplifies a lot of the routing we need to use going forward.


ChomickiM (approved):

ui/package.json
```diff
@@ -18,21 +18,21 @@
         "@fortawesome/free-solid-svg-icons": "^5.14.0",
         "@fortawesome/react-fontawesome": "^0.1.11",
         "bootstrap": "^4.5.2",
+        "connected-react-router": "^6.9.1",
         "enzyme": "^3.11.0",
         "history": "^4.10.1",
         "i": "^0.3.6",
-        "npm": "^6.14.8",
+        "npm": "^7.10.0",
```
bamandel, line 25: investigating this

PR #1533, opened 2021-04-26
Make confirm modal use children for the body.

psantos1113:


bamandel (approved):
Wow much better than before

jamesrichard91 (approved):

PR #1539, opened 2021-04-28
Convert all routes and navlinks to be static

bamandel:


ui/src/main/js/Navigation.js
```diff
@@ -15,145 +14,90 @@ import { SCHEDULING_INFO } from 'global/components/scheduling/SchedulingModel';
...
-    createNavItemForDescriptors(descriptorType, context, uriPrefix, header) {
-        const { descriptors } = this.props;
-        if (!descriptors) {
-            return null;
-        }
-        const descriptorList = DescriptorUtilities.findDescriptorByTypeAndContext(descriptors, descriptorType, context);
-        if (!descriptorList) {
-            return null;
-        }
+    const doesDescriptorExist = (key) => Object.prototype.hasOwnProperty.call(descriptorMap, key);
```
psantos1113, line 35: You use this in a couple places.  In the util directory should we have a DescriptorUtils.js file where we put the construction of the descriptor map code and this check function?

psantos1113 (approved):

PR #1554, opened 2021-05-05
Testing Slack failures in the audit accessor

ChomickiM:


channel/src/test/java/com/synopsys/integration/alert/channel/slack/distribution/SlackDistributionEventReceiverTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+        SlackJobDetailsAccessor slackJobDetailsAccessor = jobId -> Optional.of(slackJobDetailsModel);
+
+        slackDistributionEventReceiver = new SlackDistributionEventReceiver(gson, processingAuditAccessor, slackJobDetailsAccessor, slackChannel, slackChannelKey);
+    }
+
+    @AfterEach
+    public void cleanup() throws IOException {
+        mockSlackServer.shutdown();
+    }
```
gkillough, line 71: Should we construct/destruct this once rather than once per test?

ChomickiM: We created it this way to ensure that the MockWebServer queue is empty between tests so that if the first test was to fail for some reason it wouldn't impact the second test with responses still in queue.

gkillough: There's no way to clear the queue instead?

ChomickiM: I don't see a way immediately to remove it using MockWebServer's api. Even if we did though I believe this would be an issue for our mockSlackServer.getRequestCount assertions that we are doing in both tests

psantos1113 (approved):

PR #1557, opened 2021-05-05
Implement missing distribution channels.

psantos1113:
feat: Implement Azure boards and channel skeletons.

feat: Add MSTeams channel and fix azure.

feat: Add fields for Jira Cloud and Server.

feat: Add field mapping field.

feat: Jira make field mapping readonly.

### DIFF
--- a/ui/src/main/js/distribution/DistributionConfigurationForm.js
+++ b/ui/src/main/js/distribution/DistributionConfigurationForm.js
@@ -24,8 +24,15 @@ import * as DistributionRequestUtility from 'distribution/DistributionRequestUti
 import { AZURE_INFO } from 'global/channels/azure/AzureModel';
 import { BLACKDUCK_INFO } from 'global/providers/blackduck/BlackDuckModel';
 import { EMAIL_INFO } from 'global/channels/email/EmailModels';
+import { JIRA_CLOUD_INFO } from 'global/channels/jira/cloud/JiraCloudModel';
+import { JIRA_SERVER_INFO } from 'global/channels/jira/server/JiraServerModel';
+import { MSTEAMS_INFO } from 'global/channels/msteams/MSTeamsModel';
 import { SLACK_INFO } from 'global/channels/slack/SlackModels';
+import AzureDistributionConfiguration from 'distribution/channels/azure/AzureDistributionConfiguration';
 import EmailDistributionConfiguration from 'distribution/channels/email/EmailDistributionConfiguration';
+import JiraCloudDistributionConfiguration from 'distribution/channels/jira/cloud/JiraCloudDistributionConfiguration';
+import JiraServerDistributionConfiguration from 'distribution/channels/jira/server/JiraServerDistributionConfiguration';
+import MsTeamsDistributionConfiguration from 'distribution/channels/msteams/MsTeamsDistributionConfiguration';
 import SlackDistributionConfiguration from 'distribution/channels/slack/SlackDistributionConfiguration';
 
 const DistributionConfigurationForm = ({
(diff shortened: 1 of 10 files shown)


ui/src/main/js/distribution/DistributionConfigurationForm.js
```diff
@@ -163,8 +170,16 @@ const DistributionConfigurationForm = ({
         const channelKey = FieldModelUtilities.getFieldModelSingleValue(channelModel, DISTRIBUTION_COMMON_FIELD_KEYS.channelName);
         const providerFound = FieldModelUtilities.hasValue(channelModel, DISTRIBUTION_COMMON_FIELD_KEYS.providerName);
         setProviderHasChannelName(true);
-        if (channelKey === EMAIL_INFO.key) {
+        if (channelKey === AZURE_INFO.key) {
```
bamandel, line 173: In a setting like this, do we prefer a switch over a bunch of if statements?

psantos1113: Since a switch uses strict comparison we should change to that here.

ui/src/main/js/distribution/channels/azure/AzureDistributionConfiguration.js
```diff
@@ -0,0 +1,79 @@
...
+                errorValue={errors[AZURE_DISTRIBUTION_FIELD_KEYS.comment]}
+            />
+            <TextInput
+                id={AZURE_DISTRIBUTION_FIELD_KEYS.project}
+                label="Azure Project"
+                description="The project name or id in Azure Boards."
+                name={AZURE_DISTRIBUTION_FIELD_KEYS.project}
+                required
+                onChange={FieldModelUtilities.handleChange(data, setData)}
+                value={FieldModelUtilities.getFieldModelSingleValueOrDefault(data, AZURE_DISTRIBUTION_FIELD_KEYS.project)}
```
bamandel, line 33: I'm seeing a lot of references to *OrDefault which is supposed to take in 3 arguments but only 2 are passed in. Was this by design? Additionally as a note (A bug I ran into), I tried using this function previously but ran into issues with the default being shown but not actually populating the fieldmodel object as it's just a getter.

ui/src/main/js/distribution/channels/jira/cloud/JiraCloudDistributionConfiguration.js
```diff
@@ -0,0 +1,118 @@
...
+                label="Re-open Transition"
+                description="If a transition is listed (case sensitive), it will be used when re-opening an issue. This will happen when Alert receives an ADD/UPDATE operation from a provider. Note: This must be in the 'To Do' status category."
+                name={JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.reopenWorkflow}
+                readOnly={readonly}
+                onChange={FieldModelUtilities.handleChange(data, setData)}
+                value={FieldModelUtilities.getFieldModelSingleValueOrDefault(data, JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.reopenWorkflow)}
+                errorName={FieldModelUtilities.createFieldModelErrorKey(JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.reopenWorkflow)}
+                errorValue={errors[JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.reopenWorkflow]}
+            />
+            {!readonly && (
```
bamandel, line 85: Jira cloud here seems to hide the collapsiblePane where Jira server doesn't. Perhaps this check should be removed?

ui/src/main/js/distribution/channels/azure/AzureDistributionConfiguration.js
psantos1113: This was a copy paste error I will search for it.

bamandel (approved):

PR #1578, opened 2021-05-13
Fix Jira Configure Plugin Field

gkillough:


alert-database/src/main/resources/liquibase/6.5.0/fix-jira-fields.xml
```diff
@@ -0,0 +1,30 @@
...
+        <insert schemaName="alert" tableName="descriptor_fields">
+            <column name="descriptor_id" valueComputed="GET_DESCRIPTOR_ID('channel_jira_server')"/>
+            <column name="field_id" valueComputed="GET_FIELD_ID('jira.server.configure.plugin')"/>
+        </insert>
+        <delete schemaName="alert" tableName="descriptor_fields">
+            <where>
+                descriptor_id = GET_DESCRIPTOR_ID('channel_jira_cloud')
+                AND field_id = GET_FIELD_ID('jira.server.configure.plugin')
+            </where>
+        </delete>
```
jamesrichard91, line 28: Will this cause any issues with the current Jira Coud or Jira Server configurations?

gkillough: No. This field only exists for the sake of the autogenerated UI. No value is ever stored. Also, this is just redefining the relationship between the descriptor and the field, not actually deleting the field.

PR #1579, opened 2021-05-14
Bm fix distribution

bamandel:
Cleaned up some of the Distribution code to to get it to a working state.
**This does not include any button permission fixes, that still needs to be done**

### DIFF
--- a/ui/src/main/js/common/input/EndpointSelectField.js
+++ b/ui/src/main/js/common/input/EndpointSelectField.js
@@ -36,16 +36,6 @@ const EndpointSelectField = ({
     const [options, setOptions] = useState([]);
     const [requestErrorValue, setRequestErrorValue] = useState(null);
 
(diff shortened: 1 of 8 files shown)


ui/src/main/js/common/input/EndpointSelectField.js
```diff
@@ -36,16 +36,6 @@ const EndpointSelectField = ({
...
 
-    const emptyFieldValue = async () => {
-        const eventObject = {
-            target: {
-                name: fieldKey,
-                value: []
-            }
-        };
-        onChange(eventObject);
-    };
```
bamandel, line 47: This seemed to do nothing but break the field, if someone knows what this prevents let me know

psantos1113 (approved):

gkillough (approved):

ChomickiM (approved):

PR #1598, opened 2021-05-21
Expand support for additional Jira custom fields 

ChomickiM:
Multi-selection or array field types in Alert previously would pass string item types to Jira. For fields like Component/s or other option fields like checkboxes this  would fail due to the incorrect "items" being used. The new method will add the values to a list of key-value pair mappings and pass it into Jira correctly.


channel/src/test/java/com/synopsys/integration/alert/channel/jira/server/JiraServerExternalConnectionTest.java
```diff
@@ -0,0 +1,107 @@
...
+import com.synopsys.integration.alert.common.persistence.model.job.details.JiraServerJobDetailsModel;
+import com.synopsys.integration.alert.common.rest.ProxyManager;
+import com.synopsys.integration.alert.descriptor.api.JiraServerChannelKey;
+import com.synopsys.integration.alert.processor.api.extract.model.ProviderDetails;
+import com.synopsys.integration.alert.processor.api.extract.model.ProviderMessageHolder;
+import com.synopsys.integration.alert.processor.api.extract.model.SimpleMessage;
+import com.synopsys.integration.alert.test.common.TestProperties;
+import com.synopsys.integration.alert.test.common.TestPropertyKey;
+
+public class JiraServerExternalConnectionTest {
```
gkillough, line 39: We could add `@Tag(TestTags.CUSTOM_EXTERNAL_CONNECTION)` here.

bamandel (approved):

channel/src/main/java/com/synopsys/integration/alert/channel/jira/server/distribution/JiraServerProcessorFactory.java
```diff
@@ -56,6 +56,7 @@ public JiraServerProcessorFactory(
 
     @Override
     public IssueTrackerProcessor<String> createProcessor(JiraServerJobDetailsModel distributionDetails) throws AlertException {
+        //TODO: Review with team, should we call the JiraServerPropertiesFactory.createJiraProperties using a FieldModel?
         JiraServerProperties jiraProperties = jiraServerPropertiesFactory.createJiraProperties();
```
gkillough, line 60: The short answer is "no". 

The long answer: 
`jiraServerPropertiesFactory.createJiraProperties()` removes the repetitive work of retrieving information from the database when instantiating a `JiraServerProperties`. By retrieving this ourselves, we would just be creating "WET" code. If the expectation is that a `FieldModel` would be provided through the method signature, that would be muddling the intent of `IssueTrackerProcessorFactory`. An `IssueTrackerProcessorFactory` is part of a pre-defined workflow. The public API of `createProcessor` only demands distribution (job) details because that is the only piece of state that should be "externally controlled". The "internally controlled" state is the global state of Alert (including the global Jira Server config) which is expected to be configured and valid by the time an event reaches a channel.

channel/src/main/java/com/synopsys/integration/alert/channel/jira/common/distribution/custom/JiraCustomFieldResolver.java
```diff
@@ -92,6 +92,28 @@ protected Object convertValueToRequestObject(CustomFieldDefinitionModel fieldDef
...
+            case CUSTOM_FIELD_TYPE_OPTION_VALUE:
+                for (String optionValue : StringUtils.split(innerFieldValue)) {
+                    jsonArray.add(createJsonObject("value", optionValue));
+                }
+                break;
+            default:
+                throw new AlertRuntimeException(String.format("Unsupported item: '%s' for array field type", fieldArrayItems));
+        }
+        return jsonArray;
+    }
```
gkillough, line 115: I heard the plan was get rid of switch-cases, but as long as they are in our code, let's try to avoid `break;` statements as much as possible. It's too easy to make mistakes with them. Instead we can break this into two methods: 

1. `JsonArray createJsonArray(String innerFieldValue, String fieldArrayItems)`
2. `List<Object> createJsonArrayElements(String innerFieldValue, String fieldArrayItems)`

The second method should only have `return` and `throw` statements.

psantos1113 (approved):

PR #1599, opened 2021-05-21
Load test.properties from test-common

gkillough:
# Note: Alert developers will need to put their `test.properties` file into _test-common/src/main/resources_. 
Notice this is a _main_ resource and not a _test_ resource. This is because _test-common_ is a library and not a set of actual tests.

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/util/TestPropertiesFileGenerator.java
+++ b/src/test/java/com/synopsys/integration/alert/util/TestPropertiesFileGenerator.java
@@ -15,7 +15,7 @@ public class TestPropertiesFileGenerator {
     @Test
     @Disabled("This test is to generate the test.properties for new developers.")
     public void generatePropertiesFile() throws IOException {
-        String propertiesFileName = new File(TestResourceUtils.BASE_TEST_RESOURCE_DIR.getCanonicalFile(), TestResourceUtils.DEFAULT_PROPERTIES_FILE_LOCATION).getCanonicalPath();
+        String propertiesFileName = TestResourceUtils.createTestPropertiesCanonicalFilePath();
         System.out.println("Generating file: " + propertiesFileName + "..");
 
         File testPropertiesFile = new File(propertiesFileName);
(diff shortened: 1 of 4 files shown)


ChomickiM (approved):

src/test/java/com/synopsys/integration/alert/util/TestPropertiesFileGenerator.java
```diff
@@ -15,7 +15,7 @@
     @Test
     @Disabled("This test is to generate the test.properties for new developers.")
     public void generatePropertiesFile() throws IOException {
-        String propertiesFileName = TestResourceUtils.BASE_TEST_RESOURCE_DIR + "/" + TestResourceUtils.DEFAULT_PROPERTIES_FILE_LOCATION;
+        String propertiesFileName = new File(TestResourceUtils.BASE_TEST_RESOURCE_DIR.getCanonicalFile(), TestResourceUtils.DEFAULT_PROPERTIES_FILE_LOCATION).getCanonicalPath();
```
bamandel, line 18: Would having this method in TestResourceUtils be useful?

test-common/src/main/java/com/synopsys/integration/alert/test/common/TestProperties.java
```diff
@@ -18,7 +20,11 @@
     private String propertiesLocation;
 
     public TestProperties() {
-        propertiesLocation = TestResourceUtils.DEFAULT_PROPERTIES_FILE_LOCATION;
+        try {
+            propertiesLocation = new File(TestResourceUtils.BASE_TEST_RESOURCE_DIR.getCanonicalFile(), TestResourceUtils.DEFAULT_PROPERTIES_FILE_LOCATION).getCanonicalPath();
```
bamandel, line 24: Looks like we do use it again here, I would move this to TestResourceUtils if possible

psantos1113 (approved):

bamandel (approved):

PR #1602, opened 2021-05-21
Remove Descriptor field data

bamandel:


gkillough (approved):

web/src/main/java/com/synopsys/integration/alert/web/api/provider/ProviderNameSelectCustomFunctionAction.java
```diff
@@ -18,36 +16,25 @@
...
-        List<LabelValueSelectOption> options = descriptorMap.getDescriptorByType(DescriptorType.PROVIDER).stream()
-                                                   .map(descriptor -> descriptor.createMetaData(ConfigContextEnum.DISTRIBUTION))
-                                                   .flatMap(Optional::stream)
-                                                   .map(descriptorMetadata -> new LabelValueSelectOption(descriptorMetadata.getLabel(), descriptorMetadata.getName()))
-                                                   .sorted()
-                                                   .collect(Collectors.toList());
-        LabelValueSelectOptions optionList = new LabelValueSelectOptions(options);
-        return new ActionResponse<>(HttpStatus.OK, optionList);
+        // FIXME removed this as it's no longer being used. Should be deleted before release
+        return new ActionResponse(HttpStatus.OK, List.of());
```
jamesrichard91, line 37: We should Deprecate this method, and we can change the response to an error code rather than OK.

bamandel: Should we deprecate most custom function controllers this version since they will most likely be removed in the future?

jamesrichard91 (approved):

PR #1610, opened 2021-05-26
Ek update to bdc 55

ekerwin:


alert-platform/build.gradle
```diff
@@ -12,11 +12,11 @@ dependencies {
...
 
-        api 'com.synopsys.integration:integration-rest:8.0.2'
-        api 'com.synopsys.integration:integration-common:24.3.0'
-        api 'com.synopsys.integration:blackduck-common:54.2.3'
-        api 'com.synopsys.integration:int-jira-common:2.3.0'
-        api 'com.synopsys.integration:phone-home-client:4.0.4'
+        api 'com.synopsys.integration:integration-rest:9.0.2'
+        api 'com.synopsys.integration:integration-common:25.0.0'
+        api 'com.synopsys.integration:blackduck-common:55.0.0'
+        api 'com.synopsys.integration:int-jira-common:3.0.0-SNAPSHOT'
```
gkillough, line 18: Don't forget to update this after it is released.

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
```diff
@@ -73,11 +75,10 @@ public void createOrUpdateBlackDuckIssue(BlackDuckProviderIssueModel issueModel,
     }
 
     private void performRequest(HttpUrl httpUrl, HttpMethod httpMethod, IssueRequest issueRequest) throws IntegrationException {
-        Request request = new BlackDuckRequestBuilder(new Request.Builder())
-                              .url(httpUrl)
+        BlackDuckResponseRequest request = new BlackDuckRequestBuilder()
                               .method(httpMethod)
-                              .bodyContent(new StringBodyContent(gson.toJson(issueRequest)))
-                              .build();
+                              .bodyContent(new StringBodyContent(gson.toJson(issueRequest), BodyContentConverter.DEFAULT))
```
gkillough, line 80: I'm surprised there's no constructor with this default converter.

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/BlackDuckMessageComponentVersionUpgradeGuidanceService.java
```diff
@@ -31,26 +31,6 @@ public BlackDuckMessageComponentVersionUpgradeGuidanceService(ComponentService c
     }
 
     public List<LinkableItem> requestRemediationItems(ComponentVersionView componentVersionView) throws IntegrationException {
```
jamesrichard91, line 33: Why not just delete this method?

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/web/PolicyNotificationFilterCustomFunctionAction.java
```diff
@@ -102,15 +102,16 @@ private boolean isJobFilterableByPolicy(Collection<String> notificationTypes) {
     }
 
     private BlackDuckPageResponse<PolicyRuleView> retrievePolicyRules(BlackDuckServicesFactory blackDuckServicesFactory, int pageNumber, int pageSize, String searchTerm) throws IntegrationException {
+        ApiDiscovery apiDiscovery = blackDuckServicesFactory.getApiDiscovery();
```
bamandel, line 105: This is an interesting class, it looks like it is used to create full URL's for different data in BD. Originally, we would just reference these URL's using the provided static fields and then merge them with the BD URL, you can see on the left where we do exactly that using blackDuckApiClient.getUrl. Why do we need to do this at all? Can the client not perform this ApiDiscovery task itself? Seems like an unnecessary step users will have to do.

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/issue/BlackDuckProviderIssueHandler.java
ekerwin: Me too.

PR #1630, opened 2021-06-04
Move slack channel into subproject

ChomickiM:


channel-slack/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+dependencies {
+    implementation platform(project(':alert-platform'))
+
+    implementation project(':alert-common')
+    implementation project(':api-common-model')
+    implementation project(':api-channel')
+    implementation project(':api-processor')
+    implementation project(':api-descriptor')
+
+    implementation 'org.springframework:spring-web'
```
gkillough, line 12: Is this dependency necessary?

gkillough (approved):

channel-slack/build.gradle
ChomickiM: I added it early on to fix up some tests since it was included in the channels subproject. Just removed it and tests are  still passing so I'll remove this dependency.

PR #1640, opened 2021-06-11
Fix issue tracker errors displayed as success

ChomickiM:
Given our  refactor  in 6.5.0 we no longer generate AlertFieldExceptions for validation failures. Since the MessageResults that we now return do not have an error or warning associated with them the UI will not generate any failure messages. This workaround will create "empty" AlertFieldStatus errors that will tell the UI there was an error and where it failed in the banner.

Note that due to these changes we are no longer able to display field errors beneath each field when running test configuration.


channel-api/src/main/java/com/synopsys/integration/alert/channel/api/issue/action/IssueTrackerTestAction.java
```diff
@@ -144,4 +149,9 @@ private ProjectIssueModel createPlaceholderProjectIssueModel(Long blackDuckConfi
         );
     }
 
+    //TODO: This addresses that hasErrors is set in the UI but this should be improved upon since it is not associated with a field
```
gkillough, line 152: Let's make sure this finds its way into a Jira ticket. "Provide a better mechanism for reporting errors from test-actions" or something like that.

gkillough (approved):

jamesrichard91 (approved):

bamandel (approved):

PR #1644, opened 2021-06-14
Updating description for the Jira Disable Plugin Check.

jamesrichard91:


bamandel:
Did these show in the UI? You should need to update the UI description in its respective React page in 6.5.0+

PR #1658, opened 2021-06-18
saml want assertions

bamandel:


bamandel:
Just want to verify that the DB changes look OK and that there isn't any other places SAML would need to be changed. I believe only the metadata needs to be updated

component/src/main/java/com/synopsys/integration/alert/component/authentication/descriptor/AuthenticationUIConfig.java
```diff
@@ -46,6 +46,7 @@
     private static final String LABEL_LDAP_GROUP_SEARCH_FILTER = "LDAP Group Search Filter";
     private static final String LABEL_LDAP_GROUP_ROLE_ATTRIBUTE = "LDAP Group Role Attribute";
     private static final String LABEL_SAML_ENABLED = "SAML Enabled";
+    private static final String LABEL_SAML_WANT_ASSERTIONS_SIGNED = "Sign Assertions";
```
jamesrichard91, line 49: Are the changes to this class even needed at this point?

bamandel: Not necessarily, but I added them for now while they still exist just in case

alert-database/src/main/resources/liquibase/6.6.0/saml-field-update.xml
```diff
@@ -0,0 +1,34 @@
...
+        <insert schemaName="ALERT" tableName="DESCRIPTOR_FIELDS">
+            <column name="DESCRIPTOR_ID" valueComputed="GET_DESCRIPTOR_ID('component_authentication')"/>
+            <column name="FIELD_ID" valueComputed="GET_FIELD_ID('settings.saml.want.assertions.signed')"/>
+        </insert>
+        <insert schemaName="ALERT" tableName="FIELD_CONTEXTS">
+            <column name="FIELD_ID" valueComputed="GET_FIELD_ID('settings.saml.want.assertions.signed')"/>
+            <column name="CONTEXT_ID" valueComputed="GET_CONTEXT_ID('GLOBAL')"/>
+        </insert>
+    </changeSet>
+    <changeSet id="bmandel" author="saml-want-assertions-field-default">
```
gkillough, line 25: `id` and `author` are switched

bamandel: lol, good find

gkillough (approved):

jamesrichard91 (approved):

PR #1688, opened 2021-07-14
Bm jira cloud summary

bamandel:
Let me know if you guys would prefer a code review, just want to make sure I made changes in the expected places


api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
```diff
@@ -25,17 +21,24 @@
...
-    public static JiraCustomFieldReplacementValues trivial(LinkableItem provider) {
-        return new JiraCustomFieldReplacementValues(provider.getLabel(), DEFAULT_REPLACEMENT, null, null, null, null);
+    public static MessageReplacementValues trivial(String providerLabel, String projectName) {
+        return new MessageReplacementValues(
+            providerLabel,
+            projectName,
+            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE,
+            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE,
+            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE,
+            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE
```
gkillough, line 31: Is there a difference between these being `DEFAULT_REPLACEMENT` and `null`? I have a feeling behavior will change.

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
```diff
@@ -25,17 +21,24 @@
...
         String providerName,
         String projectName,
-        @Nullable String projectVersionName,
-        @Nullable String componentName,
-        @Nullable String componentVersionName,
-        @Nullable String severity
+        String projectVersionName,
+        String componentName,
+        String componentVersionName,
+        String severity
```
ChomickiM, line 41: These values a can still be null if they are not provided. Do we no longer need the @Nullable annotation on them?

channel-jira-cloud/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/descriptor/JiraCloudDistributionUIConfig.java
```diff
@@ -71,12 +71,13 @@ public JiraCloudDistributionUIConfig(Gson gson) {
         ConfigField resolveWorkflow = new TextInputConfigField(JiraCloudDescriptor.KEY_RESOLVE_WORKFLOW_TRANSITION, LABEL_RESOLVE_WORKFLOW_TRANSITION, DESCRIPTION_RESOLVE_WORKFLOW_TRANSITION);
         ConfigField openWorkflow = new TextInputConfigField(JiraCloudDescriptor.KEY_OPEN_WORKFLOW_TRANSITION, LABEL_OPEN_WORKFLOW_TRANSITION, DESCRIPTION_OPEN_WORKFLOW_TRANSITION)
                                        .applyRequiredRelatedField(resolveWorkflow.getKey());
+        ConfigField issueSummary = new TextInputConfigField(JiraCloudDescriptor.KEY_ISSUE_SUMMARY, "", "");
```
gkillough, line 74: Label? Description?

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
ChomickiM: Example in JiraIssueCreator::113, we could still be passing in null values since we don't use the "trivial" static method

channel-jira-cloud/src/main/java/com/synopsys/integration/alert/channel/jira/cloud/descriptor/JiraCloudDistributionUIConfig.java
gkillough: Oh nvm. This tech debt it confusing...

bamandel: I could just not add the field here as well if this makes it too confusing.

gkillough (changes requested):
I think adding the default replacement values might be a bug. Don't merge until that is checked.  

Also, it would have been great for this to be broken up into at least 3 PRs separating the concerns: 
- One for refactoring the replacement values class (responsible for a lot of the extra files in this PR) 
-  At least one for adding/storing the field 
- One for the UI

ui/src/main/js/page/channel/jira/cloud/JiraCloudDistributionConfiguration.js
```diff
@@ -85,6 +85,17 @@ const JiraCloudDistributionConfiguration = ({
                 errorName={FieldModelUtilities.createFieldModelErrorKey(JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.reopenWorkflow)}
                 errorValue={errors.fieldErrors[JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.reopenWorkflow]}
             />
+            <TextInput
+                id={JIRA_CLOUD_DISTRIBUTION_FIELD_KEYS.issueSummary}
+                label="Issue Summary"
+                description="The summary to use for each issue created. Can use the following variables to populate data from the message content {{providerName}}, {{projectName}}, {{projectVersion}}, {{componentName}}, {{componentVersion}}, {{severity}}"
```
ChomickiM, line 91: We have these documented in confluence currently for the Jira custom fields. As we need to add more and more replacement values custom fields we would need to change this description each time as well. Perhaps we need a new section in confluence to track these changes and direct customers there? https://synopsys.atlassian.net/wiki/spaces/INTDOCS/pages/520781913/Configuring+Distribution+Jobs+in+Alert#Jira-Server

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
bamandel: If you check master the original functionality would use null initially in this method then change it to none a couple steps later. So we were always pushing none up to Jira
```
    private static String replaceFieldValues(String originalFieldValue, JiraCustomFieldReplacementValues replacementValues) {
        String replacedFieldValue = StringUtils.replace(originalFieldValue, REPLACEMENT_PROVIDER_NAME, replacementValues.getProviderName());
        replacedFieldValue = StringUtils.replace(replacedFieldValue, REPLACEMENT_PROJECT_NAME, replacementValues.getProjectName());
        replacedFieldValue = StringUtils.replace(replacedFieldValue, REPLACEMENT_PROJECT_VERSION, replacementValues.getProjectVersionName().orElse(JiraCustomFieldReplacementValues.DEFAULT_REPLACEMENT));
        replacedFieldValue = StringUtils.replace(replacedFieldValue, REPLACEMENT_COMPONENT_NAME, replacementValues.getComponentName().orElse(JiraCustomFieldReplacementValues.DEFAULT_REPLACEMENT));
        replacedFieldValue = StringUtils.replace(replacedFieldValue, REPLACEMENT_COMPONENT_VERSION, replacementValues.getComponentVersionName().orElse(JiraCustomFieldReplacementValues.DEFAULT_REPLACEMENT));
        replacedFieldValue = StringUtils.replace(replacedFieldValue, REPLACEMENT_SEVERITY, replacementValues.getSeverity().orElse(JiraCustomFieldReplacementValues.DEFAULT_REPLACEMENT));
        return replacedFieldValue;
    }
```
Since this was originally intended for fields, perhaps there's a better alternative? Do we want to insert 'None' into the variables when none is found or empty? Empty might be confusing imo

ui/src/main/js/page/channel/jira/cloud/JiraCloudDistributionConfiguration.js
bamandel: Yes this is definitely something that we should be concerned about especially as we add more fields as you suggested. I will add an update to the documentation for this and perhaps as we add more, we remove them from the description when it gets too long.

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
gkillough: Ok, this probably makes sense because if a field is required, it might not be able to be created without a value.

gkillough (approved):

ChomickiM (approved):

PR #1698, opened 2021-07-16
Bm jira server custom summary

bamandel:


ChomickiM (approved):

channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
```diff
@@ -40,6 +42,7 @@ public JiraServerRestConfig createJiraServerConfig() throws IssueTrackerExceptio
         jiraServerConfigBuilder.setAuthPassword(password);
         jiraServerConfigBuilder.setAuthUsername(username);
         jiraServerConfigBuilder.setProxyInfo(proxyInfo);
+        jiraServerConfigBuilder.setTrustCert(trustCert);
```
gkillough, line 45: What's this about?

gkillough (approved):
I thought we were trying to distance ourselves from the "trust cert" logic? Everything else seems good.

channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/JiraServerProperties.java
bamandel: This was added as jira server wouldn't use the trustcert value passed into Alert when communicating causing pkix errors. Seemed like a bug to me as well as one that was preventing me from properly testing this functionality.

PR #1718, opened 2021-07-29
Updating the readme files with the external database requirements

jamesrichard91:


deployment/helm/synopsys-alert/README.md
```diff
@@ -500,6 +501,42 @@ For the on-premise database deployment a second Persistent Volume must be create
...
+        - DELETE
+        - EXECUTE
+    - The Alert User role should have the following privileges on all objects in the alert schema (or the schema/database when relevant):
+        - SELECT
+        - INSERT
+        - UPDATE
+        - DELETE
+        - EXECUTE
+
 - On the external database create a user to own the Alert database
```
jamesrichard91, line 540: Are the steps from line 540 - 554 still correct? They seem  to conflict with our documented requirements.

gkillough: @psantos1113 Is DB ownership required? It doesn't seem to be based on our liquibase. We just need to be able to perform the necessary operations in our liquibase scripts and at runtime.

deployment/helm/synopsys-alert/README.md
```diff
@@ -15,6 +15,7 @@ Helm Charts for Synopsys Alert
   - [Custom Certificates](#custom-certificates)
   - [Persistent Storage](#persistent-storage)
   - [External Postgres Database](#external-postgres-database)
+    - [External Postgres Database Requirements](#external-postgres-database-requirements)
```
psantos1113, line 18: Appears to be an extra whitespace character here.

deployment/helm/synopsys-alert/README.md
psantos1113: They don't have to create a separate user or database for Alert.  It isn't required based on what you've added to the document.  If they have an existing database and user with the permissions, then that can be used.  The steps to create a user and a database were here to help simplify the setup process for a user of Alert and separate the Alert data from any other database data.  Ultimately it's the database administrators choice.

PR #1721, opened 2021-08-02
Fixed Azure OAuth validation error

bamandel:
This was a regression from 6.5.0. Not sure what changed to break it but made similar changes from when this was originally fixed.


web/src/main/java/com/synopsys/integration/alert/web/api/config/ConfigActions.java
```diff
@@ -197,13 +197,13 @@ public ConfigActions(AuthorizationManager authorizationManager, DescriptorAccess
     protected ValidationActionResponse validateWithoutChecks(FieldModel resource) {
         if (!encryptionUtility.isInitialized() && !settingsDescriptorKey.getUniversalKey().equals(resource.getDescriptorName())) {
             ValidationResponseModel validationResponseModel = ValidationResponseModel.generalError(EncryptionSettingsValidator.ENCRYPTION_MISSING);
-            return new ValidationActionResponse(validationResponseModel);
+            return new ValidationActionResponse(HttpStatus.BAD_REQUEST, validationResponseModel);
```
gkillough, line 200: This might be changing behavior.

gkillough: Although I can see now that this is the bug fix. Maybe a 500 rather than 400? Since it's not a problem with the request, but with the state of the server?

PR #1723, opened 2021-08-02
Pass BlackDuckServicesFactory to create services needed by extractors and update ComponentPolicy

ChomickiM:
For IALERT-1850

This PR is part 1 of 2 for changes in order to get the policy category. Right now, this information is only available in PolicyRuleView which we did not previously use. The changes proposed in this PR is to add the PolicyRuleService in Alert to get the PolicyRuleView. 

To accomplish this, we've refactored BlackDuckMessageBomComponentDetailsCreatorFactory.createBomComponentDetailsCreator to instead accept a BlackDuckServicesFactory and create a BlackDuckMessageBomComponentDetailsCreator with the required services including BlackDuckApiClient.

Discussing with Paulo we feel that this simplifies the, *NotificationMessageExtractor classes that call this creator. They would not be required to create a PolicyRuleService and instead simplifies that code to only be called once in the factory. 

I am hoping to get feedback around this change and see if the team is okay with this approach.

The second part of this PR is adding in the "category" field into ComponentPolicy to demonstrate how this new service would be used.  Implementing this change will happen in a following PR.


provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/BlackDuckMessageBomComponentDetailsCreator.java
```diff
@@ -148,10 +154,20 @@ private ComponentVulnerabilities retrieveComponentVulnerabilities(ProjectVersion
...
 
+    private Optional<PolicyRuleView> getPolicyRuleView(String policyName) {
+        Optional<PolicyRuleView> policyRuleView;
+        try {
+            policyRuleView = policyRuleService.getPolicyRuleViewByName(policyName);
+        } catch (IntegrationException e) {
+            return Optional.empty();
+        }
+        return policyRuleView;
+    }
```
gkillough, line 169: This logic seems like it would fit better in `BlackDuckComponentPolicyDetailsCreator`. 

Also, the Alert team prefers `fetch` or `retrieve` instead of `get` as a prefix for this type of method.

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/policy/BlackDuckComponentPolicyDetailsCreator.java
```diff
@@ -30,11 +32,12 @@ public BlackDuckComponentPolicyDetailsCreator(BlackDuckPolicySeverityConverter p
         this.policySeverityConverter = policySeverityConverter;
     }
 
-    public ComponentPolicy toComponentPolicy(ComponentPolicyRulesView componentPolicyRulesView) {
+    public ComponentPolicy toComponentPolicy(ComponentPolicyRulesView componentPolicyRulesView, Optional<PolicyRuleView> policyRuleView) {
```
gkillough, line 35: Taking something optional as a method parameter is ill-advised. I suspect if you moved the policy retrieval logic into this class this could be removed.

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/BlackDuckMessageBomComponentDetailsCreatorFactory.java
```diff
@@ -27,8 +29,10 @@ public BlackDuckMessageBomComponentDetailsCreatorFactory(
         this.policyDetailsCreator = policyDetailsCreator;
     }
 
-    public BlackDuckMessageBomComponentDetailsCreator createBomComponentDetailsCreator(BlackDuckApiClient blackDuckApiClient) {
-        return new BlackDuckMessageBomComponentDetailsCreator(blackDuckApiClient, vulnerabilityDetailsCreator, policyDetailsCreator);
+    public BlackDuckMessageBomComponentDetailsCreator createBomComponentDetailsCreator(BlackDuckServicesFactory blackDuckServicesFactory) {
+        BlackDuckApiClient blackDuckApiClient = blackDuckServicesFactory.getBlackDuckApiClient();
+        PolicyRuleService policyRuleService = blackDuckServicesFactory.createPolicyRuleService();
```
bamandel, line 34: Instead of passing in BlackDuckServicesFactory we generally want to pass in BlackDuckApiClient and PolicyRuleService. Since we don't use the object to perform any function or modify it in anyway we don't really need the factory. It should also make testing simpler.

ChomickiM: The reasoning for this was that the extractors would then need to also call the BlackDuckServicesFactory.createPolicyRuleService() in each of the extractor classes such as VulnerabiliyNotificationMessageExtractor. Rather than call that 4 times across those classes we only call it once here.

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/processor/message/service/policy/BlackDuckComponentPolicyDetailsCreator.java
```diff
@@ -30,11 +32,12 @@ public BlackDuckComponentPolicyDetailsCreator(BlackDuckPolicySeverityConverter p
...
         ComponentConcernSeverity componentConcernSeverity = policySeverityConverter.toComponentConcernSeverity(componentPolicyRulesView.getSeverity().name());
         boolean overridden = ProjectVersionComponentPolicyStatusType.IN_VIOLATION_OVERRIDDEN.equals(componentPolicyRulesView.getPolicyApprovalStatus());
         boolean vulnerabilityPolicy = isVulnerabilityPolicy(componentPolicyRulesView);
-        return new ComponentPolicy(componentPolicyRulesView.getName(), componentConcernSeverity, overridden, vulnerabilityPolicy, componentPolicyRulesView.getDescription());
+        String category = policyRuleView.map(ruleView -> ruleView.getCategory().name()).orElse(null);
```
bamandel, line 39: This can be mapped a second time instead of using a lambda.
```suggestion
        String category = policyRuleView.map(PolicyRulesView::getCategory).map(PolicyRuleCategoryType::name).orElse(null);
```

bamandel (approved):

bamandel (approved):

gkillough (approved):

PR #1729, opened 2021-08-03
NamedDistributionJobDetailsModel wrapper added

bamandel:
Added a wrapper to pass jobName along with the job details for future improvements

### DIFF
--- a/api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/IssueTrackerChannel.java
+++ b/api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/IssueTrackerChannel.java
@@ -14,6 +14,7 @@ import com.synopsys.integration.alert.api.channel.issue.model.IssueTrackerRespon
 import com.synopsys.integration.alert.api.common.model.exception.AlertException;
 import com.synopsys.integration.alert.common.message.model.MessageResult;
 import com.synopsys.integration.alert.common.persistence.model.job.details.DistributionJobDetailsModel;
+import com.synopsys.integration.alert.common.persistence.model.job.details.NamedDistributionJobDetailsModel;
 import com.synopsys.integration.alert.processor.api.extract.model.ProviderMessageHolder;
 
 /**
@@ -30,8 +31,8 @@ public abstract class IssueTrackerChannel<D extends DistributionJobDetailsModel,
     }
 
     @Override
-    public MessageResult distributeMessages(D distributionDetails, ProviderMessageHolder messages) throws AlertException {
-        IssueTrackerProcessor<T> processor = processorFactory.createProcessor(distributionDetails);
+    public MessageResult distributeMessages(NamedDistributionJobDetailsModel<D> distributionDetails, ProviderMessageHolder messages) throws AlertException {
+        IssueTrackerProcessor<T> processor = processorFactory.createProcessor(distributionDetails.getDistributionJobDetailsModel());
         IssueTrackerResponse<T> issueTrackerResponse = processor.processMessages(messages);
 
         responsePostProcessor.postProcess(issueTrackerResponse);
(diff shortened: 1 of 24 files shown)


api-channel/src/main/java/com/synopsys/integration/alert/api/channel/DistributionEventHandler.java
```diff
@@ -25,19 +28,24 @@
...
-        Optional<D> details = jobDetailsAccessor.retrieveDetails(event.getJobId());
-        if (details.isPresent()) {
+        Optional<D> detailsOptional = jobDetailsAccessor.retrieveDetails(event.getJobId());
+        if (detailsOptional.isPresent()) {
+            D details = detailsOptional.get();
+            String jobName = jobAccessor.getJobById(details.getJobId()).map(DistributionJobModel::getName).orElse("UNKNOWN");
+            NamedDistributionJobDetailsModel<D> namedDistributionJobDetailsModel = new NamedDistributionJobDetailsModel<>(details, jobName);
             try {
-                channel.distributeMessages(details.get(), event.getProviderMessages());
+                channel.distributeMessages(namedDistributionJobDetailsModel, event.getProviderMessages());
```
bamandel, line 48: Main change

api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/IssueTrackerChannel.java
```diff
@@ -30,8 +31,8 @@ protected IssueTrackerChannel(IssueTrackerProcessorFactory<D, T> processorFactor
     }
 
     @Override
-    public MessageResult distributeMessages(D distributionDetails, ProviderMessageHolder messages) throws AlertException {
-        IssueTrackerProcessor<T> processor = processorFactory.createProcessor(distributionDetails);
+    public MessageResult distributeMessages(NamedDistributionJobDetailsModel<D> distributionDetails, ProviderMessageHolder messages) throws AlertException {
+        IssueTrackerProcessor<T> processor = processorFactory.createProcessor(distributionDetails.getDistributionJobDetailsModel());
```
gkillough, line 35: Do people not want this on issue-tracker channels? Maybe included with the _create_ comment?

api-channel/src/main/java/com/synopsys/integration/alert/api/channel/DistributionEventHandler.java
```diff
@@ -25,19 +28,24 @@
...
 
     @Override
     public final void handle(DistributionEvent event) {
-        Optional<D> details = jobDetailsAccessor.retrieveDetails(event.getJobId());
-        if (details.isPresent()) {
+        Optional<D> detailsOptional = jobDetailsAccessor.retrieveDetails(event.getJobId());
+        if (detailsOptional.isPresent()) {
+            D details = detailsOptional.get();
+            String jobName = jobAccessor.getJobById(details.getJobId()).map(DistributionJobModel::getName).orElse("UNKNOWN");
+            NamedDistributionJobDetailsModel<D> namedDistributionJobDetailsModel = new NamedDistributionJobDetailsModel<>(details, jobName);
```
gkillough, line 46: I really struggle with making two DB calls here. Is there no chance that `JobDetailsAccessor::retrieveDetails` could return a `NamedDistributionJobDetailsModel<D>`? The entire point of `JobDetailsAccessor` is so that we don't have to retrieve the entire job.

bamandel: That was my original approach until I realized that there's a JobDetailsAccessor object for each specific channel which only queries the jobs specific table (Which doesn't have my data). So I'd need to join tables or do 2 queries still. this seemed like the best way to make the fewest amount of changes. Is there are better spot to retrieve this data than something like DefaultAzureBoardsJobDetailsAccessor?

api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/IssueTrackerChannel.java
bamandel: They do, haven't looked into the best place to make this change for issue trackers but will probably fix this in my next PR which actually adds the feature we want.

api-channel/src/main/java/com/synopsys/integration/alert/api/channel/DistributionEventHandler.java
gkillough: Those accessors are designed for this class. If this class needs new things, those accessors should change IMO.

PR #1730, opened 2021-08-04
Policy category field implementation

ChomickiM:
Continuation of https://github.com/blackducksoftware/blackduck-alert/pull/1723

This PR is the implementation for the changes made to PolicyCategory. We've added the new custom field in Jira for {{policyCategory}} populated by the field with the same name in policy management in BlackDuck.

### DIFF
--- a/api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
+++ b/api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
@@ -22,11 +22,13 @@ public class MessageReplacementValues {
     private final String componentName;
     private final String componentVersionName;
     private final String severity;
+    private final String policyCategory;
 
     public static MessageReplacementValues trivial(String providerLabel, String projectName) {
         return new MessageReplacementValues(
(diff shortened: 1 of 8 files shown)


api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
```diff
@@ -40,14 +42,16 @@ public MessageReplacementValues(
         @Nullable String projectVersionName,
         @Nullable String componentName,
         @Nullable String componentVersionName,
-        @Nullable String severity
+        @Nullable String severity,
+        @Nullable String policyCategory
     ) {
```
gkillough, line 47: It doesn't need to happen now, but this class has enough nullable fields that it should probably have a builder.

ChomickiM: That may be a good idea for a future improvement for this class. Not including this change, we have a request for 4 additional custom fields in 6.7.0

bamandel (approved):

gkillough (approved):

PR #1731, opened 2021-08-04
Clean project input causing issues with validation

gkillough:


alert-common/src/main/java/com/synopsys/integration/alert/common/action/api/AbstractJobResourceActions.java
```diff
@@ -219,4 +237,27 @@ private boolean checkContextAndDescriptorKey(FieldModel fieldModel, BiFunction<C
...
     }
 
+    // FIXME More tech debt until we fix the Jobs API
+    private void correctProjectsField(JobFieldModel jobFieldModel) {
+        List<JobProviderProjectFieldModel> projects = jobFieldModel.getConfiguredProviderProjects();
+        if (null == projects) {
+            projects = List.of();
+        }
+
+        String projectFieldKey = "channel.common.configured.project";
```
bamandel, line 247: Should this project have access to this static key?

gkillough: No, which is why this is a temporary fix :'(

bamandel (approved):

PR #1738, opened 2021-08-05
Update MessageReplacementValues to use a builder object

ChomickiM:


api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValuesBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static MessageReplacementValues trivial(String providerLabel, String projectName) {
+        return new MessageReplacementValuesBuilder(providerLabel, projectName)
+                   .projectVersionName(MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE)
+                   .componentName(MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE)
+                   .componentVersionName(MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE)
+                   .severity(MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE)
+                   .policyCategory(MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE)
+                   .build();
+    }
```
gkillough, line 20: Since the constructor takes the two required params, I don't think this static initializer is still needed. In the `build()` method, we should have a `defaultIfBlank()` method to handle replacing values with the default as needed.

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
```diff
@@ -23,20 +23,8 @@
...
-            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE,
-            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE,
-            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE,
-            MessageReplacementValues.DEFAULT_NOTIFICATION_REPLACEMENT_VALUE
-        );
-    }
-
-    public MessageReplacementValues(
+    
+    protected MessageReplacementValues(
```
bamandel, line 27: Should this be package private since we don't use the child functionality that protected offers?

bamandel (approved):

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
gkillough, line 27: I would almost prefer that this be private and the builder was a `public static final` inner-class. Then this class could have a static initializer for the builder.

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/custom/MessageReplacementValues.java
```diff
@@ -82,4 +70,73 @@ public String getProjectName() {
...
+            return this;
+        }
+
+        public Builder policyCategory(String policyCategory) {
+            this.policyCategory = policyCategory;
+            return this;
+        }
+
+        private String defaultIfBlank(String replacementField) {
+            if (!StringUtils.isBlank(replacementField)) {
```
gkillough, line 135: `StringUtils` has an `isNotBlank(...)` method as well. I think that would be slightly more readable here. Or we could remove the negation and swap the returns. Any time we can avoid negation is a win for readability.

gkillough (approved):

PR #1743, opened 2021-08-06
Demo: Validator with explicit returns

gkillough:
## DO NOT MERGE 

Making our validation methods `void` muddles intent and makes the code less expressive. For someone new attempting to use the `ConfigurationFieldValidator`, the method provides no insight into what will happen if validation fails. 

Will it throw a runtime exception? 
Will it log an error? 
Will it add to some data structure that the method tells me nothing about? (In this case, it was this one). 

By not providing more information in the method signature like a return type or an exception declaration, the only way for an unfamiliar developer to know what this method does is to actually look at the implementation details. The problem of lore-based code has plagued Alert in many places and we can avoid future headaches by writing expressive, maintainable code. 

This PR is not meant to be merged (as usages of this class have not been updated), but it is meant to be a contrast between more obscure and less obscure code.

There are plenty of ways these methods could be improved further (e.g. single return-paths, validation results object, etc), but a bare minimum should be to return the results of the operation which the method performs.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/validator/ConfigurationFieldValidator.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/validator/ConfigurationFieldValidator.java
@@ -12,7 +12,7 @@ import java.net.URL;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/validator/ConfigurationFieldValidator.java
```diff
@@ -51,76 +50,82 @@ public static ConfigurationFieldValidator fromJobFieldModel(JobFieldModel jobFie
 
     public ConfigurationFieldValidator(Map<String, FieldValueModel> mapToValidate) {
         this.fieldMap = mapToValidate;
-        statuses = new HashSet<>();
     }
 
-    public Set<AlertFieldStatus> getValidationResults() {
-        return statuses;
-    }
```
gkillough, line 59: Nothing about the other methods tells me I will need to call this at some point. With a more descriptive class name it would put me on the right track, but ultimately it would still be lore-based.

PR #1767, opened 2021-08-10
Message replacement component usage/license

ChomickiM:
Add two new message replacement values as well as cleaning up some tests to better reflect the classes they are supposed to be testing.


gkillough (approved):

ui/src/main/js/common/input/FieldMappingField.js
```diff
@@ -57,7 +57,7 @@ const FieldMappingField = ({
 
     const createNewRow = () => {
         const { fieldName, fieldValue } = fieldMappingRow;
-        const valueOptions = ['{{providerName}}', '{{projectName}}', '{{projectVersion}}', '{{componentName}}', '{{componentVersion}}', '{{severity}}', '{{policyCategory}}'];
+        const valueOptions = ['{{providerName}}', '{{projectName}}', '{{projectVersion}}', '{{componentName}}', '{{componentVersion}}', '{{componentUsage}}', '{{componentLicense}}', '{{severity}}', '{{policyCategory}}'];
```
bamandel, line 60: If we add anymore of these we may want to start referencing them to the docs or offer a simpler method of knowing what options they have.

bamandel (approved):

PR #1775, opened 2021-08-11
feat: Concrete Global Email Configuration Controller

rottebds:
In broad strokes, the creation of a concrete global email configuration endpoint resulted in the creation of four classes:

**EmailGlobalConfigResponse**: a data class that represents the json structure requested by the endpoint.
**EmailGlobalConfigController**: the REST controller that accepts and responds with EmailGlobalConfigResponse.
**EmailGlobalConfigActions**: the functionality of the controller, as per ConfigController / ConfigActions.
**EmailGlobalConfigResponseTransformer**: which handles transforming between the Response and the FieldModel.


channel-email/src/main/java/com/synopsys/integration/alert/channel/email/web/EmailGlobalConfigResponse.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Copyright (c) 2021 Synopsys, Inc.
+ *
+ * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All rights reserved worldwide.
+ */
+package com.synopsys.integration.alert.channel.email.web;
+
+import com.synopsys.integration.alert.common.rest.model.Config;
+
+public class EmailGlobalConfigResponse extends Config {
```
gkillough, line 12: Because this model is used in both requests and responses, I would be in favor of using the suffix "RESTModel" or something to indicate that the object is used for both.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/web/EmailGlobalConfigController.java
```diff
@@ -0,0 +1,61 @@
...
+import com.synopsys.integration.alert.common.rest.AlertRestConstants;
+import com.synopsys.integration.alert.common.rest.ResponseFactory;
+import com.synopsys.integration.alert.common.rest.api.BaseResourceController;
+import com.synopsys.integration.alert.common.rest.api.TestController;
+import com.synopsys.integration.alert.common.rest.api.ValidateController;
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+
+@RestController
+@RequestMapping(AlertRestConstants.EMAIL_CONFIGURATION_PATH)
+public class EmailGlobalConfigController implements BaseResourceController<EmailGlobalConfigResponse>, TestController<EmailGlobalConfigResponse>, ValidateController<EmailGlobalConfigResponse> {
```
gkillough, line 23: I wonder if we could also create a `ChannelGlobalConfigController` superclass that implements the relevant interfaces and handles all of the calls to `ResponseFactory.createContentResponseFromAction(...)`. Not for this PR, but I wonder if there is ever a case where we don't do that.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/web/EmailGlobalConfigResponse.java
rottebds: Updated to Model to be consistent with the rest of Alert.

Where did Model come from? I'm seeing the term Resource a lot in RESTful documentation, but not so much Model.

gkillough: I have no preference if it is called, "model" or "resource". I just want to make it clear that the object represents more than _responses_ since it is also used for requests.

PR #1781, opened 2021-08-13
Add missing builder method for ComponentItemBuilder

ChomickiM:
While working on upgrade guidance changes I noticed that in the following code we were creating a List<LinkableItem>. componentAttributes but did nothing with this list. I believe this was overlooked and have added in the builder step to apply the componentAttributes.

This change impacts the email attachment json file where the attributes was previously left blank.


gkillough (approved):
Good catch

PR #1782, opened 2021-08-16
Create ComponentUpgradeGuidance and implement it as a custom field replacment

ChomickiM:
Removed upgrade guidance from additional attributes and created a new object ComponentUpgradeGuidance to hold the values. This ticket was to track refactoring the places it was previously used and updating the tests.

Lastly, we implemented the changes in the  MessageValueReplacementResolver, MessageReplacementValues and its builder, and in JiraIssueCreator.

UI updates to refer to the docs will be made in a separate PR.

### DIFF
--- a/api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/model/IssueBomComponentDetails.java
+++ b/api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/model/IssueBomComponentDetails.java
@@ -14,6 +14,7 @@ import org.jetbrains.annotations.Nullable;
 import com.synopsys.integration.alert.common.message.model.LinkableItem;
 import com.synopsys.integration.alert.processor.api.extract.model.project.AbstractBomComponentDetails;
 import com.synopsys.integration.alert.processor.api.extract.model.project.ComponentPolicy;
+import com.synopsys.integration.alert.processor.api.extract.model.project.ComponentUpgradeGuidance;
 import com.synopsys.integration.alert.processor.api.extract.model.project.ComponentVulnerabilities;
 
 public class IssueBomComponentDetails extends AbstractBomComponentDetails {
@@ -28,6 +29,7 @@ public class IssueBomComponentDetails extends AbstractBomComponentDetails {
             List.of(),
             UNKNOWN_LICENSE,
             UNKNOWN_USAGE,
+            ComponentUpgradeGuidance.none(),
             List.of(),
             ""
         );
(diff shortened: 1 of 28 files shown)


api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/model/IssueBomComponentDetails.java
```diff
@@ -28,6 +29,7 @@ public static IssueBomComponentDetails fromSearchResults(LinkableItem component,
             List.of(),
             UNKNOWN_LICENSE,
             UNKNOWN_USAGE,
+            ComponentUpgradeGuidance.none(),
```
gkillough, line 32: If this is not required in this class, then let's not include it in the abstract class.

api-processor/src/main/java/com/synopsys/integration/alert/processor/api/extract/model/project/AbstractBomComponentDetails.java
```diff
@@ -24,6 +24,7 @@
 
     private final LinkableItem license;
     private final String usage;
+    private final ComponentUpgradeGuidance componentUpgradeGuidance;
```
gkillough, line 27: Based on the fact that only one implementation uses this field, it should not be in the abstract class, but rather in the implementation.

api-channel-issue-tracker/src/main/java/com/synopsys/integration/alert/api/channel/issue/model/IssueBomComponentDetails.java
ChomickiM: It's still required for the other static method, fromBomComponentDetails

api-processor/src/main/java/com/synopsys/integration/alert/processor/api/extract/model/project/ComponentUpgradeGuidance.java
```diff
@@ -0,0 +1,37 @@
...
+        this.longTermUpgradeGuidance = longTermUpgradeGuidance;
+    }
+
+    public Optional<LinkableItem> getShortTermUpgradeGuidance() {
+        return Optional.ofNullable(shortTermUpgradeGuidance);
+    }
+
+    public Optional<LinkableItem> getLongTermUpgradeGuidance() {
+        return Optional.ofNullable(longTermUpgradeGuidance);
+    }
```
gkillough, line 35: I would much rather have `ComponentUpgradeGuidance` be optional on a component than have its fields be optional (if possible).

ChomickiM: I don't believe it would make sense in this case since the short term and long term guidance could be optional. For example, it's possible there is no short term guidance but there is long term guidance or vice versa.

gkillough (approved):

PR #1783, opened 2021-08-16
fix: Resolve custom field ids to field names

rottebds:


ChomickiM (approved):
I'm not the biggest fan of changing response from Jira, but I'm not entirely sure how else we can support the customer request without it.

api-channel-jira/src/main/java/com/synopsys/integration/alert/api/channel/jira/distribution/JiraErrorMessageUtility.java
```diff
@@ -33,12 +34,18 @@ public JiraErrorMessageUtility(Gson gson) {
...
         try {
             List<String> responseErrors = extractErrorsFromResponseContent(restException.getHttpResponseContent(), issueCreatorFieldKey, issueCreatorEmail);
             if (!responseErrors.isEmpty()) {
-                message += " | Details: " + StringUtils.join(responseErrors, ", ");
+                String responseErrorString = StringUtils.join(responseErrors, ", ");
+                if (responseErrorString.contains("customfield_")) {
+                    for (String customFieldId : jiraCustomFieldResolver.getCustomFieldIds()) {
+                        responseErrorString = responseErrorString.replace(customFieldId, jiraCustomFieldResolver.resolveCustomFieldIdToName(customFieldId));
+                    }
+                }
```
gkillough, line 47: Could this code move into a separate component that has its own `JiraCustomFieldResolver`?

gkillough (approved):

PR #1789, opened 2021-08-18
sslMode added to Alert startup

bamandel:
This gives users the ability to connect to the DB from Alert using SSL if desired.

- Allows users to add Env vars for docker deployment
- Allows users to add Env vars for Helm deployment
- Added cert paths in the case that users want full verification
- Helm allows secrets to be used instead of plain text env vars


deployment/helm/synopsys-alert/templates/alert.yaml
```diff
@@ -72,6 +72,23 @@ spec:
               key: ALERT_POSTGRES_ADMIN_PASSWORD_FILE
               name: {{ .Release.Name }}-db-creds
         {{- end }}
+        {{- if (eq .Values.postgres.sslConfig.sslUseFiles true) }}
+        - name: ALERT_DB_SSL_KEY_PATH
+          valueFrom:
+            secretKeyRef:
+              key: {{ required ".Values.postgres.sslSecrets.sslKeyKey is missing." .Values.postgres.sslSecrets.sslKeyKe}}
```
psantos1113, line 79: This seems to be a typo 
`.Values.postgres.sslSecrets.sslKeyKe`

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -47,6 +47,17 @@ cfssl:
 postgres:
   registry: "docker.io/centos" # override the docker registry at container level
   isExternal: false # false for running Postgres as a container and true for using External Postgres database
+  ssl: true # If true, Alert uses SSL for external Postgres connection
+  sslConfig:
+    sslUseFiles: false # If true, Alert will expect to find ssl certs for communicating with the External Postgres database
+    sslKeyFilePath: "" # If using files for ssl, specify ssl key path
```
psantos1113, line 53: Why would the paths be used over the secrets?  Is this to map to a path in the volume?  I would assume that we just need secrets defined with the certificate contents.

deployment/helm/synopsys-alert/templates/_helpers.tpl
```diff
@@ -32,3 +32,31 @@ Environs for Alert Config Map
...
+{{- end -}}
+{{- end -}}
+
+{{/*
+Use SSL Files for External Postgres
+*/}}
+{{- define "postgresSSLFiles" -}}
+{{- if and (eq .Values.postgres.isExternal true) (eq .Values.postgres.ssl true) -}}
+{{- if (eq .Values.postgres.sslConfig.sslUseFiles true) -}}
+ALERT_DB_SSL_KEY_PATH: {{ .Values.postgres.sslConfig.sslKeyFilePath }}
```
psantos1113, line 57: If we are using secrets and we probably should just require secrets be created for those files, then do we need the file path options in the templates?  We look for the secrets in the templates below.  When would the file paths be used over the secret use case?

deployment/docker-swarm/external-db/docker-compose.local-overrides.yml
```diff
@@ -14,6 +14,10 @@ services:
 #      - ALERT_DB_HOST=alertdb
 #      - ALERT_DB_PORT=5432
 #      - ALERT_DB_NAME=alertdb
+#      - ALERT_DB_SSL_MODE=require
```
gkillough, line 17: Is using "require" a new development? I thought "allow" was the preferred default?

bamandel: 'Allow' is still the default. This setting it commented out meaning no value results in 'allow'. If they uncomment, this sets it to 'require' changing the default

deployment/helm/synopsys-alert/values.yaml
bamandel: I may not have mentioned it, but the cert related settings ask for the file paths not the actual content https://www.postgresql.org/docs/12/libpq-connect.html (search for sslkey). Since it's their paths, should I remove the helm secret functionality?

deployment/helm/synopsys-alert/templates/_helpers.tpl
bamandel: I kept the non-secret use case as we do something similar with DB credentials. This could have been for backwards compatibility but I wasn't sure so I added the ability to add them directly or to use secrets. If it was only for backwards compatibility, I can certainly use secrets or the other.

deployment/helm/synopsys-alert/values.yaml
```diff
@@ -47,6 +47,17 @@ cfssl:
 postgres:
   registry: "docker.io/centos" # override the docker registry at container level
   isExternal: false # false for running Postgres as a container and true for using External Postgres database
+  ssl: true # If true, Alert uses SSL for external Postgres connection
```
ChomickiM, line 50: is this supposed to be true by default going forward?

deployment/helm/synopsys-alert/templates/_helpers.tpl
psantos1113: The DB credentials are also strings.  We simply need the text in the case of credentials.  Here we need files stored.  Secrets are going to be stored on the system as files in the secrets directory.  For Helm I would prefer to just support secrets to have less to maintain and support.  These are SSL related files they should be treated like passwords.  Unless anyone else strongly objects.

ChomickiM (approved):

deployment/helm/synopsys-alert/values.yaml
psantos1113: secrets are stored as files.  So you need a path to the secrets file.  That's why I push for secrets.  Secrets are stored as files in a specific directory on the pod.

deployment/helm/synopsys-alert/values.yaml
bamandel: I think it would make sense if it were false for Alert. Will change that unless others object

psantos1113 (approved):

PR #1802, opened 2021-08-26
Improve Purge Task

gkillough:


alert-database/src/main/java/com/synopsys/integration/alert/database/audit/AuditEntryRepository.java
```diff
@@ -36,4 +37,10 @@
     )
     Optional<AuditEntryEntity> findMatchingAudit(Long notificationId, UUID commonConfigId);
 
+    @Query("DELETE FROM AuditEntryEntity audit"
+               + " WHERE audit.id NOT IN (SELECT relation.auditEntryId FROM com.synopsys.integration.alert.database.audit.AuditNotificationRelation relation)"
```
gkillough, line 41: Joins are more efficient than sub-queries, so this query should be updated.

gkillough: This can't be done with JPQL

PR #1807, opened 2021-08-31
Adding additional logging for notification processing

jamesrichard91:


jamesrichard91:
Should we create a new Logger for these to make them easier to filter out from the rest of the logs?

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/DistributionEventReceiver.java
```diff
@@ -43,6 +45,9 @@ public final void handleEvent(DistributionEvent event) {
         Optional<D> details = jobDetailsAccessor.retrieveDetails(event.getJobId());
         if (details.isPresent()) {
             try {
+                if (logger.isDebugEnabled()) {
```
jamesrichard91, line 48: These if's use the normal logger throughout the PR so that we only log these messages when the User has enabled debug logging for all of Alert.

alert-common/src/main/java/com/synopsys/integration/alert/common/logging/AlertLoggerFactory.java
```diff
@@ -31,4 +32,18 @@ public static Logger getLogger(String name) {
...
+    public static Logger getNotificationLogger(Class<?> clazz) {
+        Logger logger = LoggerFactory.getLogger(clazz);
+        Logger auditLogger = LoggerFactory.getLogger(ALERT_NOTIFICATION_LOGGER);
+
+        return new AlertCompositeLogger(auditLogger, logger);
+    }
+
+    public static Logger getNotificationLogger(String name) {
+        Logger logger = LoggerFactory.getLogger(name);
+        Logger auditLogger = LoggerFactory.getLogger(ALERT_NOTIFICATION_LOGGER);
```
gkillough, line 44: `auditLogger` -> `notificationLogger`

gkillough (commented):
The if-statements everywhere bloat the code. Is there any way we can abstract them? Maybe add a `debugIfEnabled` method to the composite logger?

channel-api/src/main/java/com/synopsys/integration/alert/channel/api/DistributionEventReceiver.java
```diff
@@ -61,12 +66,12 @@ public final String getDestinationName() {
     }
 
     protected void handleAlertException(AlertException e, DistributionEvent event) {
-        logger.error("An exception occurred while handling the following event: {}", event, e);
+        logger.error("An exception occurred while handling the following event: {}. {}", event.getEventId(), e);
```
gkillough, line 69: Using `{}` for the exception is unnecessary. When the last parameter is a `Throwable`, SLF4J will log the stack trace. https://www.baeldung.com/slf4j-log-exceptions

jamesrichard91:
> The if-statements everywhere bloat the code. Is there any way we can abstract them? Maybe add a `debugIfEnabled` method to the composite logger?

I think that is a good idea, I don't know the best way to do that. I used those if's because the streams will slow things down and we only want to do that if we are going to log the messages and that will only happen if debug logging is enabled.

gkillough:
> > The if-statements everywhere bloat the code. Is there any way we can abstract them? Maybe add a `debugIfEnabled` method to the composite logger?
> 
> I think that is a good idea, I don't know the best way to do that. I used those if's because the streams will slow things down and we only want to do that if we are going to log the messages and that will only happen if debug logging is enabled.

At the very least, we could make a private method for `NotificationProcessor` so that we aren't calling identical code twice.

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessor.java
```diff
@@ -55,19 +62,33 @@ public NotificationProcessor(
 
     public void processNotifications(List<AlertNotificationModel> notifications, List<FrequencyType> frequencies) {
         try {
+            if (logger.isDebugEnabled()) {
+                List<Long> notificationIds = notifications.stream()
+                                                 .map(AlertNotificationModel::getId)
+                                                 .collect(Collectors.toList());
+                String joinedIds = StringUtils.join(notificationIds, ", ");
+                notificationLogger.debug("Processing notifications: {}", joinedIds);
```
psantos1113: We may want some similar easy to search for keywords in the log messages like Start processing notifications:
Then we can search in the logs for a simple pattern `Start processing  notifications:`
 and a corresponding `Finished processing notifications:`. Just might make it easier to debug if we have a simple pattern to the debug messages.
```suggestion
                notificationLogger.debug("Start processing notifications: {}", joinedIds);
```

processor-api/src/main/java/com/synopsys/integration/alert/processor/api/NotificationProcessor.java
```diff
@@ -55,19 +62,33 @@ public NotificationProcessor(
...
+            }
             processAndDistribute(notifications, frequencies);
             notificationAccessor.setNotificationsProcessed(notifications);
+            if (logger.isDebugEnabled()) {
+                List<Long> notificationIds = notifications.stream()
+                                                 .map(AlertNotificationModel::getId)
+                                                 .collect(Collectors.toList());
+                String joinedIds = StringUtils.join(notificationIds, ", ");
+                notificationLogger.debug("These notifications have been processed: {}", joinedIds);
```
psantos1113: The second part of the pair of messages to make it easier to correlate the start and finish.
```suggestion
                notificationLogger.debug("Finished processing notifications: {}", joinedIds);
```

provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/task/accumulator/BlackDuckAccumulator.java
```diff
@@ -129,15 +132,22 @@ private void storeNotifications(List<NotificationView> notifications) {
...
     private void write(List<AlertNotificationModel> contentList) {
         logger.info("Writing {} notifications...", contentList.size());
-        notificationAccessor.saveAllNotifications(contentList);
+        List<AlertNotificationModel> savedNotifications = notificationAccessor.saveAllNotifications(contentList);
+        if (logger.isDebugEnabled()) {
+            List<Long> notificationIds = savedNotifications.stream()
+                                             .map(AlertNotificationModel::getId)
+                                             .collect(Collectors.toList());
+            String joinedIds = StringUtils.join(notificationIds, ", ");
+            notificationLogger.debug("Saving notifications: {}", joinedIds);
```
psantos1113, line 149: You might want to change this to `Saved Notifications` just to be clear it's what was written to the database.

psantos1113 (approved):

PR #1810, opened 2021-09-02
Fixing the project owner only email handling

jamesrichard91:


provider/src/main/java/com/synopsys/integration/alert/provider/blackduck/BlackDuckProviderDataAccessor.java
```diff
@@ -81,7 +81,7 @@ public BlackDuckProviderDataAccessor(ConfigurationAccessor configurationAccessor
                 BlackDuckApiClient blackDuckApiClient = blackDuckServicesFactory.getBlackDuckApiClient();
 
                 ProjectView foundProject = blackDuckApiClient.getResponse(new HttpUrl(projectHref), ProjectView.class);
-                convertBlackDuckProjects(List.of(foundProject), blackDuckApiClient);
+                return convertBlackDuckProjects(List.of(foundProject), blackDuckApiClient).stream().findFirst();
```
jamesrichard91, line 84: This return was missing, so the method was only ever returning Optional.empty()

bamandel (approved):

PR #1813, opened 2021-09-09
Ps missing notification fix

psantos1113:
Update the query to add a sort order on the job creation time.

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
@@ -39,7 +39,7 @@ public interface DistributionJobRepository extends JpaRepository<DistributionJob
     Page<DistributionJobEntity> findByChannelDescriptorNamesAndSearchTerm(@Param("channelDescriptorNames") Collection<String> channelDescriptorNames, @Param("searchTerm") String searchTerm, Pageable pageable);
 
     //TODO: Determine if its possible to pass in a RequestModel directly into  the JPA repository methods
-    @Query(value = "SELECT DISTINCT jobEntity FROM DistributionJobEntity jobEntity "
+    @Query(value = "SELECT jobEntity FROM DistributionJobEntity jobEntity "
                        + "    LEFT JOIN jobEntity.blackDuckJobDetails blackDuckDetails ON jobEntity.jobId = blackDuckDetails.jobId "
                        + "    LEFT JOIN blackDuckDetails.blackDuckJobNotificationTypes notificationTypes ON jobEntity.jobId = notificationTypes.jobId "
                        + "    LEFT JOIN blackDuckDetails.blackDuckJobPolicyFilters policyFilters ON jobEntity.jobId = policyFilters.jobId "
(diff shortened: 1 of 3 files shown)


src/test/java/com/synopsys/integration/alert/database/api/ProcessingJobAccessorTestIT.java
```diff
@@ -0,0 +1,240 @@
...
+        CREATED_JOBS.forEach(jobAccessor::deleteJob);
+        CREATED_JOBS.clear();
+        configurationAccessor.deleteConfiguration(providerConfigId);
+    }
+
+    @Test
+    public void testMatchingEnabledJobsUniqueJobsPerPage() throws InterruptedException {
+        int expectedNumOfJobs = 1000;
+        Set<UUID> previousJobIdSet = new HashSet<>();
+        createJobs(createDistributionJobModels(List.of(VulnerabilitySeverityType.LOW.name()), 100));
```
psantos1113, line 94: createJobs taken from the JobNotificationMapperTestIT class to be able to quickly reproduce the problem.

src/test/java/com/synopsys/integration/alert/database/api/ProcessingJobAccessorTestIT.java
```diff
@@ -0,0 +1,240 @@
...
+        createJobs(createDistributionJobModels(List.of(VulnerabilitySeverityType.LOW.name()), 100));
+        createJobs(createDistributionJobModels(List.of(VulnerabilitySeverityType.LOW.name()), 100));
+        createJobs(createDistributionJobModels(List.of(VulnerabilitySeverityType.LOW.name()), 100));
+        createJobs(createDistributionJobModels(List.of(VulnerabilitySeverityType.LOW.name()), 100));
+        createJobs(createDistributionJobModels(List.of(VulnerabilitySeverityType.LOW.name()), 100));
+
+        List<DetailedNotificationContent> notifications = new ArrayList<>();
+        notifications.addAll(createVulnerabilityNotificationWrappers(List.of(VulnerabilitySeverityType.LOW.name()), PROJECT_NAME_1));
+        notifications.addAll(createVulnerabilityNotificationWrappers(List.of(VulnerabilitySeverityType.HIGH.name()), PROJECT_NAME_2));
```
psantos1113, line 107: createVulnerabilityNotificationWrapper take from JobNotificationMapperTestIT.  We may want to add some utilities to make it easier to create jobs and save notifications.  Just wrote this to test the issue.  We can clean it up more in the main branch.

alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
```diff
@@ -61,8 +61,9 @@
                        + "            OR policyFilters.policyName IN (:policyNames)"
                        + "          )"
                        + "    )"
+                       + "GROUP BY jobEntity.jobId ORDER BY jobEntity.createdAt ASC"
```
gkillough, line 64: I believe we can remove `DISTINCT` from the beginning of this query now. I would also prefer that the `GROUP BY` and `ORDER BY` statements are on separate lines. Finally, because of _String Concatenation_, we should prefix each line with at least one space character.

alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
```diff
@@ -61,8 +61,9 @@
...
     )
-    Page<DistributionJobEntity> findMatchingEnabledJobsByFilteredNotifications(
+    Page<DistributionJobEntity> findMatchingEnabledJobsByFilteredNotificationsGroupByJobIdOrderByCreatedAt(
```
gkillough, line 66: Because we use a custom query, I don't think this method name being this verbose is necessary. Perhaps something like `findAndSortMatchingEnabledJobsByFilteredNotifications` (which is already pretty long).

gkillough (approved):

PR #1820, opened 2021-09-14
Test coverage for api-processors

ChomickiM:
Adding coverage for the JobNotificationProcessorTest and NotificationContentProcessorTest. There exists some duplication setting up extractors, content, and views. As mentioned in the meeting it may be worthwhile to investigate future test improvements to create test utilities for the extractor and content classes.

### DIFF
--- a/api-processor/build.gradle
+++ b/api-processor/build.gradle
@@ -10,7 +10,6 @@ dependencies {
     implementation 'com.synopsys.integration:blackduck-common', rootProject.ext.blackduckCommonExcludes
     implementation 'org.springframework:spring-context'
 
-    testImplementation project(':test-common')
     testImplementation project(':test-common-blackduck')
     testImplementation project(':provider-blackduck')
 }
--- a/api-processor/build.gradle
+++ b/api-processor/build.gradle
@@ -11,5 +11,6 @@ dependencies {
     implementation 'org.springframework:spring-context'
 
     testImplementation project(':test-common')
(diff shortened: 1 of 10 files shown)


test-common/build.gradle
```diff
@@ -3,6 +3,8 @@ dependencies {
 
     implementation platform(project(':alert-platform'))
     implementation project(':alert-common')
+    implementation project(':provider-blackduck')
```
ChomickiM, line 6: provider-blackduck is required for RuleViolationUniquePolicyNotificationContent

api-processor/build.gradle
```diff
@@ -11,5 +11,6 @@ dependencies {
     implementation 'org.springframework:spring-context'
 
     testImplementation project(':test-common')
```
gkillough, line 13: We can remove this since `test-common-blackduck` declares it as an `api` dependency.

test-common-blackduck/src/main/java/com/synopsys/integration/alert/test/common/processor/api/BlackDuckResponseTestUtility.java
```diff
@@ -0,0 +1,97 @@
+package com.synopsys.integration.alert.test.common.processor.api;
```
gkillough, line 1: This package should probably be `com.synopsys.integration.alert.test.common.blackduck` now.

PR #1822, opened 2021-09-14
Concrete Email Channel

rottebds:


src/main/java/com/synopsys/integration/alert/update/UpdateEmailService.java
```diff
@@ -100,16 +118,15 @@ private boolean wasEmailAlreadySentForVersion(String updateVersion) {
                    .isPresent();
     }
 
-    private void handleSendAndUpdateDatabase(EmailProperties emailProperties, Map<String, Object> templateFields, String emailAddress) throws AlertException {
+    private void handleSendAndUpdateDatabase(Properties javamailProperties, String smtpFrom, String smtpHost, int smtpPort, boolean smtpAuth, String smtpUsername, String smtpPassword, Map<String, Object> templateFields, String emailAddress) throws AlertException {
```
jamesrichard91, line 121: I know you didn't come up with this method name, but can we change it to be `handleSend`? because the database update does not occur in this method

rottebds: Can do!

PR #1828, opened 2021-09-20
Disitribution controller with basic audit info

bamandel:
Beginning of the new API for retrieving audit info


alert-database/src/main/java/com/synopsys/integration/alert/database/distribution/DistributionRepository.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.database.job.DistributionJobEntity;
+
+@Component
+public interface DistributionRepository extends JpaRepository<DistributionJobEntity, UUID> {
+
+    @Query(value = "SELECT new com.synopsys.integration.alert.database.distribution.DistributionWithAuditEntity(job.jobId, job.enabled, job.name, job.channelDescriptorName, job.distributionFrequency, audit.timeLastSent, audit.status) "
+                       + " FROM DistributionJobEntity job"
+                       + " LEFT JOIN AuditEntryEntity audit ON job.jobId = audit.commonConfigId"
```
gkillough, line 23: A distribution job can have multiple audit entries, so we should make sure we order by `time_last_sent` descending. We will also need to deduplicate these carefully to ensure we don't accidentally end up with the wrong statuses.

PR #1829, opened 2021-09-20
api-processor summarize test coverage

ChomickiM:
Added coverage for the ProjectMessageSummarizer under api-processor. The test is broken into several parts to ensure coverage and covering several other cases.

- We are testing the multiple message reasons to ensure the summary and description contain the correct relevant information. (Project Status, Project Version Status, Component Update, Component Concern) 
- For components we are also verifying that the correct information is used for the label in the LinkableItem
- Lastly we are verifying if there are multiple operations with the same ComponentConcernSummaryGrouping that they are correctly added to the value in the LinkableItem

To avoid matching specific strings I have been using the String.contains method to verify if key pieces of information are present in the Strings.

### DIFF
--- a/api-processor/src/test/java/com/synopsys/integration/alert/processor/api/summarize/ProjectMessageSummarizerTest.java
+++ b/api-processor/src/test/java/com/synopsys/integration/alert/processor/api/summarize/ProjectMessageSummarizerTest.java
@@ -41,6 +41,7 @@ public class ProjectMessageSummarizerTest {
 
         ProcessedProviderMessage<SimpleMessage> summarizedSimpleMessage = projectMessageSummarizer.summarize(processedProviderMessage);
         SimpleMessage simpleMessage = summarizedSimpleMessage.getProviderMessage();
+        printSimpleMessage(simpleMessage);
 
         testProjectStatus(simpleMessage);
         assertEquals(1, simpleMessage.getDetails().size());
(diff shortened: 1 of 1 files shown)


api-processor/src/test/java/com/synopsys/integration/alert/processor/api/summarize/ProjectMessageSummarizerTest.java
```diff
@@ -0,0 +1,323 @@
...
+        ComponentConcern vulnerabilityConcern3 = ComponentConcern.vulnerability(ItemOperation.DELETE, "CVE-246", ComponentConcernSeverity.MINOR_MEDIUM, "https://vuln-rul");
+
+        BomComponentDetails bomComponentDetails = createBomComponentDetails(List.of(policyConcern, vulnerabilityConcern1, vulnerabilityConcern2, vulnerabilityConcern3));
+
+        ProjectMessage projectMessage = ProjectMessage.componentConcern(providerDetails, commonProject, commonProjectVersion, List.of(bomComponentDetails));
+        ProcessedProviderMessage<ProjectMessage> processedProviderMessage = new ProcessedProviderMessage<>(Set.of(1L), projectMessage);
+
+        ProcessedProviderMessage<SimpleMessage> summarizedSimpleMessage = projectMessageSummarizer.summarize(processedProviderMessage);
+        SimpleMessage simpleMessage = summarizedSimpleMessage.getProviderMessage();
+        printSimpleMessage(simpleMessage);
```
psantos1113: You may just want to incorporate the printSimpleMessage call into the tests that actually run before the assertions. Then remove the disabled tests altogether.  The disabled tests only output the messages.  If the intent is for the printSimpleMessage to be a quick eyeball test, then I would just consider including the method call in the tests that run.  

Not a major issue here the tests overall look fine.

PR #1846, opened 2021-10-01
Disable audit resend from Details modal

ChomickiM:
If a user is missing audit "EXECUTE" permissions, we will disable the button.

### DIFF
--- a/ui/src/main/js/page/audit/Details.js
+++ b/ui/src/main/js/page/audit/Details.js
@@ -9,6 +9,7 @@ import TextArea from 'common/input/TextArea';
 import RefreshTableCellFormatter from 'common/table/RefreshTableCellFormatter';
 import * as DescriptorUtilities from 'common/util/descriptorUtilities';
 import StatusMessage from 'common/StatusMessage';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 class Details extends Component {
     constructor(props) {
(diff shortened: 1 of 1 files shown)


ui/src/main/js/page/audit/Details.js
```diff
@@ -82,14 +84,29 @@ class Details extends Component {
...
+        }
+        return (<div className="jobIconButtonDisabled"><FontAwesomeIcon icon="sync" className="alert-icon" size="lg" /></div>);
+    }
+
+    isResendAllowed() {
+        const { descriptors } = this.props;
+        if (descriptors) {
+            const descriptorList = DescriptorUtilities.findDescriptorByNameAndContext(descriptors, DescriptorUtilities.DESCRIPTOR_NAME.COMPONENT_AUDIT, DescriptorUtilities.CONTEXT_TYPE.GLOBAL);
+            if (descriptorList) {
+                return descriptorList.some((descriptor) => DescriptorUtilities.isOperationAssigned(descriptor, DescriptorUtilities.OPERATIONS.EXECUTE));
```
bamandel, line 105: I believe you can return
```suggestion
            return descriptorList && descriptorList.some((descriptor) => DescriptorUtilities.isOperationAssigned(descriptor, DescriptorUtilities.OPERATIONS.EXECUTE));
```
it'll verify the first item isn't null and then perform the check

bamandel (approved):

PR #1848, opened 2021-10-04
refactor: Change the helpers from being components.

psantos1113:


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/ConfigurationCRUDHelper.java
```diff
@@ -48,7 +48,7 @@ public ConfigurationCRUDHelper(AuthorizationManager authorizationManager) {
         return new ActionResponse<>(HttpStatus.OK, optionalResponse.get());
     }
 
-    public <T> ActionResponse<T> create(Supplier<ValidationResponseModel> validator, Callable<T> createdModelSupplier, ConfigContextEnum context, DescriptorKey descriptorKey) {
+    public <T> ActionResponse<T> create(Supplier<ValidationResponseModel> validator, Callable<T> createdModelSupplier) {
```
rottebds, line 51: I think using Callable here and on line 68 implies concurrency where there isn't any. If we want to have the createdModelSupplier able to throw exceptions, we should instead use our own ThrowingSupplier for clarity.

I also personally prefer modelCreator to createdModelSupplier, but that's splitting hairs =)

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/ConfigurationCRUDHelper.java
```diff
@@ -87,7 +87,7 @@ public ConfigurationCRUDHelper(AuthorizationManager authorizationManager) {
         }
     }
 
-    public <T> ActionResponse<T> delete(BooleanSupplier existingModelSupplier, Runnable deleteFunction, ConfigContextEnum context, DescriptorKey descriptorKey) {
+    public <T> ActionResponse<T> delete(BooleanSupplier existingModelSupplier, Runnable deleteFunction) {
```
rottebds, line 90: This is a little trickier since Java doesn't give us a functional interface equivalent to Runnable-- in Jenkins Common we have a ThrowingExecutor-- maybe we should have something similar in Alert Common? Executor may not be the best name since Executors are a thing in Java that run Runnables.

rottebds (approved):

PR #1852, opened 2021-10-05
Changing the email logic to not log in for every message.

jamesrichard91:
Log in once and re-use the transport for multiple messages

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/email/EmailMessagingService.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/email/EmailMessagingService.java
@@ -59,10 +59,10 @@ public class EmailMessagingService {
         try {
             String templateName = StringUtils.trimToEmpty(emailTarget.getTemplateName());
             Set<String> emailAddresses = emailTarget.getEmailAddresses()
-                                             .stream()
-                                             .map(String::trim)
-                                             .filter(StringUtils::isNotBlank)
-                                             .collect(Collectors.toSet());
+                .stream()
+                .map(String::trim)
+                .filter(StringUtils::isNotBlank)
+                .collect(Collectors.toSet());
             if (emailAddresses.isEmpty() || StringUtils.isBlank(templateName)) {
                 // Nothing to send
                 logger.debug("No non-blank email addresses were provided");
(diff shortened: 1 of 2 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/email/EmailMessagingService.java
```diff
@@ -168,22 +168,25 @@ private Session createMailSession(EmailProperties emailProperties) {
...
-                    Transport.send(message);
-                }
-            } catch (MessagingException e) {
-                if (recipients != null) {
-                    Stream.of(recipients).map(Address::toString).forEach(invalidRecipients::add);
+
+        try (Transport transport = getTransport(emailProperties, session)) {
+            for (Message message : messages) {
+                Address[] recipients = null;
+                try {
```
rottebds, line 175: Not a fan of this nested try. Could we instead extract a second public sendMessages method which takes a Transport and a List\<Message\>?

alert-common/src/main/java/com/synopsys/integration/alert/common/email/EmailMessagingService.java
```diff
@@ -197,19 +200,19 @@ public void sendMessages(EmailProperties emailProperties, Session session, List<
         }
     }
 
-    private void sendAuthenticated(EmailProperties emailProperties, Message message, Session session) throws MessagingException {
+    private Transport getTransport(EmailProperties emailProperties, Session session) throws MessagingException {
```
rottebds, line 203: In truth, there's little "get"-ing going on here. Maybe getAndConnectToTransport?

psantos1113 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/email/EmailMessagingService.java
```diff
@@ -167,49 +168,54 @@ private Session createMailSession(EmailProperties emailProperties) {
...
+            transport.sendMessage(message, recipients);
+        } catch (MessagingException e) {
+            Set<String> recipientAddresses = Collections.emptySet();
+            if (recipients != null) {
+                recipientAddresses = Stream.of(recipients).map(Address::toString).collect(Collectors.toSet());
+            }
+            String error = String.format("Could not send this email to the following recipients: %s. Reason: %s", recipientAddresses, e.getMessage());
+            errorMessages.add(error);
         }
+        return errorMessages;
```
rottebds, line 218: Is this Set ever going to have more than one element in it? Should we just return an Optional\<String\>?

rottebds (approved):

PR #1876, opened 2021-10-26
azure provider URL custom field

psantos1113:
If we remove urls from the azure custom properties we will need changes like this to be able find Jira tickets created by older versions of Alert.  

We will need to expand this to other custom fields as well which will make the queries very complex.


psantos1113:
Closing the PR because it is used as a reference for what we can do.  The corresponding problem that prompted this work hasn't been agreed upon for further work.  This PR can be used as a reference.

PR #1888, opened 2021-11-09
Static endpoint for alert encryption settings

ChomickiM:
This is the first PR to implement a SettingsEncryptionController, CRUD actions, and validation. This controller will be disabled until the UI component has been created.


component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
```diff
@@ -0,0 +1,60 @@
...
+import com.synopsys.integration.alert.common.rest.api.ValidateController;
+import com.synopsys.integration.alert.common.rest.model.ValidationResponseModel;
+import com.synopsys.integration.alert.component.settings.encryption.action.SettingsEncryptionCrudActions;
+import com.synopsys.integration.alert.component.settings.encryption.action.SettingsEncryptionValidationAction;
+import com.synopsys.integration.alert.component.settings.encryption.model.SettingsEncryptionModel;
+
+//TODO: Disabled until new UI component is complete
+//@RestController
+//@RequestMapping(AlertRestConstants.SETTINGS_ENCRYPTION_PATH)
+public class SettingsEncryptionController implements BaseResourceController<SettingsEncryptionModel>, ValidateController<SettingsEncryptionModel> {
```
bamandel, line 25: A bit of a heads up, depending on how the UI is implemented/will be implemented you might need a get page endpoint. If you use the getOne endpoint, where will you get the original ID from? Alternatively, we could come up with a new type of Controller that assumes it will only ever have a single row that it always needs to return/modify.

ChomickiM: In my discussions with Paulo and James we only ever plan to have a single encryption configuration. Within the SettingsEncryptionCrudActions we have a DEFAULT_ENCRYPTION_ID as a static ID. Our plan was for the UI to only ever call that same ID. Likewise, we have set validation within  the getSettingsEncryptionModel should the ID not match the static encryption ID.

ChomickiM: That said, I do think if we had a new type of controller that only ever expected a single value to return or modify, then it would simplify this somewhat and remove that ambiguity.

psantos1113 (approved):

PR #1892, opened 2021-11-12
Settings Encryption UI component

ChomickiM:
Added a new beta page for settings encryption configuration. The proxy UI will be completed in a separate PR.

### DIFF
--- a/component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
@@ -25,7 +25,6 @@ import com.synopsys.integration.alert.component.settings.encryption.action.Setti
 import com.synopsys.integration.alert.component.settings.encryption.action.SettingsEncryptionValidationAction;
 import com.synopsys.integration.alert.component.settings.encryption.model.SettingsEncryptionModel;
 
-//TODO: Disabled until new UI component is complete
 @RestController
 @RequestMapping(AlertRestConstants.SETTINGS_ENCRYPTION_PATH)
 public class SettingsEncryptionController implements ValidateController<SettingsEncryptionModel> {
--- a/component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
@@ -13,6 +13,7 @@ import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
@@ -50,7 +51,7 @@ public class SettingsEncryptionController implements ValidateController<Settings
 
     @PutMapping
     @ResponseStatus(HttpStatus.NO_CONTENT)
-    public void update(SettingsEncryptionModel resource) {
+    public void update(@RequestBody SettingsEncryptionModel resource) {
         ResponseFactory.createContentResponseFromAction(configActions.update(resource));
     }
 
--- a/component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
@@ -13,8 +13,11 @@ import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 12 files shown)


component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/web/SettingsEncryptionController.java
```diff
@@ -23,8 +26,8 @@
 import com.synopsys.integration.alert.component.settings.encryption.model.SettingsEncryptionModel;
 
 //TODO: Disabled until new UI component is complete
```
bamandel, line 28: Might want to remove the TODO

ui/src/main/js/page/settings/standalone/SettingsConfigurationStandalone.js
```diff
@@ -0,0 +1,48 @@
...
+import * as FieldModelUtilities from 'common/util/fieldModelUtilities';
+import { CONTEXT_TYPE } from 'common/util/descriptorUtilities';
+import CommonGlobalConfiguration from 'common/global/CommonGlobalConfiguration';
+import { SETTINGS_INFO } from 'page/settings/SettingsModel';
+import SettingsEncryptionConfiguration from 'page/settings/standalone/SettingsEncryptionConfiguration.js';
+
+const SettingsConfigurationStandalone = ({
+    csrfToken, errorHandler, readOnly, displaySave
+}) => {
+    const [formData, setFormData] = useState(FieldModelUtilities.createEmptyFieldModel([], CONTEXT_TYPE.GLOBAL, SETTINGS_INFO.key));
```
bamandel, line 13: I don't think you need this state, especially since it's using the old field model stuff. lastUpdated can be empty

ChomickiM: Paulo and I discussed this one and weren't entirely sure if we wanted it here or not. We think we may need to add it back in when we do the Proxy configuration later to get the time updated on the page. We left it as a reminder but I can remove it and find a solution to it in the Proxy UI config

ui/src/main/js/page/settings/standalone/SettingsEncryptionConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+
+const SettingsEncryptionConfiguration = ({
+    csrfToken, errorHandler, readonly, displaySave
+}) => {
+    const encryptionRequestUrl = `${ConfigurationRequestBuilder.ENCRYPTION_API_URL}`;
+
+    const [settingsEncryptionConfig, setSettingsEncryptionConfig] = useState({});
+    const [errors, setErrors] = useState(HttpErrorUtilities.createEmptyErrorObject());
+
+    const [passwordFromApiExists, setPasswordFromApiExists] = useState(false);
```
bamandel, line 18: I imagine you'd need state for each password field. This state is meant to track whether or not a specific field has a value according to the API and since you have 2 password fields, you'll need to track them both individually

ui/src/main/js/page/settings/standalone/SettingsEncryptionConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+            <PasswordInput
+                id={SETTINGS_FIELD_KEYS.encryptionGlobalSalt}
+                name="encryptionGlobalSalt"
+                label="Encryption Global Salt"
+                description="The salt used when encrypting sensitive fields. Must be at least 8 characters long."
+                required
+                readOnly={readonly}
+                onChange={fieldModelUtilities.handleTestChange(settingsEncryptionConfig, setSettingsEncryptionConfig)}
+                value={settingsEncryptionConfig.encryptionGlobalSalt || undefined}
+                isSet={passwordFromApiExists}
```
bamandel, line 72: You'd want to call your second password state boolean here which is based off the content returned from the fetch call

ui/src/main/js/page/settings/standalone/SettingsEncryptionConfiguration.js
```diff
@@ -0,0 +1,93 @@
...
+
+    const [passwordFromApiExists, setPasswordFromApiExists] = useState(false);
+
+    const fetchData = async () => {
+        const response = await ConfigurationRequestBuilder.createReadRequest(encryptionRequestUrl, csrfToken);
+        const data = await response.json();
+
+        if (data) {
+            setSettingsEncryptionConfig(data);
+            setPasswordFromApiExists(true);
```
bamandel, line 26: If data is found and if the password field has content, mark that the password exists and delete that field from the data. Will need to do this for each password field

ui/src/main/js/common/util/fieldModelUtilities.js
```diff
@@ -270,6 +270,7 @@ export const handleChange = (data, setData) => ({ target }) => {
     setData(newState);
 };
 
+// TODO: This name can be misleading, consider changing to handleDefaultChange to separate from old FieldModels
 export const handleTestChange = (testData, setTestData) => ({ target }) => {
```
psantos1113, line 274: maybe we should call this method createDefaultInputChangeHandler because it's a function used to create an event handler.

ui/src/main/js/page/settings/standalone/SettingsEncryptionConfiguration.js
psantos1113: Makes sense the email has only the auth password.  So there was only one field to use as an example.  Good to know for my own reference in the future.

ui/src/main/js/page/settings/standalone/SettingsEncryptionConfiguration.js
psantos1113: The password field data from the server is the masked data correct?  So we empty the value and mark it as exists so the placeholder text is rendered correct?

bamandel (approved):

PR #1895, opened 2021-11-16
Ps email static model

psantos1113:


channel-email/src/main/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalCrudActions.java
```diff
@@ -37,36 +35,30 @@ public EmailGlobalCrudActions(AuthorizationManager authorizationManager, EmailGl
         this.validator = validator;
     }
 
-    public ActionResponse<EmailGlobalConfigModel> getOne(Long id) {
-        return configurationHelper.getOne(
-            // Compiler wants this cast as it does not understand the types being used
-            (Supplier<Optional<EmailGlobalConfigModel>>) () -> configurationAccessor.getConfiguration(id).map(DatabaseModelWrapper::getModel)
-        );
```
bamandel, line 44: Interesting, it looks like removing the extra types in the middle helped resolve this casting issue. Were you able to still save values as expected when running Alert?

bamandel (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ConfigurationAccessor.java
```diff
@@ -7,26 +7,25 @@
...
 
 import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
 import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
-import com.synopsys.integration.alert.common.persistence.model.DatabaseModelWrapper;
 import com.synopsys.integration.alert.common.rest.model.AlertPagedModel;
 
 public interface ConfigurationAccessor<T extends AlertSerializableModel> {
-    Optional<DatabaseModelWrapper<T>> getConfiguration(Long id);
+    int FIRST_PAGE_NUMBER = 0;
+    int DEFAULT_PAGE_SIZE = 10;
```
bamandel, line 19: I believe these defaults are available in AlertPagedModel

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/distribution/EmailChannelMessageSender.java
```diff
@@ -94,21 +89,26 @@ public MessageResult sendMessages(EmailJobDetailsModel emailJobDetails, List<Ema
...
+        EmailGlobalConfigModel emailServerConfiguration = configurations.getModels().stream()
+            .findFirst()
             .orElseThrow(() -> new AlertConfigurationException("ERROR: Missing Email global config."));
 
+        //        FieldUtility globalConfiguration = configurationModelConfigurationAccessor.getConfigurationsByDescriptorKeyAndContext(emailChannelKey, ConfigContextEnum.GLOBAL)
+        //            .stream()
+        //            .findAny()
+        //            .map(ConfigurationModel::getCopyOfKeyToFieldMap)
+        //            .map(FieldUtility::new)
+        //            .orElseThrow(() -> new AlertConfigurationException("ERROR: Missing Email global config."));
```
bamandel, line 102: Should this comment remain?

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalCrudActions.java
psantos1113: Yes. AlertPagedModel uses generics as does DatabaseModelWrapper.  AlertPagedModel doesn't know the type inside of DatabaseModelWrapper because it is using generics as well.  Now because the type of the object inside the AlertPagedModel doesn't use generics the correct type can be inferred.  So you no longer need the cast to avoid the compiler issue with generics.

psantos1113: It works fine to save the information.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/ConfigurationAccessor.java
psantos1113: Thanks.  I found something I want to switch up.  AlertPagedModel shouldn't depend on the read controller value.  The read controller should depend on the AlertPagedModel value.  I will update this.

ui/src/main/js/application/MainPage.js
```diff
@@ -144,28 +143,14 @@ const MainPage = ({
...
-                        </BetaComponent>
-                    </BetaPage>
+                    <EmailGlobalConfiguration
+                        csrfToken={csrfToken}
+                        errorHandler={errorHandler}
+                        readonly={readOnly}
+                        displayTest={showTest}
+                        displaySave={showSave}
+                        displayDelete={showDelete}
+                    />
```
bamandel, line 153: Maybe we should put the new page as current and the old one under beta, just in case we run into issues with any of the functionality?

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/distribution/EmailChannelMessageSender.java
psantos1113: to be removed

ui/src/main/js/application/MainPage.js
psantos1113: I will add the old configuration as EmailGlobalConfigurationOld and set it up as a beta page

PR #1907, opened 2021-11-29
activemq fixes

psantos1113:
* Change the broker url to configure splitting resources between producers and consumers.
* Set the prefetch limit to 100 from 1000


src/main/java/com/synopsys/integration/alert/configuration/ActiveMQConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+        // setup consumer queue policy so that 100 events need to be acknowledged before send more events to the consumer.
+        prefetchPolicy.setQueuePrefetch(100);
+        factory.setPrefetchPolicy(prefetchPolicy);
+        String newBrokerUrl = createCustomBrokerUrl(factory);
+        factory.setBrokerURL(newBrokerUrl);
+    }
+
+    private String createCustomBrokerUrl(ActiveMQConnectionFactory factory) {
+        String brokerUrl = factory.getBrokerURL();
+        StringBuilder urlBuilder = new StringBuilder(brokerUrl.length() + BROKER_SPLIT_MEMORY_QUERY_PARAM.length() + 1);
```
jamesrichard91, line 25: This seems a little strange, can we not set "splitSystemUsageForProducersConsumers" on the BrokerService? We have to do this odd URL modification?

jamesrichard91 (approved):

src/main/java/com/synopsys/integration/alert/configuration/ActiveMQConfiguration.java
psantos1113: When you get the broker service instance and use the set method to split the memory the broker services isn't restarted so the change doesn't have any effect. The memory limits have been calculated to 1GB.  If you manually restart the broker service JMS throws errors.  The ActiveMQConnectionFactory is used to create a connection which creates the broker service.

 Reading the ActiveMQ documentation you can modify broker properties by setting them on the broker URL.  By setting it on the broker URL this ensures the BrokerService is initialized with the property and JMS doesn't throw any exceptions due to restarting.  Because this is an embedded broker, the `broker.persistent=false` query parameter is already present in the broker URL.   So that's why we have this strange code.

PR #1910, opened 2021-11-30
Static endpoint for proxy configuration

ChomickiM:
The following PR is the backend changes for implementing the Proxy configuration. In addition, I have added several tests for CRUD actions, validation, and the Controller integration test.

As a note the following are not yet implemented but will be in followup PRs. This PR is primarily focused on setting up the new controller, its dependencies, the repository changes, and liquibase migration.

- Proxy UI
- Serialization problem discovered also affecting Encryption and Email models.
- Rewriting the ProxyManager class and any place it is implemented to use the new accessor.


alert-database/src/main/java/com/synopsys/integration/alert/database/settings/proxy/NonProxyHostConfigurationPK.java
```diff
@@ -0,0 +1,33 @@
+package com.synopsys.integration.alert.database.settings.proxy;
+
+import java.io.Serializable;
+import java.util.UUID;
+
+public class NonProxyHostConfigurationPK implements Serializable {
```
psantos1113, line 6: Add a serial version ID since this is serializable

component/src/test/java/com/synopsys/integration/alert/component/settings/proxy/actions/SettingsProxyCrudActionsTest.java
```diff
@@ -0,0 +1,212 @@
...
+    private final SettingsProxyValidator settingsProxyValidator = new SettingsProxyValidator();
+    private final SettingsDescriptorKey settingsDescriptorKey = new SettingsDescriptorKey();
+
+    @Test
+    public void getOneTest() {
+        UUID uuid = UUID.randomUUID();
+        SettingsProxyConfigurationRepository settingsProxyConfigurationRepository = Mockito.mock(SettingsProxyConfigurationRepository.class);
+        NonProxyHostsConfigurationRepository nonProxyHostsConfigurationRepository = Mockito.mock(NonProxyHostsConfigurationRepository.class);
+
+        Mockito.when(settingsProxyConfigurationRepository.findById(Mockito.any())).thenReturn(Optional.of(createSettingsProxyConfigurationEntity(uuid)));
```
psantos1113, line 69: findById(Mockito.any()) you could change it to Mockito.eq(uuid) so that it matches exactly.  This forces Mockito to only create the entity for the matching UUID which you are testing.  I've been getting into the habit of using Mockito.eq to perform an exact match on the arguments rather than any().

alert-database/src/main/resources/liquibase/6.9.0/migrate-proxy-config.xml
```diff
@@ -0,0 +1,67 @@
...
+            SELECT now(),
+                   null,
+                   'initial-proxy-config',
+                   null,
+                   null,
+                   null
+            FROM GET_GLOBAL_CONFIG_SOURCE_KEY_AND_FIELD_VALUE('component_settings') configFields
+            WHERE configFields.source_key IS NOT NULL LIMIT 1;
+        </sql>
+    </changeSet>
```
jamesrichard91, line 20: If there was no proxy configuration previously, do we have any change set to delete this entry if it is all null values? Or do we need this dummy entry for some reason?

bamandel: Maybe each of these should have a precondition to check if a proxy setting exists originally? If proxy fields were shared with other fields in its descriptor then this may be more challenging

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/database/accessor/SettingsProxyConfigAccessor.java
```diff
@@ -0,0 +1,154 @@
...
+    public void deleteConfiguration(UUID configurationId) {
+        if (null != configurationId) {
+            settingsProxyConfigurationRepository.deleteById(configurationId);
+        }
+    }
+
+    private SettingsProxyModel createConfigModel(SettingsProxyConfigurationEntity proxyConfiguration) {
+        SettingsProxyModel newModel = new SettingsProxyModel();
+        String createdAtFormatted = "";
+        String lastUpdatedFormatted = "";
```
bamandel, line 108: These may not be necessary. You should be able to move the set call into the if statement and a null value should set a default once inserted into the DB

alert-database/src/main/resources/liquibase/6.9.0/migrate-proxy-config.xml
psantos1113: I believe the GET_GLOBAL_CONFIG_SOURCE_KEY_AND_FIELD_VALUE will return an empty table if the proxy config doesn't exist.  In that case no records will be inserted into the table.  Only if an old proxy configuration exists will this query insert a record into the table.

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/database/accessor/SettingsProxyConfigAccessor.java
```diff
@@ -0,0 +1,154 @@
...
+        String host = configuration.getProxyHost().orElseThrow(null);
+        Integer port = configuration.getProxyPort().orElse(null);
+        String username = configuration.getProxyUsername().orElse(null);
+        String password = configuration.getProxyPassword().map(encryptionUtility::encrypt).orElse(null);
+
+        return new SettingsProxyConfigurationEntity(configurationId, createdTime, lastUpdated, host, port, username, password, List.of());
+    }
+
+    private List<NonProxyHostConfigurationEntity> toNonProxyHostEntityList(UUID configurationId, SettingsProxyModel configuration) {
+        List<String> hostnamePatterns = configuration.getNonProxyHosts().orElse(List.of());
```
bamandel, line 146: This initial call can be converted into stream mapping that does everything below.

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/validator/SettingsProxyValidator.java
```diff
@@ -0,0 +1,56 @@
...
+    public static final String PROXY_HOST_FIELD_NAME = "proxyHost";
+    public static final String PROXY_PORT_FIELD_NAME = "proxyPort";
+    public static final String PROXY_USERNAME_FIELD_NAME = "proxyUsername";
+    public static final String PROXY_PASSWORD_FIELD_NAME = "proxyPassword";
+
+    public ValidationResponseModel validate(SettingsProxyModel model) {
+        Set<AlertFieldStatus> statuses = new HashSet<>();
+        if (model.getProxyHost().isPresent() && model.getProxyPort().isEmpty()) {
+            validateRequiredFieldIsNotBlank(statuses, model.getProxyPort().isPresent(), PROXY_PORT_FIELD_NAME);
+        }
```
bamandel, line 24: We may want to introduce a new class to help with common validation similar to what ConfigurationFieldValidator does today otherwise we'd end up with the same private method in many places. The new class could then be used in EmailGlobalConfigurationValidator.

bamandel: The new class could also return the content in the form of a validation response so you don't need to do the statuses.empty check again.

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/web/SettingsProxyController.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public SettingsProxyModel getOne(UUID id) {
+        return ResponseFactory.createContentResponseFromAction(configActions.getOne(id));
+    }
+
+    @Override
+    public AlertPagedModel<SettingsProxyModel> getPage(Integer pageNumber, Integer pageSize, String searchTerm) {
+        return ResponseFactory.createContentResponseFromAction(configActions.getPaged(pageNumber, pageSize));
+    }
```
bamandel, line 45: Do we plan on supporting pages of proxy info? If not, we should do the getOne without an ID like we did for encryption

ChomickiM: When we were creating the controller we discussed future Alert changes where we could have multiple proxies configured and the user could pick and choose which one to apply. We decided that we could implement and support pages via the API now but have it unused for now within the UI.

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/validator/SettingsProxyValidator.java
ChomickiM: While this doesn't seem like a bad idea, I think it might be out of the scope of this PR. I can create a ticket to look at the existing implementations for Email, Encryption, and Proxy to determine what would be the best way to write a common validator. Right now each validator is somewhat different and it may take additional time to investigate, come up with a solution and write new tests around all of them.

psantos1113 (approved):

alert-database/src/main/resources/liquibase/6.9.0/migrate-proxy-config.xml
```diff
@@ -0,0 +1,75 @@
...
+
+    <changeSet author="martinch" id="update-initial-proxy-config-entries">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                SELECT COUNT(host) FROM alert.configuration_proxy WHERE host = 'initial-proxy-config';
```
jamesrichard91, line 25: @psantos1113 @ChomickiM  why is this precondition different than the one below for the non-proxy hosts?

PR #1912, opened 2021-12-01
fix: Have event manager wait.

psantos1113:


alert-common/src/main/java/com/synopsys/integration/alert/common/event/BrokerServiceDependentTask.java
```diff
@@ -22,17 +22,18 @@
...
+    private BrokerServiceWaitTask brokerServiceWaitTask;
 
-    public BrokerServiceDependentTask(String taskToExecuteName, Consumer<BrokerService> taskToExecuteOnceFound) {
+    public BrokerServiceDependentTask(String taskToExecuteName, BrokerServiceWaitTask brokerServiceWaitTask, Consumer<BrokerService> taskToExecuteOnceFound) {
         this.taskToExecuteName = taskToExecuteName;
         this.taskToExecuteOnceFound = taskToExecuteOnceFound;
+        this.brokerServiceWaitTask = brokerServiceWaitTask;
     }
 
     public String getTaskToExecuteName() {
```
psantos1113, line 39: This should have a dependency on the ExecutorService so that it executes in the same thread if needed for tests.

PR #1925, opened 2021-12-07
Proxy UI implementation

ChomickiM:
Added the new  proxy UI tab to the settings page.

In addition, I've added the"name" field to  DynamicSelectInput.js. Previously "name" was used throughout the Alert UI without actually being implemented. In these cases, "id" and "name" would always match. I've made updates to make this constant for all other places in the UI not using this, and implemented a new id/name for the proxy configuration.

### DIFF
--- a/ui/src/main/js/page/settings/standalone/SettingsConfigurationStandalone.js
+++ b/ui/src/main/js/page/settings/standalone/SettingsConfigurationStandalone.js
@@ -4,9 +4,10 @@ import { Tab, Tabs } from 'react-bootstrap';
 import CommonGlobalConfiguration from 'common/global/CommonGlobalConfiguration';
 import { SETTINGS_INFO } from 'page/settings/SettingsModel';
 import SettingsEncryptionConfiguration from 'page/settings/standalone/SettingsEncryptionConfiguration.js';
+import SettingsProxyConfiguration from './SettingsProxyConfiguration';
 
 const SettingsConfigurationStandalone = ({
-    csrfToken, errorHandler, readOnly, displaySave
+    csrfToken, errorHandler, readOnly, displaySave, displayDelete
 }) => (
     <CommonGlobalConfiguration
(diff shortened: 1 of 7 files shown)


ui/src/main/js/page/settings/standalone/SettingsConfigurationStandalone.js
```diff
@@ -4,9 +4,10 @@ import { Tab, Tabs } from 'react-bootstrap';
 import CommonGlobalConfiguration from 'common/global/CommonGlobalConfiguration';
 import { SETTINGS_INFO } from 'page/settings/SettingsModel';
 import SettingsEncryptionConfiguration from 'page/settings/standalone/SettingsEncryptionConfiguration.js';
+import SettingsProxyConfiguration from './SettingsProxyConfiguration';
```
bamandel, line 7: Will want to change this to an absolute path

ui/src/main/js/page/settings/standalone/SettingsProxyConfiguration.js
```diff
@@ -0,0 +1,146 @@
...
+                id={SETTINGS_FIELD_KEYS.proxyPassword}
+                name={passwordName}
+                label="Proxy Password"
+                description="If the proxy server requires authentication, the password to authenticate with the proxy server."
+                readOnly={readOnly}
+                onChange={fieldModelUtilities.handleTestChange(settingsProxyConfig, setSettingsProxyConfig)}
+                value={settingsProxyConfig[passwordName] || undefined}
+                isSet={passwordFromApiExists}
+                errorName={passwordName}
+                errorValue={errors.fieldErrors.proxyPassword}
```
bamandel, line 111: The error key will generally match the actual field name. So you should get away with errors.fieldErrors[passwordName]

bamandel (approved):

psantos1113 (approved):

PR #1934, opened 2021-12-09
Ps save to concrete model

psantos1113:
This sets up the code to save the concrete configurations from GLOBAL field model configurations.  The implementations of Email and Proxy will be done in separate PRs.

* Create the interfaces to convert field models into concrete objects. 
* Create an interface to convert the field model to a global configuration
* Create a method in the ConfigurationAccessor to find the configuration by name
* Update existing code to use configuration by name rather than the default page
* Update tests


ChomickiM (approved):

web/src/main/java/com/synopsys/integration/alert/web/api/config/GlobalFieldModelToConcreteConversionService.java
```diff
@@ -0,0 +1,67 @@
...
+        Optional<DescriptorKey> descriptor = descriptorMap.getDescriptorKey(fieldModel.getDescriptorName());
+        if (descriptor.isEmpty() || !isGlobalConfig(fieldModel)) {
+            // not a field model that can be converted.
+            return;
+        }
+        GlobalFieldModelToConcreteSaveActions conversionAction = conversionActionMap.get(descriptor.get());
+        if (null != conversionAction) {
+            conversionAction.updateConcreteModel(fieldModel);
+        }
+    }
```
bamandel, line 61: This can be pulled into a private method that's passed some function which performs update or create.

bamandel (approved):

web/src/main/java/com/synopsys/integration/alert/web/api/config/GlobalFieldModelToConcreteConversionService.java
psantos1113: yes I will make that change going forward.

PR #1940, opened 2021-12-14
Proxy controller refactor

ChomickiM:
Update proxy configuration to use only a single configuration and remove the id field. In addition, we've created a "Unique" controller and accessor interfaces to better deal with configurations that do not have an ID.

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
@@ -15,13 +15,11 @@ import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurat
 public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
     String DEFAULT_CONFIGURATION_NAME = "default-configuration";
 
-    Optional<T> getConfiguration();
-
-    Optional<T> getConfigurationByName(String configurationName);
+    Optional<T> getConfigurationByName();
 
     T createConfiguration(T configuration) throws AlertConfigurationException;
 
     T updateConfiguration(T configuration) throws AlertConfigurationException;
 
-    void deleteConfiguration(String configurationName);
+    void deleteConfiguration();
 }
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
@@ -1,0 +1,26 @@
+/*
+ * alert-common
+ *
+ * Copyright (c) 2021 Synopsys, Inc.
+ *
+ * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All rights reserved worldwide.
+ */
+package com.synopsys.integration.alert.common.persistence.accessor;
+
+import java.util.Optional;
+
+import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
+import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
+
+public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
+    String DEFAULT_CONFIGURATION_NAME = "default-configuration";
+
+    Optional<T> getConfiguration();
+
+    T createConfiguration(T configuration) throws AlertConfigurationException;
+
+    T updateConfiguration(T configuration) throws AlertConfigurationException;
+
+    void deleteConfiguration();
+
+}
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
@@ -23,6 +23,5 @@ public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
 
     T updateConfiguration(T configuration) throws AlertConfigurationException;
 
-    void deleteConfiguration();
-
+    void deleteConfiguration(String configurationName);
 }
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
@@ -15,7 +15,7 @@ import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurat
 public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
     String DEFAULT_CONFIGURATION_NAME = "default-configuration";
 
-    Optional<T> getConfigurationByName();
+    Optional<T> getConfiguration();
 
     T createConfiguration(T configuration) throws AlertConfigurationException;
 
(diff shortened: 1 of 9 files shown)


alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Optional;
+
+import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
+import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
+
+public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
+    String DEFAULT_CONFIGURATION_NAME = "default-configuration";
+
+    Optional<T> getConfiguration();
```
jamesrichard91, line 18: Is this method used at all?

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/database/accessor/SettingsProxyConfigAccessor.java
```diff
@@ -49,14 +51,12 @@ public SettingsProxyConfigAccessor(
...
-        return settingsProxyConfigurationRepository.count();
-    }
-
-    @Override
-    @Transactional(readOnly = true)
-    public Optional<SettingsProxyModel> getConfiguration(UUID id) {
-        return settingsProxyConfigurationRepository.findById(id).map(this::createConfigModel);
+    public Optional<SettingsProxyModel> getConfiguration() {
+        Page<SettingsProxyConfigurationEntity> proxyConfigPage = settingsProxyConfigurationRepository.findAll(pageRequest);
+        return proxyConfigPage.getContent()
```
jamesrichard91, line 56: Do you want to log a warning if there are multiple proxy configs found? People can modify the database directly or we have run into issues with migration in the past that lead to multiple configs when there should only be one.

jamesrichard91: It would make debugging a lot easier

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/database/accessor/SettingsProxyConfigAccessor.java
```diff
@@ -108,10 +105,11 @@ public SettingsProxyModel updateConfiguration(UUID configurationId, SettingsProx
...
     @Transactional(propagation = Propagation.REQUIRED)
-    public void deleteConfiguration(UUID configurationId) {
-        if (null != configurationId) {
-            settingsProxyConfigurationRepository.deleteById(configurationId);
-        }
+    public void deleteConfiguration(String configurationName) {
+        settingsProxyConfigurationRepository.findByName(configurationName)
+            .stream()
+            .findFirst()
+            .ifPresent(settingsProxyConfigurationEntity -> settingsProxyConfigurationRepository.deleteById(settingsProxyConfigurationEntity.getConfigurationId()));
```
jamesrichard91, line 112: Couldn't you just create a method in `settingsProxyConfigurationRepository ` to delete by name?

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
```diff
@@ -0,0 +1,27 @@
...
+
+    Optional<T> getConfigurationByName(String configurationName);
```
psantos1113, line 20: You don't need this method.  If this is a uniqueAccessor then you don't need to pass the configuration name.  You just need getConfiguration.  The implementer can hide that it looks up 'default-configuration'

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
ChomickiM: Not anymore, I thought I could keep it in if we wanted to use it in the future but removing it probably makes more sense.

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
```diff
@@ -0,0 +1,27 @@
...
+
+    T createConfiguration(T configuration) throws AlertConfigurationException;
+
+    T updateConfiguration(T configuration) throws AlertConfigurationException;
+
+    void deleteConfiguration(String configurationName);
```
psantos1113, line 26: You don't need to pass the configuration name since this represents a single resource that isn't going to change.

jamesrichard91: I would recommend changing this to take in an ID not a name.

bamandel: The idea with this constructor is that only 1 entity can be stored and thus we always interact with that 1 entity.

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/database/accessor/SettingsProxyConfigAccessor.java
```diff
@@ -93,9 +87,12 @@ public SettingsProxyModel createConfiguration(SettingsProxyModel configuration)
 
     @Override
     @Transactional(propagation = Propagation.REQUIRED)
-    public SettingsProxyModel updateConfiguration(UUID configurationId, SettingsProxyModel configuration) throws AlertConfigurationException {
-        SettingsProxyConfigurationEntity configurationEntity = settingsProxyConfigurationRepository.findById(configurationId)
-                                                                   .orElseThrow(() -> new AlertConfigurationException(String.format("Config with id '%s' did not exist", configurationId.toString())));
+    public SettingsProxyModel updateConfiguration(SettingsProxyModel configuration) throws AlertConfigurationException {
+        SettingsProxyConfigurationEntity configurationEntity = settingsProxyConfigurationRepository.findByName(configuration.getName())
```
bamandel, line 91: This may be confusing to a new user. If they try updating the hidden name using our API it would throw this exception. We may want to stick to the default of using the default name

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.Optional;
+
+import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
+import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
+
+public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
+    String DEFAULT_CONFIGURATION_NAME = "default-configuration";
+
+    Optional<T> getConfigurationByName();
```
bamandel, line 18: I think this would be better without the byName suffix.

bamandel (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/accessor/UniqueConfigurationAccessor.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package com.synopsys.integration.alert.common.persistence.accessor;
+
+import java.util.Optional;
+
+import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
+import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
+
+public interface UniqueConfigurationAccessor<T extends AlertSerializableModel> {
+    String DEFAULT_CONFIGURATION_NAME = "default-configuration";
```
psantos1113, line 16: Now that we have this defined in ConfigurationAccessor and here we should move this into AlertConstants and remove the constant from here and ConfigurationAccessor.

psantos1113 (approved):

PR #1942, opened 2021-12-16
email field model convert

psantos1113:
With the old rest APIs convert field models into their static configuration.


ChomickiM (approved):

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/convert/EmailGlobalFieldModelConverter.java
```diff
@@ -0,0 +1,87 @@
...
+@Component
+public class EmailGlobalFieldModelConverter implements GlobalFieldModelToConcreteConverter<EmailGlobalConfigModel> {
+    private Logger logger = LoggerFactory.getLogger(getClass());
+    private static final String EMAIL_FIELD_PREFIX = "mail.smtp.";
+    public static final String AUTH_REQUIRED_KEY = EMAIL_FIELD_PREFIX + "auth";
+    public static final String EMAIL_FROM_KEY = EMAIL_FIELD_PREFIX + "from";
+    public static final String EMAIL_HOST_KEY = EMAIL_FIELD_PREFIX + "host";
+    public static final String AUTH_PASSWORD_KEY = EMAIL_FIELD_PREFIX + "password";
+    public static final String EMAIL_PORT_KEY = EMAIL_FIELD_PREFIX + "port";
+    public static final String AUTH_USER_KEY = EMAIL_FIELD_PREFIX + "user";
```
bamandel, line 35: Shouldn't these keys be available in EmailPropertyKeys? Is that not reachable from this class?

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/convert/EmailGlobalFieldModelConverter.java
```diff
@@ -0,0 +1,87 @@
...
+            globalFieldModel.getFieldValue(EMAIL_PORT_KEY)
+                .map(Integer::valueOf)
+                .ifPresent(model::setSmtpPort);
+            globalFieldModel.getFieldValue(AUTH_REQUIRED_KEY)
+                .map(Boolean::valueOf)
+                .ifPresent(model::setSmtpAuth);
+            globalFieldModel.getFieldValue(AUTH_USER_KEY)
+                .ifPresent(model::setSmtpUsername);
+            globalFieldModel.getFieldValueModel(AUTH_PASSWORD_KEY)
+                .filter(Predicate.not(FieldValueModel::getIsSet))
```
bamandel, line 69: Do we want to check isSet here? If the user passes a value in the password section, I feel that we should set that value regardless of whether or not the isSet field is true or false.

bamandel: I guess we do need to check if it's set in the case that it's empty we need to insert the original value, right?

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/convert/EmailGlobalFieldModelConverter.java
psantos1113: yes we should be able to use this instead.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/convert/EmailGlobalFieldModelConverter.java
psantos1113: I'm going to look into this.

psantos1113:
Found an issue while performing additional testing closing this PR.

PR #1944, opened 2021-12-16
Update settings configurations to use new isSet methods for password fields

ChomickiM:


component/src/main/java/com/synopsys/integration/alert/component/settings/encryption/action/SettingsEncryptionCrudActions.java
```diff
@@ -89,6 +89,8 @@ private SettingsEncryptionModel updateSettingsEncryptionModel(SettingsEncryption
     private SettingsEncryptionModel createMaskedSettingsEncryptionModel() {
         // EncryptionUtility does not return a model. A SettingsEncryptionModel with values must be created in order to obfuscate in the ConfigurationCrudHelper later.
         SettingsEncryptionModel settingsEncryptionModel = new SettingsEncryptionModel();
+        settingsEncryptionModel.setIsEncryptionPasswordSet(true);
+        settingsEncryptionModel.setIsEncryptionGlobalSaltSet(true);
         settingsEncryptionModel.setEncryptionPassword(AlertRestConstants.MASKED_VALUE);
         settingsEncryptionModel.setEncryptionGlobalSalt(AlertRestConstants.MASKED_VALUE);
```
ChomickiM, line 95: As a note, we still need the MASKED_VALUE here since we don't get a model  back from the EncryptionUtility. When obfuscate is called later this will set the isSet fields and then return a null value for the fields themselves.

bamandel: We might want to add this as a comment to the code since it's a unique case. Maybe instead of the masked value, you can set it to a different constant that makes it clear it will be replaced later.

ChomickiM: I think the comment on line 90 would cover it but I can include a direct mention to those values getting  replaced as well as creating a new constant

bamandel (approved):

psantos1113 (approved):

PR #1948, opened 2021-12-17
Proxy Manager refactor

ChomickiM:
Draft to discuss changes specific to the ProxyManager


bamandel (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/SettingsUtility.java
```diff
@@ -9,23 +9,12 @@
...
 
-import com.synopsys.integration.alert.api.common.model.exception.AlertException;
-import com.synopsys.integration.alert.common.persistence.model.ConfigurationModel;
-import com.synopsys.integration.alert.common.rest.model.FieldModel;
+import com.synopsys.integration.alert.common.rest.model.SettingsProxyModel;
 import com.synopsys.integration.alert.descriptor.api.model.DescriptorKey;
 
 public interface SettingsUtility {
-
+    //TODO: This is used in AlertStartupInitializer::initializeConfigs and is set to be Deprecated
```
psantos1113, line 16: Let's just use the DescriptorMap that's in AlerrtStartupInitializer to lookup the key where the getKey method of this interface is used.

`descriptorMap.getDescriptorKey(SettingsDescriptoKey.SETTINGS_COMPONENT);`

Then we can completely delete this interface.

psantos1113: We need this interface because it's in Alert common and it's used as a marker interface correct?

psantos1113 (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/descriptor/accessor/SettingsUtility.java
ChomickiM: SettingsDescriptorKey is unavailible in the AlertStartupInitializer so it is unfortunately still required here. Likewise, since we need to implement the getConfiguration method we cannot make this a marker interface.

PR #1967, opened 2022-01-10
Bm filter by version

bamandel:
Adds a configuration option to allow users to filter by project version name
Filters on version after filtering by project name
Adds a warning to the test action in the case a matching version is not found

### DIFF
--- a/alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
+++ b/alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
@@ -51,7 +51,7 @@ public interface DistributionJobRepository extends JpaRepository<DistributionJob
                        + "    AND blackDuckDetails.globalConfigId = :blackDuckConfigId"
                        + "    AND notificationTypes.notificationType IN (:notificationTypeSet)"
                        + "    AND jobEntity.distributionFrequency IN (:frequencies)"
-                       + "    AND (blackDuckDetails.filterByProject = false OR blackDuckDetails.projectNamePattern IS NOT NULL OR projects.projectName IN (:projectNames))"
+                       + "    AND (blackDuckDetails.filterByProject = false OR blackDuckDetails.projectNamePattern IS NOT NULL OR blackDuckDetails.projectVersionNamePattern IS NOT NULL OR projects.projectName IN (:projectNames))"
                        + "    AND ("
                        + "          ("
                        + "            coalesce(:vulnerabilitySeverities, NULL) IS NULL"
--- a/api-processor/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
+++ b/api-processor/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
@@ -11,6 +11,7 @@ import java.util.List;
 import java.util.regex.Pattern;
 
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 
 import com.synopsys.integration.alert.common.persistence.model.job.BlackDuckProjectDetailsModel;
 import com.synopsys.integration.alert.common.persistence.model.job.FilteredDistributionJobResponseModel;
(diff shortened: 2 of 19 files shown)


api-processor/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
```diff
@@ -48,21 +50,26 @@ public static boolean doesNotificationTypeMatch(FilteredDistributionJobResponseM
...
+        boolean hasMatchingProjects = filteredDistributionJobResponseModel.getProjectDetails()
+            .stream()
+            .map(BlackDuckProjectDetailsModel::getName)
+            .distinct()
+            .anyMatch(projectName::equals);
+
+        String projectVersionNamePattern = filteredDistributionJobResponseModel.getProjectVersionNamePattern();
+        if (StringUtils.isNotBlank(projectVersionNamePattern)) {
+            boolean listIsEmptyOrMatches = filteredDistributionJobResponseModel.getProjectDetails().isEmpty() || hasMatchingProjects;
+            return Pattern.matches(projectVersionNamePattern, projectVersionName) && (listIsEmptyOrMatches || matchingProjectNamePattern);
```
bamandel, line 69: Main changes

alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
```diff
@@ -51,7 +51,7 @@
                        + "    AND blackDuckDetails.globalConfigId = :blackDuckConfigId"
                        + "    AND notificationTypes.notificationType IN (:notificationTypeSet)"
                        + "    AND jobEntity.distributionFrequency IN (:frequencies)"
-                       + "    AND (blackDuckDetails.filterByProject = false OR blackDuckDetails.projectNamePattern IS NOT NULL OR projects.projectName IN (:projectNames))"
+                       + "    AND (blackDuckDetails.filterByProject = false OR blackDuckDetails.projectNamePattern IS NOT NULL OR blackDuckDetails.projectVersionNamePattern IS NOT NULL OR projects.projectName IN (:projectNames))"
```
jamesrichard91, line 54: I don't think we want `blackDuckDetails.projectVersionNamePattern IS NOT NULL` as part of the query. Since we are not comparing the version in the query, it does not matter if it is set or not at this point.

api-processor/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
```diff
@@ -48,21 +50,26 @@ public static boolean doesNotificationTypeMatch(FilteredDistributionJobResponseM
...
+        boolean matchingProjectNamePattern = (StringUtils.isNotBlank(projectNamePattern)) ? Pattern.matches(projectNamePattern, projectName) : false;
+        boolean hasMatchingProjects = filteredDistributionJobResponseModel.getProjectDetails()
+            .stream()
+            .map(BlackDuckProjectDetailsModel::getName)
+            .distinct()
+            .anyMatch(projectName::equals);
+
+        String projectVersionNamePattern = filteredDistributionJobResponseModel.getProjectVersionNamePattern();
+        if (StringUtils.isNotBlank(projectVersionNamePattern)) {
+            boolean listIsEmptyOrMatches = filteredDistributionJobResponseModel.getProjectDetails().isEmpty() || hasMatchingProjects;
```
jamesrichard91, line 68: Shouldn't `hasMatchingProjects ` be false if the list is empty?

bamandel: oh true, this could have slipped in from previous iterations

alert-database/src/main/java/com/synopsys/integration/alert/database/job/DistributionJobRepository.java
bamandel: If only the project version name pattern is set (no project selected and no project name pattern), this query will not properly read jobs from the DB. There is also a test to verify this in JobNotificationMapperTestIT.

psantos1113: If we can search by projectVersionNamePattern only then we need to treat it like projectNamePattern and filterByProject is false.  We need to include all the projects and then filter using the project version name pattern in java code just like the projectNamePattern.  So I think that is needed.

psantos1113: We're just trying to find the jobs that might match the notification so projectVersionNamePattern not null might match the notification so we want to include the job in the results.

api-processor/src/main/java/com/synopsys/integration/alert/processor/api/filter/JobNotificationFilterUtils.java
bamandel: On further inspection, this case is important and changing this will fail a check in JobNotificationFilterUtilsTest. This check is here in the case that no selected projects match the notification project or no projects were ever selected. We then use this result to ensure the notification version matches as well.

jamesrichard91: " in the case that no selected projects match the notification project or no projects were ever selected"
I'm still confused, is that not covered by `hasMatchingProjects `.

bamandel: I've updated the boolean with a comment. Because of our use cases, we care if the selected projects are empty or if they have a valid project. (true or true) If either of those are true, we then check the version pattern. If the selected projects are not empty and they do not have a valid project (false or false) then we have failed as we have to match an existing project if there is a list. Also, we could be empty and since we're empty, no valid project found (true or false) meaning the version pattern could still be true and we check.

PR #1968, opened 2022-01-11
Use default Email SMTP password when testing saved configuration

ChomickiM:
Fixes issue when attempting to test a configuration after saving. Note: Right now this will use the default configuration for the global email channel which will resolve the issue in the UI. I have left a TODO comment to find a better solution once we support multiple configurations in the UI. 

Perhaps as a future solution we could determine which configuration is "in view" and use the ID to get the correct configuration from the database.

### DIFF
--- a/channel-email/src/test/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalTestActionTest.java
+++ b/channel-email/src/test/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalTestActionTest.java
@@ -3,11 +3,14 @@ package com.synopsys.integration.alert.channel.email.action;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


channel-email/src/test/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalTestActionTest.java
```diff
@@ -91,12 +103,58 @@ public void testConfigMissingDestinationTest() {
...
+        @Tag(TestTags.DEFAULT_INTEGRATION),
+        @Tag(TestTags.CUSTOM_EXTERNAL_CONNECTION)
+    })
+    public void testSmtpPasswordMissingTestIT() {
+        AuthorizationManager authorizationManager = createAuthorizationManager(AuthenticationTestUtils.FULL_PERMISSIONS);
+        EmailGlobalConfigurationValidator validator = new EmailGlobalConfigurationValidator();
+        TestProperties testProperties = new TestProperties();
+        String emailAddress = testProperties.getProperty(TestPropertyKey.TEST_EMAIL_RECIPIENT);
+
+        assumeTrue(testProperties.getOptionalProperty(TestPropertyKey.TEST_EMAIL_SMTP_PASSWORD).isPresent());
```
ChomickiM, line 144: Adding an assumeTrue check for the SMTP password and skip if it is not set.  Without it we will not be correctly test the problem when attempting to test a configuration with an obfuscated password

bamandel (approved):

psantos1113 (approved):

PR #1986, opened 2022-01-20
Create a concrete model for Jira Server

ChomickiM:
Small commit to create the model for all of the other Jira Server changes to come. Mostly adding you guys to this PR for awareness as well as checking if anything is missing or should be added. Since the "Install Plugin Remotely" is its own action, I thought it doesn't really fit to add anything for it into the model.

### DIFF
--- a/channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/model/JiraServerGlobalConfigModel.java
+++ b/channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/model/JiraServerGlobalConfigModel.java
@@ -1,0 +1,75 @@
+package com.synopsys.integration.alert.channel.jira.server.model;
+
+import java.util.Optional;
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.synopsys.integration.alert.common.rest.model.ConfigWithMetadata;
+import com.synopsys.integration.alert.common.rest.model.Obfuscated;
+
+public class JiraServerGlobalConfigModel extends ConfigWithMetadata implements Obfuscated<JiraServerGlobalConfigModel> {
+    private String url;
+    private String userName;
+    private Boolean isPasswordSet;
+    private String password;
+    private Boolean disablePluginCheck;
+
+    @Override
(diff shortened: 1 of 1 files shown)


channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/model/JiraServerGlobalConfigModel.java
```diff
@@ -0,0 +1,75 @@
...
+        jiraServerGlobalConfigModel.setUserName(userName);
+        jiraServerGlobalConfigModel.setDisablePluginCheck(disablePluginCheck);
+
+        jiraServerGlobalConfigModel.setIsPasswordSet(StringUtils.isNotBlank(password));
+        jiraServerGlobalConfigModel.setPassword(null);
+
+        return jiraServerGlobalConfigModel;
+    }
+
+    public Optional<String> getUrl() {
```
bamandel, line 36: How do we feel about this optional approach for all our fields? I'm not a fan of a required field potentially being missing. I think this approach may hide other potential issues in the code base

bamandel: I understand that we did it this way for email but want to be sure we're happy with the decision

ChomickiM: If we removed the optionals, wouldn't places like the accessors require its own validation? Right now those are validated as part of the helper classes before the accessors are called. These models don't have constructors and are created field by field using the setters so wouldn't we run the risk of potential NPEs without the optionals?

psantos1113: We will need to add tests to the action class in the future to validate that null values aren't allowed going forward.  We will need to change email, proxy, and encryption related classes in the future as well.

psantos1113 (approved):

PR #1995, opened 2022-01-25
jira environment support

psantos1113:


channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandlerFactory.java
```diff
@@ -119,7 +119,10 @@ private Properties updateConfiguration() {
         obfuscatedModel.getSmtpPort().map(String::valueOf).ifPresent(value -> properties.put(EMAIL_PORT_KEY, value));
         obfuscatedModel.getSmtpFrom().ifPresent(value -> properties.put(EMAIL_FROM_KEY, value));
         obfuscatedModel.getSmtpAuth().map(String::valueOf).ifPresent(value -> properties.put(AUTH_REQUIRED_KEY, value));
-        obfuscatedModel.getSmtpPassword().ifPresent(value -> properties.put(AUTH_PASSWORD_KEY, value));
+        Boolean passwordSet = obfuscatedModel.getIsSmtpPasswordSet();
```
psantos1113, line 122: This is to avoid a Sonar issue to use raw types in an if statement.

channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/environment/JiraServerEnvironmentVariableHandlerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+        environmentVariableUtility.getEnvironmentValue(PASSWORD_KEY)
+            .ifPresent(configModel::setPassword);
+
+        JiraServerGlobalConfigModel obfuscatedModel = configModel.obfuscate();
+        properties.put(URL_KEY, obfuscatedModel.getUrl());
+        if (StringUtils.isNotBlank(obfuscatedModel.getUserName())) {
+            properties.put(USERNAME_KEY, obfuscatedModel.getUserName());
+        }
+        obfuscatedModel.getDisablePluginCheck().map(String::valueOf).ifPresent(value -> properties.put(DISABLE_PLUGIN_KEY, value));
+        Boolean passwordSet = obfuscatedModel.getPasswordSet().orElse(Boolean.FALSE);
```
psantos1113, line 78: This is also to avoid a Sonar issue using raw types in the if statement.

ChomickiM (approved):

channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/environment/JiraServerEnvironmentVariableHandlerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+
+        if (Boolean.TRUE.equals(passwordSet)) {
+            properties.put(PASSWORD_KEY, AlertRestConstants.MASKED_VALUE);
```
bamandel, line 81: This line is a bit concerning to me. Is there a way we could modify this to where we only work with classes that implement obfuscate? That way we can call it in the generic class ensuring we are always working with obfuscated models. My concern is someone forgetting to manually call obfuscate when implementing another channel and accidentally leaking private data in the logs.

psantos1113: I would like to change the object returned by the environment handler into something that better represents the data than just properties.  Especially now that this has come up.  That would mean changes to both the email channel and jira channel so I want to do that in a different PR. 
Seemed a bit out of scope for this PR.

PR #1999, opened 2022-01-26
Environment variable refactor to add variables

psantos1113:


api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentProcessingResult.java
```diff
@@ -0,0 +1,77 @@
...
+    public boolean hasValues() {
+        return !variableValues.isEmpty();
+    }
+
+    public Set<String> getVariableNames() {
+        return variableNames;
+    }
+
+    public Optional<String> getVariableValue(String variableName) {
+        return Optional.ofNullable(variableValues.getOrDefault(variableName, null));
```
bamandel, line 45: Is this redundant? Wouldn't a map return null if there was not variable?

psantos1113: yes.  I was going to return a different default but didn't need to.

ChomickiM (approved):

PR #2002, opened 2022-01-28
testcontainers config to start the container before the tests

psantos1113:
Test containers can be configured for JDBC through the URLS.  However testcontainers needs a few dependencies to be able to start the containers at the beginning of each test.
https://www.testcontainers.org/modules/databases/jdbc/

You still need the correct dependencies with spring boot to start the container before each test:
https://github.com/Playtika/testcontainers-spring-boot

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/processing/JmsNotificationReceiverTestIT.java
+++ b/src/test/java/com/synopsys/integration/alert/processing/JmsNotificationReceiverTestIT.java
@@ -12,6 +12,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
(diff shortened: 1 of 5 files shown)


src/test/java/com/synopsys/integration/alert/processing/JmsNotificationReceiverTestIT.java
```diff
@@ -65,8 +66,9 @@
 @WebAppConfiguration
 @AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
 @ActiveProfiles("alertdb")
+@SpringBootTest
 @TestExecutionListeners({ DependencyInjectionTestExecutionListener.class, DirtiesContextTestExecutionListener.class, TransactionalTestExecutionListener.class, DbUnitTestExecutionListener.class })
-public class JmsNotificationReceiverTestIT {
+class JmsNotificationReceiverTestIT {
```
jamesrichard91, line 71: Was this needed? Why shouldn't this or the method below be public?

PR #2005, opened 2022-01-28
CRUD action and accessor improvements

ChomickiM:
This change is targeted at several changes and updates to the ConfigurationCrudHelper

- Create will now determine if a configuration by name already exists before attempting to create
- Better error handling for expected accessor errors
- By request, audit logging has been included for whenever an action begins and before it exits

The accessors have also been updated:

- Will now verify if a configuration already exists before attempting to create it. Otherwise it will throw an AlertConfigurationException
- Modified places using the accessor create to handle the new exception.


alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/ConfigurationCrudHelper.java
```diff
@@ -27,95 +27,142 @@
...
+
     public ConfigurationCrudHelper(AuthorizationManager authorizationManager, ConfigContextEnum context, DescriptorKey descriptorKey) {
         this.authorizationManager = authorizationManager;
         this.context = context;
         this.descriptorKey = descriptorKey;
     }
 
     public <T extends Obfuscated<T>> ActionResponse<T> getOne(Supplier<Optional<T>> modelSupplier) {
+        String actionName = "getOne";
+        logger.trace(ACTION_CALLED_TEMPLATE, descriptorKey.getUniversalKey(), actionName);
```
bamandel, line 50: Do we want trace or debug? Since this is the config API, it shouldn't spam the logs too much and will probably be helpful for debugging

bamandel: this is in regards to all of the logs in this class

ChomickiM: I bumped it down to trace from debug by an earlier request. It does still make a pretty decent number of log messages.  Each page we navigate to using the new controller will create at least 2 messages due to the get request we make.

ChomickiM: This logger will write to the audit file regardless of log level. So if this is  to be primarily used by us it shouldn't matter what level it's at assuming we're checking that file.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalCrudActions.java
```diff
@@ -46,6 +46,7 @@ public EmailGlobalCrudActions(AuthorizationManager authorizationManager, EmailGl
     public ActionResponse<EmailGlobalConfigModel> create(EmailGlobalConfigModel resource) {
         return configurationHelper.create(
             () -> validator.validate(resource),
+            () -> configurationAccessor.getConfigurationByName(resource.getName()).isPresent(),
```
bamandel, line 49: This warrants a new method in the accessor to determine if the config exists (using SQL) as opposed to retrieving the entire config and checking if it's present.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/database/accessor/EmailGlobalConfigAccessor.java
```diff
@@ -83,7 +83,11 @@ public long getConfigurationCount() {
 
     @Override
     @Transactional(propagation = Propagation.REQUIRED)
-    public EmailGlobalConfigModel createConfiguration(EmailGlobalConfigModel configuration) {
+    public EmailGlobalConfigModel createConfiguration(EmailGlobalConfigModel configuration) throws AlertConfigurationException {
+        Optional<EmailConfigurationEntity> configurationEntity = emailConfigurationRepository.findByName(configuration.getName());
```
bamandel, line 87: Another place the exist method would help

bamandel (commented):
Looks like we will want to add the exists by name method to each of our new accessors

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/api/ConfigurationCrudHelper.java
psantos1113: I think we want this at trace because it is logging that we entered and exited the method.  I like this in the audit log.  In our UI we have the auto-refresh so if it was at debug we might see a number of get requests because the user is just sitting on a page in the UI with autorefresh.  Might distract from when a user actually clicks on save, test, or delete

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/action/EmailGlobalCrudActions.java
psantos1113: yes we should add to the ConfigurationAccessor interface existsConfigurationByName(String name) and then add to each configuration repository existsByName.  Should be a more efficient call.

component/src/main/java/com/synopsys/integration/alert/component/settings/proxy/database/accessor/SettingsProxyConfigAccessor.java
```diff
@@ -52,11 +52,17 @@ public SettingsProxyConfigAccessor(
...
+    public boolean doesConfigurationExist() {
+        return settingsProxyConfigurationRepository.existsByName(AlertRestConstants.DEFAULT_CONFIGURATION_NAME);
+    }
+
     @Override
     @Transactional(propagation = Propagation.REQUIRED)
     public SettingsProxyModel createConfiguration(SettingsProxyModel configuration) throws AlertConfigurationException {
-        Optional<SettingsProxyModel> existingConfiguration = getConfiguration();
-        if (existingConfiguration.isPresent()) {
+        boolean foo = doesConfigurationExist(); //TODO
```
bamandel, line 64: Just want to point this out incase you forgot/missed it

PR #2012, opened 2022-02-03
Updated JiraServerPropertiesFactory to use the concrete model instead of dynamic

bamandel:
**Added password to the required arguments list.** I made this change as Password is technically required even tho it is null when obfuscated. The number of tests and classes that were forced to add the password after object creation should help show that this change is necessary.

### DIFF
--- a/channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/model/JiraServerGlobalConfigModel.java
+++ b/channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/model/JiraServerGlobalConfigModel.java
@@ -25,15 +25,15 @@ public class JiraServerGlobalConfigModel extends ConfigWithMetadata implements O
         // For serialization
     }
 
-    public JiraServerGlobalConfigModel(String id, String name, String url, String userName) {
+    public JiraServerGlobalConfigModel(String id, String name, String url, String userName, String password) {
         super(id, name);
         this.url = url;
         this.userName = userName;
+        this.password = password;
     }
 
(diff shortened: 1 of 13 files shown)


channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/model/JiraServerGlobalConfigModel.java
```diff
@@ -25,15 +25,15 @@ public JiraServerGlobalConfigModel() {
         // For serialization
     }
 
-    public JiraServerGlobalConfigModel(String id, String name, String url, String userName) {
+    public JiraServerGlobalConfigModel(String id, String name, String url, String userName, String password) {
```
gkillough, line 28: What if you had two classes related through polymorphism? One with the password, one without? Use the one with the password for _write_ operations and the one without for _read_?

ChomickiM, line 28: If we are passing the password this way, would we want it to be @Nullable since the UI would return a password with the value "null"?

psantos1113: We would need an internal model that contains the password for when the job executes. Right now this model encompasses all use cases.  Maybe the obfuscate method should return a different type as you've suggested and The implementation of the Obfuscate method declares the obfuscated type that removes the password.  We would have to make some changes to the utility classes to support this.  I think we should take this up as a future enhancement.

psantos1113: We return null today for the email password field.  we have the same constraint and concern with the email channel global configuration.

PR #2014, opened 2022-02-08
jira job selection

psantos1113:


bamandel:
In the future I believe we could modify the table select to not be field model specific and only allow a single select. That could take some time though, so this looks good!

ui/src/main/js/page/channel/jira/server/JiraServerDistributionConfiguration.js
```diff
@@ -6,16 +6,50 @@ import CheckboxInput from 'common/component/input/CheckboxInput';
 import TextInput from 'common/component/input/TextInput';
 import CollapsiblePane from 'common/component/CollapsiblePane';
 import FieldMappingField from 'common/component/input/FieldMappingField';
+import { DISTRIBUTION_COMMON_FIELD_KEYS } from 'page/distribution/DistributionModel';
+import EndpointSelectField from '../../../../common/component/input/EndpointSelectField';
+import { createReadRequest } from '../../../../common/util/configurationRequestBuilder';
```
ChomickiM, line 11: We should use the absolute paths here

ui/src/main/js/page/channel/jira/server/JiraServerDistributionConfiguration.js
```diff
@@ -6,16 +6,50 @@ import CheckboxInput from 'common/component/input/CheckboxInput';
...
 
 const JiraServerDistributionConfiguration = ({
-    data, setData, errors, readonly
+    csrfToken, data, setData, errors, readonly
 }) => {
+    const readRequest = () => {
+        const apiUrl = '/alert/api/configuration/jira_server?pageNumber=0&pageSize=25';
```
ChomickiM, line 17: I know this is probably very edge case, but would we need to state anywhere that we have an upper limit of 25 Jira Servers?

ChomickiM (approved):

ui/src/main/js/page/channel/jira/server/JiraServerDistributionConfiguration.js
psantos1113: Thanks noticed that and fixed it.

ui/src/main/js/page/channel/jira/server/JiraServerDistributionConfiguration.js
psantos1113: I added to the description we limit to the first 25 Jira servers

ui/src/main/js/common/component/input/GlobalConfigurationSelectInput.js
```diff
@@ -0,0 +1,465 @@
...
+    const hasError = getValue().find((value) => value.missing);
+    return (
+        <ValueContainer {...props}>
+            {hasError && error}
+            {children}
+        </ValueContainer>
+    );
+};
+
+const GlobalConfigurationSelectInput = (props) => {
```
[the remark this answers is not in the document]
psantos1113: I removed it.  Maybe in the future we will create something like it.

PR #2027, opened 2022-02-23
Proxy and Email environment variable fixes

ChomickiM:
Note, this is to remedy some of the issues that are outright failing when attempting to set up via the environment variables. This PR isn't intended to fix the problem with partial configurations or validating environment variables.


component/src/main/java/com/synopsys/integration/alert/component/settings/environment/ProxySettingsEnvironmentHandlerFactory.java
```diff
@@ -57,12 +58,13 @@ public EnvironmentVariableHandler build() {
     }
 
     private Boolean isConfigurationMissing() {
-        return configAccessor.getConfiguration().isPresent();
+        return configAccessor.getConfiguration().isEmpty();
```
bamandel, line 61: Would this be a case for the exists method?

ChomickiM: the exists method for SettingsProxyConfigAccessor wasn't created until 6.10.0. My idea was when merging into master  to update this in master to use the exists method.

bamandel (approved):

PR #2028, opened 2022-02-25
Email environment handler validation

ChomickiM:


channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandlerFactory.java
```diff
@@ -113,6 +122,17 @@ private Properties updateConfiguration() {
...
 
+        ValidationResponseModel validationResponseModel = validator.validate(configModel);
+        if (validationResponseModel.hasErrors()) {
+            logger.error("Error inserting startup values");
+            Map<String, AlertFieldStatus> errors = validationResponseModel.getErrors();
+            for (Map.Entry<String, AlertFieldStatus> error : errors.entrySet()) {
+                AlertFieldStatus status = error.getValue();
+                logger.error("Field: '{}' failed with the error: {}", status.getFieldName(), status.getFieldMessage());
+            }
+            return properties;
```
bamandel, line 133: We don't end up inserting any config because of the errors right? We should add a warning here stating that the email config was unable to be saved.

ChomickiM: I can add that log message after the final field error. Do you want that logged at WARN or ERROR like the others?

bamandel: Sorry had missed the initial error message at the top. You're probably fine with that but will want to add the message from the ValidationResponseModel object to the log message.

PR #2032, opened 2022-03-01
Move ValidationResponseModel and AlertFieldStatus into api-common-model

ChomickiM:


bamandel:
This is a very large PR so it may not be too valuable to review everything. Does it only include the refactor of the class moves we spoke about? If so I'd say merge away.

ChomickiM:
Yes, the primary moves here are ValidationResponseModel  into api-common-model, and I moved the "errors" package containing AlertFieldStatus, AlerFieldStatusMessages, and FieldStatusSeverity into alert-common-model. I wanted to make sure those changes sounded good to you guys and just for awareness.

Should have probably included that in the description, my bad!

PR #2039, opened 2022-03-08
Environment Handler Refactor

ChomickiM:
Opening this as a draft. To get some feedback on the updated approach for handling environment variables for new channels.

The primary goal here was to add a validation step introduced in 6.9.0 but to improve it by requiring it at the handler level rather than each having its own implementation.

To accomplish this I took the following steps:

- Moved Obfuscated interface to be in alert-common-model which allows it to be accessed from api-environment.
- Created EnvironmentVariableHandler2 with functions and suppliers for creating the model, validating, and updating within the updateFromEnvironment method.
- Created EmailEnvironmentVariableHandlerFactory2 as a proof of concept for these changes, splitting it into 3 methods to:  supply an EmailGlobalConfigModel, a function to validate the model, and a function to create and return  an EnvironmentProcessingResult

All other changes here were made to support the existing handlers, and allow the testing of updated V2 handlers. If this draft looks good, I will replace the old V1 handlers and implement these changes for Proxy, Encryption, and Jira Server.

### DIFF
--- a/api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandler2.java
+++ b/api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandler2.java
@@ -82,16 +82,6 @@ public class EnvironmentVariableHandler2<T extends Obfuscated<T>> {
     public EnvironmentProcessingResult updateFromEnvironment() {
         boolean configurationMissing = configurationMissingCheck.getAsBoolean();
         if (configurationMissing) {
-            //return updateFunction.apply();
-        }
-
-        return EnvironmentProcessingResult.empty();
-    }
-
-    //TODO: Rename and replace with the above
-    public EnvironmentProcessingResult updateFromEnvironment2() {
-        boolean configurationMissing = configurationMissingCheck.getAsBoolean();
-        if (configurationMissing) {
             //T configurationModel = createConfiguration();
             T configurationModel = configModelSupplier.get();
             ValidationResponseModel validationResponseModel = validator.apply(configurationModel);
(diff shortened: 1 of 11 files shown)


api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandler2.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public Set<String> getVariableNames() {
+        return environmentVariableNames;
+    }
+
+    public boolean isConfigurationMissing() {
+        return configurationMissingCheck.getAsBoolean();
+    }
+
+    public EnvironmentProcessingResult updateFromEnvironment() {
```
ChomickiM, line 72: Relevant changes here. I've moved the validation logic from the EnvironmentVariableHandlerFactory and put it into  the method below.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandlerFactory2.java
```diff
@@ -0,0 +1,216 @@
...
+            this::validateConfiguration,
+            this::configureModel
+        );
+    }
+
+    private Boolean isConfigurationMissing() {
+        return !configAccessor.doesConfigurationExist();
+    }
+
+    private EmailGlobalConfigModel configureModel() {
```
ChomickiM, line 126: Relevant changes here. In the 3 methods below, I've split the updateConfiguration into 3 parts to configure the model, validate it, and update. These are passed into the build method above as suppliers/functions respectively

PR #2047, opened 2022-03-14
environment handler validation (inheritence)

ChomickiM:
An alternative approach to https://github.com/blackducksoftware/blackduck-alert/pull/2039

Using this method, we would make the EnvironmentVariableHandler an abstract class and create several protected methods that are implemented by each of the new configurations.


api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandlerV3.java
```diff
@@ -0,0 +1,71 @@
...
+            T configurationModel = configureModel();
+            ValidationResponseModel validationResponseModel = validateConfiguration(configurationModel);
+            if (validationResponseModel.hasErrors()) {
+                logger.error("Error inserting startup values: {}", validationResponseModel.getMessage());
+                for (AlertFieldStatus errorStatus : validationResponseModel.getErrors().values()) {
+                    logger.error("Field: '{}' failed with the error: {}", errorStatus.getFieldName(), errorStatus.getFieldMessage());
+                }
+                return EnvironmentProcessingResult.empty();
+            }
+            EnvironmentProcessingResult processingResult = buildProcessingResult(configurationModel);
```
bamandel, line 53: obfuscating the model here and making the variable final will help remove the possibility of another implementation accidentally causing issues here

api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandlerV3.java
```diff
@@ -0,0 +1,71 @@
...
+            saveConfiguration(configurationModel, processingResult);
```
bamandel, line 54: This would of course take the original model

api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandlerV3.java
```diff
@@ -0,0 +1,71 @@
...
+        return EnvironmentProcessingResult.empty();
+    }
+
+    protected abstract Boolean configurationMissingCheck();
+
+    protected abstract T configureModel();
+
+    protected abstract ValidationResponseModel validateConfiguration(T configModel);
+
+    protected abstract EnvironmentProcessingResult buildProcessingResult(T configModel);
```
bamandel, line 67: changing the name to obfuscatedConfigModel and making it final would be beneficial here

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandler.java
```diff
@@ -0,0 +1,202 @@
...
+        if (Boolean.TRUE.equals(obfuscatedModel.getIsSmtpPasswordSet())) {
+            builder.addVariableValue(AUTH_PASSWORD_KEY, AlertConstants.MASKED_VALUE);
+        }
+
+        return builder.build();
+    }
+
+    @Override
+    protected void saveConfiguration(EmailGlobalConfigModel configModel, EnvironmentProcessingResult processingResult) {
+        if (processingResult.hasValues()) {
```
bamandel, line 152: Could this always be the case? If so it may be worth pulling it up into the parent class

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandler.java
```diff
@@ -0,0 +1,202 @@
...
+            if (environmentVariableUtility.hasEnvironmentValue(additionalPropertyName)) {
+                String javamailPropertyName = convertVariableNameToJavamailPropertyKey(additionalPropertyName);
+                String value = environmentVariableUtility.getEnvironmentValue(additionalPropertyName).orElse(null);
+                additionalProperties.put(javamailPropertyName, value);
+            }
+        }
+        configuration.setAdditionalJavaMailProperties(additionalProperties);
+    }
+
+    public static String convertVariableNameToJavamailPropertyKey(String environmentVariableName) {
```
bamandel, line 197: We generally have our public static methods towards the top of the class but we will want to verify if this method should live here and be public static.

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandler.java
ChomickiM: Jira has it slightly different here because it also checks to make sure the configuration by name doesn't exist. We could pull out the first part for the hasValues, but keep the second part for Jira

component/src/main/java/com/synopsys/integration/alert/component/settings/environment/EncryptionSettingsEnvironmentVariableHandler.java
```diff
@@ -0,0 +1,70 @@
...
+        Optional<String> salt = environmentVariableUtility.getEnvironmentValue(ENCRYPTION_SALT_KEY);
+
+        // The encryption utility will read the environment variables.  We just want to be able to log if the variables are set.
+        if (password.isPresent()) {
+            builder.addVariableValue(ENCRYPTION_PASSWORD_KEY, AlertConstants.MASKED_VALUE);
+        }
+
+        if (salt.isPresent()) {
+            builder.addVariableValue(ENCRYPTION_SALT_KEY, AlertConstants.MASKED_VALUE);
+        }
```
bamandel, line 61: Can all be condensed into a couple optional.ifPresent calls

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandler.java
bamandel: Ya it certainly looks like we can pull the first part out judging by the other handlers

channel-email/src/main/java/com/synopsys/integration/alert/channel/email/environment/EmailEnvironmentVariableHandler.java
ChomickiM: @psantos1113 could you weigh in on this? The only other place I see this used is in the tests for  the original class. I'm fine with making this private and creating a private method in the test to do the same functionality.

bamandel (approved):

PR #2055, opened 2022-03-21
Environment handler cleanup

ChomickiM:
Cleaning up the tech debt we introduced in 6.10.0. At the same time, I've also added some missing tests for other handlers.


api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandler.java
```diff
@@ -35,15 +35,39 @@ public String getName() {
     }
 
     public boolean isConfigurationMissing() {
-        return configurationMissingCheck.getAsBoolean();
+        return configurationMissingCheck();
     }
 
     public EnvironmentProcessingResult updateFromEnvironment() {
-        boolean configurationMissing = configurationMissingCheck.getAsBoolean();
+        boolean configurationMissing = configurationMissingCheck();
```
psantos1113, line 42: Is this when the configuration exists? I would change the variable to configurationExists. 
Also consider short circuiting sooner:
```suggestion
        boolean configurationMissing = !configurationMissingCheck();
        if(configurationMissing) {
            return EnvironmentProcessingResult.empty();
        }
        
        T configurationModel = configureModel();
        ValidationResponseModel validationResponseModel = validateConfiguration(configurationModel);
        ....
        return processingResult;
```

psantos1113: This isn't needed here.  Resolving conversation.

PR #2067, opened 2022-03-30
Proxy model required fields

ChomickiM:
I had to make significant changes to the validator to support the new model for email, this will more closely follow the process we use in email and jira.

### DIFF
--- a/component/src/main/java/com/synopsys/integration/alert/component/settings/convert/ProxyConfigurationModelConverter.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/settings/convert/ProxyConfigurationModelConverter.java
@@ -10,6 +10,7 @@ package com.synopsys.integration.alert.component.settings.convert;
 import java.util.ArrayList;
 import java.util.Optional;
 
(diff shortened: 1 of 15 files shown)


component/src/main/java/com/synopsys/integration/alert/component/settings/convert/ProxyConfigurationModelConverter.java
```diff
@@ -42,32 +43,32 @@ public ProxyConfigurationModelConverter(SettingsProxyValidator validator) {
 
     @Override
     public Optional<SettingsProxyModel> convert(ConfigurationModel globalConfigurationModel) {
-        Optional<SettingsProxyModel> convertedModel = Optional.empty();
-        SettingsProxyModel model = new SettingsProxyModel();
-        model.setName(AlertRestConstants.DEFAULT_CONFIGURATION_NAME);
-        try {
```
ChomickiM, line 48: Since proxy port must be handled prior to creating the model we no longer need the try-catch block

PR #2072, opened 2022-03-31
CollapsiblePane - Convert from class component to functional component

bseifert14:
_To Do:_ 

- [x] Add Testing!

**Fix:**  
Convert the `CollapsiblePane.js` component from a class to functional.  This component doesn't need any state variables as it really is just a toggle switch that determines whether something (in this case, a couple of forms) should be expanded or not.

**Testing:**  
There are four areas where `CollapsiblePane.js` is called: in Authentication (LDAP & SAML), Distribution configuration when 'Jira Cloud' or 'Jira Server' is selected in the Channel field.


bseifert14:
I also added the `icons.js` file in this MR.  The intention is to create a centralized import of the icons used in this repo.  This will 1) Reduce the size of the repo (not by much - but anything counts), 2) Help to create a standardized `Icon` functional component (reduce lines of code), and 3) Help immensely with the testing process

psantos1113 (approved):

ChomickiM (approved):

PR #2100, opened 2022-05-11
Policy notification performance test

ChomickiM:
This PR is to add a new way of testing large number of notifications in blackduck. To run this test we must do the following:

- Set alert.performance.blackduck.project.count in your test.properties
- In LargeNotificationTest run the following:
   - If projects need to be created: createProjectsAndNotificationsTest
   - If projects need to be cleaned up: deleteProjectsAndVersionsTest followed by createProjectsAndNotificationsTest to create a new set of projects on blackduck.
- Run largePolicyNotificationTest


src/main/java/com/synopsys/integration/alert/configuration/ApplicationConfiguration.java
```diff
@@ -93,6 +95,13 @@ public TaskScheduler taskScheduler() {
         threadPoolTaskScheduler.setPoolSize(Runtime.getRuntime().availableProcessors());
         return threadPoolTaskScheduler;
     }
+
+    //TODO: Required during 6.11.0 performance investigation but should be removed in the future in favor of a thread pool
+    @Bean
+    public TaskExecutor taskExecutor() {
+        return new SyncTaskExecutor();
+    }
+
```
ChomickiM, line 104: There are potential issues when we are running without Sync (too many messages distributed can lead to distribution channel failures). After speaking with Paulo, we will want to add this into master for the time being while we work on other performance improvements.

This should be revisited at a later time.

DanaMaxfield (approved):

PR #2101, opened 2022-05-13
IALERT-3051 - Only perform provider validation when configuring provider

DanaMaxfield:
In the current implementation of Alert, we perform the full provider validation, which includes checking user/role permissions in BD, when configuring a provider, configuring a distribution, and on each launch of the accumulator.

Changes:
- New canConnect() method which checks the return http status code of connecting to BD using an API token.
- Use canConnect() first when communicating with an instance of BD
- Only call isApiTokenValid() from provider
- Update tests to reflect changes
- Add full coverage for canConnect()

### DIFF
--- a/provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/BlackDuckDistributionFieldModelTestAction.java
+++ b/provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/BlackDuckDistributionFieldModelTestAction.java
@@ -126,12 +126,9 @@ public class BlackDuckDistributionFieldModelTestAction extends FieldModelTestAct
             boolean foundResult = false;
             while (!foundResult && currentPage < projectsByProviderConfigId.getTotalPages()) {
                 List<String> providerProjects = projectsByProviderConfigId.getModels();
-                foundResult = providerProjects.stream().anyMatch(href ->
-                                                                     iteratePagesAndCheck(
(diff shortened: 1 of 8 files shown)


psantos1113 (approved):

provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/BlackDuckDistributionFieldModelTestAction.java
```diff
@@ -125,11 +127,11 @@ private Optional<AlertFieldStatus> validatePatternMatchesProjectVersion(
...
-                    iteratePagesAndCheck(
-                        versionCurrentPage -> blackDuckDataAccessor.getProjectVersionNamesByHref(providerConfigId, href, versionCurrentPage),
-                        versionNames -> versionNames.stream().anyMatch(versionName -> compiledProjectVersionPattern.matcher(versionName).matches()),
-                        Boolean.FALSE
-                    ).isEmpty());
+                                                                     iteratePagesAndCheck(
+                                                                         versionCurrentPage -> blackDuckDataAccessor.getProjectVersionNamesByHref(providerConfigId, href, versionCurrentPage),
+                                                                         versionNames -> versionNames.stream().anyMatch(versionName -> compiledProjectVersionPattern.matcher(versionName).matches()),
+                                                                         Boolean.FALSE
+                                                                     ).isEmpty());
```
ChomickiM, line 134: I think that save actions may have broken here and added some spacing?

PR #2117, opened 2022-06-02
Use paging on notification processing

psantos1113:
use paging to get notifications.  Process a page of notifications into a message holder.  With each page reduce the notifications into one list.  Lastly distribute the final message holder containing all messages.


bseifert14 (approved):
I've looked through the code. Albeit I'm not an expert in java, the logic of this PR makes sense and the code does look good (again I'm no expert)

ChomickiM (approved):

PR #2131, opened 2022-06-10
test: Add new property and test to create projects only

DanaMaxfield:
* Add new property 'TEST_PERFORMANCE_BLACKDUCK_START_PROJECT_NUMBER' to control what the starting project number should be
* Add createTestProjects to only create the projects without triggering a notification

The thinking/intention behind this change is to allow us to configure a job in Jenkins to run createTestProjects multiple times simultaneously. I tested created 10K projects and it took 230. I then ran the job 5 times at once, with different values for the new property, and each creating 2K projects. The total runtime was 97 minutes.

If this is approved, I will create a Jenkins job that will run this based on some information from the user:
- BD server to run against
- BD API key
- Number of jobs to execute
- Number of projects to create in each job
- Starting project number

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/performance/LargeNotificationTest.java
+++ b/src/test/java/com/synopsys/integration/alert/performance/LargeNotificationTest.java
@@ -49,6 +49,7 @@ import com.synopsys.integration.exception.IntegrationException;
 class LargeNotificationTest {
     private static final JiraServerChannelKey CHANNEL_KEY = new JiraServerChannelKey();
     private static final int DEFAULT_NUMBER_OF_PROJECTS_TO_CREATE = 10;
+    private static final int DEFAULT_STARTING_PROJECT_NUMBER = 1;
     private static final String PERFORMANCE_POLICY_NAME = "PerformanceTestPolicy";
     private static final String DEFAULT_JOB_NAME = "JiraPerformanceJob";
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/synopsys/integration/alert/performance/LargeNotificationTest.java
```diff
@@ -91,6 +92,32 @@ public void init() {
             .orElse(DEFAULT_NUMBER_OF_PROJECTS_TO_CREATE);
     }
 
+    @Test
+    @EnabledIfEnvironmentVariable(named = "ALERT_RUN_PERFORMANCE", matches = "true")
+    void createTestProjects() throws IntegrationException {
```
psantos1113, line 97: This is good.  What I found out is that the trigger of the notifications adds a component to the project.  If we don't have  components the performance tests won't work if we rely on the policy to trigger a bunch of notifications.  So we may need to have a method to add components to the projects as well.  A bit out of scope for this PR but something to be aware of.

DanaMaxfield: So the original test we used looped twice, once to find/create the project, and once to find/create **then** trigger. I will adjust this to do that find/create then trigger.

DanaMaxfield: @psantos1113 I updated the code to do the create + trigger in one loop. Run time went from 67 minutes per 2K -> 97 minutes per 2K. And that is with 5 simultaneous jobs each making 2K.

psantos1113 (approved):

PR #2143, opened 2022-06-28
Update build to work in ARM64 environment

DanaMaxfield:
** This address IALERT-3028, IALERT-3040, and IALERT-3098

IALERT-3028/IALERT-3040
* Update images to latest minor version on current major version which supports ARM64.
* The dependency on the eclipse-temurin image was eliminated as they do not have a Java11 that supports ARM64 AND apk now carries openjdk11 as a package.
* This was tested on my new ARM64 laptop, my old laptop, and the build server.

IALERT-3098


psantos1113 (approved):

DanaMaxfield:
Updated the base image for rabbitmq. When/If a 3.10.6 is released, the 3.10 base image will be updated to be 3.10.6, but the name will remain the same.

ui/build.gradle
```diff
@@ -42,7 +42,7 @@ task npmInstallFromCache(type: Exec) {
     outputs.cacheIf { true }
     ignoreExitValue = true
 
-    commandLine 'npm', 'install', '--cache', "${project.buildDir}/npm_cache"
+    commandLine 'npm', 'install', '--force', '--cache', "${project.buildDir}/npm_cache"
```
DanaMaxfield, line 45: Why force?

bseifert14: I just did a deep dive into this and would like to overly share what I learned as I think it's important to understand the whole situation.  Within NPM packages there are five types of dependencies; **I'll focus on three of the five in this explanation**, but invite you to read about the other two within this link [NPM Dependency Overview](https://betterprogramming.pub/package-jsons-dependencies-in-depth-a1f0637a3129).  
  

1. `dependencies`: packages our project depends on. These are required for our project to run; a realtime example is to look at package.json's "dependencies".
2. `devDependencies`: packages needed during the development phase but otherwise not required for production.  These are not required for our project to run, rather they are there for testing, linting, etc.  These would be used as we build our app up in a local environment to ensure that there would be no issues in our production environment. A realtime example is to look at package.json's "devDependencies".
3. `peerDepencies`: packages used by the dependencies of the 1. `dependencies`.  I know that sounds confusing.  This bodes the question of what is a transitive dependency and what is a peer dependency.  I'm going to try and explain the difference below without upsetting the open source community gods.  
  
A dependency will **require** it's respective transitive dependencies in order to successfully install/run the package. 
A transitive dependency will **recommend** it's respective peer dependencies in order to successfully install/run the transitive package. A peer dependency essentially says, 'Hey, I'm package "superCoolPackage". I recommend you have this peerDependency as I know that "superCoolPackage" will run best with this peerDependency's version.'
  
This brings us to the `--force` flag.  In NPM's version 3 to 6, an `npm install` would automatically ignore peerDependencies.  This was all fine and dandy, except I have read that many developers found this to be unsafe as it could potentially break a build.  In NPM >=7, they've removed this and given the user the option to define either `--force` or `--legacy-peer-deps`.  
`--legacy-peer-deps`: is the old way (NPM v3 to v6) where it ignores the peer dependencies.  
`--force`: will forcefully install these peer dependencies.  
  
I have added this `--force` flag as it (in some way or another) covers us if there are any install issues of our npm build.  The compromise for that safety, is that the build time will take a little bit longer.  As I go forward updating these packages in further PR's I hope to get to a point where we can remove this flag, but for now I think it's important to be safe rather than sorry.  
  
Finally, I want to include some of the resources that helped me understand the situation:  
[NPM peerDependencies](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependencies)  
[NPM legacy-peer-deps](https://docs.npmjs.com/cli/v7/using-npm/config#legacy-peer-deps)
[NPM force](https://docs.npmjs.com/cli/v6/commands/npm-install)
[Simple explanation of peerDeps](https://flaviocopes.com/npm-peer-dependencies/)
[Peer vs Transitive Deps](https://medium.com/angular-in-depth/npm-peer-dependencies-f843f3ac4e7f)  
[Great StackOverflow explanation](https://stackoverflow.com/a/66620869)

bseifert14 (approved):

PR #2149, opened 2022-07-05
pattern filtering to match on version

psantos1113:
If the project name pattern or selected project matches check if the version name matches.  If we aren't filtering by project at all then the job applies to the project and short circuit.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -1,6 +1,6 @@
 {
     "name": "blackduck-alert",
-    "version": "6.11.0-SIGQA1",
+    "version": "6.11.0-SIGQA2-SNAPSHOT",
     "description": "Alert UI",
     "keywords": [
         "blackduck",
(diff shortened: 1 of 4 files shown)


ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "blackduck-alert",
-    "version": "6.11.0-SIGQA1",
```
DanaMaxfield, line 3: Shouldn't this have been updated by the 'Next Snapshot' stage? Is this a bug in the build library?

DanaMaxfield (approved):

ChomickiM (approved):

ui/package.json
psantos1113: The gradle task to build the UI updates the version to the gradle version which is updated with the Next Snapshot stage in the build.

PR #2150, opened 2022-07-05
IALERT-3099: Send the event to indicate jobs are mapped to notifications

psantos1113:
The DailyTask was not sending an event to indicate jobs were mapped to notifications to begin the process for sending messages to the channels.  Also the time range isn't calculated correctly.  It would only run the daily task if notifications came in since it started running.

### DIFF
--- a/component/src/main/java/com/synopsys/integration/alert/component/tasks/ProcessingTask.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/tasks/ProcessingTask.java
@@ -15,6 +15,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.scheduling.TaskScheduler;
 
+import com.synopsys.integration.alert.api.event.EventManager;
 import com.synopsys.integration.alert.api.task.StartupScheduledTask;
 import com.synopsys.integration.alert.api.task.TaskManager;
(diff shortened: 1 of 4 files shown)


component/src/main/java/com/synopsys/integration/alert/component/tasks/ProcessingTask.java
```diff
@@ -34,29 +36,28 @@ public abstract class ProcessingTask extends StartupScheduledTask {
     private final NotificationMappingProcessor notificationMappingProcessor;
     private final JobAccessor jobAccessor;
     private final FrequencyType frequencyType;
+    private final EventManager eventManager;
 
     private OffsetDateTime lastRunTime;
 
-    public ProcessingTask(TaskScheduler taskScheduler, TaskManager taskManager, NotificationAccessor notificationAccessor, NotificationMappingProcessor notificationMappingProcessor,
-        JobAccessor jobAccessor, FrequencyType frequencyType) {
+    protected ProcessingTask(
```
ChomickiM, line 43: We should break these out into their own lines

ChomickiM (approved):

component/src/main/java/com/synopsys/integration/alert/component/tasks/ProcessingTask.java
psantos1113: reformatted

PR #2154, opened 2022-07-10
IALERT-3029 - Update to latest CGP version

DanaMaxfield:
Update builds to use latest CGP version

Comprehensive build run successfully


ChomickiM:
Is there a reason that we've removed the EULA?

DanaMaxfield:
@ChomickiM 
> Is there a reason that we've removed the EULA?

I don't remember the specifics, but this was came from outside of the team that the EULA should be removed

ChomickiM (approved):

psantos1113 (approved):

PR #2168, opened 2022-08-01
IALERT-3105: Support new blackduck roles in 2022.7.0

ChomickiM:


provider-blackduck/src/main/java/com/synopsys/integration/alert/provider/blackduck/action/BlackDuckGlobalFieldModelTestAction.java
```diff
@@ -77,18 +77,34 @@ public MessageResult testConfig(String configId, FieldModel fieldModel, FieldUti
...
 
         BlackDuckApiTokenValidator blackDuckAPITokenValidator = new BlackDuckApiTokenValidator(blackDuckProperties);
         if (!blackDuckAPITokenValidator.isApiTokenValid()) {
-            throw AlertFieldException.singleFieldError(BlackDuckDescriptor.KEY_BLACKDUCK_API_KEY, "User permission failed. Cannot read notifications from Black Duck.");
+            throw AlertFieldException
+                .singleFieldError(
+                    "User permission failed. Cannot read notifications from Black Duck.",
+                    BlackDuckDescriptor.KEY_BLACKDUCK_API_KEY,
+                    "User permission failed. Cannot read notifications from Black Duck."
+                );
```
DanaMaxfield, line 107: How come the change in which constructor to use?
Why is the message and fieldError the same thing?

ChomickiM: This was an earlier oversight that was causing no message to be created when an  exception was thrown. We would attempt to log the message in the response model but since no message was created we would see a "null" in  the message body in the Alert logs.

In this case, the message and the field error would be the same so I just duplicated it in both places. Functionally,  there should be no difference, this is just to clean up some ugly "null" messages in the log.

DanaMaxfield (approved):

PR #2170, opened 2022-08-03
Refactor to Synchronous messaging

psantos1113:
Convert back to synchronous messaging but search for if the issue has already been created.

Need to update the usage of the audit entry success and failure to use the message queue.


channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/distribution/delegate/JiraServerIssueCreator.java
```diff
@@ -94,6 +109,19 @@ protected IssueCreationRequestModel createIssueCreationRequest(IssueCreationMode
         );
     }
 
+    @Override
+    protected List<JiraSearcherResponseModel> searchForIssue(IssueCreationModel alertIssueCreationModel) {
+        String query = alertIssueCreationModel.getQueryString().orElse(null);
```
ChomickiM, line 114: Is there potential for a NPE here and line 117? It appears in other places we check if the query is there with StringUtils.isBlank

ChomickiM (approved):

channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/distribution/delegate/JiraServerIssueCreator.java
psantos1113: It's not really possible but I will make a change to address it.

PR #2178, opened 2022-08-17
Add tests to api-database-job for JobSubTask entities

psantos1113:
Add test into the api-database-job project for the JobSubTask entities and accessor.
Create in test-common a class that can act as a mock JPA repository to hold data inside a map datastructure.


test-common/src/main/java/com/synopsys/integration/alert/test/common/database/MockRepositoryContainer.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @Override
+    public <S extends T> boolean exists(@NotNull Example<S> example) {
+        return false;
+    }
+
+    @Override
+    public <S extends T, R> R findBy(@NotNull Example<S> example, @NotNull Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction) {
+        return null;
+    }
```
ChomickiM, line 215: Do we want to implement these methods as well or should we refactor as needed in the future?

psantos1113: No we don't want to implement these.  I will add a comment to this section.

PR #2199, opened 2022-09-23
IALERT-3147: Make Azure Boards thread safe

psantos1113:
- Make the OAuth credential store thread safe by providing a lock for reads and updates
- Check if the OAuth fields have changed before updating in the database.  If no changes have occurred do not write and flush the same data to the database causes an error in JPA when flush is called due to no changes between the cache and the database.
- Make the custom field manager thread safe when installing custom fields by check if the custom field has been added to the workitemtype before attempting to add the custom field to the work item type.

### DIFF
--- a/channel-azure-boards/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/storage/AzureBoardsCredentialDataStore.java
+++ b/channel-azure-boards/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/storage/AzureBoardsCredentialDataStore.java
@@ -11,6 +11,8 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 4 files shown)


DanaMaxfield (approved):

channel-azure-boards/src/main/java/com/synopsys/integration/alert/channel/azure/boards/oauth/storage/AzureBoardsCredentialDataStore.java
```diff
@@ -47,48 +49,94 @@ protected AzureBoardsCredentialDataStore(AzureBoardsCredentialDataStoreFactory d
...
-                throw new IOException("Cannot update the Azure Boards global configuration", e);
+        lock.lock();
+        try {
+            if (null != key && null != value) {
+                boolean changed;
+                ConfigurationModel defaultConfig = retrieveConfiguration();
+                Map<String, ConfigurationFieldModel> keyToFieldMap = defaultConfig.getCopyOfKeyToFieldMap();
+                changed = isFieldValueChanged(keyToFieldMap, AzureBoardsDescriptor.KEY_OAUTH_USER_EMAIL, key);
+                changed = changed || isFieldValueChanged(keyToFieldMap, AzureBoardsDescriptor.KEY_ACCESS_TOKEN, value.getAccessToken());
+                changed = changed || isFieldValueChanged(keyToFieldMap, AzureBoardsDescriptor.KEY_REFRESH_TOKEN, value.getRefreshToken());
```
ChomickiM, line 99: Should this be one long logical expression rather than broken up, or was this for readability?

psantos1113: For readability and in case we need to debug an issue you know which field has changed.

ChomickiM (approved):

PR #2212, opened 2022-10-07
refactor: Update swarm to use same PG image as helm

DanaMaxfield:


docker/blackduck-alert-db/Dockerfile
```diff
@@ -10,9 +10,14 @@ LABEL com.blackducksoftware.integration.alert.vendor="Black Duck Software, Inc."
 
 ENV DOCKER_ENTRYPPOINT_INITD_DIR /docker-entrypoint-initdb.d
 
+USER root
```
psantos1113, line 13: We don't like to set the user because with openshift the customer can configure the user that runs the container and owns it.  So we don't want to run as the root user.  The recommendation we have had in the past is to not set the user in the container because customer's will want to configure the user due to security policies.

docker/blackduck-alert-db/Dockerfile
```diff
@@ -10,9 +10,14 @@ LABEL com.blackducksoftware.integration.alert.vendor="Black Duck Software, Inc."
...
 RUN mkdir -p -m 774 $DOCKER_ENTRYPPOINT_INITD_DIR
 
 COPY alertdb-healthcheck.sh /usr/local/bin/alertdb-healthcheck.sh
+COPY blackduck-alert-db-cmd.sh /usr/local/bin/blackduck-alert-db-cmd.sh
 COPY create_extension.sql $DOCKER_ENTRYPPOINT_INITD_DIR/create_extension.sql
 
-RUN chmod 774 /usr/local/bin/alertdb-healthcheck.sh
+RUN chmod 754 /usr/local/bin/alertdb-*.sh
+USER postgres
```
psantos1113, line 21: see the comment above.

PR #2226, opened 2022-10-24
IALERT-3183 Azure OAuth actions and callback

ChomickiM:
- Implementing the OAuth action in the AzureBoardsGlobalConfigController
- Creating a new AzureOAuthCallbackController and AzureBoardsOAuthCallbackAction to handle navigating back to Alert after performing OAuth handshake.
- Tests are still WIP
- The callback will test the connection and return you to the edit page for the configuration of the Azure job the authenticate was initiated from.


bseifert14 (approved):

channel-azure-boards/src/main/java/com/synopsys/integration/alert/channel/azure/boards/web/AzureOAuthCallbackControllerLegacy.java
```diff
@@ -0,0 +1,176 @@
...
+import com.synopsys.integration.azure.boards.common.service.project.AzureProjectService;
+import com.synopsys.integration.azure.boards.common.service.project.TeamProjectReferenceResponseModel;
+import com.synopsys.integration.rest.proxy.ProxyInfo;
+
+/**
+ * @deprecated This class is deprecated in favor of the AzureOAuthCallbackController. The RestController for this class has
+ * been disabled and is not an active endpoint. This class remains as reference for the previous authorization endpoint,  but
+ * should be removed in 8.0.0.
+ */
+@Deprecated(forRemoval = true)
```
psantos1113, line 53: I believe this class can be removed.  The REST API URI is the same the internal implementation has changed but the same REST API Endpoint is exposed.  That would not be a breaking change so in the future delete this class.

psantos1113 (approved):

PR #2227, opened 2022-10-24
IALERT_3149 - Support for PostgreSQL 14.X

DanaMaxfield:
**Swarm and Helm changes**
* Update PG version to latest 14
* Update Dockerfile to eliminate logic that is handled by Alert Docker image already
* Update Alert entrypoint to check connection for PG Admin user, and perform additional status checks for psql commands

**Helm only changes**
* Update to use same blackduck-alert-db container as Swarm
* Update for imageTag and registry
* Update environment variables and pathing to reflect different image settings
* Add liveness probe
* Update pullPolicy to be consistent across pods
* Update README.md to reflect changes above
* Updates for consistency and spacing

**Swarm only changes**
* Comment out defaults in docker-compose.local-overrides.yml as these are now in entry point script

**Changes specifically done for upgrading**
* Update build to handle version migrating from
* Update PG Dockerfile to copy in new scripts AND copy in Postgres binaries required for migration
* Add new alertdb-docker-entrypoint.sh script, which handles launching migration AND starting Postgres
* Add new alertdb-pg-migrate.sh, which handles migrating Postgres


psantos1113 (approved):
Will need to delete the file src/test/java/alert-example-db.dump.

PR #2230, opened 2022-10-26
IALERT-2954 AzureBoardsEnvironmentVariableHandler integration test

mau-synopsys:
TEST:
-Move JiraServerEnvironmentVariableHandlerTestIT to jire.server package
-Added AzureBoardsEnvironmentVariableHandlerTestIT

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/channel/azure/boards/environment/AzureBoardsEnvironmentVariableHandlerTestIT.java
+++ b/src/test/java/com/synopsys/integration/alert/channel/azure/boards/environment/AzureBoardsEnvironmentVariableHandlerTestIT.java
@@ -1,0 +1,87 @@
+package com.synopsys.integration.alert.channel.azure.boards.environment;
+
+import com.synopsys.integration.alert.api.common.model.AlertConstants;
+import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
+import com.synopsys.integration.alert.channel.azure.boards.database.accessor.AzureBoardsGlobalConfigAccessor;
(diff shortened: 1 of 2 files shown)


DanaMaxfield (approved):

src/test/java/com/synopsys/integration/alert/channel/azure/boards/environment/AzureBoardsEnvironmentVariableHandlerTestIT.java
```diff
@@ -0,0 +1,87 @@
...
+    @BeforeEach
+    @AfterEach
+    public void cleanup() {
+        azureBoardsGlobalConfigAccessor.getConfigurationByName(AlertRestConstants.DEFAULT_CONFIGURATION_NAME)
+            .map(Config::getId)
+            .map(UUID::fromString)
+            .ifPresent(azureBoardsGlobalConfigAccessor::deleteConfiguration);
+        mockEnvironment = new MockEnvironment();
+        environmentVariableUtility = new EnvironmentVariableUtility(mockEnvironment);
+    }
```
ChomickiM, line 47: We should perhaps split this into an init and cleanup methods rather than calling the same one.

psantos1113: agreed this should be two methods because the AfterEach the test creates an instance of MockEnvironment and EnvironmentVariableUtility initialized which isn't needed after each test.

psantos1113 (approved):

PR #2252, opened 2022-11-10
IALERT-3195 old rest endpoint update fix

ChomickiM:
Pushing changes to support update requests by the old field model converters. The change here is that we will now pass in an @Nullable existingGlobalConfig Id that whenever an update is passed in we can pass in a UUID. This is required by the validator to ignore duplicate names when the ID is the same, indicating that an update is being performed.

Test improvements will follow in a future PR.

### DIFF
--- a/channel-azure-boards/src/test/java/com/synopsys/integration/alert/channel/azure/boards/convert/AzureBoardsGlobalConfigurationModelConverterTest.java
+++ b/channel-azure-boards/src/test/java/com/synopsys/integration/alert/channel/azure/boards/convert/AzureBoardsGlobalConfigurationModelConverterTest.java
@@ -42,7 +42,7 @@ class AzureBoardsGlobalConfigurationModelConverterTest {
     void validConversionTest() {
         ConfigurationModel configurationModel = createDefaultConfigurationModel();
         AzureBoardsGlobalConfigurationModelConverter converter = new AzureBoardsGlobalConfigurationModelConverter(validator);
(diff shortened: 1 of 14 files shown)


channel-azure-boards/src/test/java/com/synopsys/integration/alert/channel/azure/boards/convert/AzureBoardsGlobalConfigurationModelConverterTest.java
```diff
@@ -42,7 +42,7 @@ public void init() {
     void validConversionTest() {
         ConfigurationModel configurationModel = createDefaultConfigurationModel();
         AzureBoardsGlobalConfigurationModelConverter converter = new AzureBoardsGlobalConfigurationModelConverter(validator);
-        Optional<AzureBoardsGlobalConfigModel> model = converter.convertAndValidate(configurationModel);
+        Optional<AzureBoardsGlobalConfigModel> model = converter.convertAndValidate(configurationModel, null);
```
psantos1113, line 45: We seem to be missing a test where the configuration id isn't null for an existing configuration.

psantos1113: This can be added after this is merged in. It's an additional test case that didn't exist already.  Will approve for now

PR #2271, opened 2022-11-18
AzureBoards string limit test failures

mau-synopsys:
Use AzureBoardsSearchPropertiesUtil max_string_value_length as source of truth


DanaMaxfield (approved):
This is great @mau-synopsys, thank you for looking into this!

PR #2276, opened 2022-11-29
DistributionRepostory IT tests for policy and vulnerability filtering

mau-synopsys:
Test cases for the following scenarios:

Policy match, Sev match -> returns matches
Policy match, Sev empty -> returns matches
Policy empty, Sev match -> returns matches
Policy empty, Sev empty -> returns matches (AKA no filtering)

Policy NOT match, Sev empty -> returns none
Policy empty, Sev NOT match -> returns none
Policy NOT match, Sev NOT match -> returns none

### DIFF
--- a/src/test/java/com/synopsys/integration/alert/database/job/DistributionJobRepositoryTestIT.java
+++ b/src/test/java/com/synopsys/integration/alert/database/job/DistributionJobRepositoryTestIT.java
@@ -1,0 +1,336 @@
+package com.synopsys.integration.alert.database.job;
+
+import com.synopsys.integration.alert.common.enumeration.FrequencyType;
+import com.synopsys.integration.alert.common.enumeration.ProcessingType;
+import com.synopsys.integration.alert.common.persistence.model.job.DistributionJobModel;
+import com.synopsys.integration.alert.common.persistence.model.job.DistributionJobRequestModel;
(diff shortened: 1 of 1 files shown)


src/test/java/com/synopsys/integration/alert/database/job/DistributionJobRepositoryTestIT.java
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    @Test
+    @Transactional
+    void findAndSortReturnsNotificationsOnNoFilters() {
+        long providerAndBDConfigId = 1L;  // This is also notification.providerConfigId
+
+        // Create a JobsDetailEntity and the DistributionJobEntity
+        FrequencyType frequency = FrequencyType.REAL_TIME;
+        EmailJobDetailsModel emailJobDetailsModel = createEmailJobDetails();
```
ChomickiM, line 53: If we can we should try to  encapsulate some of the duplicate code we find to try and cut down on the overall class size. Some of these values can either be made as private class members or just passed directly into createDistributionJobRequestModel, ex. we don't need to save an object of Frequency.REAL_TIME, we can just use  the enum directly.

ChomickiM (approved):

psantos1113 (approved):

ChomickiM (approved):

PR #2282, opened 2022-12-05
IALERT-3138 Fix for Azure boards authentication wrong port number

mau-synopsys:
Users will need to add **PUBLIC_ALERT_WEBSERVER_PORT=443** to their docker-compose.local-overrides.yml in docker-swarm and in values.yml for helm

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/AlertProperties.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/AlertProperties.java
@@ -12,16 +12,12 @@ import java.util.Optional;
 
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.web.util.UriComponentsBuilder;
 
 import com.synopsys.integration.alert.api.common.model.exception.AlertException;
 
 public class AlertProperties {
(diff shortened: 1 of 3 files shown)


psantos1113 (approved):

DanaMaxfield (commented):
`Users will need to add PUBLIC_ALERT_WEBSERVER_PORT=443 to their docker-compose.local-overrides.yml in docker-swarm and in values.yml for helm`

For values.yaml, don't we need default values in one of the templates in order to use this new setting?
For docker-compose.local-overrides.yml, we normally have a commented out example. Are we not doing this here?

psantos1113:
For the values.yaml there doesn't need to be an example because the exposedNodePort in values.yaml is already used in the helm chart to set the PUBLIC_ALERT_WEBSERVER_PORT environment variable.  An example can be added to the docker-compose.local-overrides.yml file

mau-synopsys:
> 



> `Users will need to add PUBLIC_ALERT_WEBSERVER_PORT=443 to their docker-compose.local-overrides.yml in docker-swarm and in values.yml for helm`
> 
> For values.yaml, don't we need default values in one of the templates in order to use this new setting? For docker-compose.local-overrides.yml, we normally have a commented out example. Are we not doing this here?

Im not exactly sure but @psantos1113 mentioned this is already done in helm. The value to set is exposedNodePort I believe. This is missing in docker-swarm

mau-synopsys:
Updated to leave a field for users to uncomment for setting PUBLIC_ALERT_WEBSERVER_PORT

DanaMaxfield (approved):

PR #2298, opened 2022-12-14
Create the saml config model

mau-synopsys:


component-saml/src/main/java/com/synopsys/integration/alert/component/saml/SAMLGlobalConfigModel.java
```diff
@@ -0,0 +1,69 @@
...
+        this.entityId = entityId;
+        this.entityBaseUrl = entityBaseUrl;
+        this.requireAssertionsSigned = requireAssertionsSigned;
+        this.roleAttributeMapping = roleAttributeMapping;
+
+        setCreatedAt(createdAt);
+        setLastUpdated(lastUpdated);
+    }
+
+    public Optional<Boolean> getEnabled() { return Optional.ofNullable(enabled); }
```
ChomickiM, line 54: Minor thing, I believe our code style for the public accessors is to still put them across multiple lines even if theres only a single line return. Typically the IDE will collapse them into a single line that can be expanded.

ChomickiM (approved):

psantos1113 (approved):

PR #2299, opened 2022-12-15
IALERT-3258 - Implement ConfigModel for LDAP

DanaMaxfield:


component-ldap/src/main/java/com/synopsys/integration/alert/channel/ldap/model/LDAPGlobalConfigModel.java
```diff
@@ -0,0 +1,156 @@
...
+    private String groupRoleAttribute;
+
+    public LDAPGlobalConfigModel() {
+        // For serialization
+    }
+
+    // Required
+    public LDAPGlobalConfigModel(String id) {
+        super(id, AlertRestConstants.DEFAULT_CONFIGURATION_NAME);
+    }
```
ChomickiM, line 34: 1) We can drop the required comment as it could probably be a bit misleading between this and the constructor with ALL properties passed in.

2) We should expand this constructor to cover the required fields needed for LDAP and the getters for  non-required fields should be wrapped in Optional tags below

psantos1113 (approved):

ChomickiM (approved):

PR #2300, opened 2022-12-16
IALERT-2946 SAML controller

mau-synopsys:
Depends on SAML config model PR


authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/web/SAMLGlobalConfigController.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public SAMLGlobalConfigModel create(SAMLGlobalConfigModel resource) {
+        return new SAMLGlobalConfigModel(UUID.randomUUID().toString());
+    }
+
+    @Override
+    public SAMLGlobalConfigModel getOne() {
+        return new SAMLGlobalConfigModel(UUID.randomUUID().toString());
```
psantos1113, line 31: Why not create a static member variable for the UUID

`private static final UUID DEFAULT_ID = UUID.randomUUID();`
Then just use the static in the create and get one methods since this API should always use the same UUID.  Once this uses data from the database then the static member variable can be removed.

ChomickiM:
There looks to be overlap between this PR and the previous one with  the model: https://github.com/blackducksoftware/blackduck-alert/pull/2298. We should likely merge that first and then merge this one after.

authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/web/SAMLGlobalConfigController.java
mau-synopsys: Updated

mau-synopsys:
> There looks to be overlap between this PR and the previous one with the model: #2298. We should likely merge that first and then merge this one after.

Merged down from the model PR so the changes should reflect here

ChomickiM (approved):

PR #2306, opened 2023-01-05
IALART-3280 saml crud and validator actions 

mau-synopsys:
- Hooks crud, validator actions and validator to controller and accessor
- Resolve dependencies

### DIFF
--- a/authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/AuthenticationSAMLConfigurationValidator.java
+++ b/authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/AuthenticationSAMLConfigurationValidator.java
@@ -1,0 +1,28 @@
+package com.synopsys.integration.alert.authentication.saml.validator;
+
+import com.synopsys.integration.alert.api.common.model.ValidationResponseModel;
+import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatus;
+import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatusMessages;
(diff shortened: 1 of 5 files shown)


authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/AuthenticationSAMLConfigurationValidator.java
```diff
@@ -0,0 +1,28 @@
...
+import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatusMessages;
+import com.synopsys.integration.alert.authentication.saml.model.AuthenticationSAMLConfigModel;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Component;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@Component
+public class AuthenticationSAMLConfigurationValidator {
```
ChomickiM, line 14: The name validation here, since this channel is unique, will always return a success.

Likely this class needs additional validation if it's enabled. If we still haven't identified what does and does not need to be validated, we should then add a TODO here that validations still need to be implemented.

mau-synopsys: Added a TODO comment for figure out what other validation we need. This could be additional required fields or validation on new feature fields.

ChomickiM (approved):

psantos1113 (approved):

DanaMaxfield (approved):

PR #2310, opened 2023-01-09
IALERT-3263 saml env var handler and validation changes

mau-synopsys:
-Make entity id, base url, and one of metadata pointers required
-Perform validation for above fields if saml is enabled 
-Add needed setters for config model
-Read, set config model from env variables, and build processing result


authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/SAMLConfigurationValidator.java
```diff
@@ -0,0 +1,73 @@
...
+            } else {
+                addErrorStatusIfInvalidUrl(model.getEntityBaseUrl(), "entityBaseUrl", statuses);
+            }
+
+            // Convert to empty optional for blank metadata url and file path
+            Optional<String> optionalFilteredMetadataUrl = model.getMetadataUrl().filter(StringUtils::isNotBlank);
+            Optional<String> optionalFilteredMetadataFilePath = model.getMetadataFilePath().filter(StringUtils::isNotBlank);
+            // One of url or filepath must exist
+            if (optionalFilteredMetadataUrl.isEmpty() && optionalFilteredMetadataFilePath.isEmpty()) {  // TODO: verify message
+                statuses.add(AlertFieldStatus.error("metadataUrl, metadataFilePath", "One of either fields is required."));
```
ChomickiM, line 38: I would change the message to be something more along the lines of "Either the metadata URL or metadata file path must be provided."

authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/SAMLConfigurationValidator.java
```diff
@@ -0,0 +1,73 @@
...
+            } catch (MalformedURLException e) {
+                statuses.add(AlertFieldStatus.error(fieldName, e.getMessage()));
+            }
+        }
+    }
+
+    // Could make this more generic with param field name but lint warns since it is only one value here
+    private void addErrorStatusIfInvalidMetadataFilePath(String filePath, Set<AlertFieldStatus> statuses) {
+        File file = new File(filePath);
+        if (!file.isFile()) {   // TODO: Verify if we're using AuthenticationConfigurationFieldModelValidator and FilePersistenceUtil.uploadFileExists
```
psantos1113, line 69: Use FilePersistenceUtil because the file will be uploaded to a specific location.  The test should validate that a file exists at the upload path.

authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/SAMLConfigurationValidator.java
mau-synopsys: Updated to use messages old saml validation used

authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/SAMLConfigurationValidator.java
mau-synopsys: Updated to use FilePersistenceUtil

DanaMaxfield (approved):

PR #2312, opened 2023-01-09
IALERT-3158:  Have UI use new REST API for failed audit entries

psantos1113:
This PR makes slight changes to the Audit Page to display only failed audit entries.  It will utilize the new database tables created in a previous pull request.

Changes:
- Make request to api/audit/failed
- Remove the show only sent notifications checkbox
- Update the audit page description


component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActions.java
```diff
@@ -1,167 +1,44 @@
-/*
```
psantos1113, line 1: This class was renamed.  The code that is removed here is the old code.  The old accessor is named AuditEntryActionsLegacy.

component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryActionsLegacy.java
```diff
@@ -0,0 +1,167 @@
+/*
```
psantos1113, line 1: This is the original actions class  AuditEntryActions.  There were no changes other than renaming the class.

component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryController.java
```diff
@@ -1,79 +1,42 @@
-/*
```
psantos1113, line 1: Renamed the AuditEntryController to AuditEntryControllerLegacy.  The contents of this class are the new Rest API endpoint.

component/src/main/java/com/synopsys/integration/alert/component/audit/web/AuditEntryControllerLegacy.java
```diff
@@ -0,0 +1,86 @@
+/*
```
psantos1113, line 1: This is the old controller AuditEntryController.  The only change in this file is to rename the class.

ui/src/main/js/store/actions/audit.js
```diff
@@ -3,7 +3,7 @@ import * as HTTPErrorUtils from 'common/util/httpErrorUtilities';
 import { unauthorized } from 'store/actions/session';
 import HeaderUtilities from 'common/util/HeaderUtilities';
 
-const FETCH_URL = '/alert/api/audit';
+const FETCH_URL = '/alert/api/audit/failed';
```
bseifert14, line 6: If this is for a failing request, we should label the variable to be more specific.  Something like FETCH_URL_FAIL or FETCH_URL_ERROR

psantos1113: Updated

bseifert14 (approved):

PR #2329, opened 2023-01-20
IALERT-3309 saml certs upload and validation

mau-synopsys:
- require_assertions_signed back to want_assertions_signed
- Remove enable check before validation of config model
- Endpoints for certs (encryption, signing, verification) file exists, upload
- Validation for cert files
- Code clean up to simplify some upload and verification calls
- Optional fields to nullable=true in liquibase files
- Added new fields to specify file path of cert files


authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/SAMLFileUploadValidator.java
```diff
@@ -17,10 +18,9 @@
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
DanaMaxfield, line 21: I believe we should be avoiding import * from a package.

DanaMaxfield (changes requested):

authentication-saml/src/main/java/com/synopsys/integration/alert/authentication/saml/validator/SAMLFileUploadValidator.java
mau-synopsys: Addressed!

DanaMaxfield (approved):

alert-database/src/main/resources/liquibase/6.13.0/concrete-saml-configuration-table.xml
```diff
@@ -21,16 +21,16 @@
                 <constraints nullable="true"/>
             </column>
             <column name="enabled" type="BOOLEAN" defaultValueBoolean="false">
-                <constraints nullable="false"/>
```
psantos1113, line 24: Why are all these fields now nullable?  If they are required fields for saml then the field shouldn't be nullable.

mau-synopsys: These are the fields we discussed are required. 

Metadata url OR file path required
entityBaseUrl
entityid

I previously had everything set to false so corrected the non-required ones to false here

ChomickiM (approved):

alert-database/src/main/resources/liquibase/6.13.0/concrete-saml-configuration-table.xml
psantos1113: Enabled is required as well in the context of the database.  That cannot be null.

mau-synopsys: Changed enabled, force_auth, and want_assertions_signed to nullable false

psantos1113 (approved):

PR #2346, opened 2023-02-03
IAlert-3155 accumulator fix

psantos1113:
Each notification has an href that links to the individual notification.  Use the hash of the URL to uniquely identify the notification.  If there is a duplicate hash then the notification has already been read.this prevents overlapping with the last notification search time to avoid pulling in duplicate notifications when the time window grows for pulling notifications.

- Use a SHA3-256 algorithm which is supported in Java 9+.  
- The algorithm may be accelerated by the OS and hardware from the documentation read.
- Adjust the search timestamp to avoid adding a whole second.  Rather add 1000 nanoseconds to ensure the search range won't skip notifications.
- Added a new content_id column to the raw_notification_content table
-  The majority of the changes were to the tests to update the constructor of AlertNotificationModel to include the new contentId member variable.


ChomickiM (approved):

ChomickiM:
The content ID in the notification entity will only ever be a SHA or a UUID. This may be a trivial detail to ignore, but tests are appending  "content-id-" for the String we use. Should we stick to only SHAs and UUIDs or do we not care for the purposes of unit tests?

mau-synopsys (approved):

PR #2357, opened 2023-02-10
IALERT-3041 - Don't validate model if no env vars exist

DanaMaxfield:
On startup, we run the environment variable handler, which runs configureModel() and validateConfiguration(). If the required fields for the model don't exist, we put error messages in the log which can be misleading to customers.

FIX:
Prior to running configureModel() and validateConfiguration(), check if the environment contains environment variables that pertain to the model being processed. If none are found, do not configure or validate model.


api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandler.java
```diff
@@ -34,9 +36,19 @@ public Set<String> getVariableNames() {
         return environmentVariableNames;
     }
 
+    private boolean doVariablesExistInEnvironment() {
+        for (String variableName : environmentVariableNames) {
+            if (environmentVariableUtility.getEnvironmentValue(variableName).isPresent()) {
+                return true;
+            }
+        }
```
psantos1113, line 44:
```suggestion
        boolean variablePresent = environmentVariableNames
            .stream()
            .map(environmentVariableUtility::getEnvironmentValue)
            .anyMatch(Optional::isPresent);
```

api-environment/src/main/java/com/synopsys/integration/alert/environment/EnvironmentVariableHandler.java
```diff
@@ -34,9 +36,19 @@ public Set<String> getVariableNames() {
...
+        logger.info("Did not find any environment variables configured for: {}", name);
+        return false;
```
psantos1113, line 46:
```suggestion
        if(!variablePresent) {
            logger.info("Did not find any environment variables configured for: {}", name);
        }
        return variablePresent;
```

psantos1113 (approved):

ChomickiM (approved):

PR #2359, opened 2023-02-13
IALERT-3155: accumulator timing fix

psantos1113:
- Log details about notifications that already exist due to overlaps in search times for notifications.
- Subtract 1 millisecond from search time to ensure grabbing all notifications from Black Duck.
- Add feature to diagnostics to display the counts of notifications by provider and notification type.
- Change the performance wait job to wait for the number of notifications that were written to the database.


component/build.gradle
```diff
@@ -32,6 +32,8 @@ dependencies {
     implementation 'org.springframework.amqp:spring-amqp'
     implementation 'org.springframework.amqp:spring-rabbit'
 
+    implementation 'com.synopsys.integration:blackduck-common', rootProject.ext.blackduckCommonExcludes
```
DanaMaxfield, line 35: `rootProject.ext.blackduckCommonExcludes` was removed in a PR at the end of last week. The exclude is now applied to all subprojects automatically. This should be removed.

component/src/main/java/com/synopsys/integration/alert/component/diagnostic/database/DefaultDiagnosticAccessor.java
```diff
@@ -101,7 +106,34 @@ private NotificationDiagnosticModel getNotificationDiagnosticInfo() {
...
         long numberOfNotificationsProcessed = notificationContentRepository.countByProcessed(true);
         long numberOfNotificationsUnprocessed = notificationContentRepository.countByProcessed(false);
-        return new NotificationDiagnosticModel(numberOfNotifications, numberOfNotificationsProcessed, numberOfNotificationsUnprocessed);
+        List<ProviderNotificationCounts> providerNotificationCounts = getProviderNotificationCounts();
+        return new NotificationDiagnosticModel(numberOfNotifications, numberOfNotificationsProcessed, numberOfNotificationsUnprocessed, providerNotificationCounts);
+    }
+
+    private List<ProviderNotificationCounts> getProviderNotificationCounts() {
+        Set<Long> providerConfigIds = new LinkedHashSet<>();
+        int pageSize = 100;
```
DanaMaxfield, line 115: In searching the code, I see a mix of page sizes being set as 100 or 200. Any specific reason to use 100 here vs 200?

component/src/main/java/com/synopsys/integration/alert/component/diagnostic/model/NotificationTypeCount.java
```diff
@@ -0,0 +1,23 @@
+package com.synopsys.integration.alert.component.diagnostic.model;
+
+import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
+
+public class NotificationTypeCount extends AlertSerializableModel {
+
+    private static final long serialVersionUID = -7022752228360734890L;
+    private final String notificationType;
```
ChomickiM, line 8: Should NotificationTypeCount use the NotificationType enum rather than a String?

ChomickiM: To add to this, it appears the only place we use this is in the wait task where we use the NotificationType::name to compare.

component/src/main/java/com/synopsys/integration/alert/component/diagnostic/database/DefaultDiagnosticAccessor.java
psantos1113: the size of 200 was updated for processing notifications.  Most other places it's still 100.  I will create a ticket to update this to be consistent across the board in Alert.

ChomickiM (approved):

component/src/main/java/com/synopsys/integration/alert/component/diagnostic/model/NotificationTypeCount.java
psantos1113: made the change

PR #2365, opened 2023-02-14
Ps job table refactor

psantos1113:
- Rename job_execution_status tables to job_completion_status tables.
- Rename Classes related to JobExecutionStatus JobCompletionStatus
- Rename Classes related to JobExecutionDuration to JobCompletionDuration


alert-database/src/main/resources/liquibase/6.13.0/job-execution-status.xml
```diff
@@ -4,10 +4,10 @@
     <changeSet author="psantos" id="create-job-execution-status-table">
```
ChomickiM, line 4: likely a very minor detail, but this ID could be amended to be "create-job-completion-status-table"
```suggestion
    <changeSet author="psantos" id="create-job-completion-status-table">
```

ChomickiM (approved):

PR #2385, opened 2023-02-24
IALERT-3346 - Implement validation for LDAPConfigTestModel

DanaMaxfield:
* Perform validation on the LDAPConfigTestModel prior to executing the test connection. This checks combines errors found while validating LDAPConfigModel with any while found validating LDAPConfigTestModel
* Implement JUnit test coverage for new logic

UI changes by @bseifert14 
* In pop-up when prompting for testLDAPUsername and testLDAPPassword, change name of button from "Send Test Message" -> "Test User Connection".
* On same pop-up, prevent user from clicking Test User Connection button until testLDAPUsername and testLDAPPassword are filled in.

NOTE: The UI does not send in a null if no data is supplied for the LDAP configuration. Instead, it sends a valid LDAPConfigModel, with no values. So, the null check is really only for when users hit the API directly, such as when using postman.

### DIFF
--- a/authentication-ldap/src/main/java/com/synopsys/integration/alert/authentication/ldap/validator/LDAPConfigurationValidator.java
+++ b/authentication-ldap/src/main/java/com/synopsys/integration/alert/authentication/ldap/validator/LDAPConfigurationValidator.java
@@ -10,6 +10,7 @@ import com.synopsys.integration.alert.api.common.model.ValidationResponseModel;
 import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatus;
 import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatusMessages;
 import com.synopsys.integration.alert.authentication.ldap.model.LDAPConfigModel;
+import com.synopsys.integration.alert.authentication.ldap.model.LDAPConfigTestModel;
 
 @Component
 public class LDAPConfigurationValidator {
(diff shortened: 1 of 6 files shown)


authentication-ldap/src/main/java/com/synopsys/integration/alert/authentication/ldap/validator/LDAPConfigurationValidator.java
```diff
@@ -35,4 +36,29 @@ public ValidationResponseModel validate(LDAPConfigModel ldapConfigModel) {
 
         return ValidationResponseModel.success();
     }
+
+    public ValidationResponseModel validate(LDAPConfigTestModel ldapConfigTestModel) {
+        Set<AlertFieldStatus> statuses = new HashSet<>();
+        LDAPConfigModel ldapConfigModel = ldapConfigTestModel.getLdapConfigModel();
+
+        if (null == ldapConfigModel) {
+            statuses.add(AlertFieldStatus.error("ldapConfigModel", AlertFieldStatusMessages.REQUIRED_FIELD_MISSING));
```
ChomickiM, line 45: Does an "ldapConfigModel" key exist in the UI? I see that we have fields for "testLDAPUsername" and and "testLDAPPassworrd" in AuthenticationModel.js but nothing for the model itself.

DanaMaxfield: When the test button is used, the UI passes in an LDAPConfigTestModel, which contains an LDAPConfigModel.

ChomickiM: When we create an AlertFieldStatus we're associating some error/warning message to some field in the UI. In this case the field is "ldapConfigModel". Typically its a single unit of information thats passed in a text or number field (think username/password/port/etc.).  When an error occurs we can add an AlertFieldStatus to show a red error message. This one is a bit more complex since it refers to the entire model.

In this case, I'm unable to find a field in the UI with the tag "ldapConfigModel" so I'm not sure if we post any kind of field error using it.

ChomickiM (approved):

psantos1113 (approved):

PR #2418, opened 2023-03-15
IALERT-3337: Policy filtering fix

ChomickiM:
Policy notification, if they happen against the same component, can contain multiple policies in one notification. This is filtered during mapping but as a result of message extraction further down the line these policies are re-introduced as messages. This fix is it to perform an additional layer of filtering during message extraction by looking at the job details.


DanaMaxfield (commented):
Will you be adding tests for this specific defect?

psantos1113 (approved):
A ticket has been filed in the bug tracking system the team uses to add tests around this fix.  The tests are more involved in this area of the code due to the mocking needed.  Plan is to address the issue test and in the next version cover off adding the additional tests.

PR #2441, opened 2023-04-11
IALERT-3437: Jira Server model updates

ChomickiM:
The purpose of this PR is to implement three new fields into JiraServerGlobalConfigModel in order to support Bearer authorization in the future. To support this we've added the JiraServerAuthorizationMethod enum, accessToken, and isAccessToken set fields into the model.

The remainder of the changes are refactoring the implementation of this model and updating broken tests.


channel-jira-server/src/main/java/com/synopsys/integration/alert/channel/jira/server/validator/JiraServerGlobalConfigurationValidator.java
```diff
@@ -50,7 +51,7 @@ public ValidationResponseModel validate(JiraServerGlobalConfigModel model, Strin
                 statuses.add(AlertFieldStatus.error("url", e.getMessage()));
             }
         }
-        if (StringUtils.isBlank(model.getUserName())) {
+        if (model.getUserName().isEmpty() || StringUtils.isBlank(model.getUserName().get())) {
```
mau-synopsys, line 54: I think here we should check if model.getAuthorizationMethod() is BASIC first so it doesnt fail if it's set to PAT.

mau-synopsys (approved):

PR #2446, opened 2023-04-20
IALERT-3438 jira server liquibase

mau-synopsys:
configuration_jira_server table:
- Create authorization_method column with defaultValue set to 'BASIC'
- Drop not nullable constraint on username and password


DanaMaxfield (approved):

ChomickiM (approved):

mau-synopsys:
Updated with access token column

channel-jira-server/src/test/java/com/synopsys/integration/alert/channel/jira/server/database/accessor/JiraServerGlobalConfigAccessorTest.java
```diff
@@ -36,6 +36,7 @@ class JiraServerGlobalConfigAccessorTest {
     static final String TEST_URL = "url";
     static final String TEST_USERNAME = "username";
     static final String TEST_PASSWORD = "password";
+    static final String TEST_ACCESS_TOKEN = "access_token";
```
ChomickiM, line 39: These tests may need a refactor once we've made the changes on the Alert side. We should test for username/password OR access token. In the implementations below we are using both. 

There is the potential of testing an entity that contains both user/pass and access token but I don't believe that should be the standard test scenario.

ChomickiM (approved):

channel-jira-server/src/test/java/com/synopsys/integration/alert/channel/jira/server/database/accessor/JiraServerGlobalConfigAccessorTest.java
mau-synopsys: In SAML we store information for both and the mode determines which one to use. That way if the user enters info for one mode and switches to another, the previous mode info is stored. This is assuming we have the radio button for auth method. @bseifert14 will it be the same for jira server?

PR #2455, opened 2023-05-10
IALERT-3427 keystore and 6.13.3

mau-synopsys:


alert-platform/build.gradle
```diff
@@ -16,10 +16,10 @@ dependencies {
         api 'com.sun.mail:javax.mail:1.6.2'
         api 'javax.mail:javax.mail-api:1.6.2'
 
-        api 'com.synopsys.integration:integration-rest:10.3.6'
+        api 'com.synopsys.integration:integration-rest:10.3.9-SNAPSHOT'
         api 'com.synopsys.integration:integration-common:26.0.4'
         api 'com.synopsys.integration:blackduck-common:66.0.2'
-        api 'com.synopsys.integration:int-jira-common:3.0.7'
+        api 'com.synopsys.integration:int-jira-common:3.0.8-SNAPSHOT'
```
ChomickiM, line 22: We shouldn't merge in the SNAPSHOT versions, I think these should be updated with the released versions of the libraries once the releases go out.

docker/blackduck-alert/Dockerfile
```diff
@@ -17,6 +17,7 @@ ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk
 ENV PATH $JAVA_HOME/bin:$ALERT_TAR_HOME/bin:$PATH
 ENV CERTIFICATE_MANAGER_DIR $ALERT_HOME/bin
 ENV SECURITY_DIR $ALERT_HOME/security
+ENV BLACKDUCK_ALERT_OPTS -Djavax.net.ssl.keyStore=$SECURITY_DIR/blackduck-alert.keystore -Djavax.net.ssl.keyStoreType=PKCS12 -Djavax.net.ssl.keyStorePassword=changeit
```
psantos1113, line 20: You don't want to hard code the password here.  You will need to evaluate an environment variable because the password can be set via environment variable.
Something like this:
`-Djavax.net.ssl.keyStorePassword=changeit${ALERT_KEY_STORE_PASSWORD:-changeit}`

psantos1113 (approved):

ChomickiM (approved):

DanaMaxfield (approved):

PR #2466, opened 2023-06-09
IALERT-3480: Direct user to Azure Board Edit Modal after OAuth Handshake

bseifert14:
A BIG thank you to @mau-synopsys on helping solve this one.  @mau-synopsys laid the groundwork, all that I did was add a few tweaks to re-generate the data.
  
### Issue At Hand  
When a user would attempt to Save/Authenticate an Azure Board Modal, they were redirected to a blank screen.  This is caused by a change in our user flow.  Initially, we had a page serving as a form where the user would add in Azure Board information.  They could then Save, Test, or Authenticate (via OAuth) the data.  Since we now handle this form inside of a modal, the page with the form is no longer relevant.  
  
### Testing  
To test this, attempt to authenticate an Azure Board model via OAuth and notice that upon successful authentication the modal will reopen with the relevant data still populated in the form fields.

### DIFF
--- a/ui/src/main/js/page/channel/azure/AzureBoardTable.js
+++ b/ui/src/main/js/page/channel/azure/AzureBoardTable.js
@@ -41,7 +41,7 @@ const emptyTableConfig = {
     message: 'There are no records to display for this table.  Please create an Azure Board connection to use this table.'
 };
 
-const AzureBoardTale = ({ readonly, allowDelete }) => {
+const AzureBoardTable = ({ readonly, allowDelete, location }) => {
     const dispatch = useDispatch();
     const { data } = useSelector((state) => state.azure);
     const refreshStatus = JSON.parse(window.localStorage.getItem('AZURE_BOARD_REFRESH_STATUS'));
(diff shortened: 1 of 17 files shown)


psantos1113 (approved):

ui/src/main/js/page/channel/azure/AzureBoardTable.js
```diff
@@ -57,6 +57,30 @@ const AzureBoardTale = ({ readonly, allowDelete }) => {
...
+        if (parsedUrlArray.includes('edit')) {
+            // obtain the id of the azure board that OAuth just authenticated
+            const modalDataID = parsedUrlArray.slice(-1)[0];
+            // filter the table data and set the data for the modal to the one that matches the id from the line above
+            data.models.forEach(model => {
+                if (model.id === modalDataID) {
+                    setModalData(model);
+                    setShowModal(true);
+                }
+            })
```
ChomickiM, line 81: Does the "data" contain the page of azure boards configurations or ALL azure boards configurations?

ui/src/main/js/page/channel/azure/AzureBoardsPageLayout.js
```diff
@@ -1,19 +1,24 @@
...
+    const location = useLocation();
+
+    return (
+        <div>
+            <PageHeader
+                title={AZURE_BOARDS_INFO.label}
+                description="Configure the Azure Boards instance that Alert will send issue updates to."
+                icon={['fab', 'windows']}
+            />
+            <AzureBoardTable readonly={readonly} allowDelete={allowDelete} location={location} params={params} />
```
mau-synopsys, line 18: params not needed (probably an artifact from the branch i was using?)

ui/src/main/js/page/channel/azure/AzureBoardTable.js
```diff
@@ -41,7 +41,7 @@ const emptyTableConfig = {
     message: 'There are no records to display for this table.  Please create an Azure Board connection to use this table.'
 };
 
-const AzureBoardTale = ({ readonly, allowDelete }) => {
+const AzureBoardTable = ({ readonly, allowDelete, location }) => {
```
mau-synopsys, line 44: Can location be directly retrieved in this component via useLocation? Meaning moving useLocation from AzureBoardsLayout to AzureBoardsTable

mau-synopsys (changes requested):

ui/src/main/js/page/channel/azure/AzureBoardTable.js
bseifert14: `data` here is all of the azure boards configuration that the store has.  What this is doing is it's taking the id out of the url (hacky method but it's our only option), then sifting through the data in the store to find a match on the id.  If it finds that match, it sets the data intended to be displayed in this modal (`modalData`).

ui/src/main/js/page/channel/azure/AzureBoardsPageLayout.js
bseifert14: Great catch, that was remnants of the branch you were using. I'll resolve this accordingly.

mau-synopsys (approved):

PR #2467, opened 2023-06-09
IALERT-3485: Delete Selected Rows of Data

bseifert14:
### Bug  
When a user went to delete users, roles or providers they were prompted with a modal that allowed them to unselect some of the selected data for deletion.  A bug occured where even if a user unselected data, that data would still be deleted.  The fix for this was to provide checks for if the model had the staged flag set to true.   If true, this means the user still has that given row of data selected for deletion.

### DIFF
--- a/ui/src/main/js/common/component/descriptor/DescriptorRoute.js
+++ b/ui/src/main/js/common/component/descriptor/DescriptorRoute.js
@@ -15,7 +15,7 @@ const DescriptorRoute = ({
             <Route
                 exact
                 key={urlName}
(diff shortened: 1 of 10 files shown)


mau-synopsys (approved):

ChomickiM (approved):

bseifert14:
Unsure how this first commit (`89ff1d42bf37ca06fba5a857a80baf551c820a79`) made it into another bug branch but it did.  Closing this.  It was approved and ready to be merged.

PR #2477, opened 2023-06-16
IALERT-3476: Plugin error logging fix

ChomickiM:
When returning a message from the Jira Install Plugin endpoint we would print only the general message but miss any additional information from the log message as well as the exception. This PR covers the back end fix as well as redux changes to support both message and errors fields.

### DIFF
--- a/ui/src/main/js/store/actions/jira-server.js
+++ b/ui/src/main/js/store/actions/jira-server.js
@@ -137,9 +137,11 @@ function sendJiraServerPluginSuccess(message) {
     };
 }
 
-function sendJiraServerPluginError() {
+function sendJiraServerPluginError(message, errors) {
     return {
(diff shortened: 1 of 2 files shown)


mau-synopsys (approved):

ui/src/main/js/store/actions/jira-server.js
```diff
@@ -314,7 +316,7 @@ export function installJiraServerPlugin(jiraServerModel) {
                     if (response.ok) {
                         dispatch(sendJiraServerPluginSuccess(responseData.message));
                     } else {
-                        const defaultHandler = () => sendJiraServerPluginError(responseData);
+                        const defaultHandler = () => sendJiraServerPluginError(responseData.message, responseData.error);
```
bseifert14, line 319: Nice find!

bseifert14 (approved):

PR #2479, opened 2023-06-20
[UI] IALERT-3495 Add refresh button to tables

mau-synopsys:
Also disable refresh button for 1 sec after click as a visual indicator for user and to limit click rate

### DIFF
--- a/ui/src/main/js/page/channel/jira/server/JiraServerTableActions.js
+++ b/ui/src/main/js/page/channel/jira/server/JiraServerTableActions.js
@@ -4,7 +4,7 @@ import StatusMessage from 'common/component/StatusMessage';
 import JiraServerModal from 'page/channel/jira/server/JiraServerModal';
 import JiraServerDeleteModal from 'page/channel/jira/server/JiraServerDeleteModal';
 import Button from 'common/component/button/Button';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { fetchJiraServer } from 'store/actions/jira-server';
 
 const JiraServerTableActions = ({ data, readonly, allowDelete, selected, setSelected }) => {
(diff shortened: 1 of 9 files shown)


ui/src/main/js/page/channel/jira/server/JiraServerTableActions.js
```diff
@@ -25,9 +29,16 @@ const JiraServerTableActions = ({ data, readonly, allowDelete, selected, setSele
         setStatusMessage();
         setShowDeleteModal(true);
     }
+
+    function handleRefresh() {
+        setRefreshDisabled(true);
+        setTimeout(() => setRefreshDisabled(false), 1000);
```
bseifert14, line 35: Is there any reason to setRefreshDisabled on a chron here for 1 second?

bseifert14: I think what this is trying to accomplish is to disable that button while the request processes.  A more concrete way would be to leverage the 'fetching' flag we have in the redux store.  What this does is set to true when a request is processing and false when the request has completed or errored out.  For example you'd add in this line on L22:
```
const { fetching } = useSelector((state) => state.jiraServer);
```  
  
Then you could use that fetching flag to both set the button disabled and show the loader so the user knows that it is processing.  The resulting button declared on L61 would look something like:
```
<Button onClick={handleRefresh} type="button" text="Refresh" isDisabled={fetching} showLoader={fetching} />
```  
  
One other thing is that you could remove L18, L34, and L35 if we did it this way.  Slack me if you need any clarification; Overall great work on this one!

mau-synopsys: Sounds good, Ill go with that approach. The timer was arbitrary just to make it obvious it was clicked but it would definitely be better to wait for the request to re-enable.

bseifert14 (approved):

PR #2484, opened 2023-06-23
IALERT-3502 popup style

mau-synopsys:
Recreate package-lock.json

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,8 +1,17986 @@
 {
     "name": "blackduck-alert",
-    "version": "7.0.0-SIGQA10",
-    "lockfileVersion": 1,
+    "version": "7.0.0-SIGQA11-SNAPSHOT",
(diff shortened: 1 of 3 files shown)


ui/package-lock.json
psantos1113, line 1: Should this be checked in if you're just changing the style?

ui/package.json
psantos1113, line 1: Should this be checked in if you're just changing the style?

ui/package-lock.json
mau-synopsys: the package*.json files are built from an incorrect npm version in jenkins. we thought we had updated but serv-builder was still using the old npm versions, restarted the node and planning to update to test again.

ui/package.json
mau-synopsys: Looks like serv-builder didnt upgrade package.json to sigqa11-snapshot so this was automatically adjusted during runserver

PR #2486, opened 2023-06-26
IALERT-3507/3508 Search for other fields for user, cert, provider, task, and role tables

mau-synopsys:


psantos1113 (approved):

ChomickiM (approved):

bseifert14:
Just wanted to drop a comment on the changes I added:

1. The search input field should be slightly larger (wanted to mimic the same shape as our buttons in the TableActions).
2. The search input field should only filter the table results on user pressing 'Enter' or the search icon to the right of the field.
3. The search input field should now have an 'x' to clear the input in the field
4. The `handleSearchChange()` functions in each of the tables is now just passed a `searchValue` rather than finding the evt.target.value.

bseifert14 (approved):

PR #2496, opened 2023-07-05
IALERT-3482: Modal to mimic old, set the dropdown height to 6 elements

mau-synopsys:


ChomickiM (approved):

bseifert14:
Hey Michael, I just pushed some changes that I think will fix this.  The dropdown won't open over the footer of the modal, but it will work.  Testing this branch the dropdown opened properly, but shrinking the page caused some weird ui issues (referring to the offline convo we had).  I'd say let's hold off on merging this in unless we can get the dropdown to open over the footer

PR #2524, opened 2024-01-18
IALERT-3577 Create cert and private key tables

mau-synopsys:
image

### DIFF
--- a/alert-database/src/main/resources/liquibase/7.1.0/client-certificates-tables.xml
+++ b/alert-database/src/main/resources/liquibase/7.1.0/client-certificates-tables.xml
@@ -18,6 +18,9 @@
                 <constraints nullable="false"/>
             </column>
             <column name="certificate_content" type="VARCHAR"/>
+            <column name="last_updated" type="TIMESTAMP WITH TIMEZONE" defaultValueComputed="now()">
+                <constraints nullable="false"/>
+            </column>
         </createTable>
     </changeSet>
     <changeSet author="mau" id="create-client-certificate-keys-table">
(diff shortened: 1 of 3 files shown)


alert-database/src/main/resources/liquibase/7.1.0/client-certificates-tables.xml
```diff
@@ -0,0 +1,65 @@
...
+            <column name="id" type="UUID" defaultValueComputed="uuid_generate_v4()">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="alias" type="VARCHAR(128)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="private_key_id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="certificate_content" type="VARCHAR"/>
```
psantos1113, line 20: I don't believe we want this to be nullable either.  See if you can add the constraint nullable=false for the content field as well.

alert-database/src/main/resources/liquibase/7.1.0/client-certificates-tables.xml
```diff
@@ -0,0 +1,65 @@
...
+            <column name="id" type="UUID" defaultValueComputed="uuid_generate_v4()">
+                <constraints primaryKey="true"/>
+            </column>
+            <column name="name" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="password" type="VARCHAR">
+                <constraints nullable="false"/>
+            </column>
+            <column name="key_content" type="VARCHAR"/>
```
psantos1113, line 42: I don't believe we want this to be nullable either.  See if you can add the constraint nullable=false for the content field as well.

psantos1113 (approved):

DanaMaxfield (approved):

PR #2526, opened 2024-01-23
IALERT-3578 Client cert repo and accessors

mau-synopsys:
Accessors implements UniqueConfigurationAccessor with `findByAlias` for certificates  (unique alias) and `findByName` (unique name) for certificate private keys.

```
    public static final String DEFAULT_CLIENT_CERTIFICATE_ALIAS = "default-alert-client-certificate";

    public static final String DEFAULT_CLIENT_CERTIFICATE_KEY_NAME = "default-alert-client-certificate-key";
```

Includes reformatting by BD intellij rules and removal of want_assertions_signed (leftover artifacts) from SAMLTestHelper.


alert-database/src/main/java/com/synopsys/integration/alert/database/certificates/ClientCertificateRepository.java
```diff
@@ -0,0 +1,8 @@
+package com.synopsys.integration.alert.database.certificates;
+
+import java.util.UUID;
+
+import org.springframework.data.jpa.repository.JpaRepository;
+
+public interface ClientCertificateRepository extends JpaRepository<ClientCertificateEntity, UUID> {
```
ChomickiM, line 7: We likely want some helper JPA methods here. Such as findByAlias, existsByAlias, and deleteByAlias. Similar to what we've done in SAMLConfigurationRepository. This way we can move and simplify some of the logic out of the accessor and into JPA.

alert-common/src/main/java/com/synopsys/integration/alert/common/rest/AlertRestConstants.java
```diff
@@ -3,33 +3,56 @@
  *
  * Copyright (c) 2022 Synopsys, Inc.
  *
- * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All rights reserved worldwide.
+ * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All
+ * rights reserved worldwide.
  */
 package com.synopsys.integration.alert.common.rest;
 
 public final class AlertRestConstants {
```
ChomickiM, line 11: This class may have been using the wrong save actions, I don't believe we want to insert spaces between each field

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultClientCertificateKeyAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+import com.synopsys.integration.alert.api.common.model.exception.AlertConfigurationException;
+import com.synopsys.integration.alert.common.persistence.accessor.UniqueConfigurationAccessor;
+import com.synopsys.integration.alert.common.persistence.model.ClientCertificateKeyModel;
+import com.synopsys.integration.alert.common.rest.AlertRestConstants;
+import com.synopsys.integration.alert.common.security.EncryptionUtility;
+import com.synopsys.integration.alert.common.util.DateUtils;
+import com.synopsys.integration.alert.database.certificates.ClientCertificateKeyEntity;
+import com.synopsys.integration.alert.database.certificates.ClientCertificateKeyRepository;
+
+public class DefaultClientCertificateKeyAccessor implements UniqueConfigurationAccessor<ClientCertificateKeyModel> {
```
ChomickiM, line 18: A general point to this accessor, I don't believe we should be using the AlertRestConstants.DEFAULT_CLIENT_CERTIFICATE_KEY_NAME for the name field of ClientCertificateKeyModel. Alias is specific to the content which is in the other model object. For this my thoughts are that we should perhaps use the AlertRestConstants.DEFAULT_CONFIGURATION_NAME

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ClientCertificateModel.java
```diff
@@ -0,0 +1,45 @@
...
+        this.lastUpdated = lastUpdated;
+    }
+
+    public UUID getId() {
+        return id;
+    }
+
+    public String getAlias() {
+        return alias;
+    }
```
ChomickiM, line 32: After thinking about this model, I don't think exposing the alias, nor requiring it in the constructor makes sense. This field will be ignored if passed in as part of a REST request and the results if an end user GETs the model will also not matter. As all the usages of the alias are internal to Alert we should not expose this in the API. In the future, if we are required to support multiple certificates and aliases' then we can always make an additive change to include it back in without causing breaking changes.

I think that we should remove the alias from the model, but still keep it in still in the database entity class.

alert-database/src/main/java/com/synopsys/integration/alert/database/api/DefaultClientCertificateKeyAccessor.java
mau-synopsys: Change DEFAULT_CLIENT_CERTIFICATE_KEY_NAME only

ChomickiM (approved):

psantos1113 (approved):

PR #2536, opened 2024-02-05
Dev/mc/ialert 3583 mtls client cert keystore

psantos1113:
- Create the object to import and remove the client certificates.
- Use PemSslStoreBundle to handle parsing the private key content.
- Store the client certificate in memory no need for a file.
- Clean up process of creating a keystore with the KeyStoreManager class.


buildSrc/src/main/java/com/synopsys/integration/alert/build/RunServerTask.java
```diff
@@ -168,13 +170,15 @@ public List<String> getApplicationVariables(String buildDirectory) {
         List<String> commonVariables = List.of(
             String.format("--server.ssl.key-store=%s/certs/blackduck-alert.keystore", buildDirectory),
             String.format("--server.ssl.trust-store=%s/certs/blackduck-alert.truststore", buildDirectory),
+            String.format("--alert.mtls.client.key-store=%s/certs/mtls-client.keystore", buildDirectory),
```
ChomickiM, line 173: I believe this line can be removed as we've removed the keystore file

src/test/java/com/synopsys/integration/alert/component/certificates/web/CertificateTestUtil.java
```diff
@@ -14,46 +18,75 @@
...
+    // rootCA.pem = "root-ca-cert-test.xxx.blackduck.alert.example.com"
+    // server.pem = "server-cert-test.xxx.blackduck.alert.example.com"
+    // client.pem = "client-cert-test.xxx.blackduck.alert.example.com"
+    public static final String CERTIFICATE_MTLS_CLIENT_FILE_PATH = "certificates/mtls/client.pem";
+    public static final String KEY_MTLS_CLIENT_FILE_PATH = "certificates/mtls/client.key";
     public static final String TEST_ALIAS = "test-alias";
     public static final String TRUSTSTORE_FILE_PATH = "./build/certs/blackduck-alert-test.truststore";
     public static final String TRUSTSTORE_PASSWORD = "changeit";
 
+    public static final String MTLS_KEYSTORE_FILE_PATH = "./build/certs/mtls-client-test.keystore";
```
ChomickiM, line 36: This can likely be dropped as well as I don't see other usages of it anymore

PR #2545, opened 2024-02-08
IALERT-3589 Adds ClientCertificateController, CrudActions, and ConfigurationValidator

mau-synopsys:
Tests for classes are WIP

### DIFF
--- a/component/src/main/java/com/synopsys/integration/alert/component/certificates/CertificatesDescriptor.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/certificates/CertificatesDescriptor.java
@@ -26,6 +26,11 @@ public class CertificatesDescriptor extends ComponentDescriptor {
     public static final String KEY_ALIAS = "alias";
     public static final String KEY_CERTIFICATE_CONTENT = "certificateContent";
 
+    // following fields are for client certificates
+    public static final String KEY_CLIENT_CERTIFICATE_CONTENT = "clientCertificateContent";
+    public static final String KEY_PRIVATE_KEY_PASSWORD = "keyPassword";
+    public static final String KEY_PRIVATE_KEY_CONTENT = "keyContent";
+
     @Autowired
     public CertificatesDescriptor(CertificatesDescriptorKey descriptorKey) {
         super(descriptorKey);
--- a/component/src/main/java/com/synopsys/integration/alert/component/certificates/web/ClientCertificateConfigurationValidator.java
+++ b/component/src/main/java/com/synopsys/integration/alert/component/certificates/web/ClientCertificateConfigurationValidator.java
@@ -1,0 +1,35 @@
+package com.synopsys.integration.alert.component.certificates.web;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.api.common.model.ValidationResponseModel;
+import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatus;
+import com.synopsys.integration.alert.common.persistence.model.ClientCertificateModel;
+import com.synopsys.integration.alert.component.certificates.CertificatesDescriptor;
+
+@Component
+public class ClientCertificateConfigurationValidator {
+    public ValidationResponseModel validate(ClientCertificateModel model) {
+        Set<AlertFieldStatus> statuses = new HashSet<>();
+
+        if (StringUtils.isBlank(model.getKeyPassword())) {
+            statuses.add(AlertFieldStatus.error(CertificatesDescriptor.KEY_PRIVATE_KEY_PASSWORD, "Private key password cannot be empty."));
+        }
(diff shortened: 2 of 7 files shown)


component/src/main/java/com/synopsys/integration/alert/component/certificates/CertificatesDescriptor.java
```diff
@@ -26,6 +26,11 @@ public class CertificatesDescriptor extends ComponentDescriptor {
     public static final String KEY_ALIAS = "alias";
     public static final String KEY_CERTIFICATE_CONTENT = "certificateContent";
 
+    // following fields are for client certificates
+    public static final String KEY_CLIENT_CERTIFICATE_CONTENT = "clientCertificateContent";
```
bseifert14, line 30: This looks good as is to me.  I propose the question that since this is technically an MTLS certificate, is 'client' too broad? I know we are calling that in the UI so I'm ok with what you have here.  But I do wonder if we should be more specific with something like 'mtlsCertificateContent'

ChomickiM: I think we'd still want to call it a "client certificate" as it better describes the type of certificate thats being used. mTLS is just the method of mutual authentication between the client and server. I don't believe there is explicitly an "mTLS certificate"

component/src/main/java/com/synopsys/integration/alert/component/certificates/web/ClientCertificateController.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public ClientCertificateModel getOne() {
+        return ResponseFactory.createContentResponseFromAction(configActions.getOne());
+    }
+
+    @Override
+    public void update(ClientCertificateModel resource) {
+        throw ResponseFactory.createNotImplementedException("Updates of client certificate not support, please delete and create a new configuration.");
+    }
+
```
ChomickiM, line 34: It seems that the HTTP NOT_IMPLEMENTED is a 501 error codes and I think we're discouraging throwing 5xx codes as typically it means something didn't work on the server side. Would it perhaps make more sense to use HTTP METHOD_NOT_ALLOWED 405?

mau-synopsys: Created and updated to use 405 status code.

component/src/main/java/com/synopsys/integration/alert/component/certificates/CertificatesDescriptor.java
mau-synopsys: Agreed with what Martin said. I believe alert has an always trust option? this means it can be one way as opposed to mtls

psantos1113: Yes @ChomickiM  is correct.  With mTLS (mutual TLS) there are two certificates used.  The server certificate on the server to receive the HTTP request.  The client certificate which is on the client system that is initiating the HTTP request to the server.
The certificates in use are:

- client certificate
- server certificate
- any intermediate and root certificate authority certificates used to sign both the client and server certificates which validates that the client and server can trust one another.

component/src/main/java/com/synopsys/integration/alert/component/certificates/web/ClientCertificateController.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public ClientCertificateModel getOne() {
+        return ResponseFactory.createContentResponseFromAction(configActions.getOne());
+    }
+
+    @Override
+    public void update(ClientCertificateModel resource) {
+        throw ResponseFactory.createMethodNotAllowedException("Updates of client certificate not support, please delete and create a new configuration.");
```
DanaMaxfield, line 32: Can we change this message?

Updates of client certificate are not supported. Please delete and create a new configuration.

component/src/main/java/com/synopsys/integration/alert/component/certificates/web/ClientCertificateConfigurationValidator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.stereotype.Component;
+
+import com.synopsys.integration.alert.api.common.model.ValidationResponseModel;
+import com.synopsys.integration.alert.api.common.model.errors.AlertFieldStatus;
+import com.synopsys.integration.alert.common.persistence.model.ClientCertificateModel;
+import com.synopsys.integration.alert.component.certificates.CertificatesDescriptor;
+
+@Component
+public class ClientCertificateConfigurationValidator {
+    public ValidationResponseModel validate(ClientCertificateModel model) {
```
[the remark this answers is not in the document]
mau-synopsys: Added configurationValidatorTest. Also adds the following tests to controllerIT:

`createExistingConfig
createConfigEmptyFields
createConfigThrows
deleteEmptyConfig
`

DanaMaxfield (approved):

psantos1113 (approved):

ChomickiM (approved):

PR #2552, opened 2024-02-21
Use clientCertificateContent as json field name

mau-synopsys:
To match the descriptor key value

### DIFF
--- a/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ClientCertificateModel.java
+++ b/alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ClientCertificateModel.java
@@ -2,7 +2,6 @@ package com.synopsys.integration.alert.common.persistence.model;
 
 import java.io.Serial;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.synopsys.integration.alert.api.common.model.AlertSerializableModel;
 import com.synopsys.integration.alert.api.common.model.Obfuscated;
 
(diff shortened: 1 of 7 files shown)


bseifert14 (approved):

DanaMaxfield (approved):

alert-common/src/main/java/com/synopsys/integration/alert/common/persistence/model/ClientCertificateModel.java
```diff
@@ -10,6 +11,7 @@ public class ClientCertificateModel extends AlertSerializableModel implements Ob
     private static final long serialVersionUID = -2254327445871324025L;
     private String keyPassword;
     private String keyContent;
+    @JsonProperty("clientCertificateContent")
     private String certificateContent;
```
ChomickiM, line 15: We don't need to add the JsonProperty annotation here since we're creating this as a net new property. It would likely be easier to just rename the field from "certificateContent" to "clientCertificateContent" and refactor appropriately

mau-synopsys: This would be a one line change here vs refactoring multiple files. We have examples of different property/alias name in both alert and hub. Open to the latter however.

psantos1113: This is a new feature not an existing feature.  I would change the field name to clientCertificateContent.  Yes, multiple files will change but that's ok. 7 files would change. The feature this is related to is the only feature for the release.  I would like the changes made as complete as possible before release.  If this was released code then yes I would agree with the annotation 1 line fix.  This isn't released and we would like the variable changed. We try to avoid adding the annotations unless we absolutely need to as is the case with released features where we need to maintain backwards compatibility.

mau-synopsys: Cool, just needed context on why some places we are ok with different property name but not in this instance. Renamed to clientCertificateContent

ChomickiM (approved):

psantos1113 (approved):

PR #2553, opened 2024-02-22
IALERT-3632 - Add logic to generate Swagger API spec file

DanaMaxfield:
* Create new test, SwaggerGenerateDocTestIT.testGenerateSwaggerAPISpec(), to create the needed Swagger API Spec file
* Create new test tag to allow targeting of this specific test
* Add new gradle task, swaggerApiSpec, to allow running this test directly

### DIFF
--- a/test-common/src/main/java/com/synopsys/integration/alert/test/common/TestResourceUtils.java
+++ b/test-common/src/main/java/com/synopsys/integration/alert/test/common/TestResourceUtils.java
@@ -21,9 +21,23 @@ import org.springframework.core.io.ClassPathResource;
 
 public final class TestResourceUtils {
     public static final String DEFAULT_PROPERTIES_FILE_NAME = "test.properties";
+    public static final String DEFAULT_SWAGGER_API_SPEC_FILE_NAME = "swagger.api-spec";
 
     private static final String SUB_PROJECT_NAME = "test-common";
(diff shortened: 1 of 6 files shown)


test-common/src/main/java/com/synopsys/integration/alert/test/common/TestResourceUtils.java
```diff
@@ -21,9 +21,23 @@
...
 public final class TestResourceUtils {
     public static final String DEFAULT_PROPERTIES_FILE_NAME = "test.properties";
+    public static final String DEFAULT_SWAGGER_API_SPEC_FILE_NAME = "swagger.api-spec";
 
     private static final String SUB_PROJECT_NAME = "test-common";
-    private static final File EXPECTED_BASE_TEST_RESOURCE_DIR = new File(TestResourceUtils.class.getProtectionDomain().getCodeSource().getLocation().getPath(), "../../../../src/main/resources/");
+    private static final File EXPECTED_BASE_TEST_RESOURCE_DIR = new File(
+        TestResourceUtils.class.getProtectionDomain().getCodeSource().getLocation().getPath(),
+        "../../../../"
+    );
```
DanaMaxfield, line 30: Changing of this path to remove `/src/main/resources` is done because this is a bug. `EXPECTED_BASE_TEST_RESOURCE_DIR` is used if the `buildOutputDir` is not found. But either way, found or not, we append `/src/main/resources` to the end of the path. So if `buildOutputDir` isn't found, the returned value would have been `../../../../src/main/resources/src/main/resources`

psantos1113 (approved):

ChomickiM (approved):

PR #2565, opened 2024-03-28
IALERT-3643 - Exclude snakeyaml from buildscript

DanaMaxfield:
Because snakeyaml is coming in via a Gradle plugin AND Detect gets the dependencies from each subproject individually, we need to exclude snakeyaml from buildscript.

Because we also pull in other spring dependencies directly, we still need to perform the force of the version to 2.1. I moved the force to be in section of build.gradle that already existed for subprojects, purely for clarity and consistency.

This has been run via PoP job and results confirm snakeyaml 1.33 is no longer present.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ buildscript {
         classpath "com.synopsys.integration:common-gradle-plugin:${managedCgpVersion}"
     }
 
+    // In Alert 7.2 we plan to upgrade spring-boot, which should make this exclude unnecessary
     configurations {
         classpath {
             resolutionStrategy {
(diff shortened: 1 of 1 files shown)


ChomickiM (approved):

mau-synopsys (approved):

build.gradle
```diff
@@ -11,6 +11,14 @@ buildscript {
         classpath "org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}"
         classpath "com.synopsys.integration:common-gradle-plugin:${managedCgpVersion}"
     }
+
```
psantos1113, line 14: please add a a comment that this will be removed in the future when spring is updated.

build.gradle
```diff
@@ -98,6 +106,14 @@ subprojects {
 }
 
 allprojects {
+    configurations {
```
psantos1113, line 109: please add a a comment that this will be removed in the future when spring is updated.

psantos1113 (approved):

PR #2574, opened 2024-06-05
IALERT-3600 Reconfigure BlackDuck Project Selection For Distribution Job Form

bseifert14:
# IALERT-3600

**Overview:**  
Client has requested the Project selection field, found in the Distribution Job Form, to be restructured.  Their request included that the dropdown field was not able to update the search, limit, or paging of the request that populates the data in that dropdown.  After review, I discovered that the request was hardcoded with searchQuery, limit, and pagination.  
  
**Fix Overview:**  
To fix this I reworked the field to do the following:

- A user can click on the field which will open a modal
- Inside the modal, a table will render where the user can select any or all projects
- Once the user has selected and pressed 'Save', the modal will disappear and the field will render the selected projects, to which the user can unselect a project(s) directly from the field, or they can reopen the modal and unselect/select more.

### DIFF
--- a/ui/src/main/js/common/component/table/TableSkeleton.js
+++ b/ui/src/main/js/common/component/table/TableSkeleton.js
@@ -30,10 +30,6 @@ const useStyles = createUseStyles(theme => ({
         `,
         gridTemplateRows: '45px 35px 35px 35px'
     },
-    headerContainer: {
-        width: '100%',
-        height: '45px'
-    },
     headerItem: {
         height: '25px',
         width: '75px',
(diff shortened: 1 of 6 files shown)


psantos1113 (approved):

ui/src/main/js/common/component/table/TableSkeleton.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+}));
+
+export default function TableSkeleton() {
+    const classes = useStyles();
+
+    return (
+        <div className={classes.skeletonTableContainer}>
+            <div className={classNames(classes.headerItem, classes.col1)} />
+            <div className={classNames(classes.headerItem, classes.col2)} />
```
mau-synopsys, line 102: Does classes.col2/col3 need to be defined? Same for classes.row3colx

ui/src/main/js/common/component/table/TableSkeleton.js
```diff
@@ -0,0 +1,118 @@
...
+        marginLeft: '10px'
+    },
+    row1col2: {
+        width: '200px'
+    }, 
+    row2col2: {
+        width: '200px'
+    }, 
+    row3col2: {
+        width: '200px'
```
mau-synopsys, line 92: If the widths and cols styles are all the same, can we condense these to just one?

mau-synopsys (dismissed):

ui/src/main/js/common/component/table/TableSkeleton.js
bseifert14: Great point.  I've consolidated this code, I'll resolve both this convo and the one above as they both have been fixed in latest commit.

mau-synopsys (approved):

PR #2587, opened 2024-09-11
IALERT-3554: Support Azure Flexible server

petergreen86:
Fixes IALERT-3554

Changes:

- adds `isAzure` boolean to values along with some comments
- alert.yaml template has a few conditional blocks to either include or omit the full certificate chain based on the value of `isAzure` in values

Tested against:

- Azure Flexible server (PG 14)
- GCP Postgres (PG 14) with `Require trusted client certificates` enabled

### DIFF
--- a/deployment/helm/synopsys-alert/templates/alert.yaml
+++ b/deployment/helm/synopsys-alert/templates/alert.yaml
@@ -74,12 +74,22 @@ spec:
               name: {{ .Release.Name }}-db-creds
         {{- end }}
         {{- if (eq .Values.postgres.sslUseFiles true) }}
+        {{- if (eq .Values.postgres.isAzure false)}}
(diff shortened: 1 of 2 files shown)


psantos1113:
The problem with these changes is that it is taking a Postgres feature, the SSL configuration, and adding a one off attribute for an Azure deployment of a Postgres database specifically.  The external database provider should not matter.  If the same changes are needed for another Postgres database deployment, then we would need to add another attribute to values.yaml for another one off.  That isn't the best solution here.  

The issue is that the SSL configuration values right now are an all or nothing solution and that is the problem that needs to be solved.  We have a separate issue to investigate and test various permutations of the SSL Mode configuration for Postgres.   When that investigation is done a solution will be worked on.

PR #2612, opened 2024-10-09
IALERT-3714 - Update copyright

DanaMaxfield:
**This is based off of dev/djm/IALERT-3713-optimize-imports, which is not yet merged, so flagging as a draft.**

Update copyright to new BD copyright

* Remove OLD BD copyright
* Remove Synopsys copyright
* Add in new BD copyright

Diff of commit which added copyright to previous commit:
```
3598 + *
1799 + * Copyright (c) 2024 Black Duck Software, Inc.
1799 + * Use subject to the terms and conditions of the Black Duck Software End User Software License and Maintenance Agreement. All rights reserved worldwide.
1799 + * blackduck-alert
1799 + */
1799 +/*
```


ChomickiM:
I see the first line in the copyright header no longer includes the subproject name that it's in. Is there a reason we aren't keeping that or are we supposed to be using blackduck-alert instead?

DanaMaxfield:
> I see the first line in the copyright header no longer includes the subproject name that it's in. Is there a reason we aren't keeping that or are we supposed to be using blackduck-alert instead?

That's the format used in all of the upstream libraries (using the repository name not the Gradle sub-project name) so I used that same logic here.

bseifert14 (approved):

ChomickiM (approved):

PR #2623, opened 2024-10-17
IALERT-3656: Update checkboxs for Role permission table

ChomickiM:
This change in attributes is to resolve an issue where the fields were not being reset after a permission was added to a descriptor in the role table. By switching this attribute it is more consistent with other checkbox input fields in Alert and resolves the issue of the state not being reset.


ui/src/main/js/common/component/input/Checkbox.js
```diff
@@ -20,11 +20,11 @@ const Checkbox = ({ id, name, label, placeholder, value = '', isChecked = false,
...
-            <FieldLabel label={label}/>
+        <div className={classes.checkboxField}>
+            <FieldLabel label={label} />
             <input
                 name={name}
-                aria-checked={isChecked}
+                checked={isChecked}
                 aria-label={label}
                 className={classes.input}
                 defaultChecked={isChecked}
```
mau-synopsys, line 30: We can probably remove defaultChecked as well

mau-synopsys (approved):

bseifert14:
Good find on this fix.  aria-checked is only required when the html element isn't an `<input>`.  No clue what I was thinking when I initially created this (doh!).

bseifert14 (approved):

PR #2634, opened 2024-10-31
IALERT-3765: Show Project Names in Project Filter Field

bseifert14:
# Bug 
There is a bug on the Distribution page, when a user attempts to edit a job.   If that job has projects selected to filter by, they will not render within the field.  The reason is that the data coming from the response is different than the data that we set it up as within `FieldModelUtilities`.  To counteract this bug, I've written a function, `getProjectValues()`.  In this function we check whether or not the data is set up properly to render on the screen.  If not, we correct it for the React 'Select' element.


ui/src/main/js/page/distribution/ProjectSelectModal.js
```diff
@@ -58,7 +57,7 @@ export default function ProjectSelectModal({ isOpen, handleClose, csrfToken, pro
 
         fetchProjectList();
     }, [pageNumber, pageSize, searchTerm, csrfToken, projectRequestBody]);
-
+    console.log(formData);
```
mau-synopsys, line 60: Remove console log

ChomickiM (approved):

psantos1113 (approved):

DanaMaxfield (approved):

PR #2654, opened 2025-04-11
IALERT-3817: Update old jira tickets with the new property

psantos1113:
# Problem

Duplicate issues get created by Alert when either of the following occurs:

1. Uninstall the Jira app [Alert issue property indexer](https://github.com/blackducksoftware/alert-issue-property-indexer) 0.0.5 or earlier 
2. Do not install the latest Alert issue property indexer 0.0.6 or higher  

For scenario 1. by uninstalling the older plugin the index that was created in the Jira database for the properties that Alert creates on issues is removed.  Although the JQL query string looks for both the new and old property keys if the database index is removed then issue properties cannot be searched for to find existing issues. This means that Alert can no longer find the issues previously created prior to the 8.x release.  Therefore it will create new duplicate tickets.

For scenario 2. by not installing the Alert issue property indexer there is no index for the new property key that Alert creates on issues.  Since it cannot find existing issues based on the property it will create a new Jira issue every time a notification is processed until the Alert issue property indexer is installed on the Jira instance.

# Solution
There are 2 parts to the solution in this PR.

1. Run a task for Jira server or Jira cloud to add the new property to tickets created by Alert.  When no more tickets can be found stop running the task to update older issues.
2. Allow Alert to handle updating more than 1 ticket.  We can get in a state with duplicates so now we must allow alert to update all duplicate tickets with the same updates.

Updated `IssueTrackerSearcher` to be able to find more than 1 ticket but cap the query at 100 duplicates. For scenario 1. the number of duplicate tickets is 2.  For scenario 2. there can be more but the max at 100 has been sufficient during testing so far. 

Created the class `JiraTask` and `JiraInitializer` to be able to update existing issues created with the old indexer with the new property to prevent creating duplicate tickets.


api-channel-jira/src/main/java/com/blackduck/integration/alert/api/channel/jira/lifecycle/JiraSchedulingManager.java
```diff
@@ -0,0 +1,59 @@
...
+    @Autowired
+    public JiraSchedulingManager(TaskManager taskManager) {
+        this.taskManager = taskManager;
+    }
+
+    public List<JiraTask> scheduleTasks(List<JiraTask> tasks) {
+        List<JiraTask> acceptedTasks = new ArrayList<>();
+        for (JiraTask task : tasks) {
+            logger.debug("Perform scheduling jira tasks for config with id {} and name {}", task.getConfigId(), task.getConfigName());
+            unscheduleTasks(task.getConfigId());
```
DanaMaxfield, line 26: It's picky, but can we switch lines 26 and 25? IMO the order or logging messages could be misleading to a customer.

ui/src/main/js/page/task/TaskModal.js
```diff
@@ -50,23 +50,16 @@ const TaskModal = ({ data, isOpen, toggleModal }) => {
...
-                ) : null }
+                {/* If properties array has content, Provider will be present in first position */}
+                { properties && properties.map(property => (
+                        <ReadOnlyField
+                            id={setId(property.key)}
+                            label={property.displayName}
+                            value={property.value}
+                        />
+                    ))
+                }
```
DanaMaxfield, line 62: What is this change for?

channel-jira-server/src/test/java/com/blackduck/integration/alert/channel/jira/server/convert/JiraServerGlobalConfigurationModelSaveActionsTest.java
```diff
@@ -54,6 +59,11 @@ class JiraServerGlobalConfigurationModelSaveActionsTest {
     private final FilePersistenceUtil filePersistenceUtil = new FilePersistenceUtil(alertProperties, gson);
     private final EncryptionUtility encryptionUtility = new EncryptionUtility(alertProperties, filePersistenceUtil);
     private final AuthorizationManager authorizationManager = createAuthorizationManager();
+    private final JiraSchedulingManager jiraSchedulingManager = new JiraSchedulingManager(new TaskManager());
+    private final TaskScheduler taskScheduler = Mockito.mock(TaskScheduler .class);
```
DanaMaxfield, line 63: Does the space between TaskScheduler and .class matter?

channel-jira-server/src/test/java/com/blackduck/integration/alert/channel/jira/server/action/JiraServerGlobalCrudActionsTest.java
```diff
@@ -57,14 +64,20 @@ class JiraServerGlobalCrudActionsTest {
         () -> new PermissionMatrixModel(permissions)
     );
 
+    private final TaskManager taskManager = new TaskManager();
+    private final JiraSchedulingManager jiraSchedulingManager = new JiraSchedulingManager(taskManager);
+    private final TaskScheduler taskScheduler = Mockito.mock(TaskScheduler .class);
```
DanaMaxfield, line 69: Does the space between TaskScheduler and .class matter?

DanaMaxfield (changes requested):

channel-jira-server/src/test/java/com/blackduck/integration/alert/channel/jira/server/action/JiraServerGlobalCrudActionsTest.java
psantos1113: no

ui/src/main/js/page/task/TaskModal.js
psantos1113: Tasks has a collection of metadata properties that are set on the back end.  We always assumed if we had 2 properties it was provider UUID and then the provider configuration.  I made the UI be dynamic and add fields based on the properties passed from the back end.  This works for Provider configuration and for the new tasks I have added.

api-channel-jira/src/main/java/com/blackduck/integration/alert/api/channel/jira/lifecycle/JiraSchedulingManager.java
psantos1113: Will make the change

DanaMaxfield (approved):

api-channel-issue-tracker/src/main/java/com/blackduck/integration/alert/api/channel/issue/tracker/search/IssueTrackerSearcher.java
```diff
@@ -109,39 +114,42 @@ private List<ActionableIssueSearchResult<T>> findIssuesByAllComponents(ProviderD
...
 
-    private ActionableIssueSearchResult<T> findIssueByProjectIssueModel(ProjectIssueModel projectIssueModel) throws AlertException {
-        ExistingIssueDetails<T> existingIssue = null;
-        ItemOperation searchResultOperation = ItemOperation.UPDATE;
+    private List<ActionableIssueSearchResult<T>> findIssueByProjectIssueModel(ProjectIssueModel projectIssueModel) throws AlertException {
+        List<ActionableIssueSearchResult<T>> searchResults = new LinkedList<>();
 
-        IssueTrackerSearchResult<T> searchResponse = exactIssueFinder.findExistingIssuesByProjectIssueModel(projectIssueModel);
+
+        IssueTrackerSearchResult<T> searchResponse = exactIssueFinder.findExistingIssuesByProjectIssueModel(projectIssueModel,DEFAULT_MAX_RESULTS);
```
ChomickiM, line 121: The usage of max results here is very specific so we don't introduce a performance hit from a completely unbound list of Jira issues. Perhaps we should include a comment stating why we have an upper limit of 100 in the event we need to make future changes or at least detail why this change was made.

channel-jira-cloud/src/main/java/com/blackduck/integration/alert/channel/jira/cloud/task/JiraPropertyUpdateTask.java
```diff
@@ -0,0 +1,91 @@
...
+                    // projects to query for.
+                    unScheduleTask();
+                }
+            }
+            logger.info("Jira Cloud property migrator task ended.");
+        } catch (IntegrationException e) {
+            logger.error("Error getting Jira Server Configuration.", e);
+        } catch (InterruptedException e) {
+            logger.error("Error updating Jira Server issues with new property.", e);
+            Thread.currentThread().interrupt();
```
ChomickiM, line 70: "Jira Server" should be changed to "Jira Cloud" here

ChomickiM (approved):

PR #2695, opened 2025-09-22
IALERT-3123: Unset the Preset Channel in Distribution Job Creation

bseifert14:
## Description  
IALERT-3123 deals with not showing a warning immediately when a user lands on the Distribution Job creation form.  This warning shows because the channel would be preset when the form renders on the page, but the Provider was not set.  After some talks with the Alert team, we want to proactively counteract this by not presetting the channel on form render.

## Technical Description  
In the `DistributionConfigurationForm`, we no longer preset the channel.  The preset channel is just an empty string rather than Azure Board.  I'm not sure why it was preset in the past, my assumption is that it's the first in the list of selectable options for channels.  Regardless, unsetting the channel on form render no longer renders the channel part of the form (bottom part of the form that changes when the user changes the selected channel).  Now when the form renders there is no warning displayed and the form looks like the attached screenshot.  
  
Screenshot 2025-09-22 at 10 12 24 AM

### DIFF
--- a/ui/src/main/js/page/distribution/DistributionConfigurationForm.js
+++ b/ui/src/main/js/page/distribution/DistributionConfigurationForm.js
@@ -338,7 +338,7 @@ const DistributionConfigurationForm = ({
             {label: project.name, value: project.href}
         ))
     }
-    console.log(!isOperationAssigned(descriptors[selectedChannel], OPERATIONS.EXECUTE));
+
     // TODO need to provide finer grain control with permissions.
     return (
         <>


ui/src/main/js/page/distribution/DistributionConfigurationForm.js
```diff
@@ -342,7 +338,7 @@ const DistributionConfigurationForm = ({
             {label: project.name, value: project.href}
         ))
     }
-
+    console.log(!isOperationAssigned(descriptors[selectedChannel], OPERATIONS.EXECUTE));
```
psantos1113: Doesn't this produce a linting warning by logging to the console?

PR #2745, opened 2026-03-18
IALERT-3990: Update the UI Tabs

bseifert14:
# Description
This PR handles the UI updates to our tabs.  Right now they are "out of the box" react bootstrap designed tabs.  We want them to look more like what the Hub/SCA has in their app.

# Technical Description
To accomplish this task, we needed to create a common tab component that would hold the design structure.  This component is titled, `ViewTabs`, as it essentially is a tab system that navigates views (I'm open to renaming this if that doesn't make sense).  Inside of `ViewTabs.js`, I have updated the css to look/feel like it does in the Hub/SCA (see screenshots). 

# To use this new component: 
Simply call in that component to your React file and define the id of the tab system. Then, add a react-bootstrap `Tab` as the child, with a page view as that `Tab`'s child. See below for an example:
```
import ViewTabs from 'common/component/navigation/ViewTabs';
import { Tab } from 'react-bootstrap';
...
<ViewTabs id="my-tab-system-id-here">
    <Tab eventKey={1} title="Tab Number 1">
        <div>Page Content for tab number 1 here</div>
    </Tab>
    <Tab eventKey={2} title="Tab Number 2">
        <div>Page Content for tab number 2 here</div>
    </Tab>
</ViewTabs>
```

# Current Design vs Updated Design
Screenshot 2026-03-18 at 5 57 07 PM
Screenshot 2026-03-18 at 5 56 16 PM


psantos1113 (approved):

DanaMaxfield (dismissed):
I agree with Paulo and CoPilot, can you make the imports consistent?

bseifert14:
> I agree with Paulo and CoPilot, can you make the imports consistent?

Yep! I just ran a linter that caught it all and have since updated 👍

ChomickiM (approved):

PR #2748, opened 2026-03-24
IALERT-3971: fix unscheduling tasks when channel is deleted

psantos1113:
The task for adding comments to tickets was using a task manager that was looking for JiraPropertyTask classes.  The issue is that the commenting task extends JiraTask.  This is why the task would not end.  Changed the taskmanager to use JiraTask class to unschedule all the tasks for a channel. 

As a failsafe also added a threshold limit of 120 consecutive failures and keep track of the failures.  If the threshold is exceeded then there is a log.  Each time the task fails consecutively there is a log message to indicate the number of consecutive failures. This results in a task that executes for an hour with failures will unschedule the task.  As soon as the task completes without exceptions then the failure counter is reset to 0.  This allows the task to recover from intermittent networking issues or certificate issues communicating with the channel.  If the task completes successfully and finishes it's job it will un-schedule itself as usual.

### DIFF
--- a/channel-jira-server/src/main/java/com/blackduck/integration/alert/channel/jira/server/task/JiraPropertyUpdateTask.java
+++ b/channel-jira-server/src/main/java/com/blackduck/integration/alert/channel/jira/server/task/JiraPropertyUpdateTask.java
@@ -70,11 +70,14 @@ public class JiraPropertyUpdateTask extends JiraPropertyTask {
                     unScheduleTask();
                 }
             }
+            resetConsecutiveFailures();
         } catch (IntegrationException e) {
             logger.error("Error getting Jira Server Configuration.", e);
+            checkThresholdAndIncrementFailures();
         } catch (InterruptedException e) {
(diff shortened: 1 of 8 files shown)


channel-jira-server/src/main/java/com/blackduck/integration/alert/channel/jira/server/task/JiraPropertyUpdateTask.java
```diff
@@ -70,11 +70,14 @@ protected void executeTaskImplementation() {
                     unScheduleTask();
                 }
             }
+            resetConsecutiveFailures();
         } catch (IntegrationException e) {
             logger.error("Error getting Jira Server Configuration.", e);
+            checkThresholdAndIncrementFailures();
```
psantos1113: increment the failure counter to prevent infinite execution

DanaMaxfield (approved):

ChomickiM (approved):

PR #2759, opened 2026-04-02
IALERT-3960 - Upgrade to PG18

DanaMaxfield:
Still testing


DanaMaxfield:
Upgrading to PG-17 instead of PG-18

PR #2795, opened 2026-08-06
IALERT-4057: Fix audit failure handling for issue trackers

ChomickiM:
This branch fixes a bug where Jira Cloud and Jira Server event handlers silently swallowed `AlertException` errors, preventing failures from being recorded in the Audit Failed table. 

The fix propagates exceptions from all six handlers for Jira Cloud and Jira Server (create/comment/transition)  up to the `JobSubTaskEventHandler` base class, which catches `AlertException` and publishes an `AuditFailedEvent`. This error can then be used in the Failed Audit table to either replay events that failed due to timeouts, or to inspect errors that may have occurred. 

I have also added a comment for each handler explaining the double log scenario. A second log is not produced, but we use the stack trace and error message by the upstream base class for audit messaging.

### DIFF
--- a/api-distribution/src/main/java/com/blackduck/integration/alert/api/distribution/JobSubTaskEventHandler.java
+++ b/api-distribution/src/main/java/com/blackduck/integration/alert/api/distribution/JobSubTaskEventHandler.java
@@ -11,6 +11,9 @@ import java.time.Instant;
 import java.util.UUID;
 import java.util.function.Predicate;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.blackduck.integration.alert.api.common.model.exception.AlertException;
 import com.blackduck.integration.alert.api.distribution.audit.AuditFailedEvent;
 import com.blackduck.integration.alert.api.distribution.execution.ExecutingJob;
@@ -23,6 +26,7 @@ import com.blackduck.integration.alert.common.enumeration.AuditEntryStatus;
 import com.blackduck.integration.alert.common.persistence.util.AuditStackTraceUtil;
 
 public abstract class JobSubTaskEventHandler<T extends JobSubTaskEvent> implements AlertEventHandler<T> {
+    private final Logger logger = LoggerFactory.getLogger(JobSubTaskEventHandler.class);
     private final EventManager eventManager;
     private final JobStage jobStage;
     private final ExecutingJobManager executingJobManager;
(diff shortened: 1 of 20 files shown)


sig-fisherj:
While not necessarily tied to your fix, this may be a concern.

**Stack Trace Exposed in User-Facing Audit Table (Worth Discussing)**                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  AuditStackTraceUtil.createStackTraceString(exception) stores the full Java stack trace in AuditFailedEvent. That trace is stored in the database and surfaced in Alert's "Audit Failed" table, visible to Alert end users.                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Per security guidelines: detailed error information (stack traces, internal state) should be in server-side logs, not in API responses or user-facing views. A Java stack trace with class names, packages, and call frames reveals internal implementation details that have no diagnostic value to an Alert user but provide a roadmap for an attacker.                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  This is pre-existing behavior in JobSubTaskEventHandler — not introduced by IALERT-4057. But this PR expands the scope: Jira failures were previously silently dropped; they now reach the audit table. Worth a deliberate answer: is showing the full stack trace in the Audit Failed UI intentional design? If not, this is the right moment to change createStackTraceString() to return the top-frame or a sanitized message instead.

sig-fisherj:
**JobSubTaskEventHandler Only Catches AlertException, Not All Exceptions**                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  `JobSubTaskEventHandler.handle():`                                                                                                                                                                                                                                                                                                                                                                                               
  `catch (AlertException exception) { ... /* sends AuditFailedEvent */ }`
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  If `handleEvent()` throws an unchecked exception (e.g., a NullPointerException or a runtime exception from the Jira REST client that isn't wrapped), it propagates upward unhandled. No AuditFailedEvent is sent; the failure is invisible to the audit table.                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Compare with DistributionEventHandler, which catches both:                                                                                                                                                                                                                                                                                                                                                                     
  `} catch (AlertException alertException) { handleAlertException(...) }`                                                                                                                                                                                                                                                                                                                                                        
  `} catch (Exception unknownException) { handleUnknownException(...) }`                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                 
  Again, pre-existing in JobSubTaskEventHandler. But given this PR's goal is specifically to ensure failures surface to the audit table, it's worth asking: are there code paths in the Jira handlers that could throw non-AlertException unchecked errors? If so, JobSubTaskEventHandler should be updated to mirror DistributionEventHandler's two-catch pattern.

sig-fisherj (approved):
The only comments deal with pre-existing code that was not directly modified for this PR

DanaMaxfield (approved):

bseifert14 (approved):

ChomickiM:
> While not necessarily tied to your fix, this may be a concern.
> 
> **Stack Trace Exposed in User-Facing Audit Table (Worth Discussing)**
> 
> AuditStackTraceUtil.createStackTraceString(exception) stores the full Java stack trace in AuditFailedEvent. That trace is stored in the database and surfaced in Alert's "Audit Failed" table, visible to Alert end users.
> 
> Per security guidelines: detailed error information (stack traces, internal state) should be in server-side logs, not in API responses or user-facing views. A Java stack trace with class names, packages, and call frames reveals internal implementation details that have no diagnostic value to an Alert user but provide a roadmap for an attacker.
> 
> This is pre-existing behavior in JobSubTaskEventHandler — not introduced by IALERT-4057. But this PR expands the scope: Jira failures were previously silently dropped; they now reach the audit table. Worth a deliberate answer: is showing the full stack trace in the Audit Failed UI intentional design? If not, this is the right moment to change createStackTraceString() to return the top-frame or a sanitized message instead.

The stack trace in the UI through the audit failures feature is an intended feature of the audit failures. While I agree we should likely consider sanitization of these messages and exclusively include stack traces only within the logs, this is out of the scope of this task. The audit failures feature has existed in some capacity for several years now so any changes would be major breaking changes to change this functionality today.

PR #2800, opened 2026-08-25
IALERT-4069: Replace deprecated BD vulnerability endpoints 

ChomickiM:
The latest BD 2026.7.0 removed the internal vulnerability endpoint that Alert has depended on to populate vulnerability data in outgoing notifications. Alert previously called:
```
GET .../projects/{id}/versions/{id}/components/{id}/versions/{id}/vulnerabilities
Accept: application/vnd.blackducksoftware.internal-1+json
```
The replacement is the public project-version vulnerabilities endpoint scoped to a specific component
version via the bomComponents filter:
```
GET /api/projects/{projectId}/versions/{projectVersionId}/vulnerabilities?filter=bomComponents:{base64(globalComponentVersionUrl)}
  Accept: application/vnd.blackducksoftware.bill-of-materials-8+json
```
This endpoint returns one entry per vulnerability pre-classified by severity, with remediation statuses aggregated across all origins of the filtered component version. Each entry's _meta includes a link to the Hub vulnerability detail page, which Alert uses to get vulnerability URLs for links in notifications (LinkableItems).

Some added benefits of the new endpoint are improved performance as the origin information is combined together rather than needing to make N+1 calls for each component origin. 

BlackDuckProjectVersionComponentVulnerabilitiesView (Deleted)
 - Locally maintained copy of the old internal endpoint response model. No longer needed.
 
ProjectVersionComponentVersionVulnerabilityRemediationCvssView (Deleted) 
- CVSS POJO that was used by old vulnerabilities view and Alert used to calculate severities.

BlackDuckVersionBomVulnerabilityView (New)
- Alert response model for the new endpoint extending the BlackDuckView from blackduck-common-api.

BlackDuckComponentVulnerabilityDetailsCreator
- Updated toComponentVulnerabilities to accept List<BlackDuckVersionBomVulnerabilityView> instead. severity is read directly from the view, removing all CVSS2/3/4 branching logic.
- Updated requiresRemediation to accept a List<String> of statuses; a vulnerability is included unless all statuses are in the remediated set. This resolves the need to loop over all origins individually.

BlackDuckMessageBomComponentDetailsCreator
- Replaced the per-origin vulnerability fetch loop with a single call to the new endpoint using a bomComponents filter built from the Base64 encoded component version URL. **Note:** BD does NOT current provide this endpoint with a filter that contains the already encoded component version URL. BD engineering is investigating adding a new link as an improvement containing the properly encoded value when the request to the list endpoint is made.


provider-blackduck/src/main/java/com/blackduck/integration/alert/provider/blackduck/processor/message/service/BlackDuckMessageBomComponentDetailsCreator.java
```diff
@@ -243,22 +249,49 @@ private ComponentVulnerabilities retrieveComponentVulnerabilities(ProjectVersion
...
-            UrlMultipleResponses<BlackDuckProjectVersionComponentVulnerabilitiesView> urlMultipleResponses = new UrlMultipleResponses<>(vulnerabilitiesUrl, VULNERABILITIES_LINK.getResponseClass());
-            BlackDuckMultipleRequest<BlackDuckProjectVersionComponentVulnerabilitiesView> spec = new BlackDuckRequestBuilder()
-                .commonGet()
-                .addHeader(HttpHeaders.ACCEPT, VULNERABILITIES_MEDIA_TYPE)
-                .buildBlackDuckRequest(urlMultipleResponses);
+        HttpUrl vulnerabilitiesEndpointUrl = buildProjectVersionVulnerabilitiesUrl(bomComponent.getHref());
+        // TODO: The bom component response does not yet include a pre-encoded component version URL suitable for the
+        // bomComponents filter. Until it does, the URL is encoded manually here. When Hub provides the encoded value
+        // directly this manual encoding should become a fallback for backwards compatibility.
+        String encodedComponentVersionUrl = Base64.getEncoder().encodeToString(componentVersionUrl.getBytes(StandardCharsets.UTF_8));
```
sig-fisherj, line 262: Base64 encoding correctness — verify commit 14295da used the URL-safe encoder

The initial implementation used Base64.getEncoder() (standard Base64), which emits + and / characters that are not URL-safe as query parameter values. Commit 14295da says "Update encoding method" — it needs to be either Base64.getUrlEncoder().withoutPadding() or standard Base64 followed by explicit percent-encoding. If it's still Base64.getEncoder(), this will produce malformed filter parameters whenever the component version URL contains characters whose Base64 encoding includes + or /. That's an intermittent failure that will be hard to reproduce and diagnose.

Question: Can you confirm the final encoding in 14295da is Base64.getUrlEncoder() and not Base64.getEncoder()?

provider-blackduck/src/main/java/com/blackduck/integration/alert/provider/blackduck/processor/message/service/BlackDuckMessageBomComponentDetailsCreator.java
```diff
@@ -243,22 +249,49 @@ private ComponentVulnerabilities retrieveComponentVulnerabilities(ProjectVersion
...
+            .commonGet()
+            .addHeader(HttpHeaders.ACCEPT, VULNERABILITIES_MEDIA_TYPE)
+            .addQueryParameter("filter", filterValue)
+            .buildBlackDuckRequest(urlMultipleResponses);
+
+        List<BlackDuckVersionBomVulnerabilityView> vulnerabilityViews = blackDuckApiClient.getAllResponses(spec);
+        return vulnerabilityDetailsCreator.toComponentVulnerabilities(vulnerabilityViews);
+    }
+
+    protected HttpUrl buildProjectVersionVulnerabilitiesUrl(HttpUrl bomComponentHref) throws IntegrationException {
```
sig-fisherj, line 277: What is this method trying to do? Why are we taking the BOM Component HREF and trying to derive another URL from it? There must be a better way to construct the vulnerabilities URL than deconstructing the BOM Component URL.

sig-fisherj (changes requested):

provider-blackduck/src/main/java/com/blackduck/integration/alert/provider/blackduck/processor/message/service/BlackDuckMessageBomComponentDetailsCreator.java
ChomickiM: The only href we have available at this moment is the project/version/component/version endpoint. Additionally, the links this endpoint provides do not include a link back to the project/version vulnerabilities only to the component level vulnerabilities. If BD updates the endpoint to provide a link back to the project/version/vulnerabilities we would be able to avoid a messy carve-up of the existing URL to construct the one that we need to use. 

I will be making a request to see if a future version of BD can include this link, but Alert would still need a resolution strategy to get this link for older versions of BD.

ChomickiM:
These changes are being pulled into Alert 8.4.1 rather than 9.0.0. MR for this can be found here: https://github.com/blackducksoftware/blackduck-alert/pull/2803

Resolution of the comments in this ticket will be addressed in the 8.4.1 MR. As a result, I will be closing this MR.

